PR #2, opened 2020-02-03
Initial Testing Framework

willzeng:
I'm refactoring the example ipython notebooks that @andreamari added into automated tests.

This needs to be done for:
- [X] pyquil tests
- [x] qiskit tests

This introduces separate development dependencies from core usage dependencies.

A main advantage for removing the notebooks is that because they are not automatically run and tested they tend to be difficult to maintain and avoid having go out of date. Folding the examples into automated tests addresses this issue.


mitiq/tests/pyquil_utils.py
```diff
@@ -0,0 +1,85 @@
...
+from pyquil import Program
+# Backend and Noise simulation
+from pyquil import get_qc
+from pyquil.noise import append_kraus_to_gate
+from pyquil.gates import X, Y, Z, MEASURE
+
+QVM = get_qc('1q-qvm')
+
+# Set the random seeds for testing
+QVM.qam.random_seed = 1337
```
willzeng: @andreamari FYI this is how to add a random seed to the QVM. It's more hidden than it should be to be honest...

willzeng:
Due to struggles getting QVM/quilc running through the github actions autobuild I have reduced the scope of the automatic github CI tests to only check qiskit. You can still check both locally. Added #4 to fully solve this.

README.md
```diff
@@ -6,9 +6,25 @@ A Python toolkit for implementing error mitigation on quantum computers.
 ## Contents
 
 ## Installation
+To install locally for use:
```
andreamari:
```suggestion
To install locally use:
```

PR #5, opened 2020-02-13
Unitary folding functions (pyquil and cirq)

andreamari:
This PR is about unitary folding functions:
* Global unitary folding (with an arbitrary noise factor >= 1)
* Local gate folding (with an arbitrary noise factor >= 1)

Both functions are implemented in three different versions in three different files:
* `folding_pyquil.py` (compatible with PyQuil circuits)
* `folding_cirq.py` (compatible with Cirq circuits)
* `folding_mitiq.py` (compatible with PyQuil and Cirq circuits)

The idea is that at some point we should decide which one we prefer between the 3 options.

Also two test files are present:
* `test_folding_pyquil.py` 
* `test_folding_cirq.py`


mitiq/folding_cirq.py
```diff
@@ -0,0 +1,61 @@
...
+
+    if not (stretch >= 1):
+        raise ValueError("The stretch factor must be a real number >= 1.")
+
+    out = cirq.Circuit()
+
+    if stretch <= 3:
+        # select a fraction of subindices
+        d = len(circuit)
+        f = int(d * (stretch - 1) / 2)
```
willzeng: This would be more future proof with more verbose variable names. You never know how far in the future someone will be looking at the code and some additional info in the name is helpful. Perhaps `d --> depth` and `f --> fractional_stretch` ?

mitiq/folding_cirq.py
```diff
@@ -0,0 +1,61 @@
...
+        if sampling == True:
```
willzeng: The boolean sampling kwarg indicates that there is a modularity to this function. This function could be refactored a bit to take advantage of this modularity and make it more extensible. For example:

```
def local_folding(circuit, stretch, stretcher):
    if not (stretch >= 1):
        raise ValueError("The stretch factor must be a real number >= 1.")
    if stretch <= 3:
        return stretcher(circuit)
    else:
        # recursive application for large stretching
        _ = local_folding(circuit, 3, sampling)

        return local_folding(_, stretch / 3, sampling)

def sampling_stretcher(circuit):
        d = len(circuit)
        f = int(d * (stretch - 1) / 2)
        # indices of f random gates
        sub_indices = random.sample(range(d), f)
        return append_folds(circuit, sub_indicies)

def start_stretcher(circuit):
        d = len(circuit)
        f = int(d * (stretch - 1) / 2)
        # indices of the first f gates
        sub_indices = list(range(f))
        return append_folds(circuit, sub_indicies)

def append_folds(circuit, sub_indicies):
        out = cirq.Circuit()
        # sequentially append gates to out, folding only if j is in sub_indices
        for j, gate in enumerate(circuit):
            out += gate
            if j in sub_indices:
                out += cirq.inverse(gate)
                out += gate
         return out
```

Note this may have some typos as I haven't tested it out. Written in this manner makes it easier to add different ways of selecting the indicies or different ways of inserting / appending the gates.

mitiq/folding_cirq.py
```diff
@@ -0,0 +1,61 @@
+import random
+import cirq
+
+def local_folding(circuit, stretch, sampling=False):
```
willzeng: Missing type annotations (I'll suppress this comment for the rest though it applies in other places)

mitiq/folding_cirq.py
```diff
@@ -0,0 +1,61 @@
...
+
+def unitary_folding(circuit, stretch):
+    """Applies global unitary folding and a final partial folding of the input circuit.
+    Returns a circuit of depth approximately equal to stretch*len(circuit).
+    The stretch factor can be any real number >= 1."""
+
+    if not (stretch >= 1):
+        raise ValueError("The stretch factor must be a real number >= 1.")
+
+    d, r = divmod(stretch - 1, 2)
```
willzeng: ditto on more verbose variable names. I'll suppress this comment going forward.

mitiq/folding_cirq.py
```diff
@@ -0,0 +1,61 @@
...
+
+    # global folding
+    eye = cirq.Circuit()
```
willzeng: `cirq.Circuit()` and `cirq.inverse` --> 
```
from cirq import Circuit, inverse
```
This makes it easier to look at the imports at the top and see what parts of external dependencies are used in the file.

mitiq/folding_cirq.py
```diff
@@ -0,0 +1,61 @@
...
+    for j in range(int(d)):
+        eye += cirq.inverse(circuit) + circuit
+
+    # partial folding
+    partial = int(len(circuit) * r / 2)
+    if partial != 0:
+        eye += cirq.inverse(circuit[-partial:]) + circuit[-partial:]
+
+    return circuit + eye
```
willzeng: (PEP-8 styling suggests a newline at the end of every file) Will suppress this comment going forward

mitiq/folding_mitiq.py
```diff
@@ -0,0 +1,82 @@
+import random
+
+
+def deduce_platform(circuit):
```
willzeng: This introduces some implicit dependencies on how Cirq and pyQuil manipulate programs. Without thinking this through a lot further I'd be hesitant to program in this way. Would recommend leaving this out of the commit for now as it opens up a lot of other questions. Would be happy to start a new issue/discussion if you think that this kind of abstraction is the way we should go @andreamari

mitiq/folding_pyquil.py
```diff
@@ -0,0 +1,61 @@
+import random
+from pyquil import Program
+
+def local_folding(circuit, stretch, sampling=False):
```
willzeng: Same comments from the Cirq one apply here

mitiq/tests/test_folding_cirq.py
```diff
@@ -0,0 +1,43 @@
+import numpy as np
+import random
+import cirq
+from mitiq.folding_mitiq import local_folding, unitary_folding
+
+def random_circuit(depth):
+    """Returns a single-qubit random circuit based on Pauli gates."""
+    # defines qubit
+    q = cirq.GridQubit(0, 0)
```
willzeng: `from cirq import GridQubit` etc. Will suppress this comment going forward

mitiq/tests/test_folding_cirq.py
```diff
@@ -0,0 +1,43 @@
...
+    circuit = cirq.Circuit()
+    for _ in range(depth):
+        circuit += random.choice([cirq.X(q), cirq.Y(q), cirq.Z(q)])
+    return circuit
+
+strech_vals = [1.0, 1.5, 2.0, 2.5, 3.0, 3.5, 4.5, 4.5, 5.0]
+depth = 100
```
willzeng: A good styling principle is that for global variables at the module level one should label them in all caps to distinguish their scope and prevent weird artifacts of global state.
```suggestion
STRETCH_VALS = [1.0, 1.5, 2.0, 2.5, 3.0, 3.5, 4.5, 4.5, 5.0]
DEPTH = 100
```

mitiq/tests/test_folding_cirq.py
```diff
@@ -0,0 +1,43 @@
...
+
+def test_unitary_folding():
+    print("Testing unitary folding...")
```
willzeng:
```suggestion
```
Can remove dangling print statements

willzeng: This applies to the others in this testing file. Using `pytest --pdb` one can jump into place with the debugger if there's any issues to check rather than needing the prints. This keeps the code simpler. If that isn't clear I'm happy to demo @rmlarose @andreamari :)

andreamari:
Hi Will, thanks for your comments. I should have implemented all the suggestions.
I also agree with keeping out the platform independent implementation. This could be something to keep in mind for the future but, for the moment, better to keep things as simple as possible.

mitiq/folding_cirq.py
```diff
@@ -0,0 +1,97 @@
...
+    sub_indices = list(range(reduced_depth))
+    return fold_gates(circuit, sub_indices)
+
+
+def fold_gates(circuit: Circuit, sub_indices: List[int]) -> Circuit:
+    """Applies the map G -> G G^dag G to a subset of gates of the input circuit
+    determined by sub_indices."""
+
+    out = Circuit()
+    for j, gate in enumerate(circuit):
```
rmlarose: I don't think the behavior of this function is what is intended.

When you iterate over a cirq.Circuit, you iterate over moments, which are sets of gates which happen in the same time slice. You can still apply `cirq.inverse` to a moment, but due to the name `gate`, I don't think this is the intended behavior of the function.

The test cases in [](mitiq/tests/test_folding_cirq.py) only test single qubit circuits, for which the behavior is identical, but I think this function has un-intended consequences for multi-qubit circuits.

willzeng: @rmlarose Great point. This follows on from the discussion that we had the other day on the phone. Can you please open up an issue? Please do suggest a new PR/fix into master if you have one in mind.

willzeng: My bad for missing this in code review. But luckily we have two sets of eyes!

PR #6, opened 2020-02-13
Initial Adaptive Framework

willzeng:
### Generator-Mitigator
This PR introduces the Generator-Mitigator adaptive framework.

I thought a bit more about their names and ended up sticking with them. Generator really is just an abstract way to generate new input params based on previous inputs and previous outputs. It is just an abstract object that doesn't have anything to do with error-mitigation or quantum per se.

The Mitigator is then the object that uses a Generator to adaptively run error-mitigation using whatever run_program backend is passed.

For this same reason I have moved the scale_noise not onto the Generator but onto an argument for the `.mitigate` method of the Mitigator. It could live as a property of the mitigator, but since it is only used in one place on the Mitigator method it makes more sense to just pass it through in place there.

Happy to take any suggestions on improving the structure/naming for how this is organized.

### Other stuff
This PR also includes a small folder structure refactor, adding `pyquil/` and `qiskit/` folders for the package appropriate files. This makes it easier to run the tests for each specific backend in a modular way.


andreamari:
Thanks Will!
I have only one doubt about this:
* It would be nice if one could define new extrapolation algorithms (at the classical level) by simply modifying a unique object. With the current situation, the algorithm "lives" in the generator and in the `extrapolate` function, which is called by yet another function `zne`.

I don't have a clear solution but what about if we make generator responsible for everything which is not quantum, including also the extrapolation of the final result?

willzeng:
That's a great point @andreamari If I understand your suggestion, would it be better to make the `reduce` function an attribute method of the `Generator` object?

I like that this clearly links together reductions and Generators so that you can say what reduction works for what Generator and don't apply the wrong one. For testing you can always overload the inherent method.

andreamari:
Yes, I think this is a good solution!

willzeng:
Cool. I'll keep thinking about this and submit a commit with some changes along these lines to this branch for review before merging.

mitiq/adaptive_zne.py
```diff
@@ -0,0 +1,102 @@
...
+
+class BatchedGenerator(Generator):
+    def __init__(self, scalars: List[float], instack: List[float] = None, outstack: List[float] = None):
+        """
+        Runs a series of scalar noise parameters serially.
+        :param scalars: List of scalar noise values to be executed.
+        :param instack: Running stack of noise values run so far.
+        :param outstack: Expectations calculated thus far.
+        """
+        if instack is None:
```
rmlarose: Make `instack` a keyword argument?

mitiq/adaptive_zne.py
```diff
@@ -0,0 +1,102 @@
...
+            instack = []
+        if outstack is None:
```
rmlarose: Make `outstack` a keyword argument?

mitiq/adaptive_zne.py
```diff
@@ -0,0 +1,102 @@
...
+import mitiq.qiskit.qiskit_utils as qs_utils
+
+
+class Generator(object):
+    """
+    This object adaptively generates new noise scaling parameters based on a historical stack of previous noise
+    scale parameters and previously calculated expectation values. Noise scaling parameters are stored on the `in_stack`
+    and the running list of expectation values are stored on the `out_stack`.
+    """
+    def __init__(self, instack: List[float], outstack: List[float]):
```
rmlarose: What is our style guide for `typing` and function signatures? I vote to specify return types for all functions even `__init__`, meaning this would be changed to

```python
def __init__(self, instack: List[float], outstack: List[float]) -> None:
```

mitiq/adaptive_zne.py
```diff
@@ -0,0 +1,102 @@
...
+        """
+        if instack is None:
+            instack = []
+        if outstack is None:
+            outstack = []
+        super(self.__class__, self).__init__(instack, outstack)
+        self.scalars = scalars
+
+    def step(self, instack: List[float], outstack: List[float]):
+        next_param = self.scalars[len(outstack)]
```
rmlarose: Add error handling for `IndexError`s here. Currently there are no checks on the length of `self.scalars` or `self.outstack`. An `IndexError` will be raised when `step` is called `n` times with `n >= len(self.scalars)`.

mitiq/adaptive_zne.py
```diff
@@ -0,0 +1,102 @@
...
+        self.gen = gen
+        self.run_program = run_program
+
+    def mitigate(self, pq: Program, scale_noise) -> Tuple[List[float], List[float]]:
+        """
+        Runs the generator until convergence and return the full list of noise values and expectations that have been
+        calculated.
+        :param pq: Program to mitigate.
+        """
+        gen = self.gen
```
rmlarose: I think it's more readable to leave as `self.gen` throughout the method.

mitiq/adaptive_zne.py
```diff
@@ -0,0 +1,102 @@
...
+        if gen.is_converged(gen.instack, gen.outstack):
+            return gen.instack, gen.outstack
+        else:
+            next_param = gen.step(gen.instack, gen.outstack)
+            scaled_pq = scale_noise(pq, next_param)
+            yy = self.run_program(scaled_pq)
+            gen.outstack.append(yy)
+            return self.mitigate(pq, scale_noise)
```
rmlarose: Rewrite to simplify case handling:

```python
def mitigate(self, pq: Program, scale_noise) -> Tuple[List[float], List[float]]:
        """
        Runs the generator until convergence and return the full list of noise values and expectations that have been
        calculated.
        :param pq: Program to mitigate.
        """
        gen = self.gen
        if not gen.is_converged(gen.instack, gen.outstack):
            next_param = gen.step(gen.instack, gen.outstack)
            scaled_pq = scale_noise(pq, next_param)
            yy = self.run_program(scaled_pq)
            gen.outstack.append(yy)
            return self.mitigate(pq, scale_noise)
    return gen.instack, gen.outstack

Comment (rmlarose) [reply mitiq/adaptive_zne.py]: Better variable name for `yy`.

File: mitiq/adaptive_zne.py [hunk 7]
```diff
@@ -0,0 +1,102 @@
...
+            outstack = []
+        super(self.__class__, self).__init__(instack, outstack)
+        self.scalars = scalars
+
+    def step(self, instack: List[float], outstack: List[float]):
+        next_param = self.scalars[len(outstack)]
+        self.instack.append(next_param)
+        return next_param
+
+    def is_converged(self, instack: List[float], outstack: List[float]):
```
rmlarose: Why does `is_converged` input `instack` and (especially) `outstack`? Is there a reason this shouldn't be just

```python
def is_converged(self) -> bool:
    return len(self.instack) == len(self.scalars)
```

mitiq/adaptive_zne.py
willzeng: Sounds good to me +1

mitiq/adaptive_zne.py
willzeng: Did you have in mind setting `instack=[]` as the default? The reason I didn't do that is because Python has a some unintuitive behavior for empty lists as default args. More about that is here: https://nikos7am.com/posts/mutable-default-arguments/

mitiq/adaptive_zne.py
```diff
@@ -0,0 +1,102 @@
+from pyquil import Program
+from typing import List, Tuple
+
+from mitiq.zne import richardson_extr
+import mitiq.qiskit.qiskit_utils as qs_utils
```
willzeng:
```suggestion
import mitiq.qiskit.qiskit_utils as blarg
```

andreamari:
> See inline comments.
> 
> Does anyone else think of a usual Python generator when thinking about the `Generator`? I don't know if this is reason enough to change the name, but it could be slightly confusing.

What about `Extrapolator` ?

andreamari:
Do we really need the Mitigator class?  We only use its method `mitigate` which maybe could be a simple function.

willzeng:
Yep I think Mitigator could just be a function as currently structured

mitiq/adaptive_zne.py
willzeng: added types

mitiq/adaptive_zne.py
willzeng: added

mitiq/adaptive_zne.py
willzeng: changed

mitiq/adaptive_zne.py
willzeng: The idea is that it means you can check an object to see if different conditions converge (and you see exactly what attributes of the object are checked to see convergence) without needing to modify the state of the generator

willzeng:
Have made the suggested changes, including a rename of `Generator` to `Factory`. This is ready for re-review for merge @andreamari @rmlarose

andreamari:
Thanks @willzeng for the very nice refactoring! 
From my side we can marge, however we should first solve the following two issues about the `reduce` method which I think are important: 

1. `reduce` should in general take also the scaling parameters, in addition to the expectation values, as arguments. Equivalently we could also pass `instack` and `outstack`.  Currently, your are passing only expectation values and the fact that you don't get an error is only for a very lucky combination of  constant values and default values! :-)
 
2. If you agree, I would not even specify a particular `reduce` method in `Factory` and we could instead put a not implemented template. The idea is that, `Factory` and `BatchedFactory` should represent abstract extrapolation algorithms. In a new branch I am implementing several specific algorithms (Richardson, linear fit, poly fit, etc.) in the form of child classes of `Factory` or `BatchedFactory`. I think that specific `reduce` methods should be implemented in such child classes and not in the abstract ones.

What do you think?

mitiq/adaptive_zne.py
```diff
@@ -0,0 +1,100 @@
...
+        self.instack = instack
+        self.outstack = outstack
+
+    def step(self, instack: List[float], outstack: List[float]) -> None:
+        raise NotImplementedError
+
+    def is_converged(self, instack: List[float], outstack: List[float]) -> None:
+        raise NotImplementedError
+
+    def reduce(self, expectations: List[float]) -> float:
```
rmlarose: Going with @andreamari's comment, maybe `reduce` should have an argument that specifies the method?

rmlarose:
> 1. `reduce` should in general take also the scaling parameters, in addition to the expectation values, as arguments. Equivalently we could also pass `instack` and `outstack`.  

`instack` and `outstack` are class attributes so shouldn't need to be passed in as parameters. If I understand how a `Factory` is being used correctly, I think the argument `expectations` of `reduce` will always be `outstack` as well (?).

> 2. If you agree, I would not even specify a particular `reduce` method in `Factory` and we could instead put a not implemented template. The idea is that, `Factory` and `BatchedFactory` should represent abstract extrapolation algorithms. In a new branch I am implementing several specific algorithms (Richardson, linear fit, poly fit, etc.) in the form of child classes of `Factory` or `BatchedFactory`. I think that specific `reduce` methods should be implemented in such child classes and not in the abstract ones.

An alternative, which is perhaps simpler, would be to specify a method for `reduce`. This could be a class attribute or an argument to the `reduce` method, which would then do the appropriate fitting (Richardson, linear, etc.) and return the result.

willzeng:
Definitely agreed on the points that reduce should be child classes and likely should take more general arguments. Given that you are already working on this in a new branch @andreamari I propose that we leave those changes to a new PR that you're working on and merge this one for now with the basic scaffolding (assuming the rest of the PR looks good).

What do you two think about this?

andreamari:
> `instack` and `outstack` are class attributes so shouldn't need to be passed in as parameters. If I understand how a `Factory` is being used correctly, I think the argument `expectations` of `reduce` will always be `outstack` as well (?).
> 

I am fine with both options: passing `instack` and `outstack` as arguments or simply using the internal attributes. However, for consistency, whatever decision we take should probably apply also to the methods `step` and `is_converged`.

> An alternative, which is perhaps simpler, would be to specify a method for `reduce`. This could be a class attribute or an argument to the `reduce` method, which would then do the appropriate fitting (Richardson, linear, etc.) and return the result.

Since a generic (adaptive) algorithm is determined by both the `reduce` and the `step` methods, I like the idea of having for each particular algorithm a self-contained class of `Factory` type. For the user this approach would be intuitive and simple to use: just select your favorite Factory and that's it.  Anyway, we can discuss about this in a new PR.

willzeng:
Is this good to merge? If so, can one of you (@andreamari @rmlarose approve the merge?)

PR #7, opened 2020-02-14
Adds unitary folding methods in Cirq.

rmlarose:


willzeng:
@rmlarose I like having these new methods of doing the folds. Now that we have merged #5 it would improve this PR to refactor these features into `folding_cirq.py`. I think this will end up being pretty clean and with less duplicated functionality. Do you agree?

rmlarose:
@willzeng agree completely. I will close this PR and merge code to address #8 as appropriate.

PR #10, opened 2020-02-16
Fixes indexing  of circuits in local folding methods for Cirq, adds tests

### LINKED ISSUE #8: Folding in Cirq for multi-qubit circuits
Folding a subset of gates [(fold_gates)](https://github.com/unitaryfund/mitiq/blob/master/mitiq/folding_cirq.py#L40) in Cirq iterates over moments instead of gates.

I propose we rename `fold_gates` to `fold_moments` and add another method `fold_gates` which correctly iterates over scheduled operations to fold a subset of gates. The latter is implemented in https://github.com/unitaryfund/mitiq/pull/7

rmlarose:
Fixes #8

### DIFF
--- a/mitiq/folding_cirq.py
+++ b/mitiq/folding_cirq.py
@@ -11,6 +11,21 @@ MAX_STRETCH_FACTOR = 100
 
 
 # Gate level folding
+def _fold_gate_at_index_in_moment(circuit: Circuit, moment_index: int, gate_index: int) -> None:
+    """Modifies the input circuit by replacing the gate G in (moment, index) is replaced by G G^dagger G.
+
+    Args:
+        circuit: Circuit to fold.
+        moment_index: Moment in which the gate sits in the circuit.
+        gate_index: Index of the gate within the specified moment.
+
+    Returns:
+        None
+    """
+    op = circuit[moment_index].operations[gate_index]
+    circuit.insert(moment_index, [op, inverse(op)], strategy=InsertStrategy.NEW)
+
+
 def fold_gate_at_index_in_moment(circuit: Circuit, moment_index: int, gate_index: int) -> Circuit:
     """Returns a new circuit where the gate G in (moment, index) is replaced by G G^dagger G.
 
@@ -20,9 +35,24 @@ def fold_gate_at_index_in_moment(circuit: Circuit, moment_index: int, gate_index
         gate_index: Index of the gate within the specified moment.
     """
     folded = deepcopy(circuit)
-    op = folded[moment_index].operations[gate_index]
-    folded.insert(moment_index, [inverse(op), op], strategy=InsertStrategy.NEW)
+    _fold_gate_at_index_in_moment(folded, moment_index, gate_index)
     return folded
+
+
+def _fold_gates_in_moment(circuit: Circuit, moment_index: int, gate_indices: Iterable[int]) -> None:
+    """Modifies the input circuit by applying the map G -> G G^dag G to all gates specified by
+     the input moment index and gate indices.
+
+     Args:
+         circuit: Circuit to fold.
+         moment_index: Index of moment to fold gates in.
+         gate_indices: Indices of gates within the moments to fold.
+
+     Returns:
+         None
+     """
+    for (i, gate_index) in enumerate(gate_indices):
+        _fold_gate_at_index_in_moment(circuit, moment_index + 2 * i, gate_index)  # Each fold adds two moments
 
 
 def fold_gates_in_moment(circuit: Circuit, moment_index: int, gate_indices: Iterable[int]) -> Circuit:
@@ -35,8 +65,7 @@ def fold_gates_in_moment(circuit: Circuit, moment_index: int, gate_indices: Iter
          gate_indices: Indices of gates within the moments to fold.
      """
     folded = deepcopy(circuit)
-    for (i, gate_index) in enumerate(gate_indices):
-        folded = fold_gate_at_index_in_moment(folded, moment_index + 2 * i, gate_index)  # Each fold adds two moments
+    _fold_gates_in_moment(folded, moment_index, gate_indices)
     return folded
 
 
@@ -78,7 +107,7 @@ def fold_moments(circuit: Circuit, moment_indices: List[int]) -> Circuit:
     folded = deepcopy(circuit)
     shift = 0
     for i in moment_indices:
-        folded.insert(i + shift, [inverse(circuit[i]), circuit[i]])
+        folded.insert(i + shift, [circuit[i], inverse(circuit[i])])
         shift += 2
     return folded
 
--- a/mitiq/folding_cirq.py
+++ b/mitiq/folding_cirq.py
@@ -124,8 +124,15 @@ def fold_moments(circuit: Circuit, moment_indices: List[int]) -> Circuit:
 
 
 def _fold_all_gates_locally(circuit: Circuit) -> Circuit:
-    """Replaces every gate G with G G^dag G in the circuit."""
+    """Replaces every gate G with G G^dag G by modifying the circuit in place."""
+    _fold_moments(circuit, range(len(circuit)))
 
+
+def fold_all_gates_locally(circuit: Circuit) -> Circuit:
+    """Returns a new circuit with every gate G replaced by G G^dag G."""
+    folded = deepcopy(circuit)
+    _fold_all_gates_locally(folded)
+    return folded
 
 
 def fold_gates_from_left(circuit: Circuit, stretch: float) -> Circuit:
--- a/mitiq/folding_cirq.py
+++ b/mitiq/folding_cirq.py
@@ -90,6 +90,23 @@ def fold_gates(circuit: Circuit, moment_indices: Iterable[int], gate_indices: Li
     return folded
 
 
+def _fold_moments(circuit: Circuit, moment_indices: List[int]) -> Circuit:
+    """Folds specified moments in the circuit in place.
+
+    Args:
+        circuit: Circuit to fold.
+        moment_indices: Indices of moments to fold in the circuit.
+
+    Returns:
+        None
+    """
+    shift = 0
+    for i in moment_indices:
+        circuit.insert(i + shift, [circuit[i + shift], inverse(circuit[i + shift])])
+        shift += 2
+    return circuit
+
+
 def fold_moments(circuit: Circuit, moment_indices: List[int]) -> Circuit:
     """Returns a new circuit with moments folded by mapping
 
@@ -102,11 +119,13 @@ def fold_moments(circuit: Circuit, moment_indices: List[int]) -> Circuit:
         moment_indices: List of integers that specify moments to fold.
     """
     folded = deepcopy(circuit)
-    shift = 0
-    for i in moment_indices:
-        folded.insert(i + shift, [circuit[i], inverse(circuit[i])])
-        shift += 2
+    _fold_moments(folded, moment_indices)
     return folded
+
+
+def _fold_all_gates_locally(circuit: Circuit) -> Circuit:
+    """Replaces every gate G with G G^dag G in the circuit."""
+
 
 
 def fold_gates_from_left(circuit: Circuit, stretch: float) -> Circuit:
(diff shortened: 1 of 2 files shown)


rmlarose:
Fixes #8

mitiq/folding_cirq.py
```diff
@@ -1,76 +1,197 @@
...
+        np.random.seed(kwargs.get("seed"))
 
-# ==================================================
-# Circuit level folding
-# ==================================================
+    ngates = len(list(folded.all_operations()))
+    num_to_fold = int(ngates * (stretch - 1.0) / 2.0)
 
+    for _ in range(num_to_fold):
+        # TODO: This allows for gates to be folded more than once, and folded gates to be folded.
```
willzeng: My intuition is that gates should be allowed to be sampled more than once, but that one should not sample from the new folds. Do you agree @andreamari I think this change is made doing:
```
        moment_index = np.random.choice(len(circuit))
        gate_index = np.random.choice(len(circuit[moment_index]))
```

mitiq/folding_cirq.py
```diff
@@ -1,76 +1,197 @@
...
+        #  Should this be allowed?
+        moment_index = np.random.choice(len(folded))
+        gate_index = np.random.choice(len(folded[moment_index]))
+        # TODO: It could be expensive to call fold_gate...(...) which makes a deepcopy of the circuit each call.
```
willzeng: same comment on mutability from before (am noting this todo)

mitiq/folding_cirq.py
```diff
@@ -1,76 +1,197 @@
...
+        #  ==> Possible fix: Have fold_gate_at_index_in_moment(...) modify the circuit in place.
+        folded = fold_gate_at_index_in_moment(folded, moment_index, gate_index)
 
+    return folded
+
+
+def fold_local(circuit: Circuit, stretch: float, fold_method: Callable = fold_gates_from_left, **kwargs) -> Circuit:
```
willzeng: You can add the type signature into the Callable type as follows
```suggestion
def fold_local(circuit: Circuit, stretch: float, fold_method: Callable[[Circuit, float], Circuit] = fold_gates_from_left, **kwargs) -> Circuit:
```

willzeng: Also, are the kwargs here used anywhere? I didn't catch it but may have missed it. It is one possible way to extend thing, but if we don't need this extension yet I'd recommend removing the `**kwargs`

mitiq/folding_cirq.py
```diff
@@ -1,76 +1,197 @@
...
+"""Functions to fold gates in Cirq circuits."""
+from copy import deepcopy
+from typing import (Callable, Iterable, List)
+
+import numpy as np
+
+from cirq import (Circuit, InsertStrategy, inverse)
+
+
+MAX_STRETCH_FACTOR = 100
```
willzeng: Can you say more about why you included this @rmlarose ?

mitiq/folding_cirq.py
```diff
@@ -1,76 +1,197 @@
...
+    if np.isclose(stretch, 1.0, atol=1e-2):
+        return folded
+
+    if not 1 < stretch <= MAX_STRETCH_FACTOR:
+        raise ValueError(f"The stretch factor must be a real number between 1 and {MAX_STRETCH_FACTOR}.")
+
+    while stretch > 1.:
+        this_stretch = 3. if stretch > 3. else stretch
+        # TODO: This also allows folding gates that have already been folded. Should this be allowed?
+        folded = fold_method(folded, this_stretch, **kwargs)
```
willzeng: Good point. As before I don't think it should be. The change, I think, would be to add a `circ_copy = circuit` outside the loop and then do
```suggestion
        folded = fold_method(circuit, this_stretch, **kwargs)
```

mitiq/tests/test_folding_cirq.py
```diff
@@ -1,43 +1,414 @@
...
     circuit = Circuit()
     for _ in range(depth):
-        circuit += random.choice([X(q), Y(q), Z(q)])
+        circuit += random.choice([ops.X(qubit), ops.Y(qubit), ops.Z(qubit)])
     return circuit
 
 
-STRETCH_VALS = [1.0, 1.5, 2.0, 2.5, 3.0, 3.5, 4.5, 4.5, 5.0]
-DEPTH = 100
+def _equal(circuit_one: Circuit, circuit_two: Circuit) -> bool:
```
willzeng: I'm sort of surprised this isn't built into Cirq. Maybe should upstream it? What do you think @rmlarose ?

mitiq/tests/test_folding_cirq.py
```diff
@@ -1,43 +1,414 @@
...
+    folded = fold_gate_at_index_in_moment(circ, moment_index=1, gate_index=0)
+    assert folded == Circuit(
+        [ops.H.on(qbit)] + [ops.X.on(qbit)] * 3 + [ops.Z.on(qbit)]
+    )
+    # Fold the zeroth operation in the second moment
+    folded = fold_gate_at_index_in_moment(circ, moment_index=2, gate_index=0)
+    assert folded == Circuit(
+        [ops.H.on(qbit)] + [ops.X.on(qbit)] + [ops.Z.on(qbit)] * 3
+    )
+    # Make sure the original circuit wasn't modified
```
willzeng: good check

mitiq/folding_cirq.py
rmlarose: > You can add the type signature into the Callable type as follows

I didn't know this -- nice!

rmlarose: > Also, are the kwargs here used anywhere? I didn't catch it but may have missed it. It is one possible way to extend thing, but if we don't need this extension yet I'd recommend removing the `**kwargs`

The `fold_method` could have `kwargs`, so I included `kwargs` in `fold_local`. For example, `fold_gates_at_random` has `seed` as a kwarg (for testing).

I admit the only keyword arguments there are now is a random seed for testing, but it could be good to allow for this. We might come up with some wild `fold_method`s...

mitiq/folding_cirq.py
rmlarose: I have the same intuition, not 100% sure though. I'll wait to hear Andrea's and other's thoughts, it's not too hard to change either way.

mitiq/folding_cirq.py
rmlarose: Mostly because of runtime considerations, especially with making deepcopy's every iteration. Generally I think it's not useful to stretch past a certain point, though, so we might want to consider some cutoff. The number `100` is pretty arbitrary.

mitiq/tests/test_folding_cirq.py
rmlarose: I actually think it's non-trivial to check circuit equality. I had a hunch the two checks in this function weren't completely correct, and this turns out to be true. Consider the two following circuits:

Circuit 1: H(0), H(0), H(0), X(0)
Circuit 2: H(0), X(0), X(0), X(0)

The number of operations is the same and the set of operations is the same, but the circuits aren't equal.

Fortunately, the (slightly) non-trivial way to properly check circuit equality -- by seeing if the directed acyclic graphs are isomorphic -- is already [implemented in Cirq](https://github.com/quantumlib/Cirq/blob/master/cirq/circuits/circuit_dag.py#L128). I'll update the `_equal` function to use this.

mitiq/folding_cirq.py
```diff
@@ -1,76 +1,197 @@
...
+    num_folded = 0
+    moment_shift = 0
 
-    depth = len(circuit)
-    reduced_depth = int(depth * (stretch - 1) / 2)
-    sub_indices = list(range(reduced_depth))
-    return fold_gates(circuit, sub_indices)
+    for (moment_index, moment) in enumerate(circuit):
+        for gate_index in range(len(moment)):
+            # TODO: It could be expensive to call fold_gate...(...) which makes a deepcopy of the circuit each call.
```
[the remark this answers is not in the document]
rmlarose: I think this will work. I wonder if we would ever require `mutate=False`, certainly I think users wouldn't need this. It might make sense to define this as a private function

```python
def _fold_gate_at_index_in_moment(circuit: Circuit, moment_index: int, gate_index: int) -> Circuit:
    op = circuit[moment_index].operations[gate_index]
    circuit.insert(moment_index, [inverse(op), op], strategy=InsertStrategy.NEW)
```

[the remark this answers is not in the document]
willzeng: +1 to your suggestion above

mitiq/folding_cirq.py
willzeng: true, but **kwargs can be risky as it hides function signatures. My suggestion would be to explicitly make it `seed` as the argument for now. If we end up with lots of different args then we can consider a refactor later

mitiq/folding_cirq.py
willzeng: Given that the limitation seems arbitrary, I recommend that it be removed for now.

mitiq/tests/test_folding_cirq.py
willzeng: +1 good catch

mitiq/folding_cirq.py
andreamari: My intuition is that it is better to distribute the folding gates as much as possible along the circuit, so I think it is better not to sample the same gate of the original circuit (unless `stretch >3`) AND not to sample previous folds.  
This is not so easy especially because 2 new moments are created for each gate folding.
So, before solving this issue I think it is more important to decide if this proliferation of moments is good or bad. Isn't more natural to fold a gate within its moment? In other words isn't more natural that the number of input and output moments is equal for local folding?

rmlarose:
Addresses all points except folding gates that have already been folded, as I don't think we've reached a consensus on this. (Have we?)

I think the previous solution using **kwargs in the fold_method was more elegant, but this is a working version. Another possibility is to use *args.

mitiq/folding_cirq.py
```diff
@@ -1,76 +1,234 @@
...
+    num_folded = 0
+    moment_shift = 0
 
-    depth = len(circuit)
-    reduced_depth = int(depth * (stretch - 1) / 2)
-    sub_indices = random.sample(range(depth), reduced_depth)
-    return fold_gates(circuit, sub_indices)
+    for (moment_index, moment) in enumerate(circuit):
+        for gate_index in range(len(moment)):
+            # TODO: It could be expensive to call fold_gate...(...) which makes a deepcopy of the circuit each call.
```
willzeng: Any reason not to update this to use the in place `_fold_gate_at_index_in_moment` and remove the todo?

mitiq/folding_cirq.py
```diff
@@ -1,76 +1,234 @@
...
-    Returns a circuit of depth approximately equal to stretch*len(circuit).
-    The stretch factor can be any real number >= 1."""
+    for _ in range(num_to_fold):
+        # TODO: This allows for gates to be folded more than once, and folded gates to be folded.
+        #  Should this be allowed?
+        moment_index = np.random.choice(len(folded))
+        gate_index = np.random.choice(len(folded[moment_index]))
+        # TODO: It could be expensive to call fold_gate...(...) which makes a deepcopy of the circuit each call.
+        #  ==> Possible fix: Have fold_gate_at_index_in_moment(...) modify the circuit in place.
+        folded = fold_gate_at_index_in_moment(folded, moment_index, gate_index)
```
willzeng: Any reason not to change this and remove the todo?
```suggestion
        folded = _fold_gate_at_index_in_moment(folded, moment_index, gate_index)
```

willzeng:
Additionally I opened #12 for discussion on resampling folded gates

andreamari:
I just noticed that the method `fold_moments` is not compatible with `fold_local` (it cannot be passed as an argument). So maybe could remove it and just focus on gates. Alternatively we need to implement something like `fold_moments_at_random` and `fold_moments_from_left`.

mitiq/folding_cirq.py
rmlarose: Nope -- missed this, thanks!

mitiq/folding_cirq.py
rmlarose: Likewise missed this, thanks!

mitiq/folding_cirq.py
```diff
@@ -1,76 +1,248 @@
...
+    Args:
+        circuit: Circuit to apply folding operation to.
+        moment_indices: List of integers that specify moments to fold.
+    """
+    folded = deepcopy(circuit)
+    _fold_moments(folded, moment_indices)
+    return folded
+
+
+def _fold_all_gates_locally(circuit: Circuit) -> Circuit:
```
willzeng: should this be empty

Issue #9, opened 2020-02-14
Folding functions should retain circuit properties (e.g., num shots) and device information (e.g., qubits, connectivity, etc.)


rmlarose:
Our current folding techniques for Cirq build the folded circuit from an empty circuit. Circuits in Cirq can be tied to a device which tells information about qubits, connectivity, and gate operation times. As such, I think it's more appropriate to copy the input circuit and insert gates to do the folding. Thoughts?

andreamari:
I think it's a good point. Moreover the same problem probably exists also for PyQuil circuits (e.g. `num_shots` can be a property of a circuit). 

I agree with your solution, otherwise a simple alternative is also this one: 
```empty_circuit = input_circuit[0:0]```  
which creates an empty circuit with the same properties of the input and should work for both Cirq and PyQuil circuits.

**Comment by rmlarose** [comment]
Good point about pyQuil circuits and properties. 

I tested

```python
empty_circuit = input_circuit[0:0]
```

and it didn't copy over the shots.

I'll update the name of this issue to reflect the importance for pyQuil as well.

andreamari:
It is true that `empty_circuit = input_circuit[0:0]` doesn't copy `num_shots`. But I also noticed that  `circuit_sum= input_circuit +  input_circuit` doesn't copy `num_shots` to  `circuit_sum` !
So I wonder if this is a bug or a feature.

An empty circuit in PyQuil can be correctly created in this way 
```empy_circuit = copy_everything_except_instructions()```

however even in this case `num_shots` would be lost as soon as we start making simple operations (e.g., additions).

andreamari:
I will make a PR to improve the creation of empty circuits in PyQuil using `copy_everything_except_instructions()`. I am also adding some tests associated to this issue.

The problem with `num_shots` and similar properties remains. But I am quite sure that this is a bug of PyQuil. See, e.g., the  `__add__` and `__getitem__` method of the `Program` class where it is evident that properties are not retained (https://github.com/rigetti/pyquil/blob/master/pyquil/quil.py).

willzeng:
@andreamari does the issue appear specific to num_shots instructions? If so, perhaps dedicated code should exist to pop off this instruction?

Alternatively, we could code for with the assumption that the user does not choose numshots for things to be mitigated. When you run a given Program through mitigation it isn't clear what the numshots should do as the Mitigator will have discretion to choose optimal amounts of shots to take etc

willzeng:
[...]

PR #11, opened 2020-02-19
Improves PyQuil folding functions and tests

andreamari:
1. Partially solves the issue #9, improving the creation of empty circuits retaining the input properties.

2. Implements several new tests (unitary equivalence, mutation of input and output properties)


mitiq/folding_pyquil.py
```diff
@@ -41,7 +41,7 @@ def fold_gates(circuit: Program, sub_indices: List[int]) -> Program:
     """Applies the map G -> G G^dag G to a subset of gates of the input circuit
     determined by sub_indices."""
 
-    out = Program()
+    out = circuit.copy_everything_except_instructions()
```
willzeng: neat find! I didn't know they already had that function :)

mitiq/tests/test_folding_pyquil.py
```diff
@@ -1,48 +1,112 @@
...
+)
+from copy import deepcopy
+
+
+STRETCH_VALS = [1.0, 1.5, 2.0, 2.5, 3.0, 3.5, 4.0]
+DEPTH = 50
+
+# NUM_SHOTS must be 1 in order to pass the next tests.
+# This is due to the fact that num_shots are not retained by
+# some methods of the Program class (bug?)
```
willzeng: Can you say a little more about what this means? What is an example method that loses the num_shots?

mitiq/tests/test_folding_pyquil.py
```diff
@@ -1,48 +1,112 @@
...
-
 def test_unitary_folding():
-    print("Testing unitary folding...")
     for c in STRETCH_VALS:
         circ = random_circuit(DEPTH)
+        circ_copy = deepcopy(circ)
         out = unitary_folding(circ, c)
         actual_c = len(out) / len(circ)
-        print("Input stretch: {:}    Real stretch: {}".format(c, actual_c))
+        # test lenght scaling
```
willzeng:
```suggestion
        # test length scaling
```

mitiq/tests/test_folding_pyquil.py
```diff
@@ -1,48 +1,112 @@
...
 
 def test_local_folding_nosamp():
-    print("Testing local folding (no sampling)...")
     for c in STRETCH_VALS:
         circ = random_circuit(DEPTH)
+        circ_copy = deepcopy(circ)
         out = local_folding(circ, c, stretcher=sampling_stretcher)
         actual_c = len(out) / len(circ)
-        print("Input stretch: {:}    Real stretch: {}".format(c, actual_c))
+        # test lenght scaling
```
willzeng:
```suggestion
        # test length scaling
```

mitiq/tests/test_folding_pyquil.py
```diff
@@ -1,48 +1,112 @@
...
+from mitiq.folding_pyquil import (
+    local_folding,
+    unitary_folding,
+    sampling_stretcher,
+    left_stretcher,
+)
+from copy import deepcopy
+
+
+STRETCH_VALS = [1.0, 1.5, 2.0, 2.5, 3.0, 3.5, 4.0]
```
willzeng: For testing purposes might be good to change this to use not only integer and half-integer values and not to be evenly spaced, e.g.

```suggestion
STRETCH_VALS = [1.0, 1.3, 2.0, 2.6, 3.0, 3.5, 4.0]
```

mitiq/tests/test_folding_pyquil.py
```diff
@@ -1,48 +1,112 @@
...
         out = unitary_folding(circ, c)
         actual_c = len(out) / len(circ)
-        print("Input stretch: {:}    Real stretch: {}".format(c, actual_c))
+        # test lenght scaling
         assert np.isclose(c, actual_c, atol=1.0e-1)
+        # test unitaries are equal
+        u_in = program_unitary(circ, 2)
+        u_out = program_unitary(out, 2)
+        assert (u_in == u_out).all
+        # test input is not mutated up to "_synthesized_instructions"
```
willzeng: Given that this set of testing instructions is copied between all three tests, would it be better to have this as a separate function that is then called three times?

andreamari:
> Looks good @andreamari! I made mostly cosmetic comments.
> 
> One thing I would like to understand better is what is going on with the num_shots stuff. That might be a bigger problem than we realize. Could you say a little more about it?

The problem with `num_shots` and similar properties of a `Program` is that they are not preserved by the main methods (e.g. `__add__` and `__getitem__` ) the `Program` class.  I think it is better if we discuss about this problem in the dedicated issue #9.

willzeng:
Your right @andreamari am continuing this discussion on #9 I should have checked there beforehand

mitiq/folding_pyquil.py
rmlarose: Also wasn't aware of this -- nice!

mitiq/tests/test_folding_pyquil.py
```diff
@@ -1,48 +1,112 @@
 import numpy as np
 import random
 from pyquil import Program
-from pyquil.gates import X, Y, Z
-from mitiq.folding_pyquil import local_folding, unitary_folding, sampling_stretcher, left_stretcher
```
rmlarose: More of a future point, but I think it would be good to discuss common naming conventions for (folding) functions. I generally default to function names which are verbs that (hopefully) describe what the function does, e.g. `fold_gates_from_left`, but there are different styles. Eventually we'll want consistency in folding function names, but again more of a future point.

mitiq/tests/test_folding_pyquil.py
```diff
@@ -1,48 +1,112 @@
...
 
 def random_circuit(depth: int):
-    """Returns a single-qubit random circuit based on Pauli gates."""
+    """Returns a 2-qubit random circuit based on a simple gate set."""
     prog = Program()
+    gate_set = [X(0), Y(0), Z(0), X(1), Y(1), Z(1), CNOT(0, 1), CNOT(1, 0)]
     for _ in range(depth):
-        prog += random.choice([X(0), Y(0), Z(0)])
+        prog += random.choice(gate_set)
+    prog.num_shots = NUM_SHOTS
```
rmlarose: Should this be `prog.wrap_in_numshots_loop`? Is there a difference?

mitiq/tests/test_folding_pyquil.py
rmlarose: +1

mitiq/tests/test_folding_pyquil.py
willzeng: +1 to using verbs for function names. I like that convention.

mitiq/tests/test_folding_pyquil.py
andreamari: Yes!

mitiq/tests/test_folding_pyquil.py
andreamari: Agree. I will use the same names used for Cirq

mitiq/tests/test_folding_pyquil.py
andreamari: This part is changed. That comment is removed.

andreamari: This part is changed. That comment is removed.

mitiq/tests/test_folding_pyquil.py
andreamari: In practice it's the same. But I used your suggestion because `wrap_in_numshots_loop` looks more similar to what a user would do in a real scenario.

andreamari:
I think I have applied all suggestions and solved the main open problems of PyQuil folding functions:
 
- Better names (and docs) for functions, which are now consistent with those of `folding_cirq.py`.
- Added the possibility of using a seed, in the same way as we do for Cirq.
- Fixed issue #9 (circuit properties not retained) by properly appending instructions to the output circuit.
- Removed repetitions of the same code in tests with the new function `apply_folding_tests`.
- New test for seed parameter.

mitiq/folding_pyquil.py
```diff
@@ -1,76 +1,110 @@
...
-    sub_indices = random.sample(range(depth), reduced_depth)
+    if not (1 <= stretch <= 3):
+        raise ValueError("The stretch factor must be a real number between 1 and 3.")
+
+    if seed is not None:
+        np.random.seed(seed)
+
+    ngates = len(circuit)
+    num_to_fold = int(ngates * (stretch - 1) / 2)
+    sub_indices = np.random.choice(range(ngates), num_to_fold, replace = False)
```
rmlarose:
```suggestion
    sub_indices = np.random.choice(range(ngates), num_to_fold, replace=False)
```

rmlarose: (PEP8)

mitiq/folding_pyquil.py
```diff
@@ -1,76 +1,110 @@
...
-
-def local_folding(circuit: Program, stretch: float, stretcher: Callable = left_stretcher) -> Program:
-    """Applies the map G -> G G^dag G to a subset of gates of the input circuit.
-    Returns a circuit of depth approximately equal to stretch*len(circuit).
-    The stretch factor can be any real number >= 1."""
-
+def fold_local(
+        circuit: Program,
+        stretch: float,
+        fold_method: Callable = fold_gates_from_left,
```
rmlarose:
```suggestion
        fold_method: Callable[[Callable, float, Tuple[Any]], Circuit] = fold_gates_from_left,
```

rmlarose: ^ Add type signature to `fold_method` arg.

willzeng:
LGTM once that last type signature is in place @andreamari

mitiq/folding_pyquil.py
andreamari: Ok. But isn't this the correct signature: `Callable[[Program, float, Tuple[Any]], Program]` ?

mitiq/folding_pyquil.py
andreamari: Right, thanks!

mitiq/folding_pyquil.py
willzeng: yep that looks right @andreamari

PR #16, opened 2020-02-21
add missing return statement

willzeng:


willzeng:
@rmlarose quick ping on this small edit

willzeng:
If it looks good, go ahead and merge after approval :)

willzeng:
Got it. Also the type signature should change too!

Issue #12, opened 2020-02-20
Should refolding be allowed


willzeng:
This issue has two parts:
1. In `fold_gates_at_random` gates can be sampled multiple times AND (because the new circuit is built up iteratively, folded gates can also be sampled. IMO it should be possible to resample the same gate but we should not be sampling from folded gates.

2. In `fold_local` , if you are stretching by more than 3X, then multiple folds need to be made to get to the target stretch. IMO this type of refolding should be allowed (I don't see away around it to get to the target stretch).

This extended the conversation raised in #10 What do you think @andreamari @rmlarose ?

andreamari:
1. I agree at 99%. IMO,  I would also avoid re-sampling the same gate. Let's make an extreme example: imagine the case `stretch=3`, isn't more natural to uniformly fold all the gates of the circuit? Anyway, my doubt is probably a negligible subtlety.

2. Agree.

3. I would also add a third point: when we need to fold more gates which are originally in the same moment, should we create a new moment for each new gate, or should we instead add gates "simultaneously"  (i.e., `InsertStrategy.EARLIEST` ).

willzeng:
1. I agree with @andreamari's point. We should knock out gates once they have been sampled so that they are not re-sampled.

3. Good point. I think it would be best to add gates as soon as possible otherwise the circuit length will grow much longer than we we want. @rmlarose what do you think? You know the moments in cirq a bit better than we do

rmlarose:
So `fold_gates_at_random` should really only be used with `1 < stretch < 3`, since if `stretch = 3`, we are going to fold every gate, which isn't random.

Just to summarize, if `1 < stretch < 3`, we are going to uniformly sample from "real" gates (not "virtual" = folded gates), and allow for folding real gates more than once.

rmlarose:
> 3. I would also add a third point: when we need to fold more gates which are originally in the same moment, should we create a new moment for each new gate, or should we instead add gates "simultaneously"  (i.e., `InsertStrategy.EARLIEST` ).

The reason I created a new moment for each [operation, inverse(operation)] pair was to make indexing easier. Each new fold thus adds two moments, and therefore we know how to reference moments/gates in the original circuit and in the new folded circuit.

With a different insert strategy, it becomes less trivial to keep track of this. I don't think it's impossible, but definitely not as simple.

[...]

PR #18, opened 2020-02-25
Changes behavior of fold_gates_at_random to...

rmlarose:
1. Only fold each gate in the circuit once.
2. Not fold "virtual" gates -- i.e., gates which have already been folded.

Edits and adds tests.

### DIFF
--- a/mitiq/folding_cirq.py
+++ b/mitiq/folding_cirq.py
@@ -202,10 +202,10 @@ def fold_gates_at_random(circuit: Circuit, stretch: float, seed: Optional[int] =
     # Keep track of which gates we can fold in each moment
     remaining_gate_indices = {moment: list(range(len(circuit[moment]))) for moment in range(len(circuit))}
 
-    for _ in range(num_to_fold):
-        # Any moment with at least one gate is fair game
-        remaining_moment_indices = [i for i in remaining_gate_indices.keys() if remaining_gate_indices[i]]
+    # Any moment with at least one gate is fair game
+    remaining_moment_indices = [i for i in remaining_gate_indices.keys() if remaining_gate_indices[i]]
 
+    for _ in range(num_to_fold):
         # Get a moment index and gate index from the remaining set
         moment_index = np.random.choice(remaining_moment_indices)
         gate_index = np.random.choice(remaining_gate_indices[moment_index])
@@ -219,13 +219,17 @@ def fold_gates_at_random(circuit: Circuit, stretch: float, seed: Optional[int] =
         # Remove the gate we folded from the remaining set of gates to fold
         remaining_gate_indices[moment_index].remove(gate_index)
 
+        # If there are no gates left in the moment, remove the moment index from the remaining set
+        if not remaining_gate_indices[moment_index]:
+            remaining_moment_indices.remove(moment_index)
+
     return folded
 
 
 def fold_local(
         circuit: Circuit,
         stretch: float,
-        fold_method: Callable[[Callable, float, Tuple[Any]], Circuit] = fold_gates_from_left,
+        fold_method: Callable[[Circuit, float, Tuple[Any]], Circuit] = fold_gates_from_left,
         fold_method_args: Tuple[Any] = ()) -> Circuit:
     """Returns a folded circuit by folding gates according to the input fold method.
 
--- a/mitiq/folding_cirq.py
+++ b/mitiq/folding_cirq.py
@@ -164,7 +164,7 @@ def _update_moment_indices(moment_indices: dict, moment_index_where_gate_was_fol
     return moment_indices
 
 
-def fold_gates_at_random(circuit: Circuit, stretch: float, seed: Optional[int]) -> Circuit:
+def fold_gates_at_random(circuit: Circuit, stretch: float, seed: Optional[int] = None) -> Circuit:
     """Returns a folded circuit by applying the map G -> G G^dag G to a random subset of gates in the input circuit.
 
     The folded circuit has a number of gates approximately equal to stretch * n where n is the number of gates in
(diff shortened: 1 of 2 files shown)


rmlarose:
Fixes Issue #12

mitiq/folding_cirq.py
```diff
@@ -196,11 +196,28 @@ def fold_gates_at_random(circuit: Circuit, stretch: float, seed: Optional[int])
...
+    # Keep track of which gates we can fold in each moment
+    remaining_gate_indices = {moment: list(range(len(circuit[moment]))) for moment in range(len(circuit))}
+
     for _ in range(num_to_fold):
-        # TODO: Update gate choices properly using tracked moment indices.
-        moment_index = np.random.choice(len(folded))
-        gate_index = np.random.choice(len(folded[moment_index]))
-        _fold_gate_at_index_in_moment(folded, moment_index, gate_index)
+        # Any moment with at least one gate is fair game
+        remaining_moment_indices = [i for i in remaining_gate_indices.keys() if remaining_gate_indices[i]]
```
willzeng: Rather than recalculating this every time by looping through all the keys, would it be more efficient to check in the last step if the last gate was removed and then to pop off the key?

mitiq/folding_cirq.py
```diff
@@ -196,11 +196,28 @@ def fold_gates_at_random(circuit: Circuit, stretch: float, seed: Optional[int])
...
+        gate_index = np.random.choice(remaining_gate_indices[moment_index])
+
+        # Do the fold
+        _fold_gate_at_index_in_moment(folded, moment_indices[moment_index], gate_index)
+
+        # Update the moment indices for the folded circuit
+        _update_moment_indices(moment_indices, moment_index)
+
+        # Remove the gate we folded from the remaining set of gates to fold
+        remaining_gate_indices[moment_index].remove(gate_index)
```
willzeng:
```suggestion
        remaining_gate_indices[moment_index].remove(gate_index)
        if remaining_gate_indices[moment_index] is None:
            remaining_gate_indices.pop(moment_index)
```

mitiq/tests/test_folding_cirq.py
```diff
@@ -466,6 +467,102 @@ def test_fold_gates_at_random_seed_one_qubit():
...
+    circ = Circuit(
+        [ops.H.on_each(*qreg)],
+        [ops.CNOT.on(qreg[0], qreg[1])],
+        [ops.T.on(qreg[2])],
+        [ops.TOFFOLI.on(*qreg)]
+    )
+
+    for stretch in np.linspace(1., 3., 10):
+        folded = fold_gates_at_random(circ, stretch=stretch, seed=1)
+        print(folded, "\n\n")
```
willzeng: Should replace this with an automated test before merge. An automated but less comprehensive is better than a manual one as the manual ones will quickly outpace our ability to check.

Perhaps you could use total gate count for a given gate type to ensure none are folded too many times and that folds aren't reused?

rmlarose:
> Very nice trick. For me it can be merged, after the following minor correction:
> In the annotation of `fold_local`:
> `Callable[[Callable, float, Tuple[Any]], Circuit]` , I think it should be
> `Callable[[Circuit, float, Tuple[Any]], Circuit]`

Nice catch, thanks @andreamari!

mitiq/folding_cirq.py
rmlarose: Yes! Great suggestion.

mitiq/folding_cirq.py
rmlarose: Implemented this (except with `remaining_moment_indices.remove(moment_index)` for the last line, minor typo). Moved initial calculation of `remaining_moment_indices` outside of the loop for clarity.

mitiq/tests/test_folding_cirq.py
rmlarose: Completely agree. I used a circuit in which every gate only occurs once to check no gates are folded more than once, checking total gate count :)

rmlarose:
Really great feedback @andreamari and @willzeng. Two new commits address all comments.

mitiq/tests/test_folding_cirq.py
```diff
@@ -543,25 +543,24 @@ def test_fold_random_no_repeats():
...
-    # 0: ───H───@───@───
-    #           │   │
-    # 1: ───H───X───@───
-    #               │
-    # 2: ───H───T───X───
-    qreg = LineQubit.range(3)
+    # Test circuit:
+    # 0: ───H───@───Y───@───
+    #           │       │
+    # 1: ───────X───X───@───
```
andreamari:
```suggestion
    # 1: ───H───X───X───Z───
```

mitiq/tests/test_folding_cirq.py
```diff
@@ -543,25 +543,24 @@ def test_fold_random_no_repeats():
...
+        [ops.Y.on(qreg[0])],
+        [ops.CZ.on(*qreg)]
     )
 
     for stretch in np.linspace(1., 3., 10):
         folded = fold_gates_at_random(circ, stretch=stretch, seed=1)
-        print(folded, "\n\n")
-
-    # TODO: Right now this is just a visual test. Add checks.
+        assert len(set(folded.all_operations())) == len(set(circ.all_operations()))
```
andreamari: I think that this: `assert len(set(folded.all_operations())) == len(set(circ.all_operations()))`
only checks that the number of unique gates is preserved. But this can be true even if we fold previous gates and/or if we fold virtual gates. 
A possible solution could be to focus only on the case `stretch = 3` and test that, in this case,  `fold_gates_at_random` produces the same output of `fold_gates_from_left`.

rmlarose: Woops, that's not a very good check at all :)

New commit fixes this by checking that the count of each gate in the original circuit is less than 3.

rmlarose: > A possible solution could be to focus only on the case `stretch = 3` and test that, in this case, `fold_gates_at_random` produces the same output of `fold_gates_from_left`.

Note that this is checked in the test `test_fold_random_max_stretch`.

mitiq/tests/test_folding_cirq.py
rmlarose: Do you mean change the actual circuit? Or just the comment? Note that this is default behavior for printing circuits in Cirq.

I wouldn't put another H on qubit 1 because I want every gate to occur only once for this test.

andreamari: Sorry, I was wrong about H.
However I think that the last gate should be @--Z  and not @-@.

rmlarose: @--@ and @--Z are the same for the Cirq circuit drawer.

rmlarose:
Is this ready to merge @willzeng and @andreamari?

Issue #13, opened 2020-02-21
Add linear, polynomial, and Richardson reduce methods


willzeng:
These will be added in the framework from #6 and at least will include:

- [x] Richardson extrapolation
- [x] Linear fitting
- [x] Polynomial fitting

@andreamari Am assigning to you as you are already working on this.

PR #20, opened 2020-02-26
Two different files for factories and (adaptive) zne. Adds simple fit algorithms.

andreamari:
**Main changes**
- Deleted my initial old draft `zne.py` and associated old tests.
- Renamed `adaptive-zne.py` --->  `zne.py`
- New file `factories.py` containing the abstract classes `Factory` and `BatchedFactory` and also specific child classes corresponding to different fitting strategies.
 - In particular the 3 simple algorithms, requested in the issue #13 ( `LinearFactory`, `RichardsonFactory`, `PolyFactory`) are implemented in `factories.py`
- `reduce` method is left generic (not implemented) in `Factory` and `BatchedFactory`. For this reason, the default `Factory` for the `zne` function is not `BatchedFactory` anymore but `LinearFactory`. 

**Remark**
Even if the name of this branch is `exponential-fit`, the main changes of this PR are not related to exponentials (sorry for this confusion).


mitiq/factories.py
```diff
@@ -0,0 +1,129 @@
+from pyquil import Program
+from typing import List, Tuple, Callable
+
+import mitiq.qiskit.qiskit_utils as qs_utils
+import numpy as np
+
+class Factory(object):
```
rmlarose: Maybe `BaseFactor` or `AbstractFactor`? I think either or just `Factory` is fine, just a suggestion.

mitiq/factories.py
```diff
@@ -0,0 +1,129 @@
...
+class RichardsonFactory(BatchedFactory):
+    """Factory object implementing the Richardson's extrapolation algorithm."""
+
+    def reduce(self, x: List[float], y: List[float]) -> float:
+        """ Given two lists of x and y values associated to an unknwn function y=f(x), returns 
+        the extrapolation of the function to the x=0 limit, i.e., an estimate of f(0).
+        The algorithm is based on the Richardson's extrapolation method.
+        """
+
+        def get_gammas(c: float) -> List[float]:
```
rmlarose: I would define outside of `reduce` using `@staticmethod`.

mitiq/factories.py
```diff
@@ -0,0 +1,129 @@
...
+        The algorithm determines, with a standard least squared method, the parameters 
+        (q, m) such that the line g(x) = q + m*x optimally fits the input points.
+        Returns g(0) = q.
+        """
+        # linear least squared fit
+        _ , q = np.polyfit(x, y, deg=1)
+        return q
+
+
+class PolyFactory(BatchedFactory):
```
rmlarose: Are both `LinearFactory` and `PolyFactory` necessary? In principle, `LinearFactory.reduce` is just `PolyFactor.reduce(..., order=1)`. 

Maybe it's good to have them separate conceptually, but implement `LinearFactory` as described above? Ping @willzeng for thoughts.

mitiq/tests/test_factories.py
```diff
@@ -0,0 +1,60 @@
+# test_algorithms.py
+import numpy as np
+from typing import Callable
+from mitiq.factories import Factory, RichardsonFactory, LinearFactory, PolyFactory
+
+
+
```
rmlarose: Whitespace (too much I think).

mitiq/factories.py
willzeng: I asked Andrea just the same question! We think alike :) He had a good answer though, which was that it is likely that linear fits have a more optimal algorithm then an order 1 polyfit. The separate class is meant to hold that optimization.

mitiq/factories.py
andreamari: Ok, done! Not sure if, in the body of the class, static methods should be called with `self.reduce` or with `RichardsonFactory.reduce`. I used `self.reduce` but both seem to work.

mitiq/factories.py
andreamari: I have also many comments-ideas-suggestions about Factories. So I think it is better if we discuss all these open questions to a new issue that I will create soon.

mitiq/factories.py
andreamari: Actually also RichardsonFactory is a special case of PolyFactory with `order=length(instack)` !
I would keep different names because I think this is more handy and user friendly. 
However we could change the way in which `LinearFactory` and `RichardsonFactory` are coded.
They could be children of `PolyFactory`, or we could find a way of explicitly calling  `PolyFactory.reduce` in the body of `LinearFactory` and `RichardsonFactory`. I am not sure how to do it this in practice.

rmlarose: > Actually also RichardsonFactory is a special case of PolyFactory with `order=length(instack)` !

Ahah!

> I would keep different names because I think this is more handy and user friendly.

Complete agree.

> However we could change the way in which `LinearFactory` and `RichardsonFactory` are coded.

Agree again. Same interface for the user but we can change the implementation.

> They could be children of `PolyFactory`, or we could find a way of explicitly calling `PolyFactory.reduce` in the body of `LinearFactory` and `RichardsonFactory`. I am not sure how to do it this in practice.

I have ideas but maybe not ideal -- open to other suggestions. One idea is to implement a private `_reduce` method in `PolyFactory` which takes in an order (and other arguments). This is then called by the public `reduce` method in `PolyFactory`, and also called by `LinearFactory` with `order=1`.

> I asked Andrea just the same question! We think alike :) He had a good answer though, which was that it is likely that linear fits have a more optimal algorithm then an order 1 polyfit. The separate class is meant to hold that optimization.

For optimization, I think the current approach is good. I'm not sure if the extrapolation part is much of a bottleneck, though.

mitiq/zne.py
```diff
@@ -1,147 +1,50 @@
...
+from mitiq.factories import Factory, LinearFactory
 
-import numpy as np
 
-
-# used in Mitigator.extrapolate()
-def get_gammas(c):
-    """Returns the linear combination coefficients "gammas" for Richardson's extrapolation.
-    The input is a list of the noise stretch factors.
+def mitigate(pq: QPROGRAM, fac: Factory, scale_noise: Callable[[QPROGRAM, float], QPROGRAM], run_program: Callable) -> \
```
rmlarose: I would rename this function. `mitigate` should conceptually be `zne` below: i.e., a function take in 

1. a circuit,
2.  fold function,
3. a stretch factor, and 
4. an extrapolation method, 

then return the extrapolated zero noise (i.e., mitigated) value.

Additionally, it seems like this should be a method of a `Factory`, not a standalone function. It's odd to input a `Factory`.

mitiq/factories.py
```diff
@@ -0,0 +1,130 @@
...
+
+    def is_converged(self, instack: List[float], outstack: List[float]) -> None:
+        raise NotImplementedError
+
+    def reduce(self) -> float:
+        raise NotImplementedError
+
+
+class BatchedFactory(Factory):
+    def __init__(self, scalars: List[float], instack: List[float] = None, outstack: List[float] = None) -> None:
```
rmlarose: Change type of `scalars` to be `Iterable[float]`. It's natural to do something like `scalars = np.arange(start, stop, step)`, so we should account for numpy arrays or any other iterable object of floats.

mitiq/zne.py
rmlarose: Actually, I just realized `zne` returns a function. I admit I'm confused by this as well.

willzeng: The reason to input a factory is to decouple the backend that runs programs from the method of doing ZNE. You can define one Factory for zne and then run it on a bunch of different backends with mitigate. If they were coupled then you'd need to define a Factory with a backend specified at instantiation. Conceptually I think it is nice to assume that Factory objects are totally backend independent.

`zne` returns a function so that it can be a drop in replacement for wherever a user has `run_program` in their code, e.g.

```
qvm = QVMConnection()

output = qvm.run(my_pyquil_prog)

# becomes
zne_run = zne(qvm.run) # with default args to LinearFactory
output = zne_run(my_pyquil_program)
```

mitiq/factories.py
```diff
@@ -0,0 +1,130 @@
...
+        except IndexError:
+            raise IndexError(f"BatchedFactory cannot take another step. Number of batched scalars ({len(instack)}) "
+                             f"exceeded.")
+        self.instack.append(next_param)
+        return next_param
+
+    def is_converged(self, instack: List[float], outstack: List[float]) -> bool:
+        return len(instack) == len(self.scalars)
+
+# Specific extrapolation algorithms are given below.
```
willzeng: This todo list would be easier to track on the github issues rather than in the code itself. Could it be moved to the open issues(s)?

mitiq/factories.py
```diff
@@ -0,0 +1,130 @@
...
+# Adaptive max-likelihood
+# Adaptive Bayesian
+# etc... 
+
+
+class RichardsonFactory(BatchedFactory):
+    """Factory object implementing the Richardson's extrapolation algorithm."""
+
+    @staticmethod
+    def get_gammas(c: float) -> List[float]:
```
willzeng: Would rename the `c`-->`stretches` or something else that is more readable. Folks looking at this code may not be tied to a particular literature convention

willzeng: Also should the type be
```suggestion
    def get_gammas(c: Iterable[float]) -> Iterable[float]:
```

mitiq/factories.py
```diff
@@ -0,0 +1,130 @@
...
+        b = np.zeros(order + 1)
+        b[0] = 1
+        return np.linalg.solve(A, b)
+
+    def reduce(self, x: List[float], y: List[float]) -> float:
+        """ Given two lists of x and y values associated to an unknwn function y=f(x), returns 
+        the extrapolation of the function to the x=0 limit, i.e., an estimate of f(0).
+        The algorithm is based on the Richardson's extrapolation method.
+        """
+
```
willzeng: extra line

willzeng:
```suggestion
```

mitiq/factories.py
willzeng: Given that this is just called once in the `reduce` function. Should this be its own function or should this just happen as a subroutine in `reduce`? Is there a reason for it to be made its own modular function?

mitiq/factories.py
andreamari: Thanks Ryan, I followed more or less what you suggested. PolyFactory has now a static method `static_reduce` which is called by `PolyFactory`, `LinearFactory`, and `RichardsonFactory`.

mitiq/factories.py
andreamari: `get_gammas` is not necessary anymore. In the current version `PolyFactory.static_reduce` is used.

mitiq/factories.py
andreamari: Todo list removed

mitiq/factories.py
andreamari: `get_gammas` is not necessary anymore. In the current version `PolyFactory.static_reduce` is used.

andreamari:
Thanks for your feedback, @rmlarose and @willzeng ! I should have applied all your suggestions. In particular `LinearFactory` and `RichardsonFactory` are now explicitly special cases of `PolyFactory`.

mitiq/factories.py
willzeng: addressed in #24

willzeng:
@rmlarose Take a look through and with your approve should be good to merge

Issue #24, opened 2020-02-29
Simplify the structure of Factory objects


andreamari:
After implementing some extrapolation algorithms I noticed that two aspects of the current structure of Factory classes can be confusing:
A. `instack` and `outstack` are internal variables but are also arguments of all methods.
B. `instack` and `outstack` are currently updated in different moments and by different functions.

Based also on my last call with Will, I would propose the following changes:
1. Remove `instack` and `outstack` from the arguments of all methods of Factory. Ryan had the same suggestion for the `reduce` method, I agree and I would do it for all, even for `_init_`. The motivation is that, in my opinion, `instack` and `outstack` should simply represent a record of what has been actually experimentally measured, and so they should always be initialized as empty lists.
2. `Factory.step` should not change `instack` and `outstack`. It should only output a single `next_param` .
3. A new method (e.g. `Factory.feed(param, result)`) should be used by the `mitigate` function to append new data points (`param` and `result`) to `instack` and `outstack`. This should be the only way of updating `instack` and `outstack`.

Renaming suggestions:

4. Ryan proposed to rename `Factory` to `BaseFactory` or `AbstractFactory`.
5. Will proposed to rename `Factory.step` to `Factory.fetch`.
6. I am open to better names for `Factory.feed`. `Factory.push` :-) ?

If you have other suggestions, maybe it is useful to continue the previous enumeration (7... 8...).
In this way it is easier to give feedback on specific points.

As soon as we find a reasonable consensus I can do a PR.

andreamari:
4. I am slightly in favor of the current `Factory` mainly because it makes annotations more simple. However `BaseFactory` is also ok for me.
5. +1 to `Factory.fetch` (I also like `Factory.pull` and `Factory.next` ).

willzeng:
My vote is for `Factory` as the base class, `Factory.next`, and `Factory.push`

I think `Factory.pull` usually indicates something that would change the internal state rather than just returning the next thing. `Factory.push` does actually change the internal state.

@rmlarose do you have other suggestions?

Great thoughts @andreamari ! IMO go for it on implementation

rmlarose:
Just `Factory` (and not `BaseFactory`) for the base class is good with me. (`AbstractFactory` kinda sounds like a cool progressive rock band name, though :)

However, thinking about it, the word factory is quite general -- when one first sees or hears the name, they probably won't have a good idea of what the object is. I'm definitely not saying we should rename it now with all the other naming considerations, just a thought.

[...]

PR #26, opened 2020-03-05
Refactoring of Factory classes and ZNE functions.

andreamari:
This PR introduces a significant amount of changes, mainly to simplify the structure of the code and to decrease as much as possible the risk of creating confusion.

**Main changes:**
- Factories have been changed according to the issue #24 .
- Updates doc strings for `factories.py` and `zne.py `
- Updated PyQuil tests in (`test_zne.py`).
- New Qiskit tests, now similar to PyQuil tests.

**New (or updated) functions in `zne.py`:**
- `run_factory`: runs a factory at a classical level (no quantum).
- `qrun_factory`: runs a factory quantumly (similar to the old  `mitigate`)
- `execute_with_zne`: returns a mitigated expectation value.
- `mitigate_executor`: returns a mitigated executor, i.e., a function to evaluate mitigated expectation values. (similar to the old `zne`).
- `zne_decorator`: decorates an executor. (similar to the old `zne_factory`)

**Remark about new names**
- As you can see there are many new names. I tried to choose names which indicate the action of the function as much as possible. Feel free to change them as you prefer.
- Whatever you decide, I think `zne_factory` is a bad name for a decorator (it looks like a name for a Factory but is not a Factory).


mitiq/factories.py
```diff
@@ -1,109 +1,164 @@
+"""Contains all the main classes corresponding to different zero-noise extrapolation methods."""
+
 from typing import List, Iterable
 import numpy as np
 
 
-class Factory(object):
+class Factory():
```
rmlarose: Remove parens if not inheriting from another class.

rmlarose:
```suggestion
class Factory:
```

mitiq/tests/test_factories.py
```diff
@@ -23,43 +23,48 @@ def f_non_lin(x: float) -> float:
     return A + B * x + C * x ** 2
 
 
-def apply_algorithm(algo_class: Factory, f: Callable[[float], float], order: float = None) -> float:
+def apply_algorithm(algo_object: Factory, f: Callable[[float], float]) -> float:
```
rmlarose: The fact that this is used to simplify tests suggests that something similar should be in the main library to simplify usage.

mitiq/zne.py
```diff
@@ -1,50 +1,122 @@
...
+        noise_to_expval: Function mapping noise scale values to expectation vales.
     """
-    if not fac.is_converged(fac.instack, fac.outstack):
-        next_param = fac.step(fac.instack, fac.outstack)
+    if not fac.is_converged():
+        next_param = fac.next()
+        next_result = noise_to_expval(next_param)
+        fac.push(next_param, next_result)
+
+        return run_factory(fac, noise_to_expval)
```
rmlarose: Why not just use an iterative approach here instead of recursive? In my mind, iterative is simpler to see what's happening here, but also practically the call stack could get quite large and slow down performance.

For example, an iterative approach would be something like:

```python
while not fac.is_converged() and iterations < max_iterations:
    next_param = fac.next()
    next_results = noise_to_expval(next_param)
    fac.push(nex_param, next_result)
    iterations += 1
```

mitiq/factories.py
```diff
@@ -1,109 +1,164 @@
...
-        self.instack = instack
-        self.outstack = outstack
 
-    def step(self, instack: List[float], outstack: List[float]) -> None:
+    def __init__(self) -> None:
+        """
+        Initialization arguments (e.g. noise scale factors) depend on the particular extrapolation
+        algorithm and can be added to the "__init__" method of the associated derived class.
+        """
+        self.instack = []
```
rmlarose: If the user is never supposed to change instack or outstack, they should really be private attributes.

rmlarose: e.g., `self._instack = []` and `self._outstack` = []`.

mitiq/zne.py
```diff
@@ -1,50 +1,122 @@
...
 
-        scaled_pq = scale_noise(pq, next_param)
+    return None
 
-        next_result = run_program(scaled_pq)
-        fac.outstack.append(next_result)
+# quantum version of run_factory. Similar to the old "mitigate".
+def qrun_factory(fac: Factory, qp: QPROGRAM, executor: Callable[[QPROGRAM], float], \
```
rmlarose: This is I guess the equivalent of `test_factories.apply_algorithm` (see comment there). Can this be used in `test_factories.py`?

mitiq/factories.py
andreamari: The "standard" way updating data is Factory.push(). However I would not completely exclude the the possibility of a direct access to `instack` and `outstack`. For example for transferring the data acquired by a factory into a different factory (as you suggested). Or, more simply, to just visualize the optimization history.  What about if we keep them public for the moment? This is something we can easily change in the future.

mitiq/tests/test_factories.py
andreamari: Good point! And indeed, as you suggested, we now have a function for this: `zne.run_factory`.

mitiq/zne.py
andreamari: Done as suggested. I also added `max_iterations` as an optional argument and a warning message in the case in which `max_iterations` is reached.

mitiq/zne.py
andreamari: Good point! I have just done as suggested.

mitiq/factories.py
rmlarose: I think public is good for now. In the future, we should make getter and (maybe) setter methods.

andreamari:
Thanks for your feedback! I will squash and merge.

Issue #29, opened 2020-03-05
Folding with stretch close to one returns no output


rmlarose:
Multiple easy fixes, one is to take `num_folded >= num_to_fold` as a stopping condition in, e.g., `fold_gates_from_left`, as opposed to the stopping condition `num_folded == num_to_fold`. For very small stretches, `num_to_fold` can start out at zero.

rmlarose:
I agree with this. Before I was thinking we should just round up `num_to_fold` to 1 if it's in the interval [0.5, 1.0). But we could just as well do as you suggested and round down to the nearest integer.

andreamari:
Actually I like the idea of rounding `num_to_fold` as you suggested, it's a better approximation. However I think the rounding choice is independent from this issue: even with the rounding that you proposed `num_to_fold` can be 0 if `stretch` is sufficiently close to 1.

willzeng:
Fixed by #30

PR #30, opened 2020-03-09
Fixes bug with small stretches

rmlarose:
Fixes Issue #29.

Adds a check if `num_to_fold` is zero. 

Also changes behavior to round `num_to_fold` up or down instead of always rounding down to nearest integer.

### DIFF
--- a/mitiq/folding_cirq.py
+++ b/mitiq/folding_cirq.py
@@ -123,7 +123,9 @@ def fold_gates_from_left(circuit: Circuit, stretch: float) -> Circuit:
         raise ValueError("The stretch factor must be a real number between 1 and 3.")
 
     ngates = len(list(folded.all_operations()))
-    num_to_fold = int(ngates * (stretch - 1.0) / 2.0)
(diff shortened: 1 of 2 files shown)


andreamari:
`num_to_fold = int(ngates * fractional_stretch / 2)`
Appears 3 times also in `folding_pyquil.py` (lines 31, 54, 128).  
Could you update those lines with the same rounding of `folding_cirq.py` ?

rmlarose:
Woops, good catch. I made this a function so I don't make this mistake again. 

`round` returns an int when called with one argument. From the docs:

```
Docstring:
round(number[, ndigits]) -> number

Round a number to a given precision in decimal digits (default 0 digits).
This returns an int when called with one argument, otherwise the
same type as the number. ndigits may be negative.
Type:      builtin_function_or_method
```

andreamari:
Thanks @rmlarose! From my side you can merge

willzeng:
As this is reviewed and approved I'm going to merge it.

Issue #22, opened 2020-02-28
Check for measurement gates in Cirq folding functions


rmlarose:
Folding functions should ignore measurement gates and continue to try to fold other gates. No unit tests currently check circuits with measurements in them.

andreamari:
We should think how to do the global unitary_folding in the case in which we have measurements in the middle of the circuit (final measurements can be removed and re-appended after folding). Maybe, in this case of intermediate measurements, global folding is just impossible and we should only raise an error.

rmlarose:
Will and I briefly discussed the possibility of new theory for folding circuits with intermediate measurements, but for now I think raising an error is the right thing to do. (And to ignore terminal measurements while folding.)

willzeng:
I believe this was closed by #35

PR #35, opened 2020-03-09
Adds support for folding circuits with terminal measurements

rmlarose:
To address Issue #22.


mitiq/folding_cirq.py
```diff
@@ -1,10 +1,36 @@
...
+        op: Gate operation.
+    """
+    return isinstance(op.gate, ops.measurement_gate.MeasurementGate)
+
+
+def _pop_measurements(circuit: Circuit):
+    """Removes all measurements from a circuit."""
+    measurements_generator = circuit.findall_operations(_is_measurement)
+    # Make measurements a list of lists so moment indices can be updated after gates are folded
+    measurements = []
```
willzeng:
```suggestion
    measurements = [list(m) for m in measurements_generator]
```
I love python list comprehensions :)

mitiq/folding_cirq.py
```diff
@@ -1,10 +1,36 @@
...
+    for m in measurements_generator:
+        measurements.append(list(m))
+    circuit.batch_remove(measurements)
+    return measurements
+
+
+def _add_measurements(circuit: Circuit, measurements: Iterable[Tuple[int, ops.Operation]]) -> None:
```
willzeng: Does this need to be a standalone function? Could `circuit.batch_insert(measurements) be used in place?

mitiq/folding_cirq.py
```diff
@@ -95,11 +121,21 @@ def fold_moments(circuit: Circuit, moment_indices: List[int]) -> Circuit:
     return folded
 
 
-def _fold_all_gates_locally(circuit: Circuit) -> Circuit:
+def _fold_all_gates_locally(circuit: Circuit) -> None:
```
willzeng: good catch

mitiq/folding_cirq.py
```diff
@@ -114,16 +150,27 @@ def fold_gates_from_left(circuit: Circuit, stretch: float) -> Circuit:
     Note:
         Folding a single gate adds two gates to the circuit, hence the maximum stretch factor is 3.
     """
-    folded = deepcopy(circuit)
+    if not circuit.are_all_measurements_terminal():
```
willzeng: that's a handy built-in function :)

mitiq/folding_cirq.py
```diff
@@ -150,9 +199,22 @@ def fold_gates_from_right(circuit: Circuit, stretch: float) -> Circuit:
...
     """
+    if not circuit.are_all_measurements_terminal():
+        raise ValueError(f"Input circuit contains intermediate measurements and cannot be folded.")
+
+    add = False
+    if circuit.has_measurements():
+        measurements = _pop_measurements(circuit)
+        for i in range(len(measurements)):
+            measurements[i][0] = -1  # Make sure the moment to insert into is the last in the circuit
+        add = True
```
willzeng: The above code block appears several times. Perhaps it should be a separate function?  
```
measurements = get_measurements(circuit)
folded = Circuit(reversed(reversed_folded_circuit))
if measurements != None:
    _add_measurements(folded, measurements)
```

This could also avoid the need for the `add` boolean

andreamari: Maybe we could rename `add_mesurements` to `append_measurements` and put both the index update `...  measurements[i][0] = -1` and the if condition `if measurements != None:`, in the body of the function `append_measurements`.

andreamari:
Measurements are removed in the last function `unitary_folding`.
Apart from this, everything looks very good to me! Thanks Ryan.

mitiq/folding_cirq.py
rmlarose: My thinking was to abstract one layer away from Cirq in case they change anything. That way we can keep the usage of `_add_measurements` the same but change its implementation if necessary.

mitiq/folding_cirq.py
rmlarose: Yeah since we always want measurements to be terminal, it makes sense to add the block to `_add_measurements` and rename the function to `_append_measurements`.

rmlarose:
Put the repeating block of code into `_add_measurements` and renamed it to `_append_measurements`. This simplifies the checks of whether measurements need to be appended or not, as `_append_measurements` will return an empty list of there are no terminal measurements in the circuit. Also, `_add_measurements` will do nothing when an empty list is input.

rmlarose:
I will address `unitary_folding` function in a separate issue/PR. This hasn't been touched in a while and needs more updates than just checking for measurements.

rmlarose:
I will address `unitary_folding` function in a separate issue/PR. This hasn't been touched in a while and needs more updates than just checking for measurements.

> Measurements are not removed in the last function `unitary_folding`.
> Apart from this, everything looks very good to me! Thanks Ryan.

willzeng:
Got it. Note #45 is the new issue for updating unitary folding. LGTM

PR #38, opened 2020-03-11
Adds exponential decay and exponential poly-decay factories

### LINKED ISSUE #14: Add exponential fitting
Add a reduce method that fits an exponential curve

andreamari:
Fixes #14 .

Main changes:
-  New factory `DecayFactory`: does zne using the ansatz y(x) = a + b * exp(-c * x), with c > 0.
-  New factory `PolyDecayFactory`: y(x) = a + exp(z(x)), where z(x) is a polynomial of a given order. This is useful to fit data which deviates from an ideal exponential but still has a general decay structure. 
- Associated tests in `test_factories.py`

Note: if the asymptotic value (the "a" constant of the functions above) is given, a linear fit of z(x) = log(y - a) is evaluated to make the extrapolation. Otherwise a non-linear fit is directly applied to the decay ansatz using `scipy.optimize.curve_fit`, in other words in this case we try to fit also the "a" parameter.

Note there is a "TODO:" in `factories.py`: for the moment if the `asymptote` parameter is not given, we can only fit with `order` at most `3` and the code is not very elegant. At some point we have to clean this part. If you have a simple solution in mind, we can solve this issue in this PR. Otherwise we can keep it as a "TODO".

@willzeng @rmlarose 

@tudorgt (you can have a look if you want and if you have time. These methods are perhaps not particularly sophisticated but better to start with simple things).

### DIFF
--- a/mitiq/factories.py
+++ b/mitiq/factories.py
@@ -291,7 +291,6 @@ class PolyDecayFactory(BatchedFactory):
         # CASE 2: asymptote is given.
         # Plynomial fit of z(x).
         zstack = [np.log(max(y - asymptote, eps)) for y in outstack]
-        print("zstack", zstack)
         # Get coefficients {z_j} of z(x)= z_0 + z_1*x + z_2*x**2...
         # Note: coefficients are ordered from high powers of x to low powers x.
         z_coefficients = np.polyfit(instack, zstack, deg=order)
--- a/mitiq/factories.py
+++ b/mitiq/factories.py
@@ -1,8 +1,8 @@
 """Contains all the main classes corresponding to different zero-noise extrapolation methods."""
 
-from typing import List, Iterable
+from typing import List, Iterable, Union
 import numpy as np
-
+from scipy.optimize import curve_fit
 
 class Factory:
     """
@@ -118,7 +118,7 @@ class PolyFactory(BatchedFactory):
         can be called also by other factories which are particular cases of PolyFactory,
         e.g., LinearFactory and RichardsonFactory.
         """
-        # check arguments
+        # Check arguments
         error_str = "Data is not enough: at least two data points are necessary."
         if instack is None or outstack is None:
             raise ValueError(error_str)
@@ -129,7 +129,7 @@ class PolyFactory(BatchedFactory):
                 "Extrapolation order is too high. "
                 "The order cannot exceed the number of data points minus 1."
             )
-        # get coefficients {c_j} of p(x)= c_0 + c_1*x + c_2*x**2... which best fits the data
+        # Get coefficients {c_j} of p(x)= c_0 + c_1*x + c_2*x**2... which best fits the data
         coefficients = np.polyfit(instack, outstack, deg=order)
         # c_0, i.e., the value of p(x) at x=0, is returned
         return coefficients[-1]
@@ -175,14 +175,14 @@ class DecayFactory(BatchedFactory):
     Otherwise, a non-linear fit of y(x) is perfomed.
     """
 
-    def __init__(self, scalars: Iterable[float], asymptote: float = None) -> None:
+    def __init__(self, scalars: Iterable[float], asymptote: Union[float, None] = None) -> None:
         """
         Args:
             scalars: Iterable of noise scale factors at which expectation values should be measured.
             asymptote: Infinite-noise limit (optional argument).
         """
         super(DecayFactory, self).__init__(scalars)
-        if not isinstance(asymptote, (float, None)):
+        if not (asymptote is None or isinstance(asymptote, float)):
             raise ValueError("The argument 'asymptote' must be either a float or None")
         self.asymptote = asymptote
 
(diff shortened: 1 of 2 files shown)


mitiq/factories.py
```diff
@@ -162,3 +164,147 @@ def reduce(self) -> float:
...
+
+        ansatzes = (ansatz_zero, ansatz_one, ansatz_two, ansatz_three)
+
+        if asymptote is None:
+            opt_params, _ = curve_fit(ansatzes[order], instack, outstack)
+            # Return ansatx(0)
+            return opt_params[0] + np.exp(opt_params[1])
+
+        # CASE 2: asymptote is given.
+        # Plynomial fit of z(x).
```
willzeng:
```suggestion
        # Polynomial fit of z(x).
```

mitiq/factories.py
```diff
@@ -162,3 +164,147 @@ def reduce(self) -> float:
...
+                x: float, asympt: float, z_zero: float, z_one: float, z_two: float, z_three: float
+            ) -> float:
+            """Ansatz function of order 3."""
+            return asympt + np.exp(z_zero - z_one * x - z_two * x ** 2 - z_three * x ** 3)
+
+        ansatzes = (ansatz_zero, ansatz_one, ansatz_two, ansatz_three)
+
+        if asymptote is None:
+            opt_params, _ = curve_fit(ansatzes[order], instack, outstack)
+            # Return ansatx(0)
```
willzeng:
```suggestion
            # Return ansatz(0)
```

mitiq/tests/test_factories.py
```diff
@@ -1,49 +1,111 @@
...
+    """Test of exponential decay extrapolator."""
+    algo_object = DecayFactory(X_VALS, asymptote=None)
+    run_factory(algo_object, f_decay)
+    assert np.isclose(algo_object.reduce(), f_decay(0, err=0), atol=CLOSE_TOL)
+
+
+def test_poly_decay_factory_no_asympt():
+    """Test of (almost) exponential decay extrapolator."""
+    # test that, for a decay with a non-linear exponent,
+    # order=1 is bad while ored=2 is better.
```
willzeng:
```suggestion
    # order=1 is bad while order=2 is better.
```

mitiq/tests/test_factories.py
```diff
@@ -1,49 +1,111 @@
...
+    """Test of exponential decay extrapolator."""
+    algo_object = DecayFactory(X_VALS, asymptote=A)
+    run_factory(algo_object, f_decay)
+    assert np.isclose(algo_object.reduce(), f_decay(0, err=0), atol=CLOSE_TOL)
+
+
+def test_poly_decay_factory_with_asympt():
+    """Test of (almost) exponential decay extrapolator."""
+    # test that, for a decay with a non-linear exponent,
+    # order=1 is bad while ored=2 is better.
```
willzeng:
```suggestion
    # order=1 is bad while order=2 is better.
```

mitiq/factories.py
```diff
@@ -162,3 +164,147 @@ def reduce(self) -> float:
...
+            raise ValueError(error_str)
+        if order > len(instack) - 1:
+            raise ValueError(
+                "Extrapolation order is too high. "
+                f"The order cannot exceed the number of data points minus {1 + shift}."
+            )
+
+        # CASE 1: asymptote is None.
+
+        # TODO: there must be better way of doing this. *args does not seem to work with curve_fit.
```
willzeng: I haven't tested this, but could you do a `polyfit` of the log of the data? e.g. something like
```
np.polyfit(np.log(instack), outstack, deg=order)
```
@andreamari

willzeng:
Looks really good! Made some small comments above.

mitiq/factories.py
andreamari: This is equivalent to assume a zero asymptote and in fact one can get the fit that you proposed by setting `asymptote=0` instead of `asymptote=None`. 
If data really decays close to zero for large noise, then it should work. Otherwise I am not sure.

willzeng: Ah yes you're right. Maybe @tudorgt knows of a better way to do this.

andreamari:
**A small improvement:**

Previously only decaying exponentials where supported, i.e., those converging to their asymptote from above. However expectation values could also converge to the asymptotic value from below as well. In practice the difference is just the a sign parameter in the ansatz,  which is taken into account when we compute the logarithm. 

For this reason I also renamed `DecayFactory` to `ExpFactory` and `PolyDecayFactory` to `PolyExpFactory`.

Both decaying and growing exponential functions are now tested, and everything seem to work.
Sorry for this late change, I should have done this before making the PR but I did not think about it.

mitiq/factories.py
andreamari: I think what we do is fine from a fitting point of view, it is just done in a dirty way (I hard coded 4 ansatz functions, one for each order: 0,1,2,3).

willzeng: Given that this works, I'm going to approve. If we come up with a more elegant way then it can be added to a later update

Issue #31, opened 2020-03-09
Create circuit conversion from allowed frontends (Qiskit for now) to mitiq representation (Cirq for now)


rmlarose:
For the purpose of unitary folding (which inputs Cirq circuits). In the future, we should allow for many front-end circuit types, including

- [ ] Qiskit
- [ ] pyQuil

and others. For the pre-alpha, release, we will focus on Qiskit QuantumCircuit's as the primary allowed front-end and Cirq Circuits as the underlying representation used to fold gates.

willzeng:
Since Cirq is the underlying representation we should also accept cirq front ends and back ends yes? @rmlarose I figure this is what you had in mind but would be good to make this workflow explicit as well.

rmlarose:
Yeah definitely accept Cirq frontends, even I can convert a Cirq circuit to a Cirq circuit. Also support for Google backends is easy :)

Issue #32, opened 2020-03-09
Create circuit conversion from underlying representation (Cirq for now) to allowed backends (Qiskit for now)


rmlarose:
For the purpose of running folded circuits on real backends. In the future, we should allow as many backends as possible, including

- [ ] IBM
- [ ] Rigetti
- [ ] Google
- [ ] IonQ, Honeywell, etc.

For the pre-alpha release, we will focus on converting from Cirq circuits to Qiskit circuits to submit on IBMQ processors.

PR #42, opened 2020-03-13
Adds circuit conversions.

rmlarose:
To resolve Issue #31 and Issue #32.

Early comments welcome on skeleton + any other thoughts.


rmlarose:
Ready for review by @willzeng and/or @andreamari and/or @nathanshammah. All conversions were already implemented in Cirq so the functions were easy enough to write, had to change around some other things to get everything working smoothly. Notably creating `mitiq.utils` and adding an option to ignore qubit names/types when checking for equality.

willzeng:
Thanks Ryan! @rmlarose is this branch dependent on #53 ? It looks like in addition to the conversions code here there are also changes about folding? If possible, we could review them in parallel if they are separate. If not let us know and we can review them serially.

mitiq/qiskit/conversions.py
```diff
@@ -0,0 +1,66 @@
+"""Functions to convert from Mitiq's internal circuit representation
```
willzeng: Cool! This whole module does come out rather smoothly. Seems to validate the choice of Cirq as at least a relatively easy place to start.

mitiq/utils.py
```diff
@@ -0,0 +1,74 @@
+"""Utility functions."""
+
+import random
+
+import cirq
+
+
+def random_circuit(depth: int, **kwargs) -> cirq.Circuit:
```
willzeng: Do we need to use **kwargs here? Using seed=None would make the method's type signature more explicit to users.

mitiq/utils.py
```diff
@@ -0,0 +1,74 @@
...
+    qubit = cirq.GridQubit(0, 0)
+    gates = [cirq.ops.X, cirq.ops.Y, cirq.ops.Z]
+    circuit = cirq.Circuit(
+        [random.choice(gates).on(qubit) for _ in range(depth)]
+    )
+
+    return circuit
+
+
+def _equal(
```
willzeng: This is a neat function, well done. I thought this would be much harder to implement but Cirq has lots of useful functions that you used well.

mitiq/utils.py
```diff
@@ -0,0 +1,74 @@
...
+            else the two are not equal.
+
+            If True, the qubits of both circuits must have
+            a well-defined ordering.
+    """
+    if circuit_one is circuit_two:
+        return True
+
+    if not require_qubit_equality:
+        print("Transforming qubits of circuit one...")
```
willzeng: Should these print statements become comments?

rmlarose:
> Thanks Ryan! @rmlarose is this branch dependent on #53 ? It looks like in addition to the conversions code here there are also changes about folding? If possible, we could review them in parallel if they are separate. If not let us know and we can review them serially.

Thanks @willzeng. Pull #53 should be merged before this one. The changes to folding here only affect the utility functions `_equal` and `random_circuit`. Both made sense to use in testing conversions, so I took them from `folding_cirq.py` and made them `utils.py`.

I will address Andrea's comments on #53 now, then make the changes here.

willzeng:
Roger that. Good plan.

mitiq/qiskit/conversions.py
```diff
@@ -0,0 +1,66 @@
...
+to Qiskit representations.
+"""
+
+import cirq
+from cirq.contrib.qasm_import import circuit_from_qasm
+from qiskit import QuantumCircuit
+
+
+QASM = str
```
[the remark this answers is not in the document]
rmlarose: Changed.

mitiq/utils.py
rmlarose: Made `seed` an optional arg.

mitiq/utils.py
rmlarose: The qubit equality aspect of comparing circuits is kind of annoying, but thankfully there was a relatively straightforward workaround with transforming qubits. The only requirement is that the qubits must be sortable. I think this is always the case, but if it's not there's really no way at all to "line up" qubits from one circuit to the other.

mitiq/utils.py
rmlarose: Love it when debug prints make it into the commits haha... Fixed.

rmlarose:
Changes made. Should be ready to merge after #53 if no more changes.

mitiq/qiskit/conversions.py
```diff
@@ -0,0 +1,66 @@
...
+
+    Returns
+    -------
+        Qiskit.QuantumCircuit object equivalent to the input Mitiq circuit.
+    """
+    return QuantumCircuit.from_qasm_str(_to_qasm(circuit))
+
+
+def _from_qiskit(circuit: QuantumCircuit) -> cirq.Circuit:
+    """Returns a Qiskit circuit equivalent to the input Mitiq circuit.
```
andreamari:
```suggestion
    """Returns a Mitiq circuit equivalent to the input Qiskit circuit.
```

mitiq/qiskit/conversions.py
```diff
@@ -0,0 +1,66 @@
...
+
+    Returns
+    -------
+        Mitiq circuit representation equivalent to the input Qiskit circuit.
+    """
+    return _from_qasm(circuit.qasm())
+
+
+def _from_qasm(qasm: QASMType) -> cirq.Circuit:
+    """Returns a Qiskit circuit equivalent to the input Mitiq circuit.
```
andreamari:
```suggestion
    """Returns a Mitiq circuit equivalent to the input QASM string.
```

willzeng:
@rmlarose It looks like this still has some folding stuff in the diff. Perhaps there's some rebasing on top of master that still needs to happen?

rmlarose:
@willzeng should be good now, let me know.

rmlarose:
> Thanks @rmlarose very nice! Before merging, see the minor comments about to the docstrings.

Nice catch! Thanks @andreamari.

PR #44, opened 2020-03-14
add general mitiq testing to the CI

willzeng:
I realized today that the continuous integration tests were only running the `mitiq/qiskit/tests` and not the general ones. 

This PR fixes this to run the general mitiq tests as well as the qiskit ones.

**NOTE** This change uncovers some bugs that previously were not caught by the CI!

@rmlarose @andreamari If you have suggestions on what these bugs might be or how to fix them then please do push those bug fixes directly to this branch. I propose we freeze any other merges into master until we get this resolved.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -30,4 +30,5 @@ jobs:
     - name: Test with pytest
       run: |
         pip install pytest
+        pytest mitiq/tests
         pytest mitiq/qiskit/tests


rmlarose:
There's an error in folding_cirq.py which is fixed in #35.

willzeng:
Roger that @rmlarose Have rebased on top of master after merging #35 and it now passes the updated CI.

PR #48, opened 2020-03-17
Adds Adaptive Exponential Factory (AdaExpFactory) and tests.

andreamari:
**Main changes**

- Our first adaptive extrapolation algorithm: `AdaExpFactory`. It is similar to `ExpFactory` but the choice of the noise parameters is done according to the history of the optimization steps. In particular the next noise value depends on the most recent fit of the exponent parameter (adaptive rule proposed by @tudorgt ).
- The full optimization history is logged into `self.history`, in this way the user can see which steps have been performed by the optimizer.
- New tests associated to `AdaExpFactory`.

### DIFF
--- a/mitiq/factories.py
+++ b/mitiq/factories.py
@@ -370,7 +370,7 @@ class AdaExpFactory(Factory):
         self.steps = steps
         self.scalar = scalar
         self.asymptote = asymptote
-        # Keep a log of the optimization process storing: 
+        # Keep a log of the optimization process storing:
         # noise value(s), expectation value(s), parameters, and zero limit
         self.history = []  # type: List[Tuple[List[float], List[float], List[float], float]]
 
--- a/mitiq/factories.py
+++ b/mitiq/factories.py
@@ -1,6 +1,6 @@
 """Contains all the main classes corresponding to different zero-noise extrapolation methods."""
 
-from typing import List, Iterable, Union
+from typing import List, Iterable, Union, Tuple
 import numpy as np
 from scipy.optimize import curve_fit
 
@@ -93,7 +93,7 @@ class BatchedFactory(Factory):
 
 class PolyFactory(BatchedFactory):
     """
-    Factory object implementing a zero-noise extrapolation algotrithm based on a polynomial fit.
+    Factory object implementing a zero-noise extrapolation algorithm based on a polynomial fit.
     Note: RichardsonFactory and LinearFactory are special cases of PolyFactory.
     """
 
@@ -154,7 +154,7 @@ class RichardsonFactory(BatchedFactory):
 
 
 class LinearFactory(BatchedFactory):
-    """Factory object implementing a zero-noise extrapolation algotrithm based on a linear fit."""
+    """Factory object implementing a zero-noise extrapolation algorithm based on a linear fit."""
 
     def reduce(self) -> float:
         """
@@ -167,12 +167,12 @@ class LinearFactory(BatchedFactory):
 
 class ExpFactory(BatchedFactory):
     """
-    Factory object implementing a zero-noise extrapolation algotrithm assuming an
+    Factory object implementing a zero-noise extrapolation algorithm assuming an
     exponential ansatz y(x) = a + b * exp(-c * x), with c > 0.
 
     If the asymptotic value (y(x->inf) = a) is known, a linear fit with respect
     to z(x) := log[sing(b) (y(x) - a)] is used.
-    Otherwise, a non-linear fit of y(x) is perfomed.
+    Otherwise, a non-linear fit of y(x) is performed.
     """
 
     def __init__(self, scalars: Iterable[float], asymptote: Union[float, None] = None) -> None:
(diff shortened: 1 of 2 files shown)


willzeng:
I've been a little delayed in reviewing this and so am adding @rmlarose for review as well

mitiq/factories.py
```diff
@@ -271,30 +276,33 @@ def static_reduce(instack: List[float], outstack: List[float], asymptote: Union[
...
         def ansatz_two(x: float, asympt: float, b: float, z_one: float, z_two: float) -> float:
             """Ansatz function of order 2."""
-            return asympt + b * np.exp(-z_one * x - z_two * x ** 2)
+            return asympt + b * np.exp(z_one * x + z_two * x ** 2)
         def ansatz_three(
                 x: float, asympt: float, b: float, z_one: float, z_two: float, z_three: float
             ) -> float:
             """Ansatz function of order 3."""
-            return asympt + b * np.exp(-z_one * x - z_two * x ** 2 - z_three * x ** 3)
+            return asympt + b * np.exp(z_one * x + z_two * x ** 2 + z_three * x ** 3)
```
rmlarose: To reduce the number of functions, what about the signature

```python
def ansatz(x: float, a: float, b: float, coeffs: List[float]) -> float:
```

Then, the order is determined and defined by the number of coefficients, i.e., by `len(coeffs)`. The body of the function could be

```python
return a + b * np.exp(np.polyval(coeffs, x))
```

I think [`np.polyval`](https://docs.scipy.org/doc/numpy/reference/generated/numpy.polyval.html) will correctly handle the case when `len(coeffs) = 0` but please double check.

mitiq/factories.py
```diff
@@ -319,4 +329,89 @@ def reduce(self) -> float:
...
+
+    The noise scale factors are are chosen adaptively at each step, depending on the history
+    of collected results.
+
+    If the asymptotic value (y(x->inf) = a) is known, a linear fit with respect
+    to z(x) := log[sing(b) (y(x) - a)] is used.
+    Otherwise, a non-linear fit of y(x) is performed.
+    """
+
+    SHIFT_FACTOR = 1.27846
```
rmlarose:
```suggestion
    _SHIFT_FACTOR = 1.27846
```

rmlarose: + later changes. If this is constant it shouldn't be visible outside the class.

mitiq/factories.py
```diff
@@ -319,4 +329,89 @@ def reduce(self) -> float:
...
+                "The argument 'steps' must be an integer greater or equal to 3. "
+                "If 'asymptote' is None, 'steps' must be greater or equal to 4."
+            )
+        self.steps = steps
+        self.scalar = scalar
+        self.asymptote = asymptote
+        # The variable self.history is a log of the optimization process.
+        # It is a list of tuples (instack, outstack, params, zero_limit).
+        # One tuple for each fitting step.
+        self.history = []
```
rmlarose:
```suggestion
        # Keep a log of the optimization process storing noise value(s), expectation value(s), parameters, and zero limit
        self.history = []  # type: List[Tuple[float...]]
```

rmlarose: (An attempt at shortening the comment/description.)

mitiq/tests/test_factories.py
```diff
@@ -123,3 +124,31 @@ def test_poly_exp_factory_no_asympt():
         algo_object = PolyExpFactory(X_VALS, order=2, asymptote=None)
         run_factory(algo_object, f)
         assert np.isclose(algo_object.reduce(), f(0, err=0), atol=CLOSE_TOL)
+
+def test_ada_exp_factory_with_asympt():
+    """Test of the adaptive exponential extrapolator."""
+    for f in [f_exp_down, f_exp_up]:
+        algo_object = AdaExpFactory(steps=3, scalar=2.0, asymptote=A)
+        run_factory(algo_object, f)
+        assert np.isclose(algo_object.reduce(), f(0, err=0), atol=CLOSE_TOL)
```
rmlarose: Instead of looping over `f_exp_down` and `f_exp_up` in the body of each test, we could use [`pytest.mark.parameterize`](https://docs.pytest.org/en/latest/parametrize.html).

It won't change much here but I'm in favor of using this because:

1. It will look a little cleaner.
2. Reduce code duplication.
3. Add more unit tests :)
4. Make it easier to add more functions to the test cases in the future.

Just a suggestion, if it's a hassle for now just ignore.

mitiq/factories.py
andreamari: Thanks Ryan, I have done more or less as you suggested and now it works! 
Since `curve_fit` wants only scalar variables, I used `*coeffs` as argument. The initialization values `p0` of `curve_fit` determines the `len(coeffs) ` and so `curve_fit` knows how many parameters should be optimized.

def ansatz(x: float, *coeffs: float):
            """Ansatz function of generic order."""
            # Coefficients of the polynomial to be exponentiated
            z_coeffs = coeffs[2:][::-1]
            return coeffs[0] + coeffs[1] * np.exp(x * np.polyval(z_coeffs, x))

mitiq/factories.py
andreamari: Ok, changed.

mitiq/tests/test_factories.py
andreamari: Good suggestion thanks! Now all the loops are replaced by `pytest.mark.parameterize`

andreamari:
Thanks @rmlarose for your comments, the code is much more clean now !

willzeng:
Awesome! merged.

Issue #49, opened 2020-03-19
Add continuous integration from github action on pull request from fork


nathanshammah:
In PR #47 tests are not run. Tests are run using GitHub actions. As suggested by @andreamari, it looks like this is due to the fact that #47 is a made from a fork of `unitaryfund/mitiq`.

As pointed out by @willzeng, this should be addressed by editing the continuous integration (CI) workflow. 

I can set this up in my fork, but also for the future (external collaborators or forks) I think it's most straightforward to set this up in the generic way. It may be that 
```
on: [push]
```
in [.github/workflows/ci.yml](https://github.com/unitaryfund/mitiq/blob/master/.github/workflows/ci.yml) is not enough to trigger the test. 
It may be that
```
on: [push, pull_request]
```
would work. 

If I understand correctly, there can be even more general events that trigger the CI, as a [`pull_request_review`](https://help.github.com/en/actions/reference/events-that-trigger-workflows#pull-request-review-event-pull_request_review), which looks useful.

I can change the `ci.yml` file. I also need to become familiar with [webhooks](https://developer.github.com/v3/repos/hooks/#list-hooks).

willzeng:
I think this was closed by #50 ? If that is not the case @nathanshammah please re-open

PR #50, opened 2020-03-20
Run continuous integration on pull request from fork

nathanshammah:
Change the .github workflow for continuous integration to be triggered by a pull request, as per #49.


willzeng:
Nice. That was an easy fix.

Issue #45, opened 2020-03-15
Update `unitary_folding` function in `folding_cirq.py`. 


rmlarose:
Updates include:

- Using `_get_num_to_fold`
- Checking for (terminal) measurements.
- etc.

PR #53, opened 2020-03-20
Updates unitary folding.

rmlarose:
More tests needed then ready for review.


rmlarose:
Addresses Issue #45.

mitiq/folding_cirq.py
```diff
@@ -339,26 +339,51 @@ def fold_method(circuit: Circuit, stretch: float, **kwargs):
...
+    if not circuit.are_all_measurements_terminal():
+        raise ValueError(f"Input circuit contains intermediate measurements and cannot be folded.")
+
+    folded = deepcopy(circuit)
+    measurements = _pop_measurements(folded)
+    base_circuit = deepcopy(folded)
+
+    # Determine the number of global folds and the final fractional stretch
+    num_global_folds = int(stretch // 3)
+    fractional_stretch = stretch % 3
```
andreamari: I think there is a problem here.  The integer stretch factors are = 1 + 2 n .  So they can be any odd integer number and not just multiples of 3.   So I think we need the previous 
`num_foldings, fractional_stretch = divmod(stretch - 1, 2)` or something equivalent to this.

mitiq/folding_cirq.py
```diff
@@ -339,26 +339,51 @@ def fold_method(circuit: Circuit, stretch: float, **kwargs):
...
+    # Do the global folds
+    for _ in range(num_global_folds):
+        folded += Circuit(inverse(base_circuit), base_circuit)
+
+    # Do the local folds
+    # Adjust the stretch to account for the fact that there are now more gates in the circuit
+    folded_ngates = len(list(folded.all_operations()))
+    adjusted_stretch = 1. + ngates / folded_ngates * (fractional_stretch - 1)
+    if adjusted_stretch > 1:
+        folded = fold_local(folded, adjusted_stretch, finish_fold_method, finish_fold_method_args)
```
andreamari: In my opinion since we are folding the full circuit globally, it is better to fold globally also the final fraction of the circuit.
I.e. I think we should append something equivalent to the previous `inverse(circuit[-fractional_depth:]) + circuit[-fractional_depth:]`.  Which is different form folding individual gates locally. Note that folding gates locally would produce a discontinuous behavior when passing from `stretch`=2.999.. to `stretch`=3.

mitiq/folding_cirq.py
```diff
@@ -128,10 +128,10 @@ def _get_num_to_fold(stretch: float, ngates: int) -> int:
     """Returns the number of gates to fold to acheive the desired (approximate) stretch factor.
 
     Args:
-        stretch: Floating point value to stretch the circuit by. Between 1 and 3.
+        stretch: Floating point value to stretch the circuit by.
         ngates: Number of gates in the circuit to stretch.
     """
-    return int(round(ngates * (stretch - 1.0) / 2.0))
+    return int(max(round(ngates * (stretch - 1.0) / 2.0), 0))
```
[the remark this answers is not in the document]
rmlarose: This can be changed back. It was an artifact of me miscalculating the `adjusted_stretch` that appears later in unitary folding. Good catch.

mitiq/folding_cirq.py
rmlarose: Fixed.

mitiq/folding_cirq.py
rmlarose: Even with this strategy, wouldn't you still have the degree of freedom of which "part" of the unitary to fold globally at the end?

For example, suppose the unitary is `U = A B C` where each `A`, `B`, and `C` consists of equal numbers of gates. If we need to fold 1/3 of the gates at the end, is it clear which of the following we should do?

(1) `A B C --> A A^dag A B C` (fold globally from left)
(2) `A B C --> A B B^dag B C` (fold globally from middle)
(3) `A B C --> A B C C^dag C` (fold globally from right)

rmlarose: Just for the sake of getting #42 merged, it might be good to keep the local folding methods for now and open an issue to discuss/address. I guess both local/global could be final folding methods, though maybe I'm over-complicating it.

Either way, this is an easy change which just entails something like

```python
ops = list(circuit.all_operations())
n = int(round(fractional_stretch * len(ops))
folded += cirq.Circuit([inverse(ops[-n:]), ops[-n:]])
# Or folded += cirq.Circuit([inverse(ops[:n]), ops[:n]]), etc.
```

I'll go with what you advise here @andreamari.

andreamari: Hi @rmlarose , sorry for my delayed reply. For global folding the natural choice is option (3), because ti is the only one which is consistent with integer folding. Let's make an example with more letters:  circuit = A B C D E.  A partial folding is something like this:
`(ABCDE) (E^dagD^dagC^dagB^dag)(BCDE)`,
now assume that we want to stretch a bit more, we would get:
`(ABCDE) (E^dagD^dagC^dagB^dagA^dag)(ABCDE) = circuit circuit^dag circuit,`
i.e. we converge to the correct integer circuit folding. This continuous stretching doesn't happen if we use different partial folding methods.

In conclusion I would vote for keeping it simple and not mixing local_folding with global_folding. However I also see that there is some urgency of merging. So, no problem, we can always fix this aspect in a new issue/PR.

rmlarose:
Removed the local folding @andreamari, ready to merge if all looks good.

mitiq/folding_cirq.py
rmlarose: Thanks @andreamari, makes sense! Quick change so I just added it now.

PR #58, opened 2020-03-24
Pins dependency versions

### LINKED ISSUE #55: Pin to specific Qiskit, pyQuil, and Cirq versions in requirements
Just ran into an issue where Qiskit circuit data is stored differently between v0.0.15 and v0.0.16, causing local tests to pass but the build to fail.

To avoid similar issues in the future, we should pin to specific versions now. @willzeng thoughts on what the versions should be? Latest?

Note that PR #42 pins Cirq to version 0.0.7.

willzeng:
Fixes #55 

I pinned these to the latest versions.

### DIFF
--- a/development_requirements.txt
+++ b/development_requirements.txt
@@ -1,3 +1,3 @@
-pyquil
-qiskit
+pyquil==2.18.0
+qiskit==0.16.2
 pytest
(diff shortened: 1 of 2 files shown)


nathanshammah:
Can I merge it...? We require only one approval in general?

willzeng:
For small stuff like this just one approval is fine.

For more significant PRs that have to do with real functionality there
needs to be more.

PR #288, opened 2020-08-09
[Parameter] Add implementation + tests of `add_parameter_noise`

### LINKED ISSUE #276: Add parameter noise scaling function
Supersedes #101. 

The signature of this function should be something like

```python
def param_scale_noise(circuit: QPROGRAM, scale_factor: float) -> QPROGRAM:
```

potentially with additional arguments (sigma, etc.). By convention, the input circuit should not be modified.

Assigning to @yhindy as discussed in last week's engineering meeting.

yhindy:
This pull request adds the `angle.py` module which has functionality for performing density matrix simulation of parameter noise scaling as well as hardware-type shot-based techniques. The main function that the quantum programmer can use is `add_parameter_noise`. 

This should close #276.


mitiq/angle.py
```diff
@@ -0,0 +1,256 @@
...
+    if isinstance(gate, YPowGate):
+        return YPowGate
+    if isinstance(gate, CXPowGate):
+        return CXPowGate
+    if isinstance(gate, CZPowGate):
+        return CZPowGate
+    else:
+        raise Exception(
+            "Must have circuit be made of rotation gates. \
+            Your gate {} may not be supported".format(gate))
```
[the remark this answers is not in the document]
yhindy: I think for now we only want to deal with parameterized gates. I'm not sure how the ZNE math works out if we skip gates. 

Yes not a rotation means that its not in the base gate set.

yhindy:
RFAL, implemented most changes recommended

mitiq/parameter.py
```diff
@@ -0,0 +1,257 @@
...
+
+BASE_GATES = [ZPowGate, HPowGate, XPowGate, YPowGate, CXPowGate \
+            CZPowGate]
+
+@converter
+def scale_parameters(
+    circ: Circuit,
+    scale_factor: float,
+    sigma: float,
+    seed: int = None
```
rmlarose, line 25:
```suggestion
    seed: Optional[int] = None
```

mitiq/parameter.py
```diff
@@ -0,0 +1,257 @@
+from typing import Iterable
```
rmlarose, line 1:
```suggestion
from typing import Iterable, Optional
```

mitiq/parameter.py
```diff
@@ -0,0 +1,257 @@
...
+        curr_moment = []
+        for op in moment.operations:
+            gate = copy.deepcopy(op.gate)
+            qubits = op.qubits
+            if isinstance(gate, MeasurementGate):
+                curr_moment.append(gate(*qubits))
+            else:
+                base_gate = _get_base_gate(gate)
+                param = gate.exponent * np.pi
+                rng = np.random.RandomState(seed)
```
rmlarose, line 54: Should be defined before the loop otherwise `rng.normal` will return the same value every time.

mitiq/parameter.py
```diff
@@ -0,0 +1,257 @@
...
+                noisy_op = parameter_noise_1q(noise, gate)(qubits)
+            elif len(qubits) == 2:
+                noisy_op = parameter_noise_2q(noise, gate)(qubits[0], qubits[1])
+            else:
+                raise Exception(
+                    "Gates for more than two qubits are not \
+                    supported with parameter noise scaling.")
+            curr_moment.append(noisy_op)
+        final_moments.append(Moment(curr_moment))
+    return Circuit(final_moments)
```
willzeng, line 101: Do I understand right that `scale_parameters` is the scaling method that users are going to use, while `scale_parameters_channel` is used to model the noise of a parameter noise backend for testing?

`scale_parameters` is multi-backend compatible and so seems better than the `channel` method.

@yhindy  can you say a bit more about the use case for `scale_parameters_channel`?

nathanshammah:
Continuous integration is failing at [testing](https://github.com/unitaryfund/mitiq/pull/288/checks?check_run_id=995550543)
```
Makefile:59: recipe for target 'test-all' failed
```
and [linting](https://github.com/unitaryfund/mitiq/pull/288/checks?check_run_id=995550622)
```
Makefile:16: recipe for target 'check-style' failed
```

mitiq/parameter.py
yhindy: yeah scale parameters channel is for simulation on density matrix solvers. the `scale_parameters` is multi-backend compatible. i think we discussed this and the `channel` one should be removed from `mitiq` proper.

mitiq/zne/parameter.py
```diff
@@ -0,0 +1,66 @@
...
+    return Circuit(final_moments)
+
+
+def _get_base_gate(gate):
+    for base_gate in BASE_GATES:
+        if isinstance(gate, base_gate):
+            return base_gate
+    raise Exception(
+        "Must have circuit be made of rotation gates. \
+        Your gate {} may not be supported".format(gate))
```
willzeng, line 66:
```suggestion
        "Circuit must consist only of rotation gates. \
        Your gate {} may not be supported".format(gate))
```

Also code coverage notified me that there wasn't a test for this exception. This could be added with something like https://docs.pytest.org/en/reorganize-docs/new-docs/user/pytest_raises.html:
```
class GateTypeException(Exception):
    pass

def test_gate_exeception():
    with pytest.raises(GateTypeException):
        # make a circuit with a non rotation gate and run _get_base_gate on it
```
Would recommend the GateTypeException instead of a generic Exception.

PR #318, opened 2020-08-18
Store optimal parameters after calling reduce in factories

### LINKED ISSUE #130: Add option to return optimal parameters for extrapolation/reduce methods
It may be nice to have an option to output more than just the extrapolated zero-noise value in, e.g., the return value(s) of `Factory.reduce` methods.

For example, if I wanted to fit the line `f(x) = a x + b`, it would be nice to output the optimal `a` and `b` parameters in addition to `f(0)`. This would be useful for visualization and potentially other things.

rmlarose:
Fixes #130 + other minor changes to factories, notably:

* `PolyFactory.static_reduce` removed in favor of `mitiq_polyfit`.
* `instack` and `outstack` made private attributes.
* Fixes some tests and improves docstrings.


willzeng:
Given that this touches static_reduce, I think it would be better for @andreamari to review this. Will re-assign

mitiq/zne/inference.py
```diff
@@ -536,13 +517,17 @@ class LinearFactory(BatchedFactory):
...
-        """
-        Determines, with a least squared method, the line of best fit
-        associated to the data points. The intercept is returned.
+        """Returns the zero-noise limit found by fitting a line to the input
+        data of scale factors and expectation values.
+
+        Stores the optimal parameters for the fit in `self.opt_params`.
         """
         # Richardson's extrapolation is a particular case of a polynomial fit
         # with order equal to 1.
```
andreamari, line 526:
```suggestion
        # Linear extrapolation is a particular case of a polynomial fit
        # with order equal to 1.
```

mitiq/zne/inference.py
```diff
@@ -422,67 +426,38 @@ def __init__(
...
+        expectation_values = self.get_expectation_values()
 
-        This static method is equivalent to the "self.reduce" method of
-        PolyFactory, but can be called also by other factories which are
-        particular cases of PolyFactory, e.g., LinearFactory
-        and RichardsonFactory.
+        if len(scale_factors) < 2:
+            raise ValueError(
+                "At least two data points are needed to fit a polynomial."
+            )
```
andreamari, line 448: This check seems unnecessary since we already check it at initialization. Given that it is missing in `LinearFactory` and in `RichardsonFactory` I would remove it also from here (`PolyFactory`) for consistency. 
```suggestion
```

andreamari (approved):
Thanks @rmlarose!
Only 2 minor suggestions (see above). 
For simplicity, I also directly committed a couple of `instack` --> `_instack` and `outstack` --> `_outstack`.
Apart from these minor things, LGTM.

rmlarose:
Thanks @andreamari for the doc fixes and suggestions!

PR #324, opened 2020-08-20
Add basic compilation facilities, don't relabel qubits

karalekas:
While trying to gather QCS data using mitiq, I remembered (the hard way) that there is no way to "just nativize" when using the Quil compiler, and thus we need a facility to perform nativization without optimization. Fortunately, `forest-benchmarking` had a partially-implemented utility for this (but rather than depend on it, which itself has a heavy set of dependencies, I chose to copy it over). Unfortunately, we lose a lot of the really powerful features that quilc offers by doing this, including automatic qubit rewiring. This is a draft PR because I'm going to test it on QCS to see if it works as intended. If this ends up being a nice tool, might try to get it merged into pyquil.

Additionally updates `to_quil` to keep the qubit labeling the same (which does not happen by default in Cirq) using a little `QubitOrder` hack. Maybe this could end up in Cirq?

Together, these two enhancements should allow us to perform error mitigation on QCS.

### DIFF
--- a/mitiq/mitiq_pyquil/compiler.py
+++ b/mitiq/mitiq_pyquil/compiler.py
@@ -25,6 +25,7 @@ from typing import Tuple
 
 import numpy as np
 
+from cirq import match_global_phase
 from pyquil.gates import RX, RZ, CZ, I, XY
 from pyquil.quil import Program
 from pyquil.quilbase import Gate
@@ -213,44 +214,10 @@ def _Z(q: int) -> Program:
     return p
 
 
-# This function is taken from cirq. License: apache 2.
-def match_global_phase(
-    a: np.ndarray, b: np.ndarray
-) -> Tuple[np.ndarray, np.ndarray]:
-    """
-    Phases the given matrices so that they agree on the phase of one entry.
-    To maximize precision, the position with the largest entry from one of the
-    matrices is used when attempting to compute the phase difference between
-    the two matrices.
-    :param a: The first matrix
-    :param b: The second matrix
-    :return: A tuple (a', b') where a' == b' implies a == b*exp(i t) for some t
-    """
-
-    # Not much point when they have different shapes.
-    if a.shape != b.shape:
-        return a, b  # pragma: no coverage
-
-    # Find the entry with the largest magnitude in one of the matrices.
-    k = max(np.ndindex(*a.shape), key=lambda t: abs(b[t]))
-
-    def dephase(v):
-        r = np.real(v)
-        i = np.imag(v)
-
-        # Avoid introducing floating point error when axis-aligned.
-        if i == 0:
-            return -1 if r < 0 else 1
-        if r == 0:
-            return 1j if i < 0 else -1j
-
-        return np.exp(-1j * np.arctan2(i, r))
-
-    # Zero the phase at this entry in both matrices.
-    return a * dephase(a[k]), b * dephase(b[k])
-
-
 def is_magic_angle(angle: float) -> bool:
+    """
+    Checks to see if an angle is 0, +/-pi/2, or +/-pi.
+    """
     return (
         np.isclose(np.abs(angle), pi / 2)
         or np.isclose(np.abs(angle), pi)
--- a/mitiq/mitiq_pyquil/compiler.py
+++ b/mitiq/mitiq_pyquil/compiler.py
@@ -242,16 +242,13 @@ def basic_compile(program: Program) -> Program:
 
     for inst in program:
         if isinstance(inst, Gate):
-            angle_param = None
-            if len(inst.params) > 0:
-                angle_param = inst.params[0]
-
             if inst.name == "CCNOT":
                 new_prog += _CCNOT(*inst.qubits)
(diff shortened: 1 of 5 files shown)


karalekas:
After some more thought, this should be included in the current milestone as we're likely going to use `basic_compile` in executor snippets in the preprint, and those snippets should only contain stuff from released versions of mitiq.

mitiq/mitiq_pyquil/tests/test_pyquil_compiler.py
```diff
@@ -0,0 +1,319 @@
...
+    _Z,
+    match_global_phase,
+)
+
+try:
+    from pyquil.simulation.tools import program_unitary
+
+    unitary_tools = True
+except ImportError:
+    unitary_tools = False
```
willzeng, line 73: Why is this needed? Is this to support different pyquil versions?

mitiq/mitiq_pyquil/compiler.py
```diff
@@ -0,0 +1,343 @@
...
+    """
+    p = Program()
+    p += RZ(np.pi, q)
+    return p
+
+
+# This function is taken from cirq. License: apache 2.
+def match_global_phase(
+    a: np.ndarray, b: np.ndarray
+) -> Tuple[np.ndarray, np.ndarray]:
```
willzeng, line 219: Given that we depend on cirq, could we just import this?

mitiq/mitiq_pyquil/compiler.py
```diff
@@ -0,0 +1,343 @@
...
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program.  If not, see <https://www.gnu.org/licenses/>.
+
+"""Provides a utility basic_compile to allow for running on QCS without
+using quilc, as quilc will optimize away unitary folding.
+
+NB: Copied in large part from rigetti/forest-benchmarking (Apache-2.0)
+and modified to support a larger gateset (e.g. CPHASE).
```
rmlarose, line 20:
```suggestion
NB: Copied in large part from rigetti/forest-benchmarking (Apache-2.0) [1]
and modified to support a larger gateset (e.g. CPHASE).

[1] https://github.com/rigetti/forest-benchmarking
```

mitiq/mitiq_pyquil/compiler.py
rmlarose: +1.

mitiq/mitiq_pyquil/tests/test_pyquil_compiler.py
```diff
@@ -0,0 +1,319 @@
...
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program.  If not, see <https://www.gnu.org/licenses/>.
+
+"""NB: Copied in large part from rigetti/forest-benchmarking (Apache-2.0)
+and modified to test a larger gateset.
```
rmlarose, line 17:
```suggestion
"""Tests for the basic pyQuil compiler.

NB: Copied in large part from rigetti/forest-benchmarking (Apache-2.0) [1]
and modified to test a larger gateset.

[1] [1] https://github.com/rigetti/forest-benchmarking
```

mitiq/mitiq_pyquil/tests/test_pyquil_compiler.py
```diff
@@ -0,0 +1,319 @@
...
+
+try:
+    from pyquil.simulation.tools import program_unitary
+
+    unitary_tools = True
+except ImportError:
+    unitary_tools = False
+
+
+def assert_all_close_up_to_global_phase(
```
rmlarose, line 76: Alternative: [`cirq.equal_up_to_global_phase`](https://cirq.readthedocs.io/en/stable/generated/cirq.equal_up_to_global_phase.html#cirq.equal_up_to_global_phase)

mitiq/mitiq_pyquil/tests/test_pyquil_compiler.py
karalekas: Yeah, and it was copied over from forest-benchmarking -- I'll go ahead and delete it, because we require a later version of pyquil anyways.

mitiq/mitiq_pyquil/compiler.py
karalekas: Good point!

mitiq/mitiq_pyquil/tests/test_pyquil_compiler.py
karalekas: Nice!

karalekas:
@willzeng yes i could see this ending up in pyquil at some point, but for now i think we should keep it here

karalekas:
Rebased off of master and fixed the `mypy` issues -- should be ready for final review!

PR #326, opened 2020-08-24
Add mypy type checking to CI, resolve existing issues

### LINKED ISSUE #253: Mypy is pretty upset with the QPROGRAM type
Not sure what the deal is quite yet, but running `mypy` (the static type checker) gives lots of errors of this form. Might be worth investigating?

```
mitiq/mitiq_pyquil/tests/test_zne.py:21: error: Variable "mitiq.QPROGRAM" is not valid as a type
mitiq/mitiq_pyquil/tests/test_zne.py:21: note: See https://mypy.readthedocs.io/en/latest/common_issues.html#variables-vs-type-aliases
```

karalekas:
Note the change to `_typing.py`.

Goal for this PR is to get basic mypy checks working, might follow-up with the `--strict` option in a future PR.

Will need to rebase after #324 is merged and make sure the tests pass.

Fixes #253.

### DIFF
--- a/mitiq/_about.py
+++ b/mitiq/_about.py
@@ -27,7 +27,10 @@ from scipy import __version__ as scipy_version
 
 import mitiq
 
-MITIQ_INSTALL_PATH = os.path.dirname(inspect.getsourcefile(mitiq))
+MITIQ_SOURCE_FILE = inspect.getsourcefile(mitiq)
+assert MITIQ_SOURCE_FILE
+
+MITIQ_INSTALL_PATH = os.path.dirname(MITIQ_SOURCE_FILE)
 PYTHON_VERSION = sys.version_info[0:3]
 
 
--- a/mitiq/zne/inference.py
+++ b/mitiq/zne/inference.py
@@ -812,6 +812,13 @@ class PolyExpFactory(BatchedFactory):
         )
 
 
+# Keep a log of the optimization process storing:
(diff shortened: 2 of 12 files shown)


mitiq/zne/inference.py
```diff
@@ -884,9 +882,9 @@ def __init__(
         self.max_scale_factor = max_scale_factor
         # Keep a log of the optimization process storing:
         # noise value(s), expectation value(s), parameters, and zero limit
-        self.history = (
-            []
-        )  # type: List[Tuple[List[float], List[float], List[float], float]]
+        self.history: List[
+            Tuple[List[Dict[str, float]], List[float], List[float], float]
```
rmlarose, line 886: This should probably be aliased, it's pretty hard to read.

mitiq/_about.py
```diff
@@ -27,7 +27,10 @@
 
 import mitiq
 
-MITIQ_INSTALL_PATH = os.path.dirname(inspect.getsourcefile(mitiq))
+MITIQ_SOURCE_FILE = inspect.getsourcefile(mitiq)
+assert MITIQ_SOURCE_FILE
```
rmlarose, line 31: Would this ever fail?

karalekas: `getsourcefile` has branches where it returns `None`, which is why mypy complained and wanted me to add something of this flavor

Issue #293, opened 2020-08-12
Restructure library


rmlarose:
The "Mitiq man" in the paper draft assumes the following library structure:

![image](https://user-images.githubusercontent.com/32416820/89968972-45b9ec00-dc23-11ea-8329-a4c1a4fdfbd2.png)

Shall we restructure the library to match this?

I vote yes since I think it is conceptually clear and allows us to add mitiq.pec.

The structure would be something like:

```
mitiq/
  zne/
    zne.py
    scaling/
      folding.py
      angle.py (for parameter scaling)
    inference/
      factories.py
  mitiq_qiskit/
  mitiq_pyquil/
```

I think `scaling` is a good name, maybe `noise_scaling` to be super clear, but `inference` could have alternatives, e.g. `fitting` or `extrapolation`.

rmlarose:
Closed by #298.

PR #338, opened 2020-09-01
Restructures scaling.py into a scaling module.

rmlarose:
With the addition of parameter scaling (#288), it makes sense to now have a `scaling` module instead of single file. Additionally, the old `scaling.py` was long enough to warrant a module on its own.

The structure is now

```
zne/
  scaling/
    conversions.py
    folding.py
    parameter.py
```

and one can import the five implemented noise scaling methods (`fold_left`, `fold_right`, `fold_random`, `fold_global`, `scale_parameters`) from `mitiq.zne.scaling`.

### DIFF
--- a/mitiq/conversions.py
+++ b/mitiq/conversions.py
@@ -117,16 +117,22 @@ def convert_from_mitiq(circuit: Circuit, conversion_type: str) -> QPROGRAM:
     return converted_circuit
 
 
-def converter(fold_method: Callable) -> Callable:
-    """Decorator for handling conversions."""
+def converter(noise_scaling_function: Callable) -> Callable:
+    """Decorator for handling conversions.
 
-    @wraps(fold_method)
-    def new_fold_method(circuit: QPROGRAM, *args, **kwargs) -> QPROGRAM:
+    Args:
+        noise_scaling_function: Function which inputs a cirq.Circuit, modifies
+            it to scale noise, then returns a cirq.Circuit.
+    """
+
+    @wraps(noise_scaling_function)
+    def new_scaling_function(circuit: QPROGRAM, *args, **kwargs) -> QPROGRAM:
         mitiq_circuit, input_circuit_type = convert_to_mitiq(circuit)
         if kwargs.get("return_mitiq") is True:
-            return fold_method(mitiq_circuit, *args, **kwargs)
+            return noise_scaling_function(mitiq_circuit, *args, **kwargs)
         return convert_from_mitiq(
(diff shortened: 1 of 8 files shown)


nathanshammah:
I proposed using `scaling` as a sub-package in #293, but a comment by @andreamari (+ @rmlarose) pointed out the issues coming with deep nesting, which we agreed upon, e.g., ``mitiq.zne.scaling.parameter.function`` is quite nested. I think this is too complex. Also, why would `conversions` fit into scaling?

rmlarose:
> I proposed using `scaling` as a sub-package in #293, but a comment by @andreamari (+ @rmlarose) pointed out the issues coming with deep nesting, which we agreed upon, e.g., `mitiq.zne.scaling.parameter.function` is quite nested. I think this is too complex. Also, why would `conversions` fit into scaling?

The nesting is the same as before, namely `mitiq.zne.scaling.scaling_function`. where `scaling_function` is one of

1. `fold_gates_from_left`
1. `fold_gates_from_right`
1. `fold_gates_at_random`
1. `fold_global`
1. `scale_parameters`

We can make them visible from `zne` if we want.

Conversions are needed for scaling because all noise scaling functions act on Cirq circuits but input `QPROGRAM`s. If (when) we need conversions for other components of the library, we can certainly move this module, but for now this is the only place its used.

andreamari (commented):
Thanks @rmlarose! 
Soon or later we may want `convert_to_mitiq` and `convert_from_mitiq` to be outside the `scaling` module. E.g. directly in `utils.py,` or in a new `conversions.py` file placed in the root folder.

Not sure if it is better to do it now in this PR or later. For the rest, LGTM.

nathanshammah:
I'm moving this to Milestone 6.

willzeng:
I'd vote for `conversions.py` to exist above the `/scaling` module

rmlarose:
Moved conversions.py out of scaling, + added tests for conversions. Not sure why coverage is failing (any ideas @karalekas?) but ready for re-review.

andreamari (approved):
LGTM! Thanks @rmlarose. 
The codcov error is unclear also to me.

karalekas:
@rmlarose the problem is that somehow even though the "diff" is fully covered, the overall coverage percentage for the repository is going down. I'll take a quick look -- it might be the `__init__.py` file.

karalekas:
@rmlarose It's unclear to me -- I would just ignore the failure and go ahead and merge.

rmlarose:
Thanks @karalekas. Yeah ironically coverage went down even though I added tests, wanted to make sure I wan't missing anything.

PR #339, opened 2020-09-02
PyQuil executor docs; fns for generating a QCS executor & RB programs

### LINKED ISSUE #315: Add documentation & code examples for running on Rigetti QCS
In line with the IBMQ backends documentation, we should have a section for QCS showing off how to write an executor for QCS, etc. now that pyQuil support has been implemented.

karalekas:
Adds documentation for pyQuil executors, and thus fixes #315. Additionally adds some convenience functions that I put together when collecting data on QCS, specifically a `generate_qcs_executor` function that builds an executor, and `generate_rb_program` function which builds randomized benchmarking (RB) programs using quilc.


mitiq/mitiq_pyquil/executor.py
```diff
@@ -0,0 +1,105 @@
...
+    Example expectation_fn. Computes the ground state expectation, also
+    called survival probability.
+    """
+    num_shots = len(results)
+    return (
+        num_shots - np.count_nonzero(np.count_nonzero(results, axis=1))
+    ) / num_shots
+
+
+def zz_expectation(results: np.ndarray) -> float:
```
rmlarose, line 99: Is this just for two qubit circuits (measurement data)? Should it be generalized to Zi Zj for any pair (i, j) of qubits?

docs/source/guide/guide-executors.rst
```diff
@@ -203,6 +203,74 @@ You can also include both noise models and finite sampling in your executor.
...
+
+        # compute ZZ expectation value
+        return zz_expectation(results)
+
+    program = Program()
+    program += X(0)
+    program += X(1)
+
+    # should give 1.0 with a noiseless backend
+    executor(program)
```
rmlarose, line 272: Since this code is not tested and duplicated from `executor.py`, I propose linking to the function in `executor.py` (API doc or otherwise), or adding a test here.

mitiq/mitiq_pyquil/benchmarking.py
```diff
@@ -0,0 +1,82 @@
...
+
+
+def two_qubit_gateset(q0: int, q1: int) -> List[Gate]:
+    """
+    Return the RX, RZ, CZ native gateset on two qubits.
+    """
+    return one_qubit_gateset(q0) + one_qubit_gateset(q1) + [CZ(q0, q1)]
+
+
+def generate_rb_program(
```
rmlarose, line 50: An alternative would be to generate an RB circuit in Qiskit (we have a function for this) and use conversions. This might be preferable as we would be consistent between RB circuits in Qiskit and pyQuil. We could even make it a more general function and specify the output type.

mitiq/mitiq_pyquil/tests/test_pyquil_executor.py
```diff
@@ -0,0 +1,38 @@
...
+    zz_expectation,
+)
+
+QVM = get_qc("2q-qvm")
+
+
+def test_ground_state_executor():
+    executor = generate_qcs_executor(QVM, ground_state_expectation, debug=True)
+    program = Program(X(0))
+    assert 0.0 == executor(program)
```
rmlarose, line 32: More tests for multi-qubit circuits would be good, as `ground_state_expectation` is for any number of qubits as far as I understand.

mitiq/mitiq_pyquil/executor.py
karalekas: This file is really for templates & examples -- I don't think it needs to be generalized necessarily.

rmlarose (changes requested):
Thanks @karalekas. The `generate_qcs_executor` function is great. Thanks for making it for an arbitrary observable. 

See inline comments for suggestions/requested changes. Also, doc-strings should be Google style for consistency, cf. e.g. [here](https://github.com/unitaryfund/mitiq/blob/master/mitiq/zne/zne.py#L32).

mitiq/mitiq_pyquil/benchmarking.py
karalekas: This is the function that I used for the paper, so I wanted to include it in the repo. But we can drop if you disagree.

karalekas:
@rmlarose good point re: docstrings!!

mitiq/mitiq_pyquil/tests/test_pyquil_executor.py
karalekas: Happy to test it a bit more. Once again, this is more about providing some simple examples for `expectation_fn` than for giving people a complete library of them.

docs/source/guide/guide-executors.rst
karalekas: Yeah I wanted to test it but the Aspen-8 thing causes issues obviously. I'll think on it a bit more.

mitiq/mitiq_pyquil/benchmarking.py
rmlarose: I'd say we can keep it for now and open a new issue to consolidate RB functions. What do you think?

If you agree and want to consolidate now, that's fine also, but I don't think it's blocking for this PR. We can include the function elsewhere (mitiq-examples maybe) so the code for the paper is somewhere.

mitiq/mitiq_pyquil/executor.py
rmlarose: Since the ground state function provides both a template and a (general) example, I would remove `zz_expectation`. I'll leave it to your discretion, though.

mitiq/mitiq_pyquil/benchmarking.py
karalekas: Issue opened: #385

karalekas:
@rmlarose Should be ready for final review.

docs/source/guide/guide-executors.rst
```diff
@@ -203,6 +203,84 @@ You can also include both noise models and finite sampling in your executor.
...
+.. testcode::
+
+    from pyquil import Program, get_qc
+    from pyquil.gates import MEASURE, RESET, X
+
+    from mitiq.mitiq_pyquil.compiler import basic_compile
+    from mitiq.mitiq_pyquil.executor import ground_state_expectation
+
+    # replace with qpu = get_qc("Aspen-8") to run on the Aspen-8 QPU
+    qpu = get_qc("2q-pyqvm")
```
rmlarose, line 236: Also `get_qc("Aspen-8", as_qvm=True)` would work I think? Either is fine.

rmlarose (approved):
Thanks @karalekas, LGTM.

docs/source/guide/guide-executors.rst
karalekas: Yes, but then the docs need a QVM server running, which is annoying. This version is python-only.

PR #352, opened 2020-09-10
Add static extrapolate method to all factories

### LINKED ISSUE #342: Stateful factories can lead to incorrect extrapolations if you do not know to reset them
Issue Description
-----------------

When building custom functionality with factories it can be easy to forget to reset the factory. An example use case is when data is loaded repeatedly into the same factory and then extrapolated.

How to Reproduce
----------------

```python
from mitiq.zne.inference import LinearFactory

fac = LinearFactory([1, 2, 3])
# manually set the data
mitigateds = []
for name, fac in factories.items():
    for e, s in zip(trial_expectations, scale_factors):
        fac.push({"scale_factor": s}, e)
        result = fac.reduce()
        extrapolations_dict[name] = result
        # CRUCIAL NOT TO FORGET
        fac.reset()
```

Potential Solutions
------------------
Could this behavior be prevented or at least to warn the user when they are using this pattern that they might want to reset the factory. For example there could be a warning when you add more data to a non-empty in_stack / out_stack or when you call `.reduce` on the same factory twice.

andreamari:
Description
-----------
Fix #342.

Adds a static `extrapolate` method to all factories. So, if necessary, one can directly extrapolate without instantiating a factory object.
E.g. `zero_limig = RichardsonFactory.extrapolate(scale_factors, exp_values)`

It also adds a flag `self._already_reduced` which is used to raise a warning when new data is pushed to a factory which was already "reduced".

Checklist
-----------

- [x] Contributions to mitiq should follow the [pep8 style](https://www.python.org/dev/peps/pep-0008/). You can enforce it easily with [`black`][black] style and with [`flake8`][flake8] conventions.
- [x] Functions and classes have useful [Google-style][google] docstrings and [type hints][hints] in the signature of the objects.
- [x] (Bug fix) The associated issue is referenced above using [auto-close keywords][auto-close]. If the PR fixes an issue, use the keyword fix/fixes/fixed followed by the issue id. For example, if the PR fixes issue 1184, type "Fixes #1184" (without quotes).
- [x] The [changelog][changelog] is updated, including author and PR number (@username, gh-xxx).

### DIFF
--- a/docs/source/guide/guide-zne.rst
+++ b/docs/source/guide/guide-zne.rst
@@ -442,9 +442,37 @@ corresponds to a statistical inference based on the measured data.
    Error with richardson_fac: 0.0070
    Error with poly_fac: 0.0110
 
----------------------------
+Behind the scenes, a factory object collects different expectation values at different scale factors.
+After running a factory, this information can be accessed with appropriate *get* methods. For example:
+
+.. testcode::
+
+   scale_factors = poly_fac.get_scale_factors()
+   print("Scale factors:", scale_factors)
+   exp_values = poly_fac.get_expectation_values()
+   print("Expectation values:", np.round(exp_values, 2))
+
+.. testoutput::
+
+   Scale factors: [1. 2. 3. 4.]
+   Expectation values: [0.88 0.79 0.72 0.67]
+
+If the user has manually evaluated a list of expectation values associated to a list of scale factors, the
+simplest way to estimate the corresponding zero-noise limit is to directly call the static `extrapolate` method of the 
+desired factory class (in this case initializing a factory object is unnecessary).  For example:
+
+.. testcode::
+
+   zero_limit = PolyFactory.extrapolate(scale_factors, exp_values, order=2)
+   print(f"Error with PolyFactory.extrapolate method: {abs(exact - zero_limit):.4f}")
+
+.. testoutput::
+
+   Error with PolyFactory.extrapolate method: 0.0110
+
+---------------------------------------------
 Advanced usage of a factory
----------------------------
+---------------------------------------------
 
 .. note::
    This section can be safely skipped by all the readers who are interested
--- a/docs/source/guide/guide-zne.rst
+++ b/docs/source/guide/guide-zne.rst
@@ -470,6 +470,8 @@ desired factory class (in this case initializing a factory object is unnecessary
 
    Error with PolyFactory.extrapolate method: 0.0110
 
+Both the zero-noise value and the optimal parameters of the fit can be returned from `extrapolate` by specifying `full_output = True`.
+
 ---------------------------------------------
 Advanced usage of a factory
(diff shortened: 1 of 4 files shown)


andreamari:
@rmlarose I added the `extrapolate` method which you have suggested in #342.
@karalekas This PR should also fix the factory resetting problem which you pointed out.

Once this PR and, in particular, the structure of the `extrapolate` method is approved, I will also update the docs.

mitiq/zne/inference.py
```diff
@@ -137,7 +130,17 @@ def mitiq_polyfit(
...
+        ExtrapolationError: if the deg > len(scale_factors) - 1.
     """
+
+    # This check may be overkilling.
+    if deg > len(scale_factors) - 1:
+        raise ExtrapolationError(
+            "Extrapolation order is too high. The order cannot exceed "
+            f"len(scale_factors) - 1 but order = {deg} and "
+            f"len(scale_factors) = {len(scale_factors)}."
+        )
```
rmlarose, line 142: I wouldn't include this in `mitiq_polyfit`. If anything, I would raise a warning, but top suggestion is to allow any degree.

mitiq/zne/inference.py
```diff
@@ -417,6 +435,37 @@ class PolyFactory(BatchedFactory):
...
+        Note:
+            This method computes the zero-noise limit only from the information
+            contained in the input arguments. To extrapolate from the internal
+            data of an instantiated Factory object, the bound method
+            ".reduce()" should be called instead.
+        """
+
+        opt_params = mitiq_polyfit(scale_factors, exp_values, order)
+        zero_lim = opt_params[-1]
+        return zero_lim, opt_params
```
rmlarose, line 467: Suggestion: Include a `full_output` such that if `full_output == True`, return both `zero_lim` and `opt_params`, else return only `zero_lim`.

I'm fine either way.

rmlarose (commented):
Thanks @andreamari. Overall looks good, a few points for discussion.

docs/source/guide/guide-zne.rst
```diff
@@ -442,9 +442,37 @@ corresponds to a statistical inference based on the measured data.
...
+
+.. testcode::
+
+   zero_limit = PolyFactory.extrapolate(scale_factors, exp_values, order=2)
+   print(f"Error with PolyFactory.extrapolate method: {abs(exact - zero_limit):.4f}")
+
+.. testoutput::
+
+   Error with PolyFactory.extrapolate method: 0.0110
+
```
rmlarose, line 472: Mention the `full_output` option?

rmlarose:
```suggestion

Both the zero-noise value and the optimal parameters of the fit can be returned from `extrapolate` by specifying `full_output = True`.

```

rmlarose (approved):

PR #362, opened 2020-09-16
Removes barriers in conversions from a Qiskit circuit.

### LINKED ISSUE #350: Barriers in Qiskit circuits cause conversion errors (e.g., when folding)
In most cases I think barriers could be removed without affecting the circuit execution -- should we ignore/remove barriers when converting to a Cirq circuit?

rmlarose:
Fixes #350.

### DIFF
--- a/mitiq/mitiq_qiskit/conversions.py
+++ b/mitiq/mitiq_qiskit/conversions.py
@@ -19,10 +19,26 @@ Qiskit's circuit representation.
 import cirq
 from cirq.contrib.qasm_import import circuit_from_qasm
 from qiskit import QuantumCircuit
+from qiskit.circuit.barrier import Barrier
+
 from mitiq.utils import _simplify_circuit_exponents
 
 
 QASMType = str
+
+
+def _remove_barriers(circuit: QuantumCircuit) -> QuantumCircuit:
+    """Returns a copy of the input circuit with all barriers removed.
+
+    Args:
+        circuit: Qiskit circuit to remove barriers from.
+    """
+    copy = circuit.copy()
+    for instr in copy.data:
+        gate = instr[0]
+        if isinstance(gate, Barrier):
+            copy.data.remove(instr)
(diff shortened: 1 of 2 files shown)


mitiq/mitiq_qiskit/conversions.py
```diff
@@ -72,4 +89,5 @@ def from_qasm(qasm: QASMType) -> cirq.Circuit:
     Returns:
         Mitiq circuit representation equivalent to the input QASM string.
     """
+    qasm = _remove_barriers(QuantumCircuit.from_qasm_str(qasm)).qasm()
```
andreamari, line 92: Do we need to `_remove_barriers` in the body of "from_qasm"? Isn't `_remove_barriers` inside `from_qiskit` enough?

mitiq/mitiq_qiskit/conversions.py
```diff
@@ -60,6 +76,7 @@ def from_qiskit(circuit: QuantumCircuit) -> cirq.Circuit:
     Returns:
         Mitiq circuit representation equivalent to the input Qiskit circuit.
     """
+    circuit = _remove_barriers(circuit)
```
andreamari, line 79: If we decide to keep  `_remove_barriers` in the body of `from_qasm`, maybe we can remove  `_remove_barriers` from the body of `from_qiskit` (since `from_qasm` is called one line below).
```suggestion
```

andreamari (commented):
Thanks Ryan! I left two minor (mutually exclusive) comments.

rmlarose:
Oh yeah, oops. Only one call is necessary, thanks @andreamari.

rmlarose:
@karalekas I'll wait for your review before merging. Feel free to merge if it looks good.

andreamari (approved):
Thanks LGTM

karalekas (approved):
LGTM!

PR #364, opened 2020-09-18
Redesign BatchedFactory

### LINKED ISSUE #317: Batch circuit executions in BatchedFactories

rmlarose:
WIP. Partially fixes #317.

@andreamari almost all tests of `BatchedFactory`s use `iterate`. The redesign gets rid of `iterate` because it doesn't make sense to iterate anymore if we send all jobs in one batch. Does this seem reasonable? If so I'll change these tests. 

Other feedback welcome.

To completely fix #317, we'll have to allow for a "batched executor."

### DIFF
--- a/mitiq/zne/inference/factories.py
+++ b/mitiq/zne/inference/factories.py
@@ -16,7 +16,17 @@
 """Classes corresponding to different zero-noise extrapolation methods."""
 from abc import ABC, abstractmethod
 from copy import deepcopy
-from typing import Any, Callable, Dict, List, Optional, Sequence, Tuple
+import inspect
+from typing import (
+    Any,
+    Callable,
+    Dict,
+    Iterable,
+    List,
+    Optional,
+    Sequence,
+    Tuple,
+)
 import warnings
 
 import numpy as np
@@ -215,7 +225,7 @@ class BatchedFactory(BaseFactory, ABC):
 
         # Average the expectation results
         self._outstack = [
-            np.average(res[i * num_to_average: (i + 1) * num_to_average])
+            np.average(res[i * num_to_average : (i + 1) * num_to_average])
             for i in range(len(res) // num_to_average)
         ]
 
@@ -224,17 +234,24 @@ class BatchedFactory(BaseFactory, ABC):
     def run(
         self,
         qp: QPROGRAM,
-        executor: Callable[..., float],
+        executor: Tuple[Callable[..., float], Callable[..., List[float]]],
         scale_noise: Callable[[QPROGRAM, float], QPROGRAM],
         num_to_average: int = 1,
     ) -> "BatchedFactory":
-        """Computes the expectation values at each scale factor by calling the
-        executor sequentially.
+        """Computes the expectation values at each scale factor and stores them
+        in the factory. If the executor returns a single expectation value, the
+        circuits are run sequentially. If the executor is batched and returns
+        a list of expectation values (one for each circuit), then the circuits
+        are sent to the backend as a single job. To detect if an executor is
+        batched, it must be annotated.
 
         Args:
             qp: Quantum circuit to run.
-            executor: Function which inputs a single circuit and outputs a
+            executor: A "single executor" (1) or a "batched executor" (2).
+                (1) A function which inputs a single circuit and outputs a
                 single expectation value of interest.
+                (2) A function which inputs a list of circuits and outputs a
+                list of expectation values (one for each circuit).
             scale_noise: Noise scaling function.
             num_to_average: Number of times to call scale_noise at each scale
                 factor.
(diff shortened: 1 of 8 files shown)


rmlarose:
NB: Also restructures inference.py into a module as this file was pretty long.

rmlarose:
Ready for "pre-review" and discussion before fixing tests (most failing tests are caused by `Factory` -> `BaseFactory`). What do you think of the structure @andreamari @karalekas @nathanshammah @willzeng?

The biggest change is `BatchedFactory` now defines a [`run_batched`](https://github.com/unitaryfund/mitiq/pull/364/files#diff-3107e5b9944f377f85584de86ca20b13R182) method.

andreamari (commented):
Thanks @rmlarose the overall structure looks good to me!

My comments:

- Up to now  `_instack` and `_outstack` where used only to save the data which have been actually measured by the executor not the initialization parameters. This means that when a factory is initialized they are empty, while they become simultaneously populated when running the factory.  Now instead `_instack ` is directly initialized by `__init__`  which means that, before running the factory, we have an intermediate situation in which `_instack` is full while `_outstack` is empty. 
If you think this is ok, we can leave the current structure. Otherwise an alternative option could be to define a private a method `_update_instack()` and to call it only when the measured expectation values are saved into `_outstack`. I have no strong preference here, just to double checking we are aware of this change of "unwritten convention".

- Shall we rename `iterate` to `run_classical` also in AdaptiveFactory? (To have a more uniform names).

- I left two minor code suggestions about `kwargs`.

rmlarose:
Thanks @andreamari. 
> * Up to now  `_instack` and `_outstack` where used only to save the data which have been actually measured by the executor not the initialization parameters. This means that when a factory is initialized they are empty, while they become simultaneously populated when running the factory.  Now instead `_instack ` is directly initialized by `__init__`  which means that, before running the factory, we have an intermediate situation in which `_instack` is full while `_outstack` is empty.
>   If you think this is ok, we can leave the current structure. Otherwise an alternative option could be to define a private a method `_update_instack()` and to call it only when the measured expectation values are saved into `_outstack`. I have no strong preference here, just to double checking we are aware of this change of "unwritten convention".

It's a good point and we should be consistent. I'll populate the `_instack` only when a `run` method is called. (This involves storing the scale factors and shots in the constructor, however.)


> * Shall we rename `iterate` to `run_classical` also in AdaptiveFactory? (To have a more uniform names).

I think having different names is actually good here since adaptive factories actually do iterate. 

Note that if we end up just adding a `run_batched` method (as I've done here), this could be done with the current factory structure by just defining `run_batched` in `BatchedFactory`. Then `iterate` would be a batched factory method.

> * I left two minor code suggestions about `kwargs`.

Thanks! I'll look into these, my current handling of kwargs is very hastily put together and should be made better.

mitiq/zne/inference/factories.py
```diff
@@ -195,6 +98,176 @@ def get_expectation_values(self) -> np.ndarray:
...
+    ) -> "BatchedFactory":
+        """Computes the expectation values at each scale factor by calling
+        `batched_executor` on a list of all circuits to run.
+
+        Args:
+            qp: Quantum circuit to run.
+            batched_executor: Function which inputs a list of circuits and
+                outputs a list of expectation values.
+            scale_noise: Noise scaling function.
+            scale_noise: Noise scaling function.
```
nathanshammah, line 197:
```suggestion
            scale_noise: Noise scaling function.
```

rmlarose:
Latest commit checks the executor signature and runs batched/sequential appropriately. See [here](https://github.com/unitaryfund/mitiq/blob/a928c11f8e6c6ca6960ea02ffdbbe7290ca72e3d/mitiq/zne/inference/factories.py#L265). Let me know what you think @karalekas @andreamari @willzeng.

Note that a function annotation is required with this method to do the batching. If there is no annotation, a "single executor" is assumed.

This can be spruced up a bit, but this is the basic idea. One can even try to get creative with [inspect](https://docs.python.org/3/library/inspect.html#inspect.getsourcelines) and auto generate a batched executor from a single executor, but let's keep it simple for now.

andreamari:
> Latest commit checks the executor signature and runs batched/sequential appropriately. See [here](https://github.com/unitaryfund/mitiq/blob/a928c11f8e6c6ca6960ea02ffdbbe7290ca72e3d/mitiq/zne/inference/factories.py#L265), screenshot below. Let me know what you think @karalekas @andreamari @willzeng.
> 
> image
> 
> Note that a function annotation is required with this method to do the batching. If there is no annotation, a "single executor" is assumed.
> 
> This can be spruced up a bit, but this is the basic idea. One can even try to get creative with [inspect](https://docs.python.org/3/library/inspect.html#inspect.getsourcelines) and auto generate a batched executor from a single executor, but let's keep it simple for now.

Using the annotation seems a reasonable solution to me: if you are a user really interested in batching, then you only need to remember to annotate the executor, otherwise nothing is broken.

If we use this approach, I would also support the idea of internally building a batched executor from a single-circuit executor instead of having two methods (`run` and `run_batched`) with very similar code. But this is a minor aspect.

rmlarose:
Great, thanks @andreamari. Unless anyone is opposed, I will tidy up this PR using the annotated executor approach.

We can open a separate issue in the future on generating a batched executor from a single executor, if even possible to do so in general.

willzeng:
Nice find @rmlarose! I'm glad Python can do this fairly cleanly.

By the way, it wasn't obvious to me that checking the return type is better than checking the input type, but it is! In fact this method would allow someone to write a executor with type `QProgram --> List[float]` that broke up some number of shots across several quantum computers in parallel, pretty snazzy 🎩

rmlarose:
Superseded by #378.

PR #371, opened 2020-09-25
Optimal QPR decomposition for depolarizing noise

### LINKED ISSUE #370: Compute optimal decomposition assuming depolarizing noise
As described in [Temme2017](https://arxiv.org/abs/1612.02058), we would like to compute the quasi-probability representation (QPR) of an "ideal gate", assuming that we have a basis of "noisy implementable gates" containing all all possible unitaries followed by a depolarizing noise channel. The optimal decomposition given such conditions is state in Temme2017 as probabilistically proceeding each gate with a Pauli (or no Pauli), with probability determined by the strength of the depolarizing channel. As such, the goal of this issue is to produce this analytically-derived optimal decomposition using the method described in the paper (rather than e.g. using a linear solver, which will be described in another issue).

### LINKED ISSUE #406: Add missing sub-folders to the "make test" command
Currently the command `make test` considers only `mitiq/tests` and `mitiq/benchmarking `.
Shall we also add `mitiq/zne`  (and in the future `mitiq/pec`) ?

@karalekas

karalekas:
Description
-----------
 
See #370 for more information. Part of the workflow for probabilistic error cancellation (PEC). Fixes #370 and fixes #406 

Checklist
-----------

### DIFF
--- a/mitiq/pec/decomposition/depolarizing.py
+++ b/mitiq/pec/decomposition/depolarizing.py
@@ -15,7 +15,7 @@
 
 from typing import List, Tuple
 
-from cirq import I, Operation, X, Y, Z
+from cirq import Operation, X, Y, Z
 
 NON_ID_PAULIS = [X, Y, Z]
 
@@ -82,6 +82,7 @@ def depolarizing_decomposition(
         "Optimal resource cost for error mitigation,"
         (https://arxiv.org/abs/2006.12509).
     """
+    post_ops: List[List[Operation]]
     qubits = ideal_operation.qubits
 
     # the single-qubit case: linear combination of 1Q Paulis
--- a/mitiq/pec/decomposition/optimal.py
+++ b/mitiq/pec/decomposition/optimal.py
@@ -1,0 +1,112 @@
+# Copyright (C) 2020 Unitary Fund
+#
+# This program is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program.  If not, see <https://www.gnu.org/licenses/>.
+
+from typing import List, Tuple
+
+from cirq import I, Operation, X, Y, Z
+
+PAULIS = [I, X, Y, Z]
+
+
+def depolarizing_decomposition(
+    ideal_operation: Operation, noise_level: float
+) -> List[Tuple[float, List[Operation]]]:
+    r"""As described in [Temme2017]_, optimally decompose a single-qubit
+    ``ideal_operation`` :math:`\mathcal{U}_{\beta}` into its quasi-probability
+    representation (QPR), which is a linear combination of noisy implementable
+    operations :math:`\{\eta_{\alpha} \mathcal{O}_{\alpha}\}`. Takes as an
+    assumption that the basis of implementable operations includes any desired
+    ideal operation followed by a depolarizing channel (meaning that all
+    :math:`\mathcal{O}_{\alpha} = \mathcal{D} \circ \mathcal{U}`, where
+    :math:`\mathcal{D}(\rho) =  (1 - \epsilon) \rho + \epsilon I/2`). Given
+    that assumption, it was proven in [Takagi2020]_ that this method gives an
+    optimal decomposition for a given depolarizing ``noise_level`` (we can
+    easily calculate :math:`\epsilon` from this ``noise_level`` value). For a
+    single-qubit ``ideal_operation``, the optimal decomposition is as follows:
+
+    .. math::
+         \mathcal{U}_{\beta} = \eta_1 \mathcal{O}_1 + \eta_2 \mathcal{O}_2 +
+                               \eta_3 \mathcal{O}_3 + \eta_4 \mathcal{O}_4
+
+    .. math::
+        \eta_1 =1 + \frac{3}{4} \frac{\epsilon}{1- \epsilon},
+        \qquad \mathcal{O}_1 = \mathcal{D} \circ \mathcal{I} \circ \mathcal{U}
+
+        \eta_2 =- \frac{1}{4}\frac{\epsilon}{1- \epsilon} , \qquad
+        \mathcal{O}_2 = \mathcal{D} \circ \mathcal{X} \circ \mathcal{U}
+
+        \eta_3 =- \frac{1}{4}\frac{\epsilon}{1- \epsilon} , \qquad
+        \mathcal{O}_3 = \mathcal{D} \circ \mathcal{Y} \circ \mathcal{U}
+
+        \eta_4 =- \frac{1}{4}\frac{\epsilon}{1- \epsilon} , \qquad
+        \mathcal{O}_4 = \mathcal{D} \circ \mathcal{Z} \circ \mathcal{U}
+
+    Args:
+        ideal_operation: The input ideal operation (gate + qubit) to decompose.
+        noise_level: The noise level (as a float) of the depolarizing channel.
+
+    Returns:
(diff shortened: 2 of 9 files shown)


nathanshammah:
Let's discuss before adding new items to the current milestone, also as we have quite a few open issues for 0.2.0. Glad to see this prototyped so fast and nicely!

mitiq/pec/decomposition/optimal.py
```diff
@@ -0,0 +1,112 @@
...
+
+def depolarizing_decomposition(
+    ideal_operation: Operation, noise_level: float
+) -> List[Tuple[float, List[Operation]]]:
+    r"""As described in [Temme2017]_, optimally decompose a single-qubit
+    ``ideal_operation`` :math:`\mathcal{U}_{\beta}` into its quasi-probability
+    representation (QPR), which is a linear combination of noisy implementable
+    operations :math:`\{\eta_{\alpha} \mathcal{O}_{\alpha}\}`. Takes as an
+    assumption that the basis of implementable operations includes any desired
+    ideal operation followed by a depolarizing channel (meaning that all
```
rmlarose, line 31:
```suggestion
    operations :math:`\{\eta_{\alpha} \mathcal{O}_{\alpha}\}`. 
    
    This function assumes depolarizing noise is the only noise present. 
    In particular, it assumes that the basis of implementable operations 
    includes any desired ideal operation followed by a depolarizing channel 
    (meaning that all
```

mitiq/pec/decomposition/tests/test_optimal.py
```diff
@@ -0,0 +1,81 @@
...
+def test_two_qubit_depolarizing_decomposition():
+    q0, q1 = LineQubit.range(2)
+    noise_level = 0.05
+    optimal_overhead = two_qubit_depolarizing_optimal_overhead(noise_level)
+    assert np.isclose(
+        optimal_overhead,
+        decomposition_overhead(
+            depolarizing_decomposition(CNOT(q0, q1), noise_level)
+        ),
+    )
```
rmlarose, line 81: Add tests for three or more qubit operations.

mitiq/pec/decomposition/tests/test_optimal.py
```diff
@@ -0,0 +1,81 @@
...
+
+
+def test_two_qubit_depolarizing_decomposition():
+    q0, q1 = LineQubit.range(2)
+    noise_level = 0.05
+    optimal_overhead = two_qubit_depolarizing_optimal_overhead(noise_level)
+    assert np.isclose(
+        optimal_overhead,
+        decomposition_overhead(
+            depolarizing_decomposition(CNOT(q0, q1), noise_level)
```
rmlarose, line 79: Consider testing a set of two-qubit operations.

rmlarose (commented):
Thanks @karalekas, looks good with some additional testing. See inline comments. We could also test each operation in the decomposition acts on the same set of qubits as the ideal operation. Also the single qubit decomposition is simple enough to test the full output. These may be overkill, though - up to your discretion.

mitiq/pec/decomposition/optimal.py
```diff
@@ -0,0 +1,112 @@
...
+    :math:`\mathcal{O}_{\alpha} = \mathcal{D} \circ \mathcal{U}`, where
+    :math:`\mathcal{D}(\rho) =  (1 - \epsilon) \rho + \epsilon I/2`). Given
```
andreamari, line 33: The `/2` factor after the identity is valid only for a single qubit but the function is more general. For n qubits the normalization is 1/(2^n).

mitiq/pec/decomposition/optimal.py
```diff
@@ -0,0 +1,112 @@
...
+    # the single-qubit case: linear combination of 1Q Paulis
+    if len(qubits) == 1:
+        q = ideal_operation.qubits[0]
+
+        epsilon = 4 / 3 * noise_level
+        alpha_pos = 1 + ((3 / 4) * epsilon / (1 - epsilon))
+        alpha_neg = -(1 / 4) * epsilon / (1 - epsilon)
+
+        alphas = [alpha_pos] + 3 * [alpha_neg]
+        post_ops = [[P(q)] for P in PAULIS]
```
andreamari, line 93: Here I think everything works if we assume that the identity operation is noiseless. However many simulators (e.g. Cirq) add noise even to identity operations. So maybe it is more safe to "do nothing" for `alpha_pos` instead of applying the `I` gate.

mitiq/pec/decomposition/optimal.py
```diff
@@ -0,0 +1,112 @@
...
+    # the two-qubit case: linear combination of 2Q Paulis
+    elif len(qubits) == 2:
+        q0, q1 = qubits
+
+        epsilon = 16 / 15 * noise_level
+        alpha_pos = 1 + ((15 / 16) * epsilon / (1 - epsilon))
+        alpha_neg = -(1 / 16) * epsilon / (1 - epsilon)
+
+        alphas = [alpha_pos] + 15 * [alpha_neg]
+        post_ops = [[Pi(q0), Pj(q1)] for Pi in PAULIS for Pj in PAULIS]
```
andreamari, line 104: Same comment as before. It is perhaps better to "do nothing" for the first coefficient  `alpha_pos` instead of applying `[I(q0), I(q1)]`. 
And probably the same should be done also for local identities: e.g.,  `[I(q0), X(q1)] `---> `[X(q1)]`.

andreamari (commented):
Thanks @karalekas ! I left a couple of comments.
My main comment is however this one:
**Main comment**: at some point we are going to need methods for testing that the matrix representation of the noisy linear combination returns the expected ideal operation. In the current implementation we never check that the decomposition "works as expected". However, I understand that this a non-trivial addition so we may try to add it now or keep it for a future new PR.

rmlarose (changes requested):
LGTM modulo the module name - what about `decomposition/depolarize.py` instead of `decomposition/optimal.py`? Or something else to reflect its assuming depolarizing noise.

karalekas:
Should be ready for final review!

mitiq/pec/decomposition/depolarizing.py
```diff
@@ -0,0 +1,122 @@
...
+    # the single-qubit case: linear combination of 1Q Paulis
+    if len(qubits) == 1:
+        q = ideal_operation.qubits[0]
+
+        epsilon = 4 / 3 * noise_level
+        alpha_pos = 1 + ((3 / 4) * epsilon / (1 - epsilon))
+        alpha_neg = -(1 / 4) * epsilon / (1 - epsilon)
+
+        alphas = [alpha_pos] + 3 * [alpha_neg]
+        post_ops = [[]]  # for alpha_pos, we do nothing, rather than I
```
andreamari, line 97: Nice trick

mitiq/pec/decomposition/tests/test_depolarizing.py
```diff
@@ -0,0 +1,95 @@
...
+
+
+def two_qubit_depolarizing_optimal_overhead(noise_level: float) -> float:
+    """See [Temme2017]_ for more information.
+
+        .. [Temme2017] : Kristan Temme, Sergey Bravyi, Jay M. Gambetta,
+            "Error mitigation for short-depth quantum circuits,"
+            *Phys. Rev. Lett.* **119**, 180509 (2017),
+            (https://arxiv.org/abs/1612.02058).
+    """
```
andreamari, line 56: I think the optimality was shown in:
```
     .. [Takagi2020] : Ryuji Takagi,
        "Optimal resource cost for error mitigation,"
        (https://arxiv.org/abs/2006.12509).
```
Otherwise we could keep reference to `[Temme2017]_` but remove the word `optimal`. In the tests we are actually not interested in the overhead being optimal.

andreamari (approved):
Thanks @karalekas, LGTM! 
See minor comments.
It would be good to test the validity of the decomposition using the `mitiq.pec.utils._operation_to_choi` to compare an ideal operation with its noisy decomposition. However, to speedup the process, this could be a new independent issue.

mitiq/pec/decomposition/tests/test_depolarizing.py
karalekas: Sounds good -- I'll remove optimal

karalekas:
@andreamari opened up a follow-up issue #415 to track test suite improvements

Issue #369, opened 2020-09-24
Create a type for the decomposition of a noisy operation


rmlarose:

karalekas:
This could use a description!

PR #373, opened 2020-09-25
Add PEC basic implementation assuming a decomposition dictionary is given

### LINKED ISSUE #372: Implement PEC assuming a decomposition dictionary is given
Implement basic functions for:

1. Sampling from a quasi-probability representation (QPR) of an operation.
2. Sampling from a QPR of a circuit.
3. Averaging over many sampled circuits to get a PEC mitigated result.

andreamari:
Description
-----------
Fixes #372 

Implements the basic functions related to Probabilistic Error Cancellation (PEC).
- [x] Hard-code a simple decomposition dictionary for a Paulis + CNOT operation set. To be used for testing all the functions of this PR. In a real world scenario this could be provided by the user or automatically be generated by Mitiq.
- [x] Sampling from a quasi-probability representation (QPR) of an operation.
- [x] Sampling from a QPR of a circuit.
- [x] Averaging over many sampled circuits to get a PEC mitigated result.

Checklist
-----------


andreamari:
Ready for review.
I only add @rmlarose and @karalekas as reviewers but, since this PR is about a key issue, everyone is welcome to have a look/add comments/suggest changes, etc.  (@nathanshammah, @willzeng, @crazy4pi314).

mitiq/pec/pec.py
```diff
@@ -0,0 +1,115 @@
...
+from mitiq._typing import QPROGRAM
+from mitiq.pec.utils import DecoType
+from mitiq.pec.sampling import sample_circuit
+
+
+def execute_with_pec(
+    circuit: QPROGRAM,
+    executor: Callable[[QPROGRAM], float],
+    deco_dict: DecoType,
+    num_to_average: int = 1,
```
rmlarose, line 29: Should this be an arg with PEC? I would remove it and keep only `num_samples`.

mitiq/utils.py
```diff
@@ -133,3 +145,71 @@ def _are_close_dict(dict_a: Dict[Any, Any], dict_b: Dict[Any, Any]) -> bool:
         if not np.isclose(dict_b[ka], va):
             return False
     return True
+
+
+def _max_ent_state_circuit(num_qubits: int) -> Circuit:
```
rmlarose, line 150: Why not allow an arbitrary number of qubits?

```python
top_register = cirq.LineQubit.range(n)
bot_register = cirq.LineQubit.range(n, 2 * n)

return cirq.Circuit(
  cirq.H.on_each(*top_register),
  [cirq.CNOT.on(top_register[i], bot_register[i]) for i in range(n)]
)

File: mitiq/tests/test_utils.py [hunk 3]
```diff
@@ -274,3 +291,46 @@ def test_are_close_dict():
...
+    four_state = np.array(3 * [1, 0, 0, 0, 0] + [1]) / 2.0
+    assert np.allclose(
+        _max_ent_state_circuit(2).final_state_vector(), two_state
+    )
+    assert np.allclose(
+        _max_ent_state_circuit(4).final_state_vector(), four_state
+    )
+
+
+def test_circuit_to_choi_and_operation_to_choi():
```
rmlarose, line 308: Split into two tests, one for `circuit_to_choi` and the other for `operation_to_choi`.

mitiq/pec/utils.py
```diff
@@ -0,0 +1,202 @@
...
+    LineQubit,
+)
+
+# Type definition for a decomposition dictionary.
+# Keys are ideal operations.
+# Values describe the associated decompositions.
+DecoType = Dict[Operation, List[Tuple[float, List[Operation]]]]
+
+
+def _simple_pauli_deco_dict(
```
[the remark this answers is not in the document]
rmlarose: General note: The PR for PEC types will have features to simplify this, and also a lot of the methods here will be methods of the decomposition type. So this code looks good for now but will most likely be updated.

mitiq/pec/test/test_pec_utils.py
```diff
@@ -0,0 +1,143 @@
...
+)
+from mitiq.utils import _operation_to_choi
+
+BASE_NOISE = 0.01
+DECO_DICT = _simple_pauli_deco_dict(BASE_NOISE)
+DECO_DICT_SIMP = _simple_pauli_deco_dict(BASE_NOISE, simplify_paulis=True)
+NOISELESS_DECO_DICT = _simple_pauli_deco_dict(0)
+
+
+def test_simple_pauli_deco_dict_CNOT(gate: Gate = CNOT):
```
rmlarose, line 36: Can this be parameterized with other two qubit gates? If not I would remove `gate` as an argument and just use `CNOT` in the body of the function.

mitiq/pec/test/test_pec_utils.py
```diff
@@ -0,0 +1,143 @@
...
+    deco = DECO_DICT[gate.on(*qreg)]
+    first_coefficient, first_imp_seq = deco[0]
+    assert np.isclose(c_pos * c_pos, first_coefficient)
+    assert first_imp_seq == [gate.on(*qreg)]
+    second_coefficient, second_imp_seq = deco[1]
+    assert np.isclose(c_pos * c_neg, second_coefficient)
+    assert second_imp_seq == [gate.on(*qreg), X.on(qreg[0])]
+    last_coefficient, last_imp_seq = deco[-1]
+    assert np.isclose(c_neg * c_neg, last_coefficient)
+    assert last_imp_seq == [gate.on(*qreg), Z.on(qreg[0]), Z.on(qreg[1])]
```
rmlarose, line 51: There's a decent number of checks here - I think it would be useful to add a doc string to this test (and other tests below) to state what is being tested.

mitiq/pec/test/test_pec_utils.py
```diff
@@ -0,0 +1,143 @@
...
+        first_coefficient, first_imp_seq = deco[0]
+        assert np.isclose(c_pos, first_coefficient)
+        assert first_imp_seq == [gate.on(q)]
+        second_coefficient, second_imp_seq = deco[1]
+        assert np.isclose(c_neg, second_coefficient)
+        assert second_imp_seq == [gate.on(q), X.on(q)]
+
+
+@mark.parametrize("gate", [X, Y, Z])
+def test_simplify_paulist_in_simple_pauli_deco_dict(gate: Gate):
```
rmlarose, line 71:
```suggestion
def test_simplify_paulis_in_simple_pauli_deco_dict(gate: Gate):
```

rmlarose: Typo?

mitiq/pec/test/test_pec_utils.py
```diff
@@ -0,0 +1,143 @@
...
+        deco = deco_dict[gate.on(q)]
+        _, first_imp_seq = deco[0]
+        assert first_imp_seq == [gate.on(q)]
+        _, second_imp_seq = deco[1]
+        input_times_x = {X: I, Y: Z, Z: Y}
+        assert second_imp_seq == [input_times_x[gate].on(q)]
+
+
+@mark.parametrize("gate", [X, Y, Z])
+def test_get_coefficients(gate: Gate):
```
rmlarose, line 84: This and below tests are clear enough to not warrant a doc string.

mitiq/pec/pec.py
```diff
@@ -0,0 +1,115 @@
...
+    norms = []
+
+    for _ in range(num_samples):
+        sampled_circuit, sign, norm = sample_circuit(circuit, deco_dict)
+        sampled_circuits.append(sampled_circuit)
+        signs.append(sign)
+        norms.append(norm)
+
+    # The norm of the quasi-distribution should be independent of sample
+    assert np.all(norms == norms[0])
```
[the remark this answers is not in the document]
rmlarose: I see from `sample_circuit` below that it is indeed useful to return the norm with the current setup, since it's different for different circuits but here the same circuit is passed in every time. So these comments can be ignored for now.

mitiq/pec/test/test_pec_sampling.py
```diff
@@ -0,0 +1,115 @@
...
+        imp_seq, sign, norm = sample_sequence(gate.on(*qreg), DECO_DICT)
+        assert all([isinstance(op, Operation) for op in imp_seq])
+        assert sign in {1.0, -1.0}
+        assert norm > 1
+
+
+def test_sample_circuit_types(circuit: Circuit = twoq_circ):
+    imp_circuit, sign, norm = sample_circuit(circuit, DECO_DICT)
+    assert isinstance(imp_circuit, Circuit)
+    assert sign in {1, 1}
```
rmlarose, line 64: Should be `{-1, 1}`? Or just `sign == 1`?

mitiq/pec/test/test_pec_sampling.py
```diff
@@ -0,0 +1,115 @@
...
+    num_qubits = gate.num_qubits()
+    qreg = LineQubit.range(num_qubits)
+    for _ in range(1000):
+        imp_seq, sign, norm = sample_sequence(gate.on(*qreg), DECO_DICT)
+        assert all([isinstance(op, Operation) for op in imp_seq])
+        assert sign in {1.0, -1.0}
+        assert norm > 1
+
+
+def test_sample_circuit_types(circuit: Circuit = twoq_circ):
```
rmlarose, line 61: Similar comment that argument can be removed and used in body of function if not being parameterized. (Unless there's a reason to do so.)

mitiq/pec/test/test_pec.py
```diff
@@ -0,0 +1,146 @@
...
+        num_to_average=1,
+    )
+    error_unmitigated = abs(unmitigated - 1.0)
+    error_mitigated = abs(mitigated - 1.0)
+    # For a trivial noiseless decomposition no PEC mitigation should happen
+    if deco_dict == NOISELESS_DECO_DICT:
+        assert np.isclose(unmitigated, mitigated)
+    else:
+        assert error_mitigated < error_unmitigated
+        assert np.isclose(mitigated, 1.0, atol=0.1)
```
rmlarose, line 87: Is this the expected tolerance given the decomposition and number of samples?

mitiq/pec/test/test_pec.py
```diff
@@ -0,0 +1,146 @@
...
+
+
```
rmlarose, line 89: Add test that error decreases as number of samples is increased (on average).

mitiq/pec/pec.py
```diff
@@ -0,0 +1,115 @@
...
+    results = results.reshape(-1, num_to_average)
+
+    # Average the "num_to_average" columns
+    exp_values = np.average(results, axis=1)
+
+    # Evaluate unbiased estimators [Temme2017], [Endo2018], [Takagi2020]
+    unbiased_estimators = [s * norm * val for val, s in zip(exp_values, signs)]
+
+    # Average to return the PEC estimate of the ideal expectation value
+    return np.average(unbiased_estimators)
```
rmlarose, line 115: Return the variance as well?

mitiq/pec/pec.py
```diff
@@ -0,0 +1,115 @@
...
+    results = np.array([executor(circ) for circ in to_run])
+
+    # Reshape results to "num_to_average" columns
+    results = results.reshape(-1, num_to_average)
+
+    # Average the "num_to_average" columns
+    exp_values = np.average(results, axis=1)
+
+    # Evaluate unbiased estimators [Temme2017], [Endo2018], [Takagi2020]
+    unbiased_estimators = [s * norm * val for val, s in zip(exp_values, signs)]
```
rmlarose, line 112:
```suggestion
    unbiased_estimators = [s * norm * val for s, val in zip(signs, exp_values)]
```

rmlarose: To match the ordering in the expression `s * norm * val`.

rmlarose (changes requested):
Great stuff @andreamari. This is a solid body of code to build on for PEC. Before merging, I suggest trying to address as many comments below as possible. Anything out of scope can become a new open issue. This is fine as some code here will change once the PR for PEC types (#369) is submitted.

rmlarose:
Also, I think the draft label can be removed now, correct?

mitiq/pec/pec.py
andreamari: I agree

mitiq/pec/pec.py
```diff
@@ -0,0 +1,115 @@
...
+    # The norm of the quasi-distribution should be independent of sample
+    assert np.all(norms == norms[0])
+    norm = norms[0]
+
+    # Repeat each sampled circuit "num_to_average" times
+    to_run = [circ for circ in sampled_circuits for _ in range(num_to_average)]
+
+    # TODO: think about dealing with batched executors
+    # Execute all the circuits
+    results = np.array([executor(circ) for circ in to_run])
```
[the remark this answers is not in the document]
andreamari: 1. Your first point is important. A significant trick for reducing the number of circuits could come from a simplification that you once suggested: for low noise most of the samples are just equal to the input circuit while the few others are usually different. So we could exploit this fact in some way.
2. I am not sure how to do it but I agree. I think that a PEC equivalent of the current BatchedFactory could be useful object here especially for storing the data. But there could be other functional solutions. 

I think that we need a dedicated issue + discussion dedicated to the batching.

mitiq/utils.py
```diff
@@ -133,3 +145,71 @@ def _are_close_dict(dict_a: Dict[Any, Any], dict_b: Dict[Any, Any]) -> bool:
...
+    The density matrix completely characterizes the quantum channel induced by
+    the input operation tree (including the effect of noise if present).
+
+    Args:
+        circuit: The input circuit.
+    Returns:
+        The density matrix of the Choi state associated to the input circuit.
+    """
+    circuit = Circuit(operation_tree)
+    return _circuit_to_choi(circuit)
```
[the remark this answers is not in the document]
andreamari: It is true that for the moment they are used only for testing.
However I expect that we are going to need very soon the Choi matrix or other equivalent matrix representations. E.g. when we will deal with tomography, decompositions, SDPs etc. For example, if we have at disposal noise model and we want to build the `real_matrix` of a `NoisyOperation`, then we could use `_circuit_to_choi`.

mitiq/pec/test/test_pec_sampling.py
andreamari: Oh, it should be `{-1, 1}`.  This shows that negative signs are very unlikely for small noise :-)

mitiq/pec/test/test_pec.py
andreamari: I think so.  But I will check if it is possible to reduce it without risking test failures.

mitiq/pec/pec.py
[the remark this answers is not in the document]
andreamari: As decided at the engineering meeting, we will add support for batching in future. So I mark this comment as resolved.

mitiq/pec/pec.py
andreamari: I added an `full_output` option for returning the standard deviation of the sampled pec values.  I used the standard dev. instead of the variance just for symmetry with respect to ZNE factories.

Note also that, as written in the docstring, this is the "width" of the sampling distribution. To estimate the error between the `pec_value` and the ideal expectation value one should
divide the standard deviation by the square root of the `num_samples`.

I am not sure if this is clear enough for the user.

andreamari:
@karalekas and @rmlarose this is ready for a second review.

mitiq/pec/pec.py
rmlarose: I guess it depends on an understanding of PEC. I would divide it by sqrt(num_samples) and return that since this is the quantity of interest in most cases.

rmlarose (approved):
Thanks @andreamari for addressing the comments. Looks good to merge - let's keep track of the TODO's and perhaps make new issues for changes/improvements (e.g., dividing the variance by sqrt(num_samples) and returning that).

mitiq/utils.py
```diff
@@ -133,3 +145,72 @@ def _are_close_dict(dict_a: Dict[Any, Any], dict_b: Dict[Any, Any]) -> bool:
...
```
karalekas, line 216: why did you choose to put these three functions in `mitiq/utils.py` instead of `mitiq/pec/utils.py`?

mitiq/pec/pec.py
```diff
@@ -0,0 +1,106 @@
...
+import numpy as np
+from mitiq._typing import QPROGRAM
+from mitiq.pec.utils import DecoType
+from mitiq.pec.sampling import sample_circuit
+
+
+def execute_with_pec(
+    circuit: QPROGRAM,
+    executor: Callable[[QPROGRAM], float],
+    deco_dict: DecoType,
```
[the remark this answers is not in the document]
karalekas: Also (this is not really a blocking review comment but a question) -- just as `execute_with_zne` which has a default factory, should we consider defaulting this to something (e.g. the depolarizing decomposition in #371)? Might be unreasonable, but just a thought.

karalekas (changes requested):
A couple thoughts!

mitiq/pec/pec.py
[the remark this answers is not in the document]
andreamari: It would make sense to consider the depolarizing noise as default. However the problem is that we wouldn't know how to choose the strength of the depolarizing parameter. So, I guess we cannot set a default decomposition here.

mitiq/utils.py
andreamari: The reason was that the Choi matrix is a general tool unrelated PEC. However, since for the moment we use it only for PEC, I decided to move the 3 new functions `mitiq/pec/utils.py` as suggested.

mitiq/pec/pec.py
```diff
@@ -0,0 +1,106 @@
...
+
+    # TODO: add support for batched executors
+    # Execute all the circuits
+    exp_values = [executor(circ) for circ in sampled_circuits]
+
+    # Evaluate unbiased estimators [Temme2017], [Endo2018], [Takagi2020]
+    unbiased_estimators = [norm * s * val for s, val in zip(signs, exp_values)]
+
+    if full_output:
+        return np.average(unbiased_estimators), np.std(unbiased_estimators)
```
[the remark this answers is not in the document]
andreamari: I used the `full_output` options for 3 reasons which are however not very strong:

1. For symmetry with `execute_with_zne`.
2. Because `execute_with_pec` is the entry point for a non-expert user which could be confused by additional outputs.
3. Because we may want to return many results in the future. (In some way to be discussed #414).

I created new issue  #414  about this and in particular about how to report raw results of a PEC process.

mitiq/pec/pec.py
andreamari: Agree. I replaced the previous `pec_std` with `pec_error` which is evaluated as:
`pec_error = np.std(unbiased_estimators) / np.sqrt(num_samples)`

andreamari:
@crazy4pi314 , I think I have implemented all your suggestions apart from removing `full_output` (see comment above about it). If you are not convinced please let me know.

I have also created three new issues which emerged in this PR: #412 #413 #414

karalekas:
@andreamari just FYI -- I rebased this off of master and force pushed

karalekas (approved):

karalekas:
This LGTM!

PR #378, opened 2020-10-01
Redesign BatchedFactory (v2)

### LINKED ISSUE #317: Batch circuit executions in BatchedFactories

rmlarose:
Fixes #317. Supersedes #364.

### DIFF
--- a/mitiq/zne/inference.py
+++ b/mitiq/zne/inference.py
@@ -277,12 +277,6 @@ class BaseFactory(ABC):
         """Returns the expectation values computed by the factory."""
         return np.array(self._outstack)
 
-    def _get_keyword_args(self) -> List[Dict[str, Any]]:
-        params = deepcopy(self._instack)
-        for d in params:
-            _ = d.pop("scale_factor")
-        return params
-
     def __eq__(self, other):
         if self._already_reduced != other._already_reduced:
             return False
@@ -378,20 +372,6 @@ class BatchedFactory(BaseFactory, ABC):
             )
         return len(self._outstack) == len(self._scale_factors)
 
-    def _batch_populate_instack(self) -> None:
-        """Populates the instack with all parameters at which to compute
-        expectation values.
-        """
-        if self._shot_list:
-            self._instack = [
-                {"scale_factor": scale, "shots": shots}
-                for scale, shots in zip(self._scale_factors, self._shot_list)
-            ]
-        else:
-            self._instack = [
-                {"scale_factor": scale} for scale in self._scale_factors
-            ]
-
     def run_batched(
         self,
         qp: QPROGRAM,
@@ -446,7 +426,7 @@ class BatchedFactory(BaseFactory, ABC):
 
         Args:
             qp: Quantum circuit to run.
-            executor: A "single executor" (1) or a "batched executor" (2).
+            executor: A "sequential executor" (1) or a "batched executor" (2).
                 (1) A function which inputs a single circuit and outputs a
                 single expectation value of interest.
                 (2) A function which inputs a list of circuits and outputs a
@@ -463,8 +443,7 @@ class BatchedFactory(BaseFactory, ABC):
             Tuple[float],
             Iterable[float],
         ):
-            self.run_batched(qp, executor, scale_noise, num_to_average)
-            return self
+            return self.run_batched(qp, executor, scale_noise, num_to_average)
 
         def _noise_to_expval(scale_factor, **exec_params) -> float:
             """Evaluates the quantum expectation value for a given
@@ -498,6 +477,26 @@ class BatchedFactory(BaseFactory, ABC):
             for _ in range(num_to_average):
                 to_run.append(scale_noise(circuit, scale_factor))
         return to_run
+
+    def _batch_populate_instack(self) -> None:
+        """Populates the instack with all parameters at which to compute
+        expectation values.
+        """
+        if self._shot_list:
+            self._instack = [
+                {"scale_factor": scale, "shots": shots}
+                for scale, shots in zip(self._scale_factors, self._shot_list)
+            ]
+        else:
+            self._instack = [
+                {"scale_factor": scale} for scale in self._scale_factors
+            ]
+
+    def _get_keyword_args(self) -> List[Dict[str, Any]]:
+        params = deepcopy(self._instack)
+        for d in params:
+            _ = d.pop("scale_factor")
+        return params
 
     def __eq__(self, other):
         return BaseFactory.__eq__(self, other) and np.allclose(
--- a/mitiq/zne/inference.py
+++ b/mitiq/zne/inference.py
@@ -16,7 +16,18 @@
 """Classes corresponding to different zero-noise extrapolation methods."""
 from abc import ABC, abstractmethod
 from copy import deepcopy
-from typing import Callable, Dict, List, Optional, Sequence, Tuple, Union
+import inspect
+from typing import (
+    Any,
+    Callable,
+    Dict,
+    Iterable,
+    List,
+    Optional,
+    Sequence,
+    Tuple,
+    Union,
+)
 import warnings
 
 import numpy as np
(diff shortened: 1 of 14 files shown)


andreamari:
Thanks @rmlarose . With the current structure the batched factory and the adaptive one have almost exactly the same methods. So what about if:

1. We go back to the old structure in which a `Factory` is adaptive by default i.e.: move the `run` method of `AdaptiveFactory` to `BaseFactory`. Remove `AdaptiveFactory` and perhaps rename `BaseFactory--->`Factory`.

2. We keep the current tree (`BaseFactory`, `BatchedFactory`, `AdaptiveFactory`) but move all the methods which are related to adaptivity (i.e. push, next, iterate, is_converged) to `AdaptiveFactory` +  re-write BatchedFactory in a more direct way without using "adaptive methods".

3. We keep the current structure of this PR. This would also be ok for me (point 1 and 2 are just minor proposals).

mitiq/zne/inference.py
```diff
@@ -397,12 +378,183 @@ def is_converged(self) -> bool:
...
+            """Evaluates the quantum expectation value for a given
+            scale_factor and other executor parameters."""
+            expectation_values = []
+            for _ in range(num_to_average):
+                scaled_qp = scale_noise(qp, scale_factor)
+                expectation_values.append(executor(scaled_qp, **exec_params))
+            return np.average(expectation_values)
+
+        return self.iterate(
+            _noise_to_expval, max_iterations=len(self._scale_factors) + 1
```
karalekas, line 479: Afaict this is identical to `AdaptiveFactory.run()`, but with `max_iterations` set to `len(self._scale_factors) + 1`. As such, if you follow @andreamari's advice (that I talk about in the previous comment above), this can be replaced by a call to `self.run` with the appropriate args, thus removing the "code smell".

mitiq/zne/inference.py
```diff
@@ -397,12 +378,183 @@ def is_converged(self) -> bool:
...
+        self.reset()
+        executor_annotation = inspect.getfullargspec(executor).annotations
+        if executor_annotation.get("return") in (
+            List[float],
+            Sequence[float],
+            Tuple[float],
+            Iterable[float],
+        ):
+            self.run_batched(qp, executor, scale_noise, num_to_average)
+            return self
```
karalekas, line 467: I believe this can just be `return self.run_batched...` as that method returns `self`.

mitiq/zne/inference.py
```diff
@@ -397,12 +378,183 @@ def is_converged(self) -> bool:
...
+                (1) A function which inputs a single circuit and outputs a
+                single expectation value of interest.
+                (2) A function which inputs a list of circuits and outputs a
+                list of expectation values (one for each circuit).
+            scale_noise: Noise scaling function.
+            num_to_average: Number of times to call scale_noise at each scale
+                factor.
+        """
+        self.reset()
+        executor_annotation = inspect.getfullargspec(executor).annotations
```
karalekas, line 459: Should we somehow notify or warn the user if they are using `run_batched` but this annotation is empty? So that they know we're defaulting to non-batched?

mitiq/zne/inference.py
```diff
@@ -397,12 +378,183 @@ def is_converged(self) -> bool:
...
+        """Computes the expectation values at each scale factor and stores them
+        in the factory. If the executor returns a single expectation value, the
+        circuits are run sequentially. If the executor is batched and returns
+        a list of expectation values (one for each circuit), then the circuits
+        are sent to the backend as a single job. To detect if an executor is
+        batched, it must be annotated.
+
+        Args:
+            qp: Quantum circuit to run.
+            executor: A "single executor" (1) or a "batched executor" (2).
```
karalekas, line 449: Non-blocking comment: I think we could come up a clearer adjective than "single" to describe (1).

mitiq/zne/inference.py
```diff
@@ -397,12 +378,183 @@ def is_converged(self) -> bool:
...
+            batched_executor: Function which inputs a list of circuits and
+                outputs a list of expectation values.
+            scale_noise: Noise scaling function.
+            scale_noise: Noise scaling function.
+            num_to_average: Number of times to call scale_noise at each scale
+                factor.
+        """
+        self.reset()
+        self._batch_populate_instack()
+        kwargs = self._get_keyword_args()
```
karalekas, line 416: Code cleanliness comment -- looks like this method is only used by the `BatchedFactory`, and as such could be moved to this class definition.

karalekas (changes requested):
Looking good! Made some suggestions in line with @andreamari's idea (1). Haven't fully considered (2), so if anyone feels strongly that that is a better alternative, I can think on it more, but so far this seems cleaner to me. As an aside, this is definitely going to conflict with #380, but I'm happy to deal with that on my end after this is merged. P.S. Don't forget to add to the changelog! :)

rmlarose:
Thanks @andreamari and @karalekas. Just addressing the big comment now: I agree (2) is the best option - this is how it was two commits back. However, this causes the Qiskit utils tests to scale because of the noise scaling issue. (#379 is related.) To keep it manageable, how about:

1. Merge this (after addressing other comments).
1. Fix the tests/scaling method in Qiskit utils. Related to #379.
1. Implement the change suggested by @andreamari's point (2).

What do you think?

karalekas:
Ok -- if we go with (2), do we have to rename `Factory` -> `BaseFactory`? And, regardless of the (1) vs. (2) choice, I think some of my comments are still applicable.

rmlarose:
> Thanks @rmlarose . With the current structure the batched factory and the adaptive one have almost exactly the same methods. So what about if:
> 
> 1. We go back to the old structure in which a `Factory` is adaptive by default i.e.: move the `run` method of `AdaptiveFactory` to `BaseFactory`. Remove `AdaptiveFactory` and perhaps rename `BaseFactory--->`Factory`.
> 2. We keep the current tree (`BaseFactory`, `BatchedFactory`, `AdaptiveFactory`) but move all the methods which are related to adaptivity (i.e. push, next, iterate, is_converged) to `AdaptiveFactory` +  re-write BatchedFactory in a more direct way without using "adaptive methods".
> 3. We keep the current structure of this PR. This would also be ok for me (point 1 and 2 are just minor proposals).

Going into more detail: I'm against (1) for the following:

* This will be nearly identical to (2) because `BatchedFactory.run` will have to be overridden in the same way it is now.
* (1) makes the structure much more rigid than with `BaseFactory`, `BatchedFactory`, and `AdaptiveFactory`. The latter will make it easier to make changes in the future and is better design in my opinion. E.g., with the proposed change, `iterate` would no longer be a method of a `BatchedFactory`, only an `AdaptiveFactory`. It's easy to do this now with the `BaseFactory` - notably harder without.
* I'm pretty indifferent w.r.t `BaseFactory` vs. `Factory` (just as a name). However, I would argue for `BaseFactory` because (i) its clear that its a base class and (ii) it used to be called `Factory` but now has a different purpose.

mitiq/zne/inference.py
rmlarose: Moved `_get_keyword_args` to `BatchedFactory` instead of `BaseFactory`.

mitiq/zne/inference.py
rmlarose: I would say no, since:

1. If they call `run` with a non-annotated executor, `run_batched` will never be called.
1. If they call `run_batched` with a non-annotated but batched executor, we will throw a warning but shouldn't.
1. If they call `run_batched` with a non-annotated, non-batched executor, I'm willing to call that user error.

mitiq/zne/inference.py
rmlarose: Yes it is identical. I'm proposing to not have a common run method between `BatchedFactory` and `AdaptiveFactory` (see below), so this will eventually become outdated.

mitiq/zne/inference.py
```diff
@@ -146,20 +157,7 @@ def mitiq_polyfit(
...
-
-    Specific zero-noise extrapolation algorithms, adaptive or non-adaptive,
-    are derived from this class.
-
-    A Factory object is not supposed to directly perform any quantum
-    computation, only the classical results of quantum experiments are
-    processed by it.
-    """
-
+class BaseFactory(ABC):
```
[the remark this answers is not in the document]
rmlarose: See my comment below.

mitiq/zne/inference.py
rmlarose: How about "sequential executor"?

rmlarose:
Moved to #381.

PR #380, opened 2020-10-01
Stricter mypy checking

### LINKED ISSUE #331: Enable strict mypy checking
A follow-on from #326. Using the `--strict` flag with `mypy` produces another 98 errors that will need to be squashed.

karalekas:
Description
-----------

Fixes #331 in line with the discussion in that issue.

Checklist
-----------

### DIFF
--- a/mitiq/zne/inference.py
+++ b/mitiq/zne/inference.py
@@ -170,7 +170,9 @@ class Factory(ABC):
         self.opt_params: List[float] = []
         self._already_reduced = False
 
-    def push(self, instack_val: Dict[str, float], outstack_val: float) -> "Factory":
+    def push(
+        self, instack_val: Dict[str, float], outstack_val: float
+    ) -> "Factory":
         """Appends "instack_val" to "self._instack" and "outstack_val" to
         "self._outstack". Each time a new expectation value is computed this
         method should be used to update the internal state of the Factory.
--- a/mitiq/zne/inference.py
+++ b/mitiq/zne/inference.py
@@ -16,7 +16,7 @@
 """Classes corresponding to different zero-noise extrapolation methods."""
 from abc import ABC, abstractmethod
 from copy import deepcopy
-from typing import Callable, Dict, List, Optional, Sequence, Tuple, Union
+from typing import Any, Callable, Dict, List, Optional, Sequence, Tuple, Union
 import warnings
 
 import numpy as np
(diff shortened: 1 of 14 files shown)


karalekas:
I think that we should ignore codecov here.

karalekas:
Will likely need to fix some merge conflicts after #381 is merged.

mitiq/zne/inference.py
```diff
@@ -302,7 +304,9 @@ def _noise_to_expval(scale_factor, **exec_params) -> float:
 
         return self.iterate(_noise_to_expval, max_iterations)
 
-    def __eq__(self, other):
+    def __eq__(self, other: Any) -> bool:
+        if not isinstance(other, Factory):
+            return False
```
willzeng, line 309: Does it make sense to allow comparisons to things that aren't Factory type objects? Should it be
```suggestion
    def __eq__(self, other: Factory) -> bool:
        if not isinstance(other, Factory):
            return False
```

willzeng: This applies to a few other __eq__ methods below but I won't duplicate the comment

willzeng (approved):
One small question on this that is not blocking. I also agree that we can ignore that the coverage slightly goes down here because of some branching on type checks.

Nicely done! Great to have this as part of our CI @karalekas

mitiq/zne/inference.py
karalekas: That's a great question! It should in fact be `Any`, and if you change it to something else here's what happens:
```
mitiq/zne/inference.py:757: error: Argument 1 of "__eq__" is incompatible with supertype "object"; supertype defines the argument type as "object"
mitiq/zne/inference.py:757: note: This violates the Liskov substitution principle
mitiq/zne/inference.py:757: note: See https://mypy.readthedocs.io/en/stable/common_issues.html#incompatible-overrides
mitiq/zne/inference.py:757: note: It is recommended for "__eq__" to work with arbitrary objects, for example:
mitiq/zne/inference.py:757: note:     def __eq__(self, other: object) -> bool:
mitiq/zne/inference.py:757: note:         if not isinstance(other, ExpFactory):
mitiq/zne/inference.py:757: note:             return NotImplemented
mitiq/zne/inference.py:757: note:         return <logic to compare two ExpFactory instances>
```
The equality comparison method is defined by `object` to accept any type as `other` and thus it violates the [Liskov substitution principle](https://en.wikipedia.org/wiki/Liskov_substitution_principle) to change it (this is one of the tenets of good object-oriented SWE, part of a collection called [SOLID](https://en.wikipedia.org/wiki/SOLID)).

willzeng:
Sounds good

nathanshammah:
This SOLID tenets, is it something we’d like to add to guidelines for contributions? Looks interesting and I was not aware of them, thanks @karalekas . https://github.com/unitaryfund/mitiq/pull/380#discussion_r498998800

PR #381, opened 2020-10-01
Redesign factory objects for batching

### LINKED ISSUE #317: Batch circuit executions in BatchedFactories

rmlarose:
Fixes #317.


rmlarose:
Supersedes #378 because I botched the git revert.

As discussed, coverage intentionally decreases until we fix the bad tests of Qiskit utils. Fixing Issue #382 will increase coverage again.

Ready for final review @andreamari @karalekas.

docs/source/guide/guide-zne.rst
```diff
@@ -287,13 +287,13 @@ The typical tasks of a factory are:
 
 3. Given the history of noise scale factors and results, evaluate the associated zero-noise extrapolation.
 
-The structure of the :class:`.Factory` class is adaptive by construction, since the choice of the next noise
+The structure of the :class:`.BaseFactory` class is adaptive by construction, since the choice of the next noise
```
karalekas, line 290:
```suggestion
The structure of the :class:`.Factory` class is adaptive by construction, since the choice of the next noise
```

mitiq/zne/zne.py
```diff
@@ -49,7 +49,7 @@ def execute_with_zne(
 
     if not isinstance(factory, Factory):
         raise TypeError(
-            f"Argument `factory` must be of type mitiq.factories.Factory "
+            f"Argument `factory` must be of type mitiq.factories.BaseFactory "
```
karalekas, line 52:
```suggestion
            f"Argument `factory` must be of type mitiq.factories.Factory "
```

karalekas (approved):
Two tiny holdovers. Didn't do a super thorough re-check of everything, but if it follows our discussion earlier it's good for me. The follow-up PR offers chances to fix things too if anything slipped through.

andreamari:
@rmlarose : I made some further changes see last commit:

- removed `next` and `is_converged` from `BatchedFactory` since I think they are not necessary
- Tried to simplify the structure by merging `run_batched` into the `run` method of `BatchedFactory`

What do you think? If I made something wrong, feel free to revert.

andreamari:
@rmlarose I got where is the bug. Will fix soon

andreamari:
This is now ready for re-review.
@rmlarose

mitiq/zne/inference.py
```diff
@@ -420,11 +422,18 @@ def run_classical(
...
+
+        if kw_list:
+            self._outstack = [
+                scale_factor_to_expectation_value(scale_factor, **kw_list[i])
+                for i, scale_factor in enumerate(self._scale_factors)
+            ]
+        else:
+            self._outstack = [
+                scale_factor_to_expectation_value(scale_factor)
+                for i, scale_factor in enumerate(self._scale_factors)
```
rmlarose, line 435: `enumerate` isn't necessary

mitiq/zne/inference.py
```diff
@@ -454,38 +459,20 @@ def _batch_populate_instack(self) -> None:
...
-        except IndexError:
-            raise IndexError(
-                "BatchedFactory cannot take another step. "
-                "Number of batched scale_factors"
-                f" ({len(self._scale_factors)}) exceeded."
-            )
-        return in_params
+        # Check if keywords are absent
+        if params is [{} for _ in params]:
+            return None
```
rmlarose, line 482: Why not return the list of empty dictionaries? This is perfectly fine and gets rid of a check for keyword args in line 395.

mitiq/zne/inference.py
```diff
@@ -157,6 +158,57 @@ def mitiq_polyfit(
...
+    def batched_executor(
+        batch: List[QPROGRAM], batched_kwargs: Optional[List[Dict]] = None,
+    ) -> List[float]:
+
+        if batched_kwargs is None:
+            return [executor(circ) for circ in batch]
+        results = []
+        for circ, kwargs in zip(batch, batched_kwargs):
+            results.append(executor(circ, **kwargs))
+        return results  # type: ignore
```
rmlarose: Also can be simplified by allowing `batched_kwargs` to be a list of empty dictionaries. See comment in `_get_keyword_args`.

mitiq/zne/tests/test_inference.py
```diff
@@ -275,7 +275,7 @@ def test_run_sequential_and_batched(factory, batched):
     # Compute expectation values at all the scale factors
     if batched:
 
-        def executor(circuits, kwargs) -> List[float]:
+        def executor(circuits, **kwargs) -> List[float]:
```
rmlarose, line 278: Since this is a batched executor, the argument `kwargs` is a list of keyword arguments, so the signature should be 

```python
def executor(circuits, kwargs) -> List[float]
```

Maybe `kwargs` can be renamed to make this clearer.

rmlarose (commented):
Thanks @andreamari. My main reservation is a conceptual one. Namely, `_make_batched_executor` does not actually make a batched executor - the second one returned is still a "single" or "sequential" executor because the circuits are run one-by-one. 

It may be useful to have such a function as a standalone, but a case for this isn't clear to me currently. For this reason I would argue it's more appropriate for this function, or whichever function handles the batching (`run_batched` before), to be a method of `BatchedFactory`.

mitiq/zne/inference.py
andreamari: Thanks. This part is now removed.

mitiq/zne/inference.py
andreamari: Thanks, good point.

mitiq/zne/tests/test_inference.py
andreamari: In the last commit I just removed the extra argument.
Note: I tried to write the code such that it is possible to define a batched executor without the `kwargs_list` argument, since this is "used" by a factory only in the particular case in which `shot_list` is used.

andreamari:
Thanks @rmlarose!

- As suggested I removed the external `_make_executor_batched` function. I agree it could have been confusing.

- Now there is an internal `_is_executor_batched` method which is used in the body of the `run` method which does not mutate the executor but just detects its type.

- I changed a bit the `_get_keywords method` such that it directly takes into account the `num_to_average` repetitions. In this way, the output is directly compatible with the list of circuits returned by `_generate_circuits`.

- I tried to explain in the docs of `run` how to deal with keywords of a batched executor via a `kwargs_list` argument. Since this is actually relevant only if the `shot_list` option is used, I tried to allow the user to also define a betched executor without the  `kwargs_list` whenever this is possible. This point is a bit delicate and there could be a better/different solution.

Feel free to directly make changes if necessary.

rmlarose:
Thanks @andreamari! Had a quick look and I like these changes. I will now work on fixing the qiskit utils tests (#382) so we can merge this.

> * As suggested I removed the external `_make_executor_batched` function. I agree it could have been confusing.
> * Now there is an internal `_is_executor_batched` method which is used in the body of the `run` method which does not mutate the executor but just detects its type.

Great!

> * I changed a bit the `_get_keywords method` such that it directly takes into account the `num_to_average` repetitions. In this way, the output is directly compatible with the list of circuits returned by `_generate_circuits`.

This is much better imo, thanks for this.

> * I tried to explain in the docs of `run` how to deal with keywords of a batched executor via a `kwargs_list` argument. Since this is actually relevant only if the `shot_list` option is used, I tried to allow the user to also define a betched executor without the  `kwargs_list` whenever this is possible. This point is a bit delicate and there could be a better/different solution.

I don't have ideas yet but I'll think about this.

andreamari:
Note sure why tests are not running

rmlarose:
> Note sure why tests are not running

It may be due to the conflicts.

rmlarose:
@andreamari latest commit fixes the Qiskit utils tests. I'll try to resolve the conflicts now, which may be a bit involved due to the changes for strict mypy checking (#380).

rmlarose:
Ready for re-re-final-re-review @andreamari @karalekas

mitiq/mitiq_qiskit/tests/test_zne_mitiq_qiskit.py
```diff
@@ -49,103 +41,113 @@ def measure(circuit, qid) -> QuantumCircuit:
     """
     # Ensure that we have a classical register of enough size available
     if len(circuit.clbits) == 0:
-        reg = ClassicalRegister(qid + 1, "creg")
+        reg = ClassicalRegister(qid + 1, "cbits")
```
rmlarose, line 44: Note `creg` is a keyword in Qasm. This was causing bugs in the conversions.

mitiq/zne/inference.py
```diff
@@ -202,6 +237,262 @@ def get_expectation_values(self) -> np.ndarray:
...
+                of circuits executed for each scale factor.
+
+        Returns:
+            The output list of keyword dictionaries.
+        """
+        params = deepcopy(self._instack)
+        for d in params:
+            _ = d.pop("scale_factor")
+
+        # Repeat each keyward num_to_average times
```
rmlarose, line 477:
```suggestion
        # Repeat each keyword num_to_average times
```

mitiq/mitiq_qiskit/tests/test_zne_mitiq_qiskit.py
andreamari: nice catch!

andreamari (approved):

andreamari:
@karalekas, you approved this PR but then we made other changes. Could you do a final check on this?

CHANGELOG.md
```diff
@@ -10,6 +10,8 @@
 
 ## Version 0.3.0 (In Development)
 
+- Redesign factories, batch runs in BatchedFactory, fix Qiskit utils tests (@rmlarose, @andreamari, gh-381).
```
karalekas, line 13: this is in the wrong spot -- will need to resolve merge conflicts when moving to "All Changes" section

mitiq/zne/__init__.py
```diff
@@ -14,3 +14,4 @@
 # along with this program.  If not, see <https://www.gnu.org/licenses/>.
 
 from mitiq.zne.zne import execute_with_zne, mitigate_executor, zne_decorator
+from mitiq.zne import scaling
```
karalekas, line 17: i dont understand what this is for

mitiq/zne/inference.py
```diff
@@ -202,6 +237,262 @@ def get_expectation_values(self) -> np.ndarray:
...
+        # Reshape "res" to have "num_to_average" columns
+        res = np.array(res).reshape((-1, num_to_average))
+
+        # Average the "num_to_average" columns
+        self._outstack = np.average(res, axis=1)
+
+        return self
+
+    def run_classical(
+        self, scale_factor_to_expectation_value: Callable[..., float]
```
karalekas, line 406: is this `scale_factor_to_expectation_value` the same as `noise_to_expval` in `AdaptiveFactory`? if so, (1) why do they have different names and (2) why isnt this method defined on `Factory` as an `@abstractmethod`

mitiq/zne/inference.py
```diff
@@ -202,6 +237,262 @@ def get_expectation_values(self) -> np.ndarray:
...
+
+        # Reshape "res" to have "num_to_average" columns
+        res = np.array(res).reshape((-1, num_to_average))
+
+        # Average the "num_to_average" columns
+        self._outstack = np.average(res, axis=1)
+
+        return self
+
+    def run_classical(
```
karalekas, line 405: renaming this from `iterate` is a backwards-incompatible change. I would recommend adding back an `iterate` method to `Factory` that (1) raises a `DeprecationWarning` saying that `iterate` will be removed in `v0.4.0`, and (2) then calls `run_classical`

karalekas (changes requested):
A couple small things.

mitiq/zne/inference.py
rmlarose: Renamed to `scale_factor_to_expectation_value` and defined the abstract method.

mitiq/zne/inference.py
rmlarose: Deprecated iterate.

requirements.txt
```diff
@@ -6,6 +6,9 @@ cirq-unstable==0.9.0.dev20200813185757
 pyquil~=2.18.0
 qiskit~=0.16.2
 
+# to handle deprecations
+deprecation~=2.1.0
```
karalekas, line 10: this is very cool, but it will need to be in `install_requires` of the `setup.py` (otherwise `mitiq` will break), and I'm not sure that it really makes life so much easier that it's worth the additional dependency

karalekas (changes requested):
Also, what's up with coverage? Can we make it pass?

requirements.txt
rmlarose: What's the alternative for handling deprecation that doesn't require an additional dependency?

karalekas: Typically one would just use the `warnings` library to raise a `DeprecationWarning` like we do in pyQuil [here](https://github.com/rigetti/pyquil/blob/6f5c9b1a3aced2b5cec5d81859d81c0e339d16fb/pyquil/device/_main.py#L106). The `deprecations` decorator likely does exactly this, but with a nicely formatted string with the version numbers indicating when to expect the feature to be removed (which you can do yourself with the string argument to `warn`). Lmk if you have any more qs!

rmlarose:
> Also, what's up with coverage? Can we make it pass?

Not sure why codecov is failing now @karalekas. The comment says coverage increased. 

> Merging #381 into master will increase coverage by 1.32%.
> The diff coverage is 93.25%.

karalekas (approved):
For some reason it wants the diff to have 93.55% coverage instead of the 93.25% coverage that it currently has. I'm fine with letting that go in the interest of moving forward. LGTM!

nathanshammah:
Catching up with this now. This is great @rmlarose, and great review and additions from @andreamari and @karalekas.

PR #401, opened 2020-10-07
Add plot methods to factories

### LINKED ISSUE #333: Add a "plot()" method to the Factory class.
Issue Description
-----------------
Very often one is interested in qualitatively "seeing" the best fit generated by a ZNE factory.
Currently it is possible to do this by hand: by extracting data from the factory and plotting it.

Proposed Solution
-----------------
Maybe it could nice if Factory objects had a self.plot() method which directly produces a plot of the best fit with a some nice and fixed style.  Perhaps we could also include additional data in the plot legend  (e.g. zne limit, error, etc. ). In this way one would get a compact visualization of the full error mitigation process.

This idea is inspired by the `lmfit` package in which a similar feature exists. See, e.g., https://lmfit.github.io/lmfit-py/examples/example_emcee_Model_interface.html

**Possible drawback**:  need to add `matplotlib` to the requirements.

rmlarose:
WIP. Fixes #333. cc @crazy4pi314 

Currently `PolyFactory.plot()` produces something like this:

image

Remaining items to do:

* Implement plot for other factories.
    - Note: `RichardsonFactory` and `LinearFactory` should really inherit from `PolyFactory`. This would finish their plot methods.
* Spurce up plots, add features, args, etc.


andreamari:
I would suggest to wait for the features implemented in #403 .
In particular in #403 there is a `get_extrapolation_curve()` which together with the existing `get_scale_factors` and `get_expectation_values`, could allow to define a unique `plot()` method in the base `Factory` class which in principle should work for all the other factories. @rmlarose what do you think?

rmlarose:
> I would suggest to wait for the features implemented in #403 .
> In particular in #403 there is a `get_extrapolation_curve()` which together with the existing `get_scale_factors` and `get_expectation_values`, could allow to define a unique `plot()` method in the base `Factory` class which in principle should work for all the other factories. @rmlarose what do you think?

Sounds useful and seems reasonable to me.

rmlarose:
Superseded by #464.

PR #403, opened 2020-10-09
Add fit errors to factories and add get methods for several results including the fit curve. 

andreamari:
Description
-----------
This is a draft PR. The current changes are:

1. `mitiq_polyfit` and `mitiq_curve_fit` now output also the covariance matrix (i.e. parameter uncertainty) in addition to the fit parameters.

2. Several fit results (optimal parameters, fit errors, zne limit, zne curve) are stored into factory objects in private variables.

3. The results stored in private variables can be accessed by the user by get methods. All the get methods including the old ones are:
      - `get_scale_factors()`
      - `get_expectation_values()`
      - `get_optimal_parameters()`
      - `get_parameters_covariance()` 
      - `get_zero_noise_limit()` 
      - `get_zero_noise_limit_error()`
      - `get_extrapolation_curve()`

**What is still left to do?**

1. If necessary, add more tests.
2. The zero_noise_limit_error is deduced from the covariance matrix error (i.e. from the error on the fit parameters). This is easy for polynomial extrapolation since the zero-noise limit is one of the fit parameters. This more complicated for exponential fits in which error propagation (from parameters to zne limit) should be used. Currently `get_zero_noise_limit_error()` for non-polynomial factories still need to be implemented (it always returns `None`).
3. Update the docs. 

**NOTE:** 
The method `get_extrapolation_curve()` introduced in this PR could be very useful for the scope of #401 

Checklist
-----------

- [ ] Contributions to mitiq should follow the [pep8 style](https://www.python.org/dev/peps/pep-0008/). You can enforce it easily with [`black`](https://black.readthedocs.io/en/stable/index.html) style and with [`flake8`](http://flake8.pycqa.org) conventions.
- [ ] Please add tests to cover your changes, if applicable, and make sure that all new and existing tests pass locally.


andreamari:
As we discussed, I would pass this PR this to @crazy4pi314 but I am always available for questions and help.

andreamari:
Thanks @crazy4pi314 !  
With the last commits, I tried to complete the error propagation part  (and also fixed some failing tests).

Since my new error propagation formulas could be wrong, I left commented some of your previous formulas with additional comments about the technical details.
- If you are convinced by the new formulas please delete the old (commented) ones. 
- If you are not convinced, please let me know! It is very likely to make mistakes on these calculations.

Once we agree on the error propagation stuff, we should try to improve the test coverage. I know that you already improved it but
the net percentage is still negative.

Updating the docs is also a missing final step, but maybe it could become new issue independent from this PR.

crazy4pi314:
Thanks for the help and feedback @andreamari! 💖

I think you are probably right on the propagation of uncertainty modifications, but I have been thinking about the validity of the overall approach.  I documented my concern and proposed solution in #418 and think this is def out of scope for 0.3.

I will continue to improve code coverage and address the docs build issues for the additional snippet I added to the docs to make sure this makes release.

crazy4pi314 (approved):
lgtm! Thanks for the help @andreamari  💖

PR #405, opened 2020-10-14
Adds note on batched executors to zne docs.

### LINKED ISSUE #404: Add docs/examples on batched executors
After merging #381, we should now note how to use "batched executors" in `BatchedFactory`s. This could be as simple as adding the signature of a batched executor

```python
def batched_executor(circuits: List[QPROGRAM]) -> List[float]:
```

as well as a few explanatory sentences to the docs.

rmlarose:
Fixes #404.


andreamari (approved):
Thanks Ryan!

PR #408, opened 2020-10-26
Adds types for PEC

### LINKED ISSUE #369: Create a type for the decomposition of a noisy operation

rmlarose:
Fixes #369.

Note 1: Forked from PR #373, relevant files are in pec/types. 

Note 2: I will follow up with additional PRs on other types to make reviewing more manageable.


mitiq/pec/types/test_types.py
```diff
@@ -0,0 +1,249 @@
...
+    assert noisy_op.real_matrix is not real
+
+
+@pytest.mark.parametrize(
+    "qubit",
+    (cirq.LineQubit(0), cirq.GridQubit(1, 2), cirq.NamedQubit("Qubit")),
+)
+def test_init_with_operation(qubit):
+    ideal_op = cirq.H.on(qubit)
+    real = np.array([[0.51, 0.49], [0.46, -0.45]]) / np.sqrt(2)
```
andreamari, line 47: Here and for all the tests below, the size of the real matrix should be d^2 x d^2 instead of d x d.

andreamari: Maybe you could use: 
`real = np.eye(4)` (this is for sure a valid CPTP operation :-) ) 
or
`real = np.kron([[0.51, 0.49], [0.46, -0.45]], [[0.51, 0.49], [0.46, -0.45]]) `

andreamari (changes requested):
Thanks @rmlarose this looks good and useful!
There is a problem with the choice and the dimension of the real matrix associated with a `NoisyOperation`. See related comments.

andreamari (approved):
Thanks @rmlarose! LGTM!
Note: there are some conflicts so tests did not run.

rmlarose:
Thanks @andreamari. I'll wait for one more review from @karalekas and/or @nathanshammah before merging.

rmlarose:
Now partially fixes #440 and partially fixes #441 (handling conversions in PEC) via conversions in `NoisyOperation`. These issues will be completely fixed once the new types are used in the functions described in these issues.

rmlarose:
@andreamari please take a look at the new changes when you get a chance.

mitiq/pec/types/types.py
```diff
@@ -0,0 +1,446 @@
...
+
+        if not all(
+            isinstance(op, NoisyOperation) for op in basis_expansion.keys()
+        ):
+            raise TypeError(
+                "All keys of `basis_expansion` must be "
+                "of type `NoisyOperation`."
+            )
+
+        self._basis_expansion = cirq.LinearDict(basis_expansion)
```
andreamari, line 368: Is there any advantage in using a `cirq.LinearDict` instead of a standard Python `Dict`?

andreamari (commented):
Thanks @rmlarose!  LGTM up to minor comments.
P.s. I just noticed both `NoisyOperation` and `OperationDecomposition` don't have a class docstring. Also many methods don't have a docstring but this could be ok, especially for simple and self-explanatory methods.

mitiq/pec/types/types.py
rmlarose: A few - arithmetic (`ldict1 + ldict2`), string formatting

```python
ldict = cirq.LinearDict({cirq.X: 0.5, cirq.Z: -0.5})
print(ldict)
# 0.500*X-0.500*Z
```

etc. Time will tell if they're useful or not, but I suspect they will be.

mitiq/pec/types/types.py
```diff
@@ -0,0 +1,446 @@
...
+        self._negativity = sum(abs(coeff) for coeff in self.coeffs)
```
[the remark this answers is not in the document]
rmlarose: Actually - shouldn't we accept any `basis_expansion` since we normalize it to a distribution here?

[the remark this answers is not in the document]
andreamari: If the coefficients don't sum up to 1.0 (without taking the absolute value) it means that we are representing some weird non-trace preserving channel. However, I agree that we should probably not enforce it.
If the user, for some strange reason, wants to use some unconventional representation we can allow it.

andreamari (approved):
LGTM thanks!

mitiq/pec/types/types.py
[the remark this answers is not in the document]
rmlarose: Sounds good, thanks @andreamari

PR #410, opened 2020-10-27
Fix a collection of small build & docs issues

### LINKED ISSUE #341: Simplify and partly automate the release process
Issue Description
-----------------
- [x] First, let's check that the instructions in the docs are correct and consequential to the detail, cc double check with @karalekas (#388)
- [ ] Second, some parts of the process of making a new release could be simplified. Also with some automation. 

Proposed Solution
-----------------
- [x] With regards to copying the pdf file, #340 has simplified the task, also thanks now to the RTD versions and pdfs downloadable there. Another option could be to remove the folder `docs/pdf` with the `mitiq.pdf` file that needs to be copied there from `docs/buld/latex/`, and instead tweak the `.gitignore` so that it ignores all the `build/` folder but for the `build/latex/mitiq.pdf` file. This seems doable with the pattern + `!` [as prefix](https://stackoverflow.com/questions/2415873/exceptions-in-gitignore), i.e.  `!build/latex/mitiq.pdf`
- [ ] We could add a shortcut to the makefile with instructions to build the pdf, just like `make docs` from root is equivalent to `cd docs/ && make html && doctest`. 

Superseded by other issues
----------------------------
- @karalekas has mentioned automatic releases on testpypi for every commit merged to master. I would propose something midways, with a command to be triggered only when really requested by a mantainer, e.g., with a special label such as ``release`` added to a PR that triggers the action when merged (?). Maybe we can see if this is easy with Github Actions. 
- It would be nice if Github Actions or some other CI could be triggered by a new TestPyPI release, and install it on several environments.

### LINKED ISSUE #409: Use ~= instead of == for Cirq dependency
This grabs a [compatible release](https://www.python.org/dev/peps/pep-0440/#compatible-release) which will pull in bugfixes, rather than forcing a pin to a particular patch version.

karalekas:
Description
-----------

Fixes #409 and fixes #341. Partially addresses #406.

Checklist
-----------


andreamari (approved):
LGTM. Thanks Peter!

PR #411, opened 2020-10-27
First pass at PMT

### LINKED ISSUE #256: Add Poor Man's Tomography as a module for obtaining `sigma` that goes into ZNE
When performing ZNE with parameter noise scaling, one needs the base level of noise `sigma_0` to perform noise scaling. We need to add a module for performing Poor Man's Tomography. This module will be designed as follows:

Input: 
Backend

Output:
`sigma_0` representing base level of noise.

Things to be worked out:
1. Do we want to do PMT on each qubit independently and get `sigma_0`s for everyone? 
2. Where will this live in the repository?

yhindy:
Description
-----------
Add poor man's tomography module. 

Fixes #256 

Checklist
-----------


mitiq/utils.py
```diff
@@ -133,3 +133,50 @@ def _are_close_dict(dict_a: Dict[Any, Any], dict_b: Dict[Any, Any]) -> bool:
...
+        """
+        rotation_angle = 2*np.pi / depth
+        moments: List[ops.Moment] = []
+        for _ in range(depth):
+            operations = []
+            num_qubits = gate(rotation_angle).num_qubits()
+            assert num_qubits == len(qubits)
+            operations.append(gate(rotation_angle)(*qubits))
+            moments.append(ops.Moment(operations))
+        return Circuit(moments)
```
willzeng, line 176: This function seems self contained enough for it to make sense to pull it out of the `_poor_mans_tomography` function. Accessing the circuit generation itself will be useful for testing.

mitiq/utils.py
```diff
@@ -133,3 +133,50 @@ def _are_close_dict(dict_a: Dict[Any, Any], dict_b: Dict[Any, Any]) -> bool:
...
+
+    circuit = PMT_circuit([qubit], depth, base_gate)
+    expectation = executor(circuit)
+    Q = (1 - np.power(2*expectation-1, 1/depth))/2
+    sigma = -0.5*np.log(1 - 2*Q)
+    return sigma
```
willzeng, line 182:
```suggestion
    return sigma

```

mitiq/utils.py
```diff
@@ -133,3 +133,50 @@ def _are_close_dict(dict_a: Dict[Any, Any], dict_b: Dict[Any, Any]) -> bool:
...
+def _get_base_gate(gate: EigenGate) -> EigenGate:
+    BASE_GATES = [ZPowGate, HPowGate, XPowGate, YPowGate, CXPowGate, CZPowGate]
+
+    for base_gate in BASE_GATES:
+        if isinstance(gate, base_gate):
+            return cast(EigenGate, base_gate)
+    raise GateTypeException(
+        "Must have circuit be made of rotation gates. "
+        "Your gate {} may not be supported".format(gate)
+    )
```
willzeng, line 147: I'm not entirely clear on this method. One function is to check that the gate type is supported, but why does the `cast` need to happen? If the `gate` is already an instance (or at least an instance of a subtype) of the base_gate why is the cast needed?

Apologies if this is obvious, but it may at least be worth an explanatory comment on the cast for future reference

mitiq/utils.py
```diff
@@ -133,3 +133,50 @@ def _are_close_dict(dict_a: Dict[Any, Any], dict_b: Dict[Any, Any]) -> bool:
...
+
+
+def _poor_mans_tomography(
+        executor: Callabe[..., float],
```
willzeng, line 151:
```suggestion
        executor: Callable[..., float],
```

willzeng (changes requested):
Thanks @yhindy I made some initial comments. A few other general comments for this are:

1. Perhaps we can rename poor mans tomography to something more specific. How about Angle Calibration? Or Parameter Calibration? Am open to other suggestions
2. We should add documentation and testing for these methods.

yhindy:
I've done the minor cleanup changes you requested Will and added a test for the circuit generation. Will still need to add a test for the entire process basically setting up a circuit with some noise and seeing if we can recover the true value.

Will think on another name in the meantime.

yhindy:
Hey so added some docstrings and cleaned up the formatting and such. @willzeng do you think this is in a state that (pending new name) we can merge? Can do a follow on PR for the documentation / examples seems like we should quickly chat about where this info will live in the docs, but in the meantime would like to get something in.

mitiq/utils.py
```diff
@@ -147,6 +149,68 @@ def _are_close_dict(dict_a: Dict[Any, Any], dict_b: Dict[Any, Any]) -> bool:
...
+
+
+def _poor_mans_tomography(
+        executor: Callable[..., float],
+        gate: Gate,
+        qubit: int,
+        depth: int = 100) -> float:
+    """
+    Given an executor and a gate, determines the effective "sigma"
+    that can be used for parameter noise scaling later on.
```
willzeng, line 192:
```suggestion
    Given an executor and a gate, determines the effective
    variance in the control parameter
    that can be used for parameter noise scaling later on.
```

CHANGELOG.md
```diff
@@ -12,6 +12,7 @@
 
 ### All Changes
 - Add a new FakeNodesFactory class based on an alternative interpolation method (@elmandouh, gh-444).
+- Add angle calibration noise tomography to find base noise for parameter noise (@yhindy, gh-411).
```
willzeng, line 15:
```suggestion
- Add parameter calibration method to find base noise for parameter noise (@yhindy, gh-411).
```

willzeng (commented):
Thanks @yhindy! Broadly this looks good and we can keep it as a self contained feature with additional issues for docs and examples as other PRs.

The main outstanding change here is to update the naming throughout. Should rename from "poor mans tomography" / "pmt" to `parameter_calibration`. Am open to other naming suggestions if you have them

@rmlarose This is ready for your review as well if you'd like to add anything else.

yhindy:
@willzeng @rmlarose sorry this took so long. Should be ready to merge

yhindy:
Tracking docs issue here: https://github.com/unitaryfund/mitiq/issues/501

mitiq/zne/scaling/parameter.py
```diff
@@ -31,17 +31,16 @@
...
 
 
-BASE_GATES = [ZPowGate, HPowGate, XPowGate, YPowGate, CXPowGate, CZPowGate]
-
-
 class GateTypeException(Exception):
     pass
 
 
 def _get_base_gate(gate: EigenGate) -> EigenGate:
```
rmlarose, line 38: Is there any case where something is done to the input gate that is not checking membership in the set of base gates? If not, I would suggest to call this `_check_base_gate` or `_check_supported_gate` and simplify this to just:

```python
BASE_GATES = [ZPowGate, HPowGate, XPowGate, YPowGate, CXPowGate, CZPowGate]


def _check_supported_gate(gate: EigenGate) -> None:
    if gate not in BASE_GATES:
        <raise error>
```

mitiq/utils.py
```diff
@@ -147,6 +149,69 @@ def _are_close_dict(dict_a: Dict[Any, Any], dict_b: Dict[Any, Any]) -> bool:
...
+    moments: List[ops.Moment] = []
+    for _ in range(depth):
+        operations = []
+        num_qubits = gate().num_qubits()
+        if num_qubits != len(qubits):
+            raise CircuitMismatchException(
+                "Number of qubits does not match domain size of gate.")
+        operations.append(gate(exponent=rotation_angle)(*qubits))
+        moments.append(ops.Moment(operations))
+    return Circuit(moments)
```
rmlarose, line 182: Checking the number of qubits only needs to happen once (outside the loop). Further, I think this can be simplified to a one-liner like

```python
def _generate_pc_circuit(...):
    if len(qubits) != cirq.num_qubits(gate):
        <raise error>

    return cirq.Circuit(gate(exponent=2 * np.pi / depth).on(*qubits) for _ in range(depth))
```

This will keep the same moment structure as the current function has.

mitiq/utils.py
```diff
@@ -147,6 +149,69 @@ def _are_close_dict(dict_a: Dict[Any, Any], dict_b: Dict[Any, Any]) -> bool:
...
+    pass
+
+
+def _generate_pc_circuit(
+        qubits: Iterable,
+        depth: int,
+        gate: EigenGate) -> Circuit:
+    """
+    Generates a circuit which should be the identity. Given a rotation
+    gate R(param), it applies R(2 * pi / depth) depth times, resulting
```
rmlarose, line 162: What is `param`? The input gate doesn't have an associated parameter, right?

rmlarose: I think every non-pathological gate will have period 2pi, but this function doesn't explicitly require this. I would be explicit about this requirement in the docstring.

mitiq/utils.py
```diff
@@ -147,6 +149,69 @@ def _are_close_dict(dict_a: Dict[Any, Any], dict_b: Dict[Any, Any]) -> bool:
     return True
 
 
+class CircuitMismatchException(Exception):
+    pass
+
+
+def _generate_pc_circuit(
```
rmlarose, line 156:
```suggestion
def _generate_parameter_calibration_circuit(
```

mitiq/utils.py
```diff
@@ -147,6 +149,69 @@ def _are_close_dict(dict_a: Dict[Any, Any], dict_b: Dict[Any, Any]) -> bool:
...
+                "Number of qubits does not match domain size of gate.")
+        operations.append(gate(exponent=rotation_angle)(*qubits))
+        moments.append(ops.Moment(operations))
+    return Circuit(moments)
+
+
+def _parameter_calibration(
+        executor: Callable[..., float],
+        gate: Gate,
+        qubit: int,
```
rmlarose, line 188: Should be `cirq.Qid` not `int`, right?

rmlarose: Does this only work for single-qubit gates? The `_generate_pc_circuit` function works for arbitrary qubit gates.

rmlarose (changes requested):
Sorry @yhindy for being slow on this! Could we put the two new functions in `mitiq.utils` in `zne.scaling.parameter`? I think they're only ever used here, so for me it makes more sense to put them in parameter.py. (There's a bunch of other functions in utils.py which probably shouldn't be there - this is a separate issue.)

mitiq/utils.py
```diff
@@ -147,6 +149,69 @@ def _are_close_dict(dict_a: Dict[Any, Any], dict_b: Dict[Any, Any]) -> bool:
...
+    Generates a circuit which should be the identity. Given a rotation
+    gate R(param), it applies R(2 * pi / depth) depth times, resulting
+    in R(2*pi)
+
+    Args:
+        qubits: a list of qubits
+        depth: the length of the circuit to create
+        gate: the base gate to apply several times
+    Returns:
+        circuit: a pc circuit that can be used for profiling
```
nathanshammah, line 170:
```suggestion
        circuit: a parametrically-calibrated circuit that can be used for profiling
```

nathanshammah (changes requested):
There needs to be documentation about this. Now that the myst interface is made, a notebook could be added in the user guide more simply.

mitiq/utils.py
yhindy: the input gate does have a parameter whether it is explcitly defined by the user or not

mitiq/utils.py
yhindy: https://github.com/unitaryfund/mitiq/issues/582

mitiq/zne/scaling/parameter.py
yhindy: yeah we actually want to return the superclass of the gate. so we just get in a vanilla X gate, we want to return the XPowGate, not say that the X gate is ok

yhindy:
> There needs to be documentation about this. Now that the myst interface is made, a notebook could be added in the user guide more simply.

Yep, theres a docs issue #501. Would be great if we can get this in first though

mitiq/zne/scaling/tests/test_parameter.py
```diff
@@ -131,3 +135,52 @@ def test_gate_type():
     with pytest.raises(GateTypeException):
         forbidden_op = CSWAP(qreg[0], qreg[1], qreg[2])
         _get_base_gate(forbidden_op)
+
+
+def test_parameter_calibration():
+    def noiseless_executor_mock(circuit):
+        return 1
+    # Perfect executor should have sigma = 0
+    qubit = LineQubit.range(1)
```
rmlarose, line 144: Use `qubit = LineQubit(0)`

mitiq/zne/scaling/tests/test_parameter.py
```diff
@@ -131,3 +135,52 @@ def test_gate_type():
...
+        return 0.5
+    qubit = LineQubit.range(1)
+    gate = ops.H.on(qubit[0]).gate
+    with pytest.warns(RuntimeWarning):
+        # Runtime warning for divide by zero
+        sigma = _parameter_calibration(noisy_executor_mock,
+                                       gate,
+                                       qubit[0],
+                                       depth=10)
+    assert sigma == math.inf
```
rmlarose, line 163: Use `np.inf` instead of `math.inf` (numpy was already imported).

mitiq/zne/scaling/tests/test_parameter.py
```diff
@@ -131,3 +135,52 @@ def test_gate_type():
...
+
+
+def test_generate_parameter_calibration_circuit_failure():
+    """Tests that parameter calibration circuit generation fails because there
+    are too many qubits"""
+    n_qubits = 3
+    qubits = LineQubit.range(n_qubits)
+    depth = 10
+    # Should raise exception because too many qubits
+    with pytest.raises(Exception):
```
rmlarose, line 185:
```suggestion
    with pytest.raises(CircuitMismatchException):
```

rmlarose (approved):
Looking good! Thanks @yhindy. A few final minor sugggestions.

mitiq/zne/scaling/parameter.py
```diff
@@ -27,27 +27,91 @@
...
+    gate R(param), it applies R(2 * pi / depth) depth times, resulting
+    in R(2*pi). Requires that the gate is periodic in 2*pi.
+
+    Args:
+        qubits: a list of qubits
+        depth: the length of the circuit to create
+        gate: the base gate to apply several times, must be periodic
+                in 2*pi
+    Returns:
+        circuit: a parameter calibration circuit that can be
```
nathanshammah, line 71:
```suggestion
                in 2*pi
                
    Returns:
        circuit: a parameter calibration circuit that can be
```

mitiq/zne/scaling/parameter.py
```diff
@@ -27,27 +27,91 @@
...
+    Only works for one qubit gates for now.
+
+    Args:
+        executor: a function that takes in a quantum circuit and returns
+            an expectation value
+        gate: the quantum gate that you wish to profile
+        qubit: the index of the qubit you wish to profile
+        depth: the number of operations you would like to use to profile
+            your gate.
+    Returns:
```
nathanshammah, line 100:
```suggestion
        depth: the number of operations you would like to use to profile
            your gate.
            
    Returns:
```

nathanshammah (approved):
Thanks @yhindy. 
Some comments:
- 1. I added minor suggestions for space formatting in the dosctrings (they need to be accepted, eventually).
- 2. I pushed directly a commit adding the `parameter` module to the apidoc.rst file, so that it shows up in the API-docs.
- 3. Since `scale_factor` here has a different meaning from the unitary folding scale factor, I wonder if it does not make sense to call it in another way.
- 4. If you could add a very simple and short tutorial to the examples or brief explanation in the user guide, that would be great. Best to be done in a later PR.

nathanshammah:
The tests are failing, I'll try to fix with Black.

nathanshammah:
There seem to be two source of errors:
1. An error in test-validate action due to a trailing whitespace, in test validate. It does not get picked up automatically with black with `make format`
2. An error in test on `LineQubit, not clear why it should have been prompted by @rmlarose suggestion to go to `qubit = LineQubit(0)`, or if it is randomic:
=================================== FAILURES ===================================
__________________________ test_parameter_calibration __________________________
[gw0] linux -- Python 3.8.8 /opt/hostedtoolcache/Python/3.8.8/x64/bin/python

    def test_parameter_calibration():
        def noiseless_executor_mock(circuit):
            return 1
    
        # Perfect executor should have sigma = 0
        qubit = LineQubit(0)
>       gate = ops.H.on(qubit[0]).gate
E       TypeError: 'LineQubit' object is not subscriptable

nathanshammah (approved):
With these small edits just pushed now test pass. Going to merge it. Thanks @yhindy! If you can provide a minimal usage example for the documentation, it will be very useful for potential users.

crazy4pi314:
@all-contributors please add @yhindy for code, test, doc

PR #421, opened 2020-10-30
Preparation for 0.3.0 release

karalekas:
Self-explanatory -- does all the pre-requisite work before performing a Github & PyPI release for `v0.3.0`.


rmlarose (changes requested):
I would add that factories now support batched executors to the high-level summary, then describe the first draft PEC code.

rmlarose (approved):

PR #422, opened 2020-11-02
Add brief description of noise and error mitigation to README.

### LINKED ISSUE #398: Adding brief info about purpose of Mitiq to top of README
Issue Description
-----------------
There is a nice section in the docs that explains the purpose of the Mitiq library here: https://mitiq.readthedocs.io/en/latest/guide/guide-overview.html
Especially for folks not from a quantum research/theory background it might be nice to include a bit of this at the top of the README for the repo.

Proposed Solution
-----------------

I think a sentence or two at the top of the readme here: https://github.com/unitaryfund/mitiq#mitiq would be all that is needed. I would suggest something covering when you need error mitigation and how mitiq solves it (with links into docs for more info).
Having right at the top will make it easy for folks to understand why they might want to use the library, and to learn more about the approach from the get go.

rmlarose:
Fixes #398 and updates getting started docs.

### DIFF
--- a/README.md
+++ b/README.md
@@ -8,7 +8,14 @@
 
 # Mitiq
 
-Mitiq is a Python toolkit for implementing error mitigation techniques on quantum computers.
+Mitiq is a Python toolkit for implementing error mitigation techniques on quantum computers. 
+
+Current quantum computers
+are noisy due to interactions with the environment, imperfect gate applications, state preparation and measurement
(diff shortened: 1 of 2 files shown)


rmlarose:
Pinging for review.

crazy4pi314 (approved):
Lgtm! Thanks for adding this @rmlarose

PR #426, opened 2020-11-06
Consolidate RB, touch up pyquil utils

### LINKED ISSUE #385: Consolidate RB sequence generation functions
It would be great to have just one source of truth for these sequences, and then convert between different circuit representations. #339 is introducing a way to build these using pyQuil and quilc, and we already have a way to do so using Cirq in the benchmarks directory.

rmlarose:
Fixes #385. Also:

- Touches up pyquil utils by removing global scale noise function in favor of unitary folding. 
- Removes executor.py and puts executor code in pyquil_utils.py.
- Removes old pyquil benchmarking.py and test_pyquil_benchmarking in favor of `rb_circuits` function.


rmlarose:
I think we can safely ignore coverage for this.

mitiq/mitiq_pyquil/pyquil_utils.py
```diff
@@ -15,167 +15,114 @@
...
-    circuit.wrap_in_numshots_loop(shots)
-
-    # we want to simulate noise, so we run without compiling
-    results = QVM.run(circuit)
-    expval = (results == [0]).sum() / shots
-    return float(expval)
+    pq.define_noisy_gate("X", [0], append_kraus_to_gate(kraus_ops, npX))
+    pq.define_noisy_gate("Y", [0], append_kraus_to_gate(kraus_ops, npY))
+    pq.define_noisy_gate("Z", [0], append_kraus_to_gate(kraus_ops, npZ))
+    return pq
```
andreamari, line 50: This function adds noise only if the gate is a single qubit Pauli: X Y Z.
It was useful to scale the noise of the old RB circuits which were made only of Pauli gates.
Now RB circuits are more general and so I think we should just remove `add_depolarizing_noise`  or make it compatible with generic circuits.

mitiq/mitiq_pyquil/tests/test_zne_mitiq_pyquil.py
```diff
@@ -15,64 +15,81 @@
...
+    )
+
+
+QVM = pyquil.get_qc("1q-qvm")
+QVM.qam.random_seed = 1337
+noiseless_executor = generate_qcs_executor(
+    qc=pyquil.get_qc("1q-qvm"),
+    expectation_fn=ground_state_expectation,
+    shots=1_000,
+)
```
andreamari, line 59: All tests seem based on a `noiseless_executor` so we actually never test that mitigation works.
This may be temporarily ok, but we could add a TODO or make a new issue about adding simulated native noise on pyquil tests.

andreamari (commented):
Thanks for the good cleaning up @rmlarose ! 
I left two comments which I think are relevant but non-blocking.

mitiq/mitiq_pyquil/pyquil_utils.py
rmlarose: Ah yeah, I didn't realize this. To be a useful utility it should really be general, I agree. I will open a separate issue for this.

mitiq/mitiq_pyquil/tests/test_zne_mitiq_pyquil.py
rmlarose: The current framework is that we test errors are mitigated in the zne module, while the `mitiq_qiskit` and `mitiq_pyquil` modules only test that the pipeline works for these packages. In my opinion this is sufficient, but please open a separate issue if you see otherwise.

andreamari (approved):
Thanks!

PR #434, opened 2020-11-12
Update version to 0.4.0dev

nathanshammah:
Update version to 0.4.0dev, it was still the released v. 0.3.0.


andreamari (approved):
Thanks @nathanshammah!

PR #436, opened 2020-11-12
Remove changelog requirement from PR template

nathanshammah:
Description
-----------
Removing the requirement to update the changelog in the PR template checklist, as the team@UF agreed that this was creating compatibility issues. This task is taken upon by the release manager before each release, as described [here](https://mitiq.readthedocs.io/en/stable/release.html#update-the-changelog).


rmlarose:
I thought the consensus was to keep the current changelog requirement? (I'm still in favor of removing it.)

rmlarose (commented):
I thought the consensus was to keep the current changelog requirement? (I'm still in favor of removing it.)

PR #443, opened 2020-11-19
Adds RungeFactory, a modified version of RichardsonFactory

### LINKED ISSUE #442: Make a new ZNE Factory based on the S-Runge interpolation algorithm.
Issue Description
-----------------
Richardson extrapolation is equivalent to:
- Getting the polynomial which perfectly interpolates the measured points
- Evaluate the polynomial at the the origin.

This method is well known to be unstable when the number of measured points is large. This problem
can be related to the so-called Runge's phenomenon (https://en.wikipedia.org/wiki/Runge%27s_phenomenon)

Proposed Solution
-----------------

There are two ways of avoiding the Runge's phenomenon:

1. The first way is to choose a particular set of x values (the noise scale factors) known as [Chebyshev nodes](https://en.wikipedia.org/wiki/Chebyshev_nodes).

2. Alternatively, there exists an algorithm to interpolate a polynomial using _fake nodes_ which seems interesting [1]. This method has also a Python implementation https://github.com/pog87/FakeNodes. 

In Mitiq this could be achieved by adding a new Factory class (e.g. `RungeFactory`), which would be almost equal to `RichardsonFactory`, but where one of the two mentioned procedures is used to evaluate the zero-noise limit.

Additional References
---------------------
[1] S. De Marchi, F. Marchetti, E. Perracchione, D. Poggiali, Polynomial interpolation via mapped bases without resampling, Journal of Computational and Applied Mathematics, 364, 2020.

elmandouh:
Fixes #442 

Description
-----------

- This adds a new class that modifies `RichardsonFactory` by polynomially extrapolating over S(scale_factors). Where S(x) maps the scale factors to a set of Chebyshev-Lobatto points [1]. 

- Adds `RungeFactory` tests to `test_inference.py`.

Notes
-----------

The interval is hardcoded to begin from 1.0 to the last sample point of `scale_factors`. 

References
-----------
[1] S. De Marchi, F. Marchetti, E. Perracchione, D. Poggiali, Polynomial interpolation via mapped bases without resampling, Journal of Computational and Applied Mathematics, 364, 2020. [Link](https://www.sciencedirect.com/science/article/abs/pii/S0377042719303449)


elmandouh:
Linting issues aside, this is failing the RungeFactory tests. It should be able to pass all tests that RichardsonFactory passes.

mitiq/zne/inference.py
```diff
@@ -920,6 +920,101 @@ def reduce(self) -> float:
...
+class RungeFactory(BatchedFactory):
+    """Factory object implementing another version of Richardson extrapolation.
+    In this version the original set of nodes (scale_factors) are mapped to set
+    of fake nodes known as Chebyshev-Lobatto points. This method gives a better
+    interpolation when there are a lot of nodes, especially close to the
+    edges of the domain.
+
+    Reference:
+    S. De Marchi, F. Marchetti, E. Perracchione, D. Poggiali,
+    Polynomial interpolation via mapped bases without resampling link, JCAM.
```
rmlarose, line 932: Did you mean to add a link here? Preferably to open-access, though I only found this with a quick search: https://www.sciencedirect.com/science/article/abs/pii/S0377042719303449

mitiq/zne/inference.py
```diff
@@ -920,6 +920,101 @@ def reduce(self) -> float:
...
+        Args:
+        scale_factors: Sequence of noise scale factors at which
+                       expectation values should be measured.
+        Returns:
+            A list of fake nodes (Chebyshev-Lobatto points)
+        """
+
+        # Define the interval [a, b] we are mapping to
+        a = 1.0
+        b = scale_factors[-1]
```
rmlarose, line 1009: I don't remember if `BatchedFactory`s require at least two non-unique scale factors, but if not there should be a check here that `b > a`.

mitiq/zne/inference.py
```diff
@@ -920,6 +920,101 @@ def reduce(self) -> float:
...
+            self.get_scale_factors(),
+            self.get_expectation_values(),
+            full_output=True,
+        )
+
+        self._already_reduced = True
+        return self._zne_limit
+
+    @staticmethod
+    def _map(scale_factors):
```
rmlarose, line 997: Add function annotation. I think it should be

```python
@staticmethod
def _map(scale_factors: np.ndarray) -> np.ndarray:
```

rmlarose (commented):
Thanks @elmandouh for the nice contribution! I really like the simplicity of the implementation.

To fix the style check, run `make format` in the root directory of the repo.

I will leave it to @andreamari to check for correctness of the method. 

Another question for @andreamari: I didn't realize how similar this was to Richardson extrapolation. Because of this, an alternative to a new factory would be to simply add an argument to `RichardsonFactory.extrapolate` for mapping the scale factors. What do you think?

mitiq/zne/inference.py
elmandouh: Should be `Sequence[float]` here.

mitiq/zne/inference.py
elmandouh: I believe `BatchedFactory` raises a value error if len(scale_factors) <2. It doesn't check if they are unique or not. So I will be adding a check.

mitiq/zne/inference.py
elmandouh: Added

elmandouh:
I believe the zero noise limit should be evaluated at S(0) instead of at 0. To do this, I extrapolated inside `RungeFactory.extrapolate` instead of calling `PolyFactory`.

elmandouh:
This pull request is closed to remove unnecessary commits and make it easier to review.

PR #428, opened 2020-11-11
Adds theory subsection about PEC in docs

elmandouh:
Fixes #395  also
- Adds Ryuji Takagi, https://arxiv.org/abs/2006.12509 to refs.bib

PR #444, opened 2020-11-19
Add a new FakeNodesFactory class based on an alternative interpolation method

### LINKED ISSUE #442: Make a new ZNE Factory based on the S-Runge interpolation algorithm.

elmandouh:
Fixes #442 

Description
-----------

- This adds a new class that modifies `RichardsonFactory` by polynomially extrapolating over S(scale_factors). Where S(x) maps the scale factors to a set of Chebyshev-Lobatto points [1]. 

- Adds `RungeFactory` tests to `test_inference.py`.

Notes
-----------

- The interval is hardcoded to begin from 1.0 to the last sample point of `scale_factors`. 

- I believe the zero noise limit should be evaluated at S(0) instead of at 0. To do this, I extrapolated inside RungeFactory.extrapolate instead of calling PolyFactory.

References
-----------
[1] S. De Marchi, F. Marchetti, E. Perracchione, D. Poggiali, Polynomial interpolation via mapped bases without resampling, Journal of Computational and Applied Mathematics, 364, 2020. [Link](https://www.sciencedirect.com/science/article/abs/pii/S0377042719303449)


elmandouh:
@rmlarose The previous pr had way too many commits, so I squashed them and started a new one. Hopefully, this one will be easier to go through. You said that the correctness would be checked by @andreamari. I will also leave the comment that you wrote for him here. 

"Another question for @andreamari: I didn't realize how similar this was to Richardson extrapolation. Because of this, an alternative to a new factory would be to simply add an argument to RichardsonFactory.extrapolate for mapping the scale factors. What do you think?"

mitiq/zne/inference.py
```diff
@@ -920,6 +922,136 @@ def reduce(self) -> float:
...
 
+class RungeFactory(BatchedFactory):
+    """Factory object implementing another version of Richardson extrapolation.
+    In this version the original set of nodes (scale_factors) are mapped to set
+    of fake nodes known as Chebyshev-Lobatto points. This method gives a better
+    interpolation when there are a lot of nodes, especially close to the
+    edges of the domain.
+
+    Reference:
+        https://www.sciencedirect.com/science/article/abs/pii/S0377042719303449
```
andreamari, line 933: There is a specific formatting that we use for including references in function docs, e.g.:
https://github.com/unitaryfund/mitiq/blob/master/mitiq/pec/pec.py

It would be good if you could use a similar style here.

mitiq/zne/inference.py
```diff
@@ -920,6 +922,136 @@ def reduce(self) -> float:
...
+        # Richardson extrapolation is a particular case of a polynomial fit
+        # with order equal to the number of data points minus 1.
+        # In this version of Richardson extrapolation we map the samples to
+        # a new set of fake nodes and polynomially extrapolate.
+
+        order = len(scale_factors) - 1
+
+        # Define interval [a, b] for which the scale_factors are mapped to
+        a = 1.0
+        b = scale_factors[-1]
```
andreamari, line 973:
```suggestion
        b = max(scale_factors)
```
This is a bit safer if `scale_factors` are not ordered.

mitiq/zne/inference.py
```diff
@@ -920,6 +922,136 @@ def reduce(self) -> float:
...
+    ]:
+        # Richardson extrapolation is a particular case of a polynomial fit
+        # with order equal to the number of data points minus 1.
+        # In this version of Richardson extrapolation we map the samples to
+        # a new set of fake nodes and polynomially extrapolate.
+
+        order = len(scale_factors) - 1
+
+        # Define interval [a, b] for which the scale_factors are mapped to
+        a = 1.0
```
andreamari, line 972:
```suggestion
        # Define interval [a, b] for which the scale_factors are mapped to
        a = min(scale_factors)
```
Not sure what is the most appropriate choice for this interval. I should check the details of the algorithm to get some intuition.  For example:
`a = min(scale_factors)` or even `a=0.0` could both be reasonable choices.

mitiq/zne/inference.py
```diff
@@ -920,6 +922,136 @@ def reduce(self) -> float:
...
+            self.get_scale_factors(),
+            self.get_expectation_values(),
+            full_output=True,
+        )
+
+        self._already_reduced = True
+        return self._zne_limit
+
+    @staticmethod
+    def _map(
```
andreamari, line 1023:
```suggestion
    def _map_to_fake_nodes(
```
A more explicit name. Note: if renamed, also successive calls to this function should be updated.

mitiq/zne/inference.py
```diff
@@ -920,6 +922,136 @@ def reduce(self) -> float:
...
+            x: Union[Sequence[float], float],
+            a: float,
+            b: float
+        ) -> Sequence[float]:
+        """
+        A function that maps points to Chebyshev-Lobatto points. Based on
```
andreamari, line 1029: If I am not wrong this function works only for equally spaced points. If so, we have two options:

1. Define a more general function which works also for non-equally spaced points. There is some piece-wise linear function in the original reference that we could use.
2. Enforce `scale_factors` to be equally spaced overriding the `__init__` function and raising an error if they are not so.

mitiq/zne/tests/test_inference.py
```diff
@@ -342,6 +347,18 @@ def test_richardson_extr(test_f: Callable[[float], float]):
     assert np.isclose(fac._opt_params[-1], zne_value)
 
 
+def test_runge_extr(test_f: Callable[[float], float]):
```
andreamari, line 350:
```suggestion
@mark.parametrize("test_f", [f_lin, f_non_lin])
def test_runge_extr(test_f: Callable[[float], float]):
```
I think the decorator to loop over test functions is missing

andreamari:
Thanks @elmandouh for your second contribution to Mitiq! 

@pog87, we don't know each other, but we are trying to use the fake node interpolation about which you already have repo (FakeNode). So feel free to have a look and give some feedback if you are interested!

mitiq/zne/tests/test_inference.py
elmandouh: Yes! I am still learning how to test in python.

mitiq/zne/inference.py
elmandouh: I decided for now to go with option number 2.

mitiq/zne/inference.py
elmandouh: I agree it should be `a = 0.0`, but testing it I found that `a = 1.0` generally gave better results. I am going to assume I somehow missed something, so I will change it to `a = min(scale_factors)` for now.

pog87:
Ciao Andrea, thanks for tagging me. I will write to you in private asap.

elmandouh:
@andreamari I have addressed all your comments. Correctness is still to be well tested, currently working on that.

mitiq/zne/inference.py
elmandouh: It seems like the Lebesgue constant explodes since we are not sampling near zero. Check out theorem 2.4 and eq.9 in the paper.

mitiq/zne/inference.py
```diff
@@ -920,6 +922,153 @@ def reduce(self) -> float:
...
+    @staticmethod
+    def _is_equally_spaced(arr: Sequence[float]) -> bool:
+        """Checks if the sequence is equally spaced."""
+        diff = arr[1] - arr[0]
+        import math
+
+        for num in range(1, len(arr) - 1):
+            if not math.isclose(arr[num + 1] - arr[num], diff, rel_tol=1.0e-9):
+                return False
+        return True
```
andreamari, line 1069:
```suggestion
        diff_arr = np.diff(np.sort(arr))
        return np.allclose(diff_arr, diff_arr[0])
```

andreamari: Not tested but should work.

andreamari (commented):
Thanks @elmandouh! A few other suggestions.

elmandouh:
@andreamari It is not passing the Richardson tests. There are two cases:

- My implementation is incorrect
- Fake nodes method is not a valid method for this kind of problem

If it is the first case then I don't see where the error is, the implementation is quite simple. If it is the second, then there are a few considerations. You mentioned the possibility of keeping it and just letting the user decide wether they want to use the method or not. The fake nodes method works best (beats existing Richardson method) for deeper circuits, so we could leave it just for that case. Another thing to consider is that we are currently only dealing with depolarization noise, I am not sure what kind of behaviour we would get for other noise processes. So again it might not hurt to keep it just for that. 

Either way testing needs to be modified.

andreamari:
> @andreamari It is not passing the Richardson tests. There are two cases:
> 
> * My implementation is incorrect
> * Fake nodes method is not a valid method for this kind of problem
> 
> If it is the first case then I don't see where the error is, the implementation is quite simple. If it is the second, then there are a few considerations. You mentioned the possibility of keeping it and just letting the user decide wether they want to use the method or not. The fake nodes method works best (beats existing Richardson method) for deeper circuits, so we could leave it just for that case. Another thing to consider is that we are currently only dealing with depolarization noise, I am not sure what kind of behaviour we would get for other noise processes. So again it might not hurt to keep it just for that.
> 
> Either way testing needs to be modified.

Thanks @elmandouh !  I will check next week what is happening and check if there exists a regime in which the method works well. Since it is an independent Factory which doesn't interfere with the rest of the library, I think it doesn't harm to keep it even if it is convenient only in some particular regimes.

andreamari:
@elmandouh, I applied some minor changes to `RungeFactory` which now is renamed to `FakeNodesFactory` (I think this is more explicit). Moreover I added a new function `f_runge` to the test file, for which `FakeNodesFactory` is better than `RichardsonFactory`. 
In the doc string of `FakeNodesFactory` I added a final disclaimer which explains that the method is convenient only for very specific regimes.

What do you think of these changes?

elmandouh:
@andreamari Looks good to me!

andreamari:
@elmandouh could you add this line:
- Add a new FakeNodesFactory class based on an alternative interpolation method (@elmandouh, gh-444).
under _All Changes_ in `CHANGELOG.md` ? 
This could create a small conflict with master which should be easy to resolve.

Feel free to also add another line related to your previous contribution about the PEC theory subsection (#428).
External contributions deserve appropriate credit :-) ! 

@rmlarose , I re-ask your review since there have been some changes since your initial review.  Not urgent.

pog87:
I have a suggestion. Instead of S(x) = -cos(pi x) you can try (the rescaled version of) S_alpha(x) = sin(alpha pi x) / sin(alpha pi). With alpha = 0 this is the identity function, with alpha = 1/2 this is called the Koslov-Tal-Ezer mapping.

andreamari:
> I have a suggestion. Instead of S(x) = -cos(pi x) you can try (the rescaled version of) S_alpha(x) = sin(alpha pi x) / sin(alpha pi).
> With alpha = 0 this is the identity function,
> with alpha = 1/2 this is the previous S(x)
> It's called the Koslov-Tal-Ezer mapping.

Hi @pog87, thanks for your suggestion! Given that you are one of the authors of the paper which is at the basis of this PR we are very glad of your direct feedback. Since your suggestion seems like a self-contained enhancement, would you like to create a new issue about it? In this way we can address it in a different PR.

P.s. If you create a new issue, feel free to self-assign the issue to yourself if you are interested in making a PR about it. Otherwise, no worries, we'll take care of it.

mitiq/zne/tests/test_inference.py
```diff
@@ -817,3 +870,19 @@ def test_get_methods_of_factories():
     assert np.allclose(fac.get_scale_factors(), x_values)
     assert np.allclose(fac.get_zero_noise_limit(), zne_reduce)
     assert np.allclose(fac.get_zero_noise_limit_error(), 1.0)
+
+
+def test__FakeNodes_scale_factors_equally_spaced():
```
rmlarose, line 875:
```suggestion
def test_fakenodes_scale_factors_equally_spaced():
```

mitiq/zne/inference.py
```diff
@@ -920,6 +922,151 @@ def reduce(self) -> float:
...
+
+        self._already_reduced = True
+        return self._zne_limit
+
+    @staticmethod
+    def _map_to_fake_nodes(
+        x: Union[Sequence[float], float], a: float, b: float
+    ) -> Sequence[float]:
+        """
+        A function that maps inputs to Chebyshev-Lobatto points. Based on
```
rmlarose, line 1029:
```suggestion
        """A function that maps inputs to Chebyshev-Lobatto points. Based on
```

rmlarose (approved):
Sorry for the delay in reviewing this. Two very minor nits but otherwise looks great.

andreamari (approved):

PR #451, opened 2020-11-24
Add option to automatically deduce the number of samples in PEC.

### LINKED ISSUE #413: Add option to automatically deduce the number of PEC samples.
Pre-Request Checklist
---------------------

Issue Description
-----------------
Currently the default number of PEC samples is 1000 or it can be set manually.
It would be good if Mitiq could, optionally, deduce the appropriate number of samples to be used.

Proposed Solution
-----------------
The number of samples could be deduced from one of (or a combination of) the following things: 

- the amount of negativity of the circuit quasi-probability distribution

- the desired precision

- the actual variance of the measured samples (adaptively)

- etc.

andreamari:
Description
-----------
Fixes #413
It adds a new option for `execute_with_pec` which is called `precision`.
If `num_samples` is not given by the user, it is automatically deduced from the one-norm of the quasi-distribution of the circuit according to the formula:
`num_samples = (one_norm / precision) ** 2`

Checklist
-----------

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,7 @@
 - Add reference to review paper in docs (@willzeng, gh-423).
 - Add unitary folding API (@rmlarose, gh-429).
 - Add new get methods (for fit errors, extrapolation curve, etc.) to Factory objects (@crazy4pi314, @andreamari, gh-403).
+- Add option to automatically deduce the number of samples in PEC (@andreamari, gh-451).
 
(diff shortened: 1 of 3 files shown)


rmlarose (approved):
LGTM, thanks @andreamari. We may want to add a warning if `num_samples`, as computed from `precision`, ends up being enormous - what do you think? Could handle this in a future PR, its low priority right now.

PR #453, opened 2020-11-24
Fix PEC sampling bug.

### LINKED ISSUE #452: Sampling with integer seeds creates sampling problems
Issue Description
-----------------
There is a problem in the way we handle seeds in the `mitiq.pec` module. 
The bug appears if an integer seed is used in `sample_circuit` or `execute_with_pec` since
both functions make repeated calls to `sample_sequence`  each time **with the same seed**.

How to Reproduce
----------------
### Code Snippet

```python
from cirq import LineQubit, Circuit, X, CNOT
from mitiq.pec.sampling import sample_circuit, sample_sequence
from mitiq.pec.utils import _simple_pauli_deco_dict

DECO_DICT = _simple_pauli_deco_dict(0.7, simplify_paulis=True)
qreg = LineQubit.range(1)
oneq_circuit = Circuit([X.on(qreg[0]) for _ in range(10)])

print("Circuits sampled without seed:")
for _ in range(3):
    print(sample_circuit(oneq_circuit, DECO_DICT, random_state=None)[0])
    
print("\nCircuits sampled with seed:")
for _ in range(3):
    print(sample_circuit(oneq_circuit, DECO_DICT, random_state=4)[0])
```

### Error Output

```
Circuits sampled without seed:
0: ───Z───X───X───I───Y───X───Y───X───X───X───
0: ───Z───X───X───X───Y───X───Y───Y───X───X───
0: ───X───X───X───X───Y───I───Y───X───X───X───

Circuits sampled with seed:
0: ───Y───Y───Y───Y───Y───Y───Y───Y───Y───Y───
0: ───Y───Y───Y───Y───Y───Y───Y───Y───Y───Y───
0: ───Y───Y───Y───Y───Y───Y───Y───Y───Y───Y───
```

Possible solution
-------------------

@rmlarose , what about if we require the seed to always be of type `np.RandomState` instead of `Union[np.RandomState, int] `? 
Alternatively we should internally convert integer seeds to `np.RandomState` in *all* sampling functions and in `execute_with_pec`.
Both options are good for me.

rmlarose:
Fixes #452.

### DIFF
--- a/mitiq/pec/tests/test_pec_sampling.py
+++ b/mitiq/pec/tests/test_pec_sampling.py
@@ -126,7 +126,7 @@ def test_sample_circuit_with_seed():
     expected = sample_circuit(circ, decomp, random_state=4)[0]
 
     # Check we're not sampling the same operation every call to sample_sequence
-    assert len(list(expected.all_operations())) > 1
+    assert len(set(expected.all_operations())) > 1
 
     for _ in range(10):
         sampled = sample_circuit(circ, decomp, random_state=4)[0]
--- a/mitiq/pec/tests/test_pec_sampling.py
+++ b/mitiq/pec/tests/test_pec_sampling.py
@@ -119,6 +119,20 @@ def test_sample_sequence_bad_random_state():
         sample_sequence(X(qreg[0]), DECO_DICT, random_state="")
 
 
+def test_sample_circuit_with_seed():
+    decomp = _simple_pauli_deco_dict(0.7, simplify_paulis=True)
+    circ = Circuit([X.on(LineQubit(0)) for _ in range(10)])
+
+    expected = sample_circuit(circ, decomp, random_state=4)[0]
+
+    # Check we're not sampling the same operation every call to sample_sequence
+    assert len(list(expected.all_operations())) > 1
+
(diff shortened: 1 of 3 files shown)


mitiq/pec/tests/test_pec_sampling.py
```diff
@@ -119,6 +119,20 @@ def test_sample_sequence_bad_random_state():
...
 
 
+def test_sample_circuit_with_seed():
+    decomp = _simple_pauli_deco_dict(0.7, simplify_paulis=True)
+    circ = Circuit([X.on(LineQubit(0)) for _ in range(10)])
+
+    expected = sample_circuit(circ, decomp, random_state=4)[0]
+
+    # Check we're not sampling the same operation every call to sample_sequence
+    assert len(list(expected.all_operations())) > 1
```
andreamari, line 129:
```suggestion
    # Check we're not sampling the same operation every call to sample_sequence
    assert len(set(expected.all_operations())) > 1
```
I guess we need `set` to avoid double counting.

andreamari (approved):
Thanks @rmlarose!  Approved, but please check the suggestion before merging.

PR #457, opened 2020-11-29
Adds Collector for efficiently collecting expectation values

### LINKED ISSUE #412: Add support for batched executors in the PEC module
Issue Description
-----------------
Currently the preliminary draft of the pec module works only with single-circuit executors.
We should support batching since it is very important for efficiency, especially in PEC.

Proposed Solution
-----------------
The way to implement this should be discussed. PECFactory? Functional approach?

**Note:** An aspect to keep in mind is that in PEC most of the samples correspond to the unmodified input circuit.
So instead of repeating many copies of the same input circuit it would be more efficient to just increase the number of shots. This trick however could make the signature of executors too complex, so we need to take into account the trade-off between efficiency and simplicity.

rmlarose:
Fixes #412 by implementing the `Collector` proposal described in that issue. Still WIP but initial comments welcome.


mitiq/collector.py
```diff
@@ -0,0 +1,263 @@
...
+        return executor_annotation.get("return") in (
+            List[float],
+            Sequence[float],
+            Tuple[float],
+            Iterable[float],
+            np.ndarray,
+        )
+
+
+class CircuitCollection:
```
willzeng, line 166: This will be more of a comment rather than a suggestion, but wanted to follow up on what I mentioned on the call yesterday @rmlarose @andreamari 

I had been thinking that using a built-in Python `Counter` object i.e. https://pymotw.com/2/collections/counter.html could simplify this. However, it seems that `Counter` objects don't take a custom equality method and instead rely on the `__hash__` function on the relevant objects ([example](https://stackoverflow.com/questions/14667332/object-equallity-in-python-collections-counter)).

Unfortunately, cirq circuits don't currently hash. Since operations like this become a lot simpler if they do, perhaps we should propose upstreaming a `__hash__` method to cirq that hashes the DAG of the `Circuit`. What do you two think? I would open a separate issue for this and it wouldn't block this issue.

mitiq/collector.py
```diff
@@ -0,0 +1,263 @@
...
+        ]
+
+        # List of unique circuits.
+        self._unique: List[cirq.Circuit] = []
+
+        # Dictionary where each key refers to an index of self._raw_circuits
+        # that is the first occurrence of a unique circuit. The value of this
+        # key is the multiplicity/count (number of occurrences) of this circuit
+        # in self._raw_circuits. See CircuitCollection.unique_with_counts.
+        self._counts: Dict[int, int] = {}
```
willzeng, line 187: I would have thought that this would be of type `Dict[int, Circuit]`. Would is be simpler to replace `self._cirq_circuits` with a dict of type `Dict[int, Circuit]`?

mitiq/collector.py
```diff
@@ -0,0 +1,263 @@
...
+            found = False
+            for j, circuit in enumerate(self._unique):
+                if _equal(circ, circuit):
+                    self._counts[list(self._counts.keys())[j]] += 1
+                    found = True
+                    break
+
+            if not found:
+                self._unique.append(circ)
+                self._counts[i] = 1
```
willzeng, line 199: This loop could also be simplified if `Circuits` could be hashed. Then a built in `defaultdict` would allow:
The loop that follows could then use a DefaultDict to be:
```
from collections import defaultdict

counts_dict = defaultdict()
for circ in self._cirq_circuits:
    counts_dict[circ] += 1
```

mitiq/collector.py
```diff
@@ -0,0 +1,263 @@
...
+
+    @property
+    def all(self) -> List[QPROGRAM]:
+        return list(self._raw_circuits)
```
willzeng, line 203: Is this method used? If not, I'd propose that we leave it out

mitiq/collector.py
```diff
@@ -0,0 +1,263 @@
...
+from mitiq import QPROGRAM
+from mitiq.conversions import (
+    convert_to_mitiq,
+    UnsupportedCircuitError,
+    CircuitConversionError,
+)
+from mitiq.utils import _equal
+
+
+class Collector:
```
willzeng, line 35: Would propose that this get broken up into a series of functions. `is_batched_executor` can just stand alone as a function.

Collector can then become a function that has type 
```
def collected(executor: Callable, max_batch_size: int = 100) -> Callable[[Sequence[QProgram], List[float]]
```
The usage would then be:
```
# case where original executor is batched, i.e. takes a list of QPrograms
slow_result = executor([circs])
faster_executor = collected(executor)
fast_result = faster_executor([circs])


# case where original executor is not batched, i.e. takes a list of QPrograms
slow_result = [executor(circ) for circ in circs]

faster_executor = collected(executor)
fast_result = faster_executor([circs])
```
Thus `collected` have the same function signature as batched executors. A decorator version of `collected` could also be defined for ease of use.

The state of the `CircuitCollection` could be stored on the appropriate Factory by calling it separately.

What's nice about this approach is that it is clear that the collection is an optimization that wraps the executor. Its a neat trick that can the be applied compositionally to any executor that a user defines.

willzeng (commented):
@rmlarose Made some more detailed comments following on from our discussion yesterday. Happy to chat further! I really like the trick of dropping out duplicates that this introduces as an optimization.

mitiq/collector.py
andreamari: This Cirq class may be useful for switching between hashable and unhasbable circuit types:
https://cirq.readthedocs.io/en/latest/generated/cirq.FrozenCircuit.html

andreamari (commented):
I have a couple of general comments:

1. I like both the current Collector class and the functional approach proposed by Will.
2. If we use a function to transform an executor into a batched one, in my opinion such function should not care about making the collection of repeated circuits. I think it is much better to do this pre-processing optimization before calling the executor (maybe in a `PecFactory` which can store the collection) or, meanwhile, even directly in the function `execute_with_pec`.
Note that if the collection is created "inside the executor", the number of input circuits would not match to the number of output floats and could create confusion.

I am thinking of this workflow: 
List of PEC circuits ---->  collection of unique circuits with counts  -----> List of unique circuits to actually run on hardware --->  execution with a properly batched executor

mitiq/collector.py
rmlarose: I think `FrozenCircuit`s with a `Counter` is a nice option.

```python
circuits = [cirq.FrozenCircuit(circ) for circ in all_circuits]
counts = Counter(circuits)
```

For example,

```python
>>> circuits
[0: ───H───,
 0: ───H───,
 0: ───H───,
 0: ───H───,
 0: ───H───,
 0: ───H───,
 0: ───H───,
 0: ───X───,
 0: ───X───,
 0: ───X───,
 0: ───X───,
 0: ───X───]

>>> Counter(circuits)
Counter({0: ───H───: 7, 0: ───X───: 5})
```

rmlarose: We'd need to rely on cirq-unstable until cirq v0.10 is released, but this shouldn't be a problem.

rmlarose:
I think the debate between class and function is just a matter of taste and not that important. With the code that's already written the `collected` function proposed by @willzeng is just

```python
def collected(executor: Callable, max_batch_size: int = 100) -> Callable[[Sequence[QProgram], List[float]]:
  return Collector(executor, max_batch_size).run
```

To me the important thing that's left is dealing with executor arguments, which is a bit subtle if executor args are included for each circuit in the list to run. For the purpose of execution, both `circuit` and `args` need to be considered when creating a unique set of circuits to run. So, I'm thinking of how of how to design the `CircuitCollection` (which will probably need to be renamed) for this purpose.

mitiq/collector.py
willzeng: Nice find! That could simplify our code a lot. Do we still need a CircuitCollection object or can we now drop it and use Counter to make dictionaries?

rmlarose:
Note: Docs are failing because the TFQ example. TFQ depends on Cirq < 10 but cirq frozen circuits are in Cirq >= 10.

willzeng:
Gah dependencies!

willzeng:
@rmlarose Is this ready for re-review or are you still working on it?

rmlarose:
Ready now @willzeng @andreamari.

It's not included in this PR, but using this collected executor in `execute_with_pec` causes some tests to fail due to (I think) adding bias. In particular the failing tests are:

```text
FAILED pec/tests/test_pec.py::test_execute_with_pec_one_qubit[decomposition_dict1-circuit0] - assert 0.05234875441167286 < 0.026311278343200684
FAILED pec/tests/test_pec.py::test_execute_with_pec_one_qubit[decomposition_dict1-circuit1] - assert 0.14241544337885303 < 0.07669305801391602
FAILED pec/tests/test_pec.py::test_execute_with_pec_one_qubit[decomposition_dict2-circuit0] - assert 0.05452047117559833 < 0.026311278343200684
FAILED pec/tests/test_pec.py::test_execute_with_pec_one_qubit[decomposition_dict2-circuit1] - assert 0.11187922680612195 < 0.07669305801391602
FAILED pec/tests/test_pec.py::test_execute_with_pec_with_different_samples[1-circuit1] - assert 0.14159116444984274 < 0.0846984748302384
FAILED pec/tests/test_pec.py::test_execute_with_pec_with_different_samples[2-circuit1] - assert 0.13975510558794788 < 0.0846984748302384
FAILED pec/tests/test_pec.py::test_execute_with_pec_with_full_output[100] - AssertionError: assert False
FAILED pec/tests/test_pec.py::test_execute_with_pec_with_full_output[1000] - AssertionError: assert False
FAILED pec/tests/test_pec.py::test_precision_option_in_execute_with_pec[0.1] - assert False
FAILED pec/tests/test_pec.py::test_precision_option_in_execute_with_pec[0.01] - assert False
```

How do you think we should handle this? I have ideas but open to suggestions.

andreamari:
Before going into a detailed review:

- Are we sure that the collection of equal circuits should happen inside `generate_collected_executor` instead of being an independent step of the PEC pipeline? In this case we could avoid doing back and forth conversions between the user circuit type and Cirq type and  `generate_collected_executor` could become `generate_batched_executor` and just handle the batching.

-  Independently of the previous point, what about if `generate_collected_executor` directly replaces the `Collector` class? Is there an advantage in having both of them?
 

About failing tests: This is strange since the executor used in those tests is exact (no shot noise) and, in principle, the results should be the same. Did you try to set `force_run_all=True` just to see if the problem is due the collection par or to something else?

rmlarose:
> * Are we sure that the collection of equal circuits should happen inside `generate_collected_executor` instead of being an independent step of the PEC pipeline? In this case we could avoid doing back and forth conversions between the user circuit type and Cirq type and  `generate_collected_executor` could become `generate_batched_executor` and just handle the batching.

For extensibility it's better to have the current pattern. If we use this elsewhere (which was the original intention), it's better to not have to handle conversions separately every place this is used. If we only every use this for PEC, then I agree either case is the same, but I don' think this is the best strategy.

> * Independently of the previous point, what about if `generate_collected_executor` directly replaces the `Collector` class? Is there an advantage in having both of them?

Advantage is not spending time on unnecessary work I guess. `Collector` is the implementation and `generate_collected_executor` is the interface.

> About failing tests: This is strange since the executor used in those tests is exact (no shot noise) and, in principle, the results should be the same. Did you try to set `force_run_all=True` just to see if the problem is due the collection par or to something else?

Indeed tests pass with `force_run_all=True`. I'm looking more into this.

andreamari:
@rmlarose about the tests: I think found the problem: the order of the results does not match the order of the input circuits. 

The following `Collector.run` method works (the main fix in the last few lines but there are also couple of minor changes e.g. the use of `freeze` ad `unfreeze` methods which you may like):
```
    def run(
        self,
        circuits: Sequence[QPROGRAM],
        force_run_all: bool = False,
        **kwargs,
    ) -> List[float]:
        """Runs all input circuits using the least number of possible calls to
        the executor.

        Args:
            circuits: Sequence of circuits to execute using the executor.
            force_run_all: If True, force every circuit in the input sequence
            to be executed (if some are identical). Else, detects identical
            circuits and runs a minimal set.
        """
        if force_run_all:
            to_run = circuits
        else:
            # Convert to mitiq circuits
            # Note: Assumes all circuits are the same type.
            _, conversion_type = convert_to_mitiq(circuits[0])
            mitiq_circuits = [convert_to_mitiq(circ)[0] for circ in circuits]

            # Convert to FrozenCircuits
            frozen_circuits = [circ.freeze() for circ in mitiq_circuits]

            # Get the unique circuits and counts
            collection = Counter(frozen_circuits)
            unique_circuits = collection.keys()
            # TODO: counts should be used to scale the number of shots
            # counts = collection.values()

            # Convert back to user type before running
            to_run = [
                convert_from_mitiq(circ, conversion_type)
                for circ in unique_circuits
            ]

        if not self._can_batch:
            for circuit in to_run:
                self._call_executor(circuit, **kwargs)

        else:
            if self._max_batch_size >= len(to_run):
                self._call_executor(to_run, **kwargs)

            else:
                stop = len(to_run)
                step = self._max_batch_size
                for i in range(int(np.ceil(stop / step))):
                    batch = to_run[i * step: (i + 1) * step]
                    self._call_executor(batch, **kwargs)

        # Expand computed results to all results using counts.
        if force_run_all:
            return self._computed_results

        # Build a dictionary for the expectation values
        expval_dict = dict(zip(unique_circuits, self._computed_results))
        results = [expval_dict[key] for key in frozen_circuits]

        return results
```

**Problem:**  There are other tests that fail (those related to the `pec_error`) since this error does not scale as expected when we collect equal circuits. My suggestion would be to temporarily ignore the tests related to `pec_error`. We can restore them as soon as we are able to scale the number of shots of the user executor proportionally to the count of equal circuits. Alternatively we can set `force_run_all = True` until we fix this issue.

rmlarose:
Thanks @andreamari for noting the ordering of results, this is important and wasn't captured by the current tests.

Let me think about `freeze` and `unfreeze` more. The motivation for

```python
circuits = [
  cirq.FrozenCircuit(
      convert_to_mitiq(circ)[0].all_operations(),
      strategy=cirq.InsertStrategy.EARLIEST,
  )
  for circ in circuits
]
```

was that it removes moment structure so that, e.g.,

```
0: ───H───

1: ───H───
```

and 

```
0: ───H───────

1: ───────H───
```

are considered equal.

andreamari:
> Thanks @andreamari for noting the ordering of results, this is important and wasn't captured by the current tests.
> 
> Let me think about `freeze` and `unfreeze` more. The motivation for
> 
> ```python
> circuits = [
>   cirq.FrozenCircuit(
>       convert_to_mitiq(circ)[0].all_operations(),
>       strategy=cirq.InsertStrategy.EARLIEST,
>   )
>   for circ in circuits
> ]
> ```
> 
> was that it removes moment structure so that, e.g.,
> 
> ```
> 0: ───H───
> 
> 1: ───H───
> ```
> 
> and
> 
> ```
> 0: ───H───────
> 
> 1: ───────H───
> ```
> 
> are considered equal.

I agree, but do we want to "secretly" squash all the moment structure when wrapping the executor?
I would lean towards not squashing at the risk of having a less efficient collection.
If we decide that squashing is important, maybe we can make it more explicit e.g. by adding an early squashing step in the run function. This would avoid strange discrepancies between the case `force_run_all == True` and `force_run_all == False`.

rmlarose:
Ready for another round of review @andreamari @willzeng, PTAL.

@andreamari I'm not sure I understand what the disadvantages of squashing moments are, but this concerns checking equality which we agreed to address in #484. So I used your suggestion here but let's revisit in #484 (or here).

mitiq/conversions.py
```diff
@@ -96,10 +104,10 @@ def convert_from_mitiq(circuit: Circuit, conversion_type: str) -> QPROGRAM:
         from mitiq.mitiq_pyquil.conversions import to_pyquil
 
         conversion_function = to_pyquil
-    elif isinstance(circuit, Circuit):
+    elif conversion_type == "cirq":
 
         def conversion_function(circ: Circuit) -> Circuit:
-            return circ
+            return Circuit(circ)
```
andreamari, line 110:
```suggestion
            return circ
```
Here we could return just `circ` if we unfreeze before conversions. Alternatively `Circuit(circ, device=circ.device)` is perhaps safer than `Circuit(circ)`.

andreamari (approved):
Thanks @rmlarose! LGTM. See minor comments.

PR #459, opened 2020-11-30
Add warning for large samples in PEC

### LINKED ISSUE #454: Raise a warning if number of samples in PEC is too large.
Issue Description
-----------------
If the `precision` option is too small or if the one-norm of the circuit quasi-distribution is too large, the number of samples in `execute_with_pec` can become huge.

Proposed Solution
-----------------
Two options:

- Raise a warning if `num_samples` is beyond a certain threshold in such a way that the user is aware that the computation may be very long.

- Automatically cut `num_samples` to some default `max_samples` and raise a warning about this truncation.

sid1993:
Fixes #454

[More details of the fix.](https://github.com/unitaryfund/mitiq/issues/454#issuecomment-735456307)

- Added test for large sample in `test_pec.py`

- Tested UserWarning by removing its filter in `pytest.ini`.
warning

### DIFF
--- a/mitiq/pec/pec.py
+++ b/mitiq/pec/pec.py
@@ -17,10 +17,23 @@
 
 from typing import Optional, Callable, Union, Tuple
 import numpy as np
-from warnings import warn
+import warnings
 from mitiq._typing import QPROGRAM
 from mitiq.pec.utils import DecompositionDict
 from mitiq.pec.sampling import sample_circuit
+
+
+class LargeSampleWarning(Warning):
+    """Warning is raised when PEC sample size is greater than 10 ** 5
+    """
+
+    pass
+
+
+_LARGE_SAMPLE_WARN = (
+    "The number of PEC samples is very large. It may take several minutes."
+    " It may be necessary to reduce 'precision' or 'num_samples'."
+)
 
 
 def execute_with_pec(
@@ -105,10 +118,7 @@ def execute_with_pec(
 
     # Issue warning for very large sample size
     if num_samples > 10 ** 5:
-        warn(
-            "Sample size in PEC is very large.It may take serveral mintues.",
-            UserWarning
-            )
+        warnings.warn(_LARGE_SAMPLE_WARN, LargeSampleWarning)
 
     sampled_circuits = []
     signs = []
--- a/mitiq/pec/pec.py
+++ b/mitiq/pec/pec.py
@@ -17,6 +17,7 @@
 
 from typing import Optional, Callable, Union, Tuple
 import numpy as np
+from warnings import warn
 from mitiq._typing import QPROGRAM
 from mitiq.pec.utils import DecompositionDict
 from mitiq.pec.sampling import sample_circuit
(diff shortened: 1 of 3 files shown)


mitiq/pec/tests/test_pec.py
```diff
@@ -110,7 +110,7 @@ def test_execute_with_pec_with_different_samples(circuit: Circuit, seed):
     assert np.average(errors_more_samples) < np.average(errors_few_samples)
 
 
-@mark.parametrize("num_samples", [100, 1000])
+@mark.parametrize("num_samples", [100, 1000, 100001])
```
andreamari, line 113: Instead of changing an existing test, I would add a new test dedicated to the new warning.
There is a specific way to test that warnings, for example you can have a look at this similar test: https://github.com/unitaryfund/mitiq/blob/e96e8a1ce50a9a70f215cd35c3ce24957a60e0d7/mitiq/zne/tests/test_inference.py#L605

mitiq/pec/pec.py
```diff
@@ -102,6 +103,13 @@ def execute_with_pec(
     if not isinstance(num_samples, int):
         num_samples = int((norm / precision) ** 2)
 
+    # Issue warning for very large sample size
+    if num_samples > 10 ** 5:
+        warn(
+            "Sample size in PEC is very large.It may take serveral mintues.",
```
andreamari, line 109:
```suggestion
            "The number of PEC samples is very large. It may take several minutes."
            " It may be necessary to reduce 'precision' or 'num_samples'.",
```
We could add this suggestion for the user. 

I apologize if this change creates flake8 problems. If so, they should be easy to fix.

andreamari (commented):
Thanks @sid1993 !
I left a couple of suggestions.
Feel free to also add a new line on the CHANGELOG.md file about this PR.

sid1993:
- Updated CHANGELOG.md
- Successfully tested warning for large samples in PEC
Success_Warning
- Failure tests for no warning and incorrect warning category/message :
Failure_Test_Wrong_Category
Failure_Incorrect_Message

- Added only one test case in `test_large_sample_size_warning` as each of them is increasing the overall testing time by 1.5-2 mins. If more test cases are required, I would suggest to decrease the warning threshold to `10 ** 4`.

andreamari:
@sid1993 Thanks!
I am a bit concerned about using a 2 min test just for the warning. Is there a way to test the warning as you did but abort the execution after the warning is raised? Maybe with some timeout function or any other method?

sid1993:
> @sid1993 Thanks!
> I am a bit concerned about using a 2 min test just for the warning. Is there a way to test the warning as you did but abort the execution after the warning is raised? Maybe with some timeout function or any other method?

@andreamari : Yes, I am trying to figure out a way and will update you soon.

sid1993:
> > @sid1993 Thanks!
> > I am a bit concerned about using a 2 min test just for the warning. Is there a way to test the warning as you did but abort the execution after the warning is raised? Maybe with some timeout function or any other method?
> 
> @andreamari : Yes, I am trying to figure out a way and will update you soon.

@andreamari : Funny story :) I had multiple VMs up on my system when i tested it and that led to incorrect results. Test time is fine just  5-6 secs more.
Overall_Test_Time

sid1993:
@andreamari : I have added another test case also. Everything looks good.

andreamari (approved):
Great! Thanks @sid1993 for your nice contribution!

crazy4pi314:
@all-contributors please add @sid1993 for code, bug

Issue #447, opened 2020-11-20
PEC test failing on master


rmlarose:
Another non-deterministic failure. I think seeding is the fix for this.

```
================================== FAILURES ===================================
________ test_execute_with_pec_one_qubit[decomposition_dict1-circuit0] ________

circuit = cirq.Circuit([
    cirq.Moment(
        cirq.Z(cirq.LineQubit(1)),
    ),
    cirq.Moment(
        cirq.Z(cirq.LineQubit(1)),
    ),
])
decomposition_dict = {cirq.X(cirq.LineQubit(0)): [(1.0205479452054795, [cirq.X(cirq.LineQubit(0))]), (-0.006849315068493151, [cirq.I(cirq.L...))]), (-0.006849315068493151, [cirq.Z(cirq.LineQubit(1))]), (-0.006849315068493151, [cirq.Y(cirq.LineQubit(1))])], ...}

    @mark.parametrize("circuit", [oneq_circ, twoq_circ])
    @mark.parametrize(
        "decomposition_dict", [NOISELESS_DECO_DICT, DECO_DICT_SIMP, DECO_DICT]
    )
    def test_execute_with_pec_one_qubit(
        circuit: Circuit, decomposition_dict: DecompositionDict
    ):
        """Tests that execute_with_pec mitigates the error of a noisy
        expectation value.
        """
        unmitigated = executor(circuit)
        mitigated = execute_with_pec(
            circuit, executor, decomposition_dict=decomposition_dict
        )
        error_unmitigated = abs(unmitigated - 1.0)
        error_mitigated = abs(mitigated - 1.0)
        # For a trivial noiseless decomposition no PEC mitigation should happen
        if decomposition_dict == NOISELESS_DECO_DICT:
            assert np.isclose(unmitigated, mitigated)
        else:
>           assert error_mitigated < error_unmitigated
E           assert 0.026439811967908122 < 0.026311278343200684

mitiq\pec\tests\test_pec.py:80: AssertionError
```

rmlarose:
Actually seeding doesn't help and this test is quite brittle. @andreamari can you take a look?

andreamari:
Hi @rmlarose:
The fact that without seed we get non-deterministic failures is probably ok (the default number of sample is not very large).
The fact that the seed doesn't seem to help is likely related to this bug #452.
Once we solve #452 we should be able to correctly seed this test.

rmlarose:
[...]

PR #432, opened 2020-11-12
Add seed to PEC functions

rmlarose:
Fixes #431 and fixes #430. See #431 for explanation/justification.

PR #460, opened 2020-12-03
Seed a PEC test to avoid non-deterministic failures

andreamari:
Description
-----------
There is a test in `mitiq/pec/tests/test_pec.py` which is non-deterministic and sometimes it fails. This was also noticed in #447.
Other tests have already been seeded by @rmlarose in #432.

Changes 
-----------
- Seed the non-deterministic test in order to make it deterministic


andreamari:
One reviewer is enough for this, whoever comes first :-)

rmlarose (approved):
Thanks @andreamari, LGTM.

PR #464, opened 2020-12-04
Add a "plot()" method to the Factory class.

crazy4pi314:
Description
-----------
Adds `plot_data` and `plot_fit` as methods to all factories, as noted in #333.

**Example of `plot_data**
image

**Example of `plot_fit**
image

Checklist
-----------


mitiq/zne/tests/test_inference.py
```diff
@@ -872,6 +872,86 @@ def test_get_methods_of_factories():
...
+        fac = factory(scale_factors=X_VALS)
+    fac.run_classical(apply_seed_to_func(f_lin, SEED))
+    fac.reduce()
+    fig = fac.plot_data()
+    ax = fig.gca()
+    x_data, y_data = ax.lines[0].get_xydata().T
+    zne_curve = fac.get_extrapolation_curve()
+    np.testing.assert_equal(
+        np.any(np.not_equal(y_data, zne_curve(x_data))), True,
+    )
```
andreamari, line 892: If I understand well, I think that here we want to test that `y_data` is equal to `zne_curve(x_data)` up to numerical rounding errors.

Since testing perfect equality with real elements is problematic, what about replacing the last instruction with:
`assert np.allclose(y_data, zne_curve(x_data))` ?

Side comment: I was also confused by
```
np.testing.assert_equal(
        np.any(np.not_equal(y_data, zne_curve(x_data))), True,
    )
```
since I expected:
```
    zne_curve = fac.get_extrapolation_curve()
    np.testing.assert_equal(
        np.any(np.not_equal(y_data, zne_curve(x_data))), False,
    )
```

But, anyway, using `np.allclose...` is probably optimal.

andreamari (changes requested):
Thanks @crazy4pi314, LGTM ! Up to minor doubts (see inline comments).

mitiq/zne/inference.py
```diff
@@ -335,6 +336,55 @@ def push(
...
+            self.get_expectation_values(),
+            "o",
+            markersize=10,
+            markeredgecolor="black",
+            alpha=0.8,
+            label="Data",
+        )
+        ax.grid(True)
+        plt.xlabel("Noise scale factor")
+        plt.ylabel("Observable value")
```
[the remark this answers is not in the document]
crazy4pi314: Fine with me, I guess we assume that the executor returns an expectation value (not that we can enforce it), so I was hedging.

mitiq/zne/tests/test_inference.py
crazy4pi314: Yeah I agree, I have swapped to using the `np.allclose` as per your suggestion.

andreamari:
Thanks, everything LGTM!

- We should not forget to show this plotting feature in the docs! We do it here or open a new issue.

- Remember the change log :-)

andreamari (approved):
Thanks @crazy4pi314 !

nathanshammah:
This is great, thanks @crazy4pi314 and @elmandouh!

PR #470, opened 2020-12-07
Removes duplication of the reduce method in every (non-adaptive) factory

### LINKED ISSUE #455: Code duplication in factories
The `reduce` method is identically re-defined in every instance of a batched factory. See

- https://github.com/unitaryfund/mitiq/blob/master/mitiq/zne/inference.py#L807
- https://github.com/unitaryfund/mitiq/blob/master/mitiq/zne/inference.py#L901
- https://github.com/unitaryfund/mitiq/blob/master/mitiq/zne/inference.py#L993
- https://github.com/unitaryfund/mitiq/blob/master/mitiq/zne/inference.py#L1136
- https://github.com/unitaryfund/mitiq/blob/master/mitiq/zne/inference.py#L1462

The only unique part is the `extrapolate` method, so this should be the `abstractmethod` in batched factory and `reduce` should a method of batched factory.

elmandouh:
Fixes #455 

Description
-----------
Instead of redefining the `reduce` method in every instance of a batched factory, we instead move `reduce` to be a method of `BatchedFactory` and extrapolate an `abstractmethod` of `BatchedFactory`. Not all instances of the `extrapolate` method take the same arguments. So as per @rmlarose's suggestion, we define a method `_extrapolate_args` for each factory that returns a dictionary with the extra arguments to be passed to `extrapolate`.


rmlarose (commented):
Thanks @elmandouh. I like this to reduce code duplication, but it also makes defining custom factories more difficult (which I didn't consider before), as evidenced by the two additional methods in the clipped linear factory example in `guide-zne`. (Actually, in this example `reduce` should be `extrapolate` - there's only two methods that need to be defined.)

Because of the current design of factory objects, I'm not sure if there there's a win-win situation here with respect to reduced code duplication and ease of defining custom factories. I think the way to do this is decoupling data collection from data fitting as outlined [here](https://github.com/unitaryfund/mitiq/issues/412#issuecomment-734535461), but this received little support. As we see from the `abstractmethod`s, the defining characteristic of a factory is the `extrapolate` method, and the `reduce` or `_extrapolate_args` methods are essentially band-aids to funnel factory attributes into `extrapolate`.

Anyway, this will be good to discuss at the Mitiq meeting Friday. To summarize my thoughts for now, there's not a clear advantage of this because of the less intuitive way of defining custom factories, although I do like the reduced code duplication! (Other thoughts @andreamari?)

andreamari:
I had the same doubt of @rmlarose, the code looks much cleaner but is less intuitive, especially for an external contributor/user. 

Maybe we could directly replace all the current factory options like `self.order`, `self.asymptote`, etc.,  with a unique dictionary `self._options` which is created in the body of   `__init__` . 
Such that:
`self.order` ----becomes ---> self._options[`order`]
`self.asymptote' ----becomes ---> self._options[`asymptote`]
etc...

With this approach would avoid the re-definition of reduce for each factory without introducing a new method (_extrapolate_args).

rmlarose:
It's not a bad idea @andreamari. If we go this route, `extrapolate` would no longer be a static method, and I think `reduce` and `extrapolate` would be identical. If so, we may have come full circle to the original reason for introducing `extrapolate` as a static method - namely, it's nice to be able to use the method outside of the factory. For instance, if you use "lower level" tools to manually scale the circuit and want to manually fit the data without populating the history of a factory.

andreamari:
> It's not a bad idea @andreamari. If we go this route, `extrapolate` would no longer be a static method, and I think `reduce` and `extrapolate` would be identical. If so, we may have come full circle to the original reason for introducing `extrapolate` as a static method - namely, it's nice to be able to use the method outside of the factory. For instance, if you use "lower level" tools to manually scale the circuit and want to manually fit the data without populating the history of a factory.

I think `extrapolate` can remain exactly as usual: static and with all the explicit arguments.
At the same time, `reduce` can be defined only once in the base `Factory` class. I.e., almost copying from this PR:
```
def reduce(self) -> float:
        """Evaluates the zero-noise limit found by fitting according to
        the factory's extrapolation method.
        Returns:
            The zero-noise limit.
        """
        (
            self._zne_limit,
            self._zne_error,
            self._opt_params,
            self._params_cov,
            self._zne_curve,
        ) = self.extrapolate(  # type: ignore
            self.get_scale_factors(),
            self.get_expectation_values(),
            full_output=True,
            **self._options,
        )

        self._already_reduced = True
        return self._zne_limit
```

If a new Factory is class is defined, it is sufficient to override only the static `extrapolate` method and, if necessary, extend  __init__ to add new keys in `self._options `.

rmlarose:
Ah, ok. So instead of being required to define `_extrapolate_args`, one instead extends `__init__`. This seems better to me.

elmandouh:
@rmlarose @andreamari Thanks for the feedback, I will try to implement @andreamari's suggestion.

rmlarose (commented):
Thanks @elmandouh, LGTM once tests are fixed.

andreamari:
@elmandouh thanks for this PR! 
I just realized that in the docs say this:

```
Defining a custom factory
If necessary, the user can modify an existing extrapolation methods by subclassing one of the :ref:`built-in factories <built-in-factories>`.

Alternatively, a new adaptive extrapolation method can be derived from the abstract class :class:`.Factory`. In this case its core methods must be implemented: self.next, self.push, self.is_converged, self.reduce, etc. Typically, the self.__init__ method must be overridden.

A new non-adaptive method can instead be derived from the abstract :class:`.BatchedFactory` class. In this case it is usually sufficient to override only the self.__init__ and the self.reduce methods, which are responsible for the initialization and for the final zero-noise extrapolation, respectively.  
```

Everything is still consistent with the new approach but we should replace `self.reduce`  with `self.extrapolate`. Note that there are two `self.reduce` in the text above. 

For the rest LGTM!

elmandouh:
@andreamari Thanks for noticing that, I updated the docs.

docs/source/guide/guide-zne.rst
```diff
@@ -637,7 +637,7 @@ Typically, the ``self.__init__`` method must be overridden.
 
 A new non-adaptive method can instead be derived from the abstract :class:`.BatchedFactory` class.
 In this case it is usually sufficient to override only the ``self.__init__`` and
-the ``self.reduce`` methods, which are responsible for the initialization and for the
+the ``self.extrapolate`` methods, which are responsible for the initialization and for the
```
andreamari, line 640: 5 lines above there is also this line:
```
``self.next``, ``self.push``, ``self.is_converged``, ``self.reduce``, etc. , 
```
which should become:
```
``self.next``, ``self.push``, ``self.is_converged``, ``self.extrapolate``, etc. , 
```

andreamari (approved):
LGTM! (See small suggestion)

docs/source/guide/guide-zne.rst
elmandouh: The adaptive factory was not modified, so should it not stay `self.reduce`?

andreamari: Good point. Your are right!

docs/source/guide/guide-zne.rst
```diff
@@ -682,7 +682,13 @@ and clips the result if it falls outside its physical domain.
          self.min_expval = min_expval
          self.max_expval = max_expval
 
-      def reduce(self) -> float:
+      def extrapolate(
+         _,
```
rmlarose, line 686: What's the reason for this?

elmandouh: This can be avoided by adding the decorator `@staticmethod`. Not sure what the clearest/easiest way this should be handled.

rmlarose (approved):

rmlarose:
Thanks @elmandouh! LGTM. I botched the merge with master but opened a PR on your branch to patch it (https://github.com/elmandouh/mitiq/pull/1). Once that is merged this PR is good to go. (Or you can remove the definition of `_is_batched_executor` in inference.py - this is the patch - whatever is easier.)

PR #476, opened 2020-12-09
Add tests for plot_data() and plot_fit()

elmandouh:
Description
-----------
This adds tests for `plot_data` and `plot_fit` and fixes tiny lint error.


elmandouh:
@crazy4pi314 I have added some tests for the plot methods. I hope this is the appropriate to contributing without write access.

crazy4pi314:
This looks great, thanks!! 🥳 I don't know the best way to make sure the test doesn't fail due to a fit fail on SciPy's part, but reducing the `PolyExpFactory` order makes it a bit less likely.

PR #482, opened 2020-12-15
Corrects typo of CnotGate in qiskit.extensions

purva-thakre:
Description
-----------
Replaced `CnotGate` by `CXGate` in `pytest.mark.parametrize` for `qiskit.extensions`. There is no attribute called `CnotGate` in `qiskit.extensions`.


purva-thakre:
I think there's a typo that's causing the error. I am not sure what it is. What I thought was a typo is not causing the error. Closing the pull request until I figure it out.

PR #485, opened 2021-01-03
Qiskit package installations

purva-thakre:
Description
-----------
Fixes # 396 (partially)

- Edited `requirements.txt` with information about `qiskit-aer` and `qiskit-ibmq-provider`. Note that `qiskit-terra` is a dependency for `aer` i.e. does not need to be specified. 
- Due to above changes, deprecated `CnotGate` was changed to `CXGate` and deprecated `n_qubits` was changed to `num_qubits`. For more information, see `Table 2 Gate name changes` and `Table 3 new methods` at [this link](https://qiskit.org/documentation/release_notes.html). 

Problem with py36 unit test
-----------
When py36 unit test is run, `pip install -r requirements.txt` gets stuck with [dependency resolution backtracking](https://pip.pypa.io/en/latest/user_guide/#dependency-resolution-backtracking).  
This seems to be an issue with `pip` dependency resolver which was not resolved even when better constraints are specified like `qiskit-terra==0.17.0`. A possible fix could be specifying all dependencies in a `constraint.txt` file. 

Checklist
-----------

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -5,7 +5,7 @@ cirq==0.10.0.dev20201203012958
 # third-party integration
 pyquil~=2.18.0
 qiskit-aer>0.7
-qiskit-ibmq-provider>0.11
+qiskit-terra>0.17
 
 # to test the code
 pytest~=5.4.1
--- a/requirements.txt
+++ b/requirements.txt
@@ -4,7 +4,8 @@ cirq==0.10.0.dev20201203012958
 
 # third-party integration
 pyquil~=2.18.0
-qiskit~=0.16.2
+qiskit-aer>0.7
+qiskit-ibmq-provider>0.11
 
 # to test the code
 pytest~=5.4.1
--- a/requirements.txt
+++ b/requirements.txt
@@ -5,7 +5,7 @@ cirq==0.10.0.dev20201203012958
 # third-party integration
 pyquil~=2.18.0
 qiskit-aer>0.7
-qiskit-terra>0.17
+qiskit-ibmq-provider>0.11
 
 # to test the code
 pytest~=5.4.1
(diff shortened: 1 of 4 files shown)


requirements.txt
```diff
@@ -4,7 +4,8 @@ cirq==0.10.0.dev20201203012958
 
 # third-party integration
 pyquil~=2.18.0
-qiskit~=0.16.2
+qiskit-aer>0.7
+qiskit-ibmq-provider>0.11
```
rmlarose, line 8: Where is this required?

rmlarose: Found it! Without ibmq-provider a runtime warning is raised, but tests still pass:

```text
Document: guide/guide-qaoa-landscape
------------------------------------
.../python3.6/site-packages/qiskit/__init__.py:75: RuntimeWarning: Could not import the IBMQ provider from the qiskit-ibmq-provider package. Install qiskit-ibmq-provider or check your installation.
  RuntimeWarning)

```

rmlarose: I managed to get a working version in #487 which only installs qiskit-aer and qiskit-terra. I'm actually not sure why [guide-ibmq-backends](https://github.com/unitaryfund/mitiq/blob/master/docs/source/guide/guide-ibmq-backends.rst) passes though because this uses the ibmq provider.

purva-thakre: So, I guess the dependency resolver is trying to resolve the dependency between `aer` and `ibmq-provider` when it downloads all the versions of `terra`. 

Also, I think ` guide/guide-ibmq-backends` passes because there is no test being run for it. The next step skips to doctest summary. https://github.com/unitaryfund/mitiq/runs/1644604484?check_suite_focus=true#step:5:283

rmlarose: Thanks @purva-thakre. Its unfortunate the dependency resolver is failing here. To me, it's not a big deal to just make qiskit the requirement in light of this. This is more favorable than adding more complexity with constraints, etc. So, unless there is a simple solution (and you're willing to keep searching!) I would vote for keeping qiskit as the requirement at this point.

purva-thakre: I have been thinking of putting this on hold until a better version of dependency resolver is released. I tried specifying the dependencies exactly via `==` but the dependency resolver was still backtracking across all versions of `terra`. I don't think specifying the constraints in `constraint.txt` file will work since my plan was to use `==` when pip installed via `requirements.txt`.

rmlarose: This sounds good to me.

PR #486, opened 2021-01-04
Split requirements into core and development requirements

### LINKED ISSUE #475: Get install_requires in setup from requirements.txt
As noted in [setup.py](https://github.com/unitaryfund/mitiq/blob/master/setup.py#L36-L40), argument `install_requires` should be a subset of `requirements.txt`, but is defined independently. This should be pulled from [requirements.txt](https://github.com/unitaryfund/mitiq/blob/master/requirements.txt#L1-L3) to be more robust. (So that, e.g., if we need to bump the Cirq dependency, we only need to do it in one place.)

rmlarose:
Fixes #475.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -40,7 +40,7 @@ format:
 
 .PHONY: install
 install:
-	pip install -e .
+	pip install -e .[development]
 
 .PHONY: pdf
 pdf:
--- a/dev_requirements.txt
+++ b/dev_requirements.txt
@@ -1,0 +1,24 @@
+# third-party integration
+pyquil~=2.18.0
+qiskit~=0.16.2
+
+# to test the code
+pytest~=5.4.1
+pytest-cov~=2.8.1
+flake8~=3.7.9
+black~=19.10b0
+mypy~=0.782
+
+# to run examples
+matplotlib~=3.2.1
+notebook~=6.1.5
+
+# to build and test the documentation
+# see docs/requirements.txt for more non-essential dependencies
+sphinx~=2.4.4
+sphinxcontrib-bibtex~=1.0.0
+sphinx-copybutton~=0.2.11
+sphinx-autodoc-typehints~=1.10.3
+m2r~=0.2.1
+pybtex~=0.22.2
+pydata-sphinx-theme~=0.3.1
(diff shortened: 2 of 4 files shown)


willzeng:
This looks good to me. I want to check with @nathanshammah though as I think we previous had split requirements that we then combined. @nathanshammah is there something we're missing about them being split? If you're good with it then it looks good to me.

nathanshammah (approved):
Thanks @rmlarose and @willzeng. The change looks fine to me. The dev requirements were removed in https://github.com/unitaryfund/mitiq/pull/258 by @karalekas, there were some comments in that discussion about the different rationale that one could use ([comment](https://github.com/unitaryfund/mitiq/pull/258#issuecomment-662734783) by @rmlarose); summarizing: `github` fork being used by developers actually making a PR or by "devs" that like to be on bleeding edge but will likely just use Mitiq as is, at first. I have no strong opinion, I am just ignorant on whether the names have any impact, especially w.r.t. PyPI, e.g., `dev_requirements.txt` or `development_requirements.txt`. I guess we can check with TestPyPI at release time.

karalekas (commented):
I personally dislike the two requirements files, and I think it sort of goes against the spirit of `requirements.txt` for it to not actually include all the requirements to develop a repository, but ofc at the end of the day it's a convention. According to the Python packaging docs, the `install_requires` and `requirements.txt` are different things (read more about best practices [here](https://packaging.python.org/discussions/install-requires-vs-requirements/)). One thing that I've wanted to try that could alleviate this double definition (though it deviates somewhat from the best practices) is just to put `.` in the `requirements.txt`, which should install what is defined in `install_requires`.

rmlarose:
I'll put this up for discussion at the Mitiq meeting for a final decision. I would favor a convention without multiply defined requirements - this just seems logical to me. A standard way I've seen this done is `requirements` and `dev_requirements` as done here.

If we want to really emphasize the distinction between `install_requires` and `requirements` as Peter notes, perhaps omitting `requirements` altogether and just defining them in `install_requires` is an option (keeping `dev_requirements`).

rmlarose:
Decided to merge after Mitiq meeting, but since @karalekas wasn't there to voice his opinion I will leave time to allow for trial by combat or other refutations :-)

PR #488, opened 2021-01-04
Fix IBMQ guide

rmlarose:
There were several issues in the IBMQ guide due to doctest not actually running the code. This PR fixes them, ensures the code is tested, and adds a separate doctest target to the Makefile. 

Thanks to @purva-thakre for pointing this out in https://github.com/unitaryfund/mitiq/pull/485#discussion_r551486468!

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -32,6 +32,9 @@ dist:
 .PHONY: docs
 docs:
 	make -C docs html
+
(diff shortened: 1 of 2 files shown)


willzeng (approved):
Great catch! Thanks @rmlarose and @purva-thakre

PR #500, opened 2021-01-18
Multi-platform PEC

### LINKED ISSUE #440: Handle conversions in getting a decomposition dictionary from an assumed noise model
The signature for getting a decomposition assuming depolarizing noise (the only example we have implemented so far) is

```python
def depolarizing_decomposition(
    ideal_operation: Operation, noise_level: float
) -> List[Tuple[float, List[Operation]]]:
```

This assumes a `cirq.Operation`, but should be generalized.

## Options

1. Define a `QOPERATION` type in Mitiq, analogous to `QPROGRAM`. The `ideal_operation` argument would then be a `QOPERATION`, and the function would

- Convert the `QOPERATION` to a `cirq.Operation`.
- Get the decomposition. (What the function currently does.)
- Convert back to the input type of the `QOPERATION`, e.g., `Qiskit`, `pyQuil`, or `Cirq`.

The first and last steps could be handled with a conversions decorator in an analogous way to the [`converter`](https://github.com/unitaryfund/mitiq/blob/master/mitiq/conversions.py#L121) for unitary folding. 

The signature in this case would be something like:

```python
@mitiq.conversions.op_converter
def depolarizing_decomposition(
    ideal_operation: QOPERATION, noise_level: float
) -> List[Tuple[float, List[QOPERATION]]]:
```

2. The same as (1), but we force the type of `ideal_operation` to be a `QPROGRAM`. This makes it so that we can use the `converter` directly here. Note that the output is a list of operations (i.e., a circuit), so it doesn't seem unreasonable to assume the input is a circuit also.

The signature in this case would be

```python
@mitiq.conversions.converter
def depolarizing_decomposition(
    ideal_operation: QPROGRAM, noise_level: float
) -> List[Tuple[float, QPROGRAM]]:
```

## Comments

I'm currently in favor of (2) but would like to hear feedback.

### LINKED ISSUE #441: Handle conversions in execute with PEC
The signature for `execute_with_pec` allows for `QPROGRAM`s but assumes `cirq.Circuit`s for the `QPROGRAM` and assumes `cirq.Operation`s as keys for the decomposition dictionary.

```python
def execute_with_pec(
    circuit: QPROGRAM,
    executor: Callable[[QPROGRAM], float],
    decomposition_dict: DecompositionDict,
    num_samples: Optional[int] = None,
    full_output: bool = False,
) -> Union[float, Tuple[float, float]]:
```

This should be generalized to support `QPROGRAM`, and the `DecompositionDict` should be generalized similarly.

Related issues: #440 and #369.

rmlarose:
Fixes #440 and fixes #441.


rmlarose:
@andreamari although its still pretty ugly it may be worth a quick "pre-review." Most functions at this point have been overhauled to support `QPROGRAM`. But there are still tests that need to be re-written, some old functions that need to be deleted, etc.

andreamari:
Hi @rmlarose, I am a bit afraid of the many conversions happening inside each function but this seems to work.

An easier but less general solution would be to immediately convert the input circuit to Cirq after calling `execute_with_pec`, assume sampling functions are only compatible with Cirq operations and circuits, and then convert back to user type just before the final execution.

However I am not against your approach. In principle this is more powerful and could allow exposing sampling functions to the user instead of keeping them private.

Let me know if you need help with tests or anything else.

rmlarose:
Thanks @andreamari, good feedback.

> An easier but less general solution would be to immediately convert the input circuit to Cirq after calling execute_with_pec, assume sampling functions are only compatible with Cirq operations and circuits, and then convert back to user type just before the final execution.

We can achieve both of these. I added an initial conversion in `execute_with_pec` - after this the sampling functions merely check the circuit type but don't convert, and sampling functions are still for general `QPROGRAM`s. 

At this point all the PEC tests are passing with the new framework. Some old code needs to be phased out, and I think a few tests need to be re-written or phased out. Also there needs to be more tests with Qiskit/pyQuil, but it's already a nontrivial change so PTAL.

andreamari:
Thanks @rmlarose this PR is great! I have nothing to comment on the new code, everything seems ok to me. 

-  About `#Outdated functions`: I agree that we can get rid of them as soon as they are not used in tests anymore.
-  About old tests for sampling functions based on Choi matrices. I think we need to re-write them for the new sampling functions. Testing the equivalence of Choi matrices is the only way to be sure that sampling works as it should. If you want I can take care of this in #502, which is related to a similar issue.
- In a recent engineering meeting we decided for the new notation in which "_decomposition_" should be replaced everywhere with "_representation_".  Maybe this is a good moment to make this change since the more we wait the harder will be to fix.

rmlarose:
Thanks Andrea. I will complete the remaining items in this list modulo the Choi tests - thanks for opening another PR for this.

> * About `#Outdated functions`: I agree that we can get rid of them as soon as they are not used in tests anymore.
> * About old tests for sampling functions based on Choi matrices. I think we need to re-write them for the new sampling functions. Testing the equivalence of Choi matrices is the only way to be sure that sampling works as it should. If you want I can take care of this in #502, which is related to a similar issue.
> * In a recent engineering meeting we decided for the new notation in which "_decomposition_" should be replaced everywhere with "_representation_".  Maybe this is a good moment to make this change since the more we wait the harder will be to fix.

rmlarose:
PTAL @andreamari. Coverage can be ignored.

One note worth mentioning: Right now the [code for finding the appropriate representation](https://github.com/unitaryfund/mitiq/pull/500/files#diff-f1e26944ec77e862ce248eae681058d9579b389163cc744873c17092183ef2dcR67) is qubit-dependent. So for example if the circuit is `0: ───X───` but the representation is for `1: ───X───`, an error will be raised. (See [here](https://github.com/unitaryfund/mitiq/pull/500/files#diff-b6e9d19af3899b9d75991d0b7e7324df6b92fb9541abb2f067300d39ad90ab1aR222).) I think both options (qubit-dependent and qubit-independent) should be supported eventually, but for now the easiest thing was to require qubit equality.

andreamari:
> PTAL @andreamari. Coverage can be ignored.
> 
> One note worth mentioning: Right now the [code for finding the appropriate representation](https://github.com/unitaryfund/mitiq/pull/500/files#diff-f1e26944ec77e862ce248eae681058d9579b389163cc744873c17092183ef2dcR67) is qubit-dependent. So for example if the circuit is `0: ───X───` but the representation is for `1: ───X───`, an error will be raised. (See [here](https://github.com/unitaryfund/mitiq/pull/500/files#diff-b6e9d19af3899b9d75991d0b7e7324df6b92fb9541abb2f067300d39ad90ab1aR222).) I think both options (qubit-dependent and qubit-independent) should be supported eventually, but for now the easiest thing was to require qubit equality.

I agree! The most flexible solution could be having a `qubit_independent` flag on each `OperationRepresentation` such that when set to True, the representation becomes valid for all qubits.

andreamari (approved):

rmlarose:
Pinging @willzeng and/or @karalekas before merging, let me know if you have further comments.

willzeng:
Very excited about this! Nice work @rmlarose and @andreamari

rmlarose:
Added some multi-qubit pyquil/qiskit tests for my own sanity. PEC with qiskit is notably slower than pyquil/cirq, probably because of conversions. But hey, we can always open another issue.

Issue #508, opened 2021-01-21
Questions about our requirements


andreamari:
I have 3 questions related to our requirements:

1. Shall we remove `numpy` and `scipy`  given they are already required by `cirq` ?
2. Shall we remove `matplotlib`  from our dev. requirements for the same reason?
3. All dev. requirements are pinned. Is this a good thing? For example, shouldn't we test that Mitiq works properly with the last stable version of `qiskit` and `pyquil`?

nathanshammah:
Hi Andrea, regarding 3., I think it is recommended to have the versions pinned, for reproducibility, and upgrade them when needed. We could use a tool like Dependabot to upgrade them, https://github.blog/2020-06-01-keep-all-your-packages-up-to-date-with-dependabot/.

Regading 1 and 2 I have no clear answer, but I guess that maybe keeping them is fine as we'd like to point out that we are explicitly calling out those packages for some features (scipy for fitting and matplotlib will be called by #464). It

crazy4pi314:
I took a stab at adding a simple dependabot action that each day checks if there are updates from pip available for any of the requirements.

andreamari:
Thanks everyone for your answers! Now all is clear to me.

crazy4pi314:
I think we can close this now, feel free to re-open if there is something else to discuss!

PR #510, opened 2021-01-21
Create dependabot alerts for requirement updates

crazy4pi314:
Description
-----------
Adds a Github Action that has Dependabot check once a day if there are any updates available for any of our requirements files. See discussion in #508 

Checklist
-----------

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,19 @@
+# To get started with Dependabot version updates, you'll need to specify which
+# package ecosystems to update and where the package manifests are located.
+# Please see the documentation for all configuration options:
+# https://help.github.com/github/administering-a-repository/configuration-options-for-dependency-updates
+
+version: 2
+updates:
+  - package-ecosystem: "pip" # See documentation for possible values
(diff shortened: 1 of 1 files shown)


nathanshammah (approved):
Thank you @crazy4pi314! It looks good to me. I have just two questions in mind, but that we can find out about them later on after trying the tool: 1. How pressing will the daily updates be, as maybe we could have a weekly one or so, instead of daily. 2. If we still need to be careful about conflicts, or if that is sorted out by the software. Given by one outstanding security threat now open on the repo, it seems like Dependabot does not realize some of the constraints we are putting (a bit like the old pip and conda). I guess that if things break, one can find out in the CI anyways.

crazy4pi314:
I don't think it checking daily means we will have reports daily to look at, it only raises a warning if there is something out of date. This for sure won't do anything automatic w.r.t. code changes, only alert us when we may want to update things and if it can suggest the PR to do so. My vote would be to just add it because we unfortunately can't test it otherwise, and if we want to decrease frequency we can.

nathanshammah:
Sounds good to me, we can give it a try as it is. Thanks @crazy4pi314!

PR #515, opened 2021-01-26
add represent_operations_in_circuit and add tests

andreamari:
Description
-----------
Add functions in `mitiq.representations.depolarizing` which can be used to generate the full representation list
given an input circuit.

Checklist
-----------


mitiq/pec/representations/tests/test_depolarizing.py
```diff
@@ -175,3 +179,73 @@ def test_three_qubit_local_depolarizing_representation_error():
...
+        assert found
+
+    # The number of reps. should match the number of unique operations
+    assert len(reps) == 3
+
+
+# TODO: Once the problem with qubit names is fixed, remove the next
+# test and add "qiskit" circuit type the test above.
+@pytest.mark.skip(reason="Problem with conversions of qiskit qubit names")
+@pytest.mark.parametrize("circuit_type", ["qiskit"])
```
rmlarose, line 231: Can you elaborate on this?

mitiq/pec/representations/depolarizing.py
```diff
@@ -236,3 +236,79 @@ def represent_operation_with_local_depolarizing_noise(
...
+    """Iterates over all unique operations of the input ``ideal_circuit`` and,
+    for each of them, generates the corresponding quasi-probability
+    representation (linear combination of implementable noisy operations).
+
+    This function assumes that the tensor product of ``k`` single-qubit
+    depolarizing channels affects each implemented operation, where
+    ``k`` is the number of qubits associated to the operation.
+
+    This function internally calls
+    :func:`represent_operation_with_local_depolarizing_noise` (more datails
```
rmlarose, line 291:
```suggestion
    :func:`represent_operation_with_local_depolarizing_noise` (more details
```

mitiq/pec/representations/depolarizing.py
```diff
@@ -236,3 +236,79 @@ def represent_operation_with_local_depolarizing_noise(
...
+) -> List[OperationRepresentation]:
+    """Iterates over all unique operations of the input ``ideal_circuit`` and,
+    for each of them, generates the corresponding quasi-probability
+    representation (linear combination of implementable noisy operations).
+
+    This function assumes that the same depolarizing noise channel of strength
+    ``noise_level`` affects each implemented operation.
+
+    This function internally calls
+    :func:`represent_operation_with_global_depolarizing_noise` (more datails
```
rmlarose, line 252:
```suggestion
    :func:`represent_operation_with_global_depolarizing_noise` (more details
```

mitiq/pec/tests/test_pec.py
```diff
@@ -32,12 +31,12 @@
...
 from mitiq.pec.representations import (
-    represent_operation_with_local_depolarizing_noise
+    represent_operations_in_circuit_with_local_depolarizing_noise,
 )
 
 
-# Noisy representations of Pauli operations for testing.
-def get_pauli_representations(
+# Noisy representations of Pauli and CNOT operations for testing.
+def get_pauli_and_CNOT_representations(
```
rmlarose, line 39:
```suggestion
def get_pauli_and_cnot_representations(
```

rmlarose (changes requested):
Can you add tests for the new functions with circuits that have measurements? And possibly other operations that shouldn't be represented, but I think it's just measurements right now.

Also, can you add 

```python
from mitiq.pec.representations import (
    represent_operation_with_global_depolarizing_noise,
    represent_operation_with_local_depolarizing_noise,
    ...
)
```

to `pec/__init__.py`?

mitiq/pec/representations/tests/test_depolarizing.py
andreamari: I made an issue about this (see  #516 )

andreamari: Update after this commit: https://github.com/unitaryfund/mitiq/pull/515/commits/5c60c11e9efeffd9ee41fbdc12b342b60cd155a1

The qiskit test is now added and working fine, at the price of
`represent_operations_in_circuit_with_...`  returning always representations where  `. native_ideal` is a `cirq.Circuit`.

This is not an ideal solution but I think that, because of #516, it is hard to return the correct native operation (with correct qubit names).

andreamari:
Sorry for the small change after the review request.

mitiq/pec/representations/tests/test_depolarizing.py
rmlarose: Hmm, maybe this isn't so bad. All PEC types are supposed to be platform agnostic. 

First, I realized from this that [`OperationRepresentation.ideal`](https://github.com/unitaryfund/mitiq/blob/master/mitiq/pec/types/types.py#L410) should be rewritten like [`NoisyOperation.ideal_circuit`](https://github.com/unitaryfund/mitiq/blob/master/mitiq/pec/types/types.py#L190). Currently `OperationRepresentation._native_ideal` isn't used anywhere.

Second, with the pattern of `represent_operations_in_circuit*` functions, I think we will always be converting the input `QPROGRAM` to a `cirq.Circuit` to iterate over the operations. If this is the case, maybe `OperationRepresentation` should be redesigned a bit to support setting the native type. Something roughly like

```python
def represent_operations_in_circuit*(circuit: QPROGRAM):
    circuit, ctype = convert_to_mitiq(circuit)

    for op in set(circuit.all_operations()):
        rep = OperationRepresentation(Circuit(op), basis_expansion, native_type=ctype)
```
 What do you think? Does this fix all the issues here?

andreamari: Note that at the moment we use the fact that [`OperationRepresentation.ideal`](https://github.com/unitaryfund/mitiq/blob/master/mitiq/pec/types/types.py#L410) always returns Cirq circuits.
This is used [sampling_sequence](https://github.com/unitaryfund/mitiq/blob/74253d72da759d290c8c9cf7273108df79aeb1cb/mitiq/pec/sampling.py#L67). If we change the return type of `OperationRepresentation.ideal` we need to reconsider this sampling step.

I am not sure if adding a `native_type` argument would fix all issues. You probably have a better vision about this since it depends on how operations and `basis_expansion` are converted and stored inside `OperationRepresentation` and `NoisyOperation` . 

My only message here is that, in order to have PEC working, the function `represent_operations_in_circuit*` always returns "converted" representations associated to Cirq operations. In fact this is what `execute_with_pec` actually needs for doing PEC.

Native representations would be a plus for the user, but I think has lower priority with respect to having PEC working.

andreamari:
@rmlarose, I left the TODOs and removed only the two lines mentioning `native_ideal`. (see last commit).
Do you think we can merge this?

rmlarose:
I was waiting to merge #524 to address [this TODO](https://github.com/unitaryfund/mitiq/pull/515/files/d8fee84edb42e1fd0153bd180236d93c72e03a0b..87c65901c4c4648e603a13e6555f2b73a349e10a#diff-1af0b6b05acf6d57861152a1026ff913d08cdaeb355060a61342d84f30db3353R280). 

Do you think this is appropriate?

andreamari:
> I was waiting to merge #524 to address [this TODO](https://github.com/unitaryfund/mitiq/pull/515/files/d8fee84edb42e1fd0153bd180236d93c72e03a0b..87c65901c4c4648e603a13e6555f2b73a349e10a#diff-1af0b6b05acf6d57861152a1026ff913d08cdaeb355060a61342d84f30db3353R280).
> 
> Do you think this is appropriate? I'm leaning towards this, since it is a more permanent fix I think.

I am actually not sure if #524 will be enough to fix the TODO. Anyway this is not a critical PR and we can keep it open and also push it to the next release. If necessary (e.g. for the documentation), we can just do a manual loop over the circuit.

rmlarose:
> I am actually not sure if #524 will be enough to fix the TODO. Anyway this is not a critical PR and we can keep it open and also push it to the next release. If necessary (e.g. for the documentation), we can just do a manual loop over the circuit.

Yeah, it will take more. I would lean towards docs without the functions in this PR for now, and wait to merge this until #524 and follow-up work is done.

nathanshammah:
@rmlarose @andreamari I understand correctly that this PR is on hold until #530 is solved, or is it possible to review / merge it already? There are some merge conflicts.

andreamari:
Actually I am not very convinced anymore about adding the 2 circuit-level functions of this PR.

Apart from the circuit type problems mentioned in the comments, I think it is not a "physical" use case to add the same noise level to all gates (e.g. the noise of CNOT and single-qubit gates is usually very different). I know that we do this a lot
in our tests but tests don't need to be realistic. 

My proposal would be to close this PR and, for the moment, to keep using the operation-level functions that we already have in master. If we realize that circuit-level functions are really important we can reopen this.

@rmlarose what do you think?

andreamari:
According to the discussion of the last meeting, the decision was to merge this PR with a note in the docstrings of the new functions saying that the output `representations` are always defined with Cirq circuits.

@rmlarose , this is ready for review.

mitiq/pec/representations/tests/test_depolarizing.py
```diff
@@ -167,3 +172,80 @@ def test_three_qubit_local_depolarizing_representation_error():
         represent_operation_with_local_depolarizing_noise(
             Circuit(CCNOT(q0, q1, q2)), 0.05,
         )
+
+
+@pytest.mark.parametrize("circuit_type", ["cirq", "qiskit", "pyquil"])
```
rmlarose, line 177: Is this used?

mitiq/pec/representations/tests/test_depolarizing.py
```diff
@@ -167,3 +172,80 @@ def test_three_qubit_local_depolarizing_representation_error():
...
+
+    reps = rep_function(ideal_circuit=circ, noise_level=0.1)
+
+    for op in convert_to_mitiq(circ)[0].all_operations():
+        found = False
+        for rep in reps:
+            if _equal(rep.ideal, Circuit(op), require_qubit_equality=True):
+                found = True
+        if not isinstance(op.gate, MeasurementGate):
+            assert found
```
rmlarose, line 248:
```suggestion
        if isinstance(op.gate, MeasurementGate):
            assert not found
        else:
            assert found
```

rmlarose: Otherwise a measurement representation can be found and the test will still pass.

rmlarose (changes requested):
Two small fixes then LGTM

mitiq/pec/representations/tests/test_depolarizing.py
andreamari: Good catch! I fixed this. Now circuit_type is used.

rmlarose (approved):

PR #523, opened 2021-01-30
Link docs badge to stable instead of latest

rmlarose:


nathanshammah (approved):
Good catch @rmlarose!

PR #524, opened 2021-01-31
Preserve register structure in Qiskit conversions

### LINKED ISSUE #355: Qiskit conversions can change the number of classical registers
This is, at the least, quite confusing, and it can lead to problems with Qiskit executors since the `counts` of a job result are formatted differently depending on how many classical registers are in the circuit. 

The screenshot below shows calling the executor on a circuit with one classical register. The output counts dict has no spaces in the bitstring keys. When execute with zne is called, conversions introduce two classical registers so the output counts dict has spaces in the bitstring keys.

image

### LINKED ISSUE #516: Allow setting qiskit.QuantumRegister when converting to Qiskit
Issue Description
-----------------
In PEC we need to convert operations defined in Qiskit or PyQuil to Cirq and viceversa. 
However this can create ambiguities: for example operations acting on different Cirq qubits can be mapped to
the operations acting on the same qiskit qubit (and probably viceversa).

How to Reproduce
----------------

```python
qreg = cirq.LineQubit.range(2)
x_on_qubit_zero = cirq.Circuit(cirq.X(qreg[0]))
print("x_on_qubit_zero:")
print(x_on_qubit_zero)
qiskit_operation = mitiq.conversions.convert_from_mitiq(x_on_qubit_zero, "qiskit")
print("x_on_qubit_zero converted to qiskit:")
print(qiskit_operation)

x_on_qubit_one = cirq.Circuit(cirq.X(qreg[1]))
print("x_on_qubit_one\n", x_on_qubit_one)
qiskit_operation_one = mitiq.conversions.convert_from_mitiq(x_on_qubit_one, "qiskit")
print("x_on_qubit_one converted to qiskit:")
print(qiskit_operation)
```

OUTPUT:
```
x_on_qubit_zero:
0: ───X───
x_on_qubit_zero converted to qiskit:
        ┌───┐
q_0: |0>┤ X ├
        └───┘
x_on_qubit_one
 1: ───X───
x_on_qubit_one converted to qiskit:
        ┌───┐
q_0: |0>┤ X ├
        └───┘

```

In the example above two Cirq operations acting on different qubits are mapped to the same qiskit object.

How to Solve
----------------
I see two possible options but I am not sure:

1. Try to convert in some way qubit names or numbers directly in conversion functions. 
2. Keep conversion function unchanged but do some qubit re-mapping after converting operations in PEC.

### LINKED ISSUE #522: Folding a Qiskit circuit with unused classical registers removes classical registers
## Description

If you have a Qiskit circuit with classical registers but no measurements into these classical registers, folding functions remove the classical registers. (When there are measurements into the registers, they are kept but renamed/altered.)

## Example

```python
import qiskit
from mitiq import zne

circuit = qiskit.QuantumCircuit(qiskit.QuantumRegister(1), qiskit.ClassicalRegister(1))

print(circuit.cregs)
# [ClassicalRegister(1, 'c2')]

scaled = zne.scaling.fold_gates_from_left(circuit, scale_factor=1)
print(scaled.cregs)
# []
```

*Note*: If the circuit has a measurement, the classical register would be in the scaled circuit.

rmlarose:
Fixes #355 (registers changed when folding Qiskit circuits) and fixes #522 (unused classical registers removed when folding Qiskit circuits). Also fixes #516 which is essentially a duplicate of #355. 

* Qiskit conversions (`to_qiskit`) now have an option to set the quantum and classical registers.
* Folding functions acting on Qiskit circuits now keep the register structure of the input Qiskit circuits.

Example of new behavior:

```python
import qiskit
from mitiq import zne

qreg = qiskit.QuantumRegister(2)
creg = qiskit.ClassicalRegister(2)
circ = qiskit.QuantumCircuit(qreg, creg)

circ.h(qreg)
print(circ)
#          ┌───┐
# q0_0: |0>┤ H ├
#          ├───┤
# q0_1: |0>┤ H ├
#          └───┘
#  c0_0: 0 ═════
#               
#  c0_1: 0 ═════

scaled = zne.scaling.fold_gates_from_left(circ, scale_factor=3)
print(scaled)  # Unused classical registers are now kept + registers are the same.
#          ┌───┐┌───┐┌───┐
# q0_0: |0>┤ H ├┤ H ├┤ H ├
#          ├───┤├───┤├───┤
# q0_1: |0>┤ H ├┤ H ├┤ H ├
#          └───┘└───┘└───┘
#  c0_0: 0 ═══════════════
#                         
#  c0_1: 0 ═══════════════
```

Let me be the first to admit that the `_transform_qubits` function is pretty ugly.

### DIFF
--- a/mitiq/mitiq_qiskit/conversions.py
+++ b/mitiq/mitiq_qiskit/conversions.py
@@ -60,6 +60,10 @@ def _map_bit_index(
         bit_index = 3, new_register_sizes = [2, 3]
         returns (1, 0), meaning the mapped (qu)bit is in the 1st new register
         and has index 0 in this register.
+
+    Note:
+        The bit_index is assumed to come from a circuit with 1 or n registers
+        where n is the maximum bit_index.
     """
     max_indices_in_registers = np.cumsum(new_register_sizes) - 1
 
@@ -81,7 +85,9 @@ def _map_bit_index(
 
 
 def _map_bits(bits, new_register_sizes, new_registers):
-    """Maps (qu)bits to new registers."""
+    """Maps (qu)bits to new registers. Assumes the input ``bits`` come from
+    a single register or n registers, where n is the number of bits.
+    """
     if len(new_registers) == 0:
         return bits
 
@@ -104,7 +110,8 @@ def _transform_registers(
     """Transforms the quantum registers in the circuit to the new registers.
 
     Args:
-        circuit: Qiskit circuit with one quantum register.
+        circuit: Qiskit circuit with one quantum register and either zero
+            classical registers or 1 or n classical registers of n bits.
         new_qregs: The new quantum registers for the circuit.
         new_cregs: The new classical registers for the circuit.
 
@@ -131,12 +138,6 @@ def _transform_registers(
             f"{len(circuit.qregs)} quantum registers."
         )
 
-    if len(circuit.cregs) > 1:
-        raise ValueError(
-            "Input circuit is required to have <= 1 classical register but has"
-            f" {len(circuit.cregs)} classical registers."
-        )
-
     qreg_sizes = [qreg.size for qreg in new_qregs]
     nqubits_in_circuit = sum(qreg.size for qreg in circuit.qregs)
 
@@ -148,6 +149,13 @@ def _transform_registers(
 
     creg_sizes = [creg.size for creg in new_cregs]
     nbits_in_circuit = sum(creg.size for creg in circuit.cregs)
+
+    if len(circuit.cregs) not in (0, 1, nbits_in_circuit):
+        raise ValueError(
+            f"Input circuit is required to have 0, 1, or {nbits_in_circuit} "
+            f"classical registers but has {len(circuit.cregs)} classical "
+            f"registers."
+        )
 
     if len(creg_sizes) and sum(creg_sizes) != nbits_in_circuit:
         raise ValueError(
@@ -192,6 +200,7 @@ def to_qiskit(
     circuit: cirq.Circuit,
     qregs: Optional[List[qiskit.QuantumRegister]] = None,
     cregs: Optional[List[qiskit.ClassicalRegister]] = None,
+    add_cregs_if_cannot_transform: bool = True,
 ) -> qiskit.QuantumCircuit:
     """Returns a Qiskit circuit equivalent to the input Mitiq circuit.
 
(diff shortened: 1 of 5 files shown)


mitiq/mitiq_qiskit/conversions.py
```diff
@@ -41,6 +46,155 @@ def _remove_barriers(circuit: QuantumCircuit) -> QuantumCircuit:
...
+
+    if isinstance(new_registers[0], qiskit.QuantumRegister):
+        Bit = qiskit.circuit.Qubit
+    else:
+        Bit = qiskit.circuit.Clbit
+
+    return [Bit(new_registers[i], j) for i, j in mapped_indices]
+
+
+def _transform_registers(
```
andreamari, line 112: I am surprised this function doesn't exist in qiskit. But indeed I could not find it. 

However, I found something similar which allows to append an "old" circuit to a new circuit.
It can be done as follows:
`new_circuit.append(c.to_instruction(), new_circuit.qubits).decompose()`

In this way, one could assign a circuit with old registers to a circuit with new registers without the need to
define qubit mapping functions.

This is an optional suggestion to simplify the code but the current solution is fine for me as long as it works.

andreamari (commented):
Thanks @rmlarose.  A function is missing some annotations and docs.

The rest is good for me but I also left an optional suggestion which could be useful to simplify the code.

mitiq/mitiq_qiskit/conversions.py
rmlarose: Ah, thanks for the suggestion. Let me look into this, I prefer to use code that already exists inasmuch possible.

rmlarose: After wrestling with Qiskit for the afternoon, I think I want to retract my previous comment.

This strategy is very brittle with a new error popping up for each unit test. Probably the simplest is when you have an unused classical register:

```python
import qiskit

old = qiskit.QuantumCircuit(qiskit.QuantumRegister(1), qiskit.ClassicalRegister(1))
old.h(old.qregs[0])
print(old)

# Try to transform registers.
new = qiskit.QuantumCircuit(qiskit.QuantumRegister(1), qiskit.ClassicalRegister(1))
new.append(old.to_instruction(), qargs=new.qubits, cargs=new.clbits)
print(new.decompose())
```

which raises this error:

```python
DAGCircuitError: 'Cannot replace node of width (1 qubits, 1 clbits) with instruction of mismatched width (1 qubits, 0 clbits).'
```

This can be fixed but it requires performing checks on the instruction, and then we're essentially back to the current implementation.

There's a bunch of other errors for other unit tests, even after fixing a bunch. I'm ready to scrap this strategy unless someone else wants to take a stab at it.

andreamari (approved):
Ok for keeping the current implementation. Thanks for having tried the other option too. 
This LGTM but I think you missed my previous comment about `_map_bits`:
> Annotations, `Args: ` and `Returns:` are missing

This is a very small addition so you can merge after fixing it without the need of further review.

rmlarose:
> This LGTM but I think you missed my previous comment about `_map_bits`:
> 
> > Annotations, `Args: ` and `Returns:` are missing
> 
> This is a very small addition so you can merge after fixing it without the need of further review.

I'll add this today then merge. I'll also open an issue for a better implementation of transforming registers using Qiskit as an enhancement.

PR #525, opened 2021-01-31
Fix folding by fidelity for scale factors larger than 3

### LINKED ISSUE #521: Folding by fidelity folds unexpected gates when scale_factor > 3
## Description

Folding with `fidelities={"single": 1.0}` should not affect single qubit gates, but it does when `scale_factor > 3`.

## Example

```python
import cirq
from mitiq import zne

qreg = cirq.LineQubit.range(2)
circuit = cirq.Circuit(cirq.H(qreg[0]), cirq.CNOT(*qreg)) 
print(circuit)
# 0: ───H───@───
#           │
# 1: ───────X───

# Scale factor of <= 3 produces expected results
zne.scaling.fold_gates_at_random(circuit, scale_factor=3, fidelities={"single": 1.0})
# 0: ───H───@───@───@───
#           │   │   │
# 1: ───────X───X───X───

# Scale factor > 3 folds single qubit gates when it shouldn't
zne.scaling.fold_gates_at_random(circuit, scale_factor=4, fidelities={"single": 1.0})
# 0: ───H───H───H───@───@───@───@───@───
#                   │   │   │   │   │
# 1: ───────────────X───X───X───X───X───
```

rmlarose:
Fixes #521.

### DIFF
--- a/mitiq/zne/scaling/folding.py
+++ b/mitiq/zne/scaling/folding.py
@@ -740,10 +740,19 @@ def _fold_local(
         this_stretch = 3.0 if scale_factor > 3.0 else scale_factor
         if fold_method_args:
             folded = fold_method(
-                folded, this_stretch, *fold_method_args, squash_moments=False
+                folded,
+                this_stretch,
+                *fold_method_args,
+                squash_moments=False,
+                **{k: v for k, v in kwargs.items() if k != "squash_moments"},
             )
         else:
-            folded = fold_method(folded, this_stretch, squash_moments=False)
+            folded = fold_method(
+                folded,
+                this_stretch,
+                squash_moments=False,
+                **{k: v for k, v in kwargs.items() if k != "squash_moments"}
+            )
         scale_factor /= 3.0
 
     if not (kwargs.get("squash_moments") is False):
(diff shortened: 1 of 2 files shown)


mitiq/zne/scaling/folding.py
```diff
@@ -740,10 +740,19 @@ def fold_method(circuit: Circuit, scale_factor: float,**kwargs):
         this_stretch = 3.0 if scale_factor > 3.0 else scale_factor
         if fold_method_args:
             folded = fold_method(
-                folded, this_stretch, *fold_method_args, squash_moments=False
+                folded,
+                this_stretch,
+                *fold_method_args,
+                squash_moments=False,
+                **{k: v for k, v in kwargs.items() if k != "squash_moments"},
```
andreamari, line 747: What happens if instead of forcing `squash_moments` to be `False` for odd scale factors,  we set it according to `kwargs` ?
Are there bad consequences?

andreamari (approved):
I left a question about `squash_moments`.
Anyway this LGTM.

mitiq/zne/scaling/folding.py
rmlarose: I don't think there are bad consequences, but there is duplicated work. You only need to [squash moments once at the end](https://github.com/unitaryfund/mitiq/pull/525/files#diff-58c4c4e3cb99c8097c841f4267e46fe9702e78865350b7eb607653a332f77b52R758-R760).

andreamari: I see, good point!

PR #535, opened 2021-02-04
Run doctest in CI

rmlarose:
Fixes #533.

PR #539, opened 2021-02-07
Fix and refactor RB circuits function

### LINKED ISSUE #538: Only the last number of clliffords is used in rb_circuits function
The [`rb_circuits`](https://github.com/unitaryfund/mitiq/blob/master/mitiq/benchmarks/randomized_benchmarking.py#L36) function takes a list of integers for numbers of clifford gates, but only the last number in the list is used.

rmlarose:
Fixes #538 and changes the behavior of this function to only accept one int for `num_cliffords`. I've never had a use case for multiple values for this argument, and I've always been confused to why it needed to be a list.

Also removes the `qubit_labels` argument because (i) it was never used as far as I can tell and (ii) a proper implementation of this should accept any qubit type, but `qubit_labels` assumed `LineQubit`s.

### DIFF
--- a/docs/source/guide/guide-executors.rst
+++ b/docs/source/guide/guide-executors.rst
@@ -545,35 +545,35 @@ Below is an example to use TensorFlow Quantum to simulate a bit-flip channel:
         nM = len(circ.moments)
         nQ = len(circ.all_qubits())
 
-        # Create array of symbolic variables and reshape to natural circuit parameterization
+        # Create array of symbolic variables and reshape to natural circuit parameterization.
         h = sympy.symbols(''.join(['h_{0} '.format(i) for i in range(nM * nQ)]), positive=True)
         h_array = np.asarray(h).reshape((nQ, nM))
 
-        # Symbolicly add X gates to the input circuit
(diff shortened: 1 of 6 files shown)


rmlarose:
#535 should be merged before this (and anything else).

andreamari (commented):
This is unrelated with the changes of this PR, but I find confusing that we have some global constants:
```
CLIFFORDS = _single_qubit_cliffords()
C1 = CLIFFORDS.c1_in_xy
CFD_MAT_1Q = np.array([_gate_seq_to_mats(gates) for gates in C1])
```
which we use for the case `num_qubits==1`, while we use local variables for the case `num_qubits==2`:
       
```
cfd_matrices = _two_qubit_clifford_matrices(
    qubit1, qubit2, CLIFFORDS,  # type: ignore
)
```
I am not sure what is better (global vs local) but I think that it is cleaner if we use the same strategy for `num_qubits==1` and `num_qubits==2`.

rmlarose:
I think its relevant enough. Could be an artifact, not sure, but I'll remove the globals.

andreamari (approved):
LGTM! Thanks

PR #541, opened 2021-02-07
Only try to simplify GateOperations in simplify exponents

### LINKED ISSUE #540: Simplify exponents fails with controlled operations
There is an [assert statement](https://github.com/unitaryfund/mitiq/blob/master/mitiq/utils.py#L67) to check that the operation is a `cirq.GateOperation`, which a `cirq.ControlledOperation` is not. I don't think the function should fail because there may be operations which are not gate operations. The better behavior is to only try to simplify `GateOperation`s, i.e., skip `ControlledOperation`s and other not gate operations. 

Example:

```python
import cirq
from mitiq.mitiq_qiskit.conversions import to_qiskit

circuit = cirq.Circuit(
    cirq.ControlledGate(cirq.SWAP, num_controls=1).on(*cirq.LineQubit.range(3))
)
print(circuit)
# 0: ───@───
#       │
# 1: ───×───
#       │
# 2: ───×───

to_qiskit(circuit)
```

raises

```python
mitiq/mitiq/mitiq/utils.py in _simplify_circuit_exponents(circuit)
     65         # Iterate over operations in moment
     66         for op in moment:
---> 67             assert isinstance(op, GateOperation)
     68 

AssertionError: 
```

rmlarose:
Fixes #540.

### DIFF
--- a/mitiq/tests/test_utils.py
+++ b/mitiq/tests/test_utils.py
@@ -23,6 +23,7 @@ import cirq
 from cirq import (
     LineQubit,
     Circuit,
+    ControlledGate,
     X,
     Y,
     Z,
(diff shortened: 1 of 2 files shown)


andreamari (approved):
LGTM!

PR #543, opened 2021-02-08
Add readthedocs config file

### LINKED ISSUE #542: Readthedocs fails to build the docs
Issue Description
-----------------

Readthedocs cannot build the docs (fails the build) after we merged this [commit](https://github.com/unitaryfund/mitiq/commit/49986668a2a9d4736d1ef066931b24fce479d644
) in which we have split the requirements into two files (`requirements.txt` and `dev_requirements.txt`).

Since readthedocs only looks inside a single file (`requirements.txt`), the build fails.

How to fix
-----------------
I am not sure what is the optimal solution. Two ideas:

- Add Cirq to `dev_requirements.txt` too. And set `dev_requirements.txt` in the readthedocs administration settings instead of `requirements.txt`. I tried this, but since Cirq is missing in `dev_requirements.txt` the API is not rendered (even if the build correctly passes). 
- Add a new readthedocs [configuration file](https://docs.readthedocs.io/en/stable/config-file/v2.html).

andreamari:
Description
-----------
Fixes #542 

With this PR, readthedocs works properly.


nathanshammah (approved):
LGTM, thanks @andreamari.

PR #544, opened 2021-02-08
Updates README

### LINKED ISSUE #537: Update readme with new requirements and description
Note to self (or note to @andreamari) to update the readme before the next release. At least two things to be updated:

* Cirq version is now 0.10.something
* Two techniques now, not just ZNE

and potentially others.

andreamari:
Description
-----------
Fixes #537 .
Updates readme. 

@rmlarose, given that the versions of the required packages will change often, what about if we avoid adding the versions of Cirq, Qiskit and PyQuil in the `README.md` ? In this way there will be no risk of going out-of-sync with `requirements.txt` and `dev_requirements.txt` in the future.
Otherwise, if you think versions should be in the README.md,  we should probably use `0.10.0.dev20201203012958` for Cirq.

### DIFF
--- a/README.md
+++ b/README.md
@@ -41,11 +41,11 @@ If you'd prefer to clone and install from source, our would like to develop Miti
 
 ### Supported quantum programming libraries
 
-Mitiq can currently interface with
+Mitiq can currently interface with:
 
-* [Cirq](https://github.com/quantumlib/Cirq) >= 0.9.0, 
-* [Qiskit](https://qiskit.org/) >= 0.19.0, and 
-* [pyQuil](https://github.com/rigetti/pyquil) >= 2.18.0. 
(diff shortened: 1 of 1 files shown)


nathanshammah (approved):
LGTM. I understand the link for PEC points to the`readthedocs-config` branch as it is now working, but we should remember in the future to change it to a stable one. 

@rmlarose I am merging to move forward with the release process. You can open an issue if we want to discuss this further.

rmlarose:
> @rmlarose, given that the versions of the required packages will change often, what about if we avoid adding the versions of Cirq, Qiskit and PyQuil in the `README.md` ? In this way there will be no risk of going out-of-sync with `requirements.txt` and `dev_requirements.txt` in the future.

Yeah, probably better to list them without versions and say "see here" or something for which versions are supported.

PR #548, opened 2021-02-08
adding mitiq feature proposal template, fixes #498

### LINKED ISSUE #498: Add new issue template for feature proposal (as opposed to request)
Pre-Request Checklist
---------------------

- [x] I checked to make sure that this feature has not already been requested

Issue Description
-----------------

Currently, there isn't a clear way for the community to propose substantial new features to Mitiq. They could file a generic issue and discuss, or use the request feature template which has the implication that the dev team will implement it.

Proposed Solution
-----------------

As discussed in Engineering meeting and Mitiq community call, we can add a new issue template called something like "Feature proposal" or "Feature Design Proposal" and use a template there to help form complete and reviewable proposals. 

It was also discussed that linking to a Google doc may also work, as that has the benefit of a better commenting system.
The downside to this is that there is a potential to lose the data and that sharing permissions need to be managed by the submitter, as opposed to the Mitiq team.

Additional References
---------------------

An example of a fully public language design process can be found in the  [Q# language design](https://github.com/microsoft/qsharp-language#contributing) repo where they use a similar issue template system.

crazy4pi314:
Description
-----------
Adding an issue template for folks to propose new features for Mitiq as per #498. Happy to change any language, based on the Q# language feature proposal template.

Checklist
-----------

### DIFF
--- a/.github/ISSUE_TEMPLATE/feature_proposal.md
+++ b/.github/ISSUE_TEMPLATE/feature_proposal.md
@@ -8,7 +8,7 @@ assignees: ''
 ---
 ## Pre-Request Checklist
 
-- [ ] I checked to make sure that this feature has not already been requested
+- [ ] I checked other issues on Mitiq to make sure that this feature has not already been requested
 - [ ] I or my organization would be willing to help implement and/or test this.
 - [ ] This is **not** a breaking change to Mitiq.
 
--- a/.github/ISSUE_TEMPLATE/feature_proposal.md
+++ b/.github/ISSUE_TEMPLATE/feature_proposal.md
@@ -1,0 +1,39 @@
+---
+name: Feature proposal
+about: Propose a great new feature for Mitiq
+title: ''
+labels: 'enhancement'
+assignees: ''
+
+---
+## Pre-Request Checklist
+
+- [ ] I checked to make sure that this feature has not already been requested
+- [ ] I or my organization would be willing to help implement and/or test this.
+- [ ] This is **not** a breaking change to Mitiq.
+
+## Feature description
+
+_TODO: Insert a short outline for what the proposed modification is._
+
+## Considerations
+
+_TODO: Explain why this modification to Mitiq is desirable. Briefly summarize the benefits and drawbacks of the chosen approach compared to other ways of achieving a similar functionality._
+
+## Context
+
+_TODO: Describe briefly the most important points to consider in more detail going forward. Link to related proposals, specs, tools, papers, documentation or other related topics._
+
+## Examples
+
+_TODO: Give concrete examples for syntax and behavior to illustrate the suggestions above._
+
+### Example 1:
+
+_TODO: insert title and caption_
+
+```python
+# TODO: 
+# Insert code example that illustrates what is described above.
+# Comment your code to further elaborate and clarify the example.
+```


.github/ISSUE_TEMPLATE/feature_proposal.md
```diff
@@ -0,0 +1,39 @@
...
+about: Propose a great new feature for Mitiq
+title: ''
+labels: 'enhancement'
+assignees: ''
+
+---
+## Pre-Request Checklist
+
+- [ ] I checked to make sure that this feature has not already been requested
+- [ ] I or my organization would be willing to help implement and/or test this.
```
rmlarose, line 12: Should this be required? It seems reasonable for someone to request a feature but they can't work on it for whatever reason.

.github/ISSUE_TEMPLATE/feature_proposal.md
```diff
@@ -0,0 +1,39 @@
...
+- [ ] This is **not** a breaking change to Mitiq.
```
rmlarose, line 13: I might remove this - some features could be breaking changes, no? and some quantum-but-not-software people may not be familiar with this terminology.

.github/ISSUE_TEMPLATE/feature_proposal.md
```diff
@@ -0,0 +1,39 @@
...
+name: Feature proposal
+about: Propose a great new feature for Mitiq
+title: ''
+labels: 'enhancement'
+assignees: ''
+
+---
+## Pre-Request Checklist
+
+- [ ] I checked to make sure that this feature has not already been requested
```
rmlarose, line 11:
```suggestion
- [ ] I checked other issues on Mitiq to make sure that this feature has not already been requested
```

.github/ISSUE_TEMPLATE/feature_proposal.md
```diff
@@ -0,0 +1,39 @@
...
+- [ ] I or my organization would be willing to help implement and/or test this.
+- [ ] This is **not** a breaking change to Mitiq.
+
+## Feature description
+
+_TODO: Insert a short outline for what the proposed modification is._
+
+## Considerations
```
rmlarose, line 19: Could we rename this section to "Purpose(s) of feature" or something similar? "Considerations"  is a bit vague to me.

.github/ISSUE_TEMPLATE/feature_proposal.md
```diff
@@ -0,0 +1,39 @@
...
+
+_TODO: Explain why this modification to Mitiq is desirable. Briefly summarize the benefits and drawbacks of the chosen approach compared to other ways of achieving a similar functionality._
```
rmlarose, line 21:
```suggestion
_TODO: Explain the purpose of this feature and/or why this modification to Mitiq is desirable._
```

rmlarose: Could you clarify the second sentence? Does "other ways of achieving similar functionality" mean using other tools that already exist? Or alternative ways I thought of doing this in Mitiq?

.github/ISSUE_TEMPLATE/feature_proposal.md
```diff
@@ -0,0 +1,39 @@
...
+
+## Context
```
rmlarose, line 23: "Context" also is a bit vague to me. What about (stealing from the description) "Important details" or "Important points going forward"?

rmlarose: Also, can we put this section after examples? I think examples are nice to see after the previous section describing the purpose and can elaborate on anything that perhaps wasn't clear there.

rmlarose (changes requested):
Generally looks good! A few very nit-picking comments because I think it's important to have this as clear & useful for both us & contributors as possible. Let me know what you think.

purva-thakre:
It could be a good idea to have a link to a general Google Doc design template similar to [what this doc has](https://docs.google.com/document/d/1MENciPA3Gri__ZDCkA3PwdtfYV8BMk8pCXTLwnHX9qw/edit?usp=sharing). Otherwise a new proposed design doc will look like an itemized list which I admit might be difficult to comment on as newer problems are identified or as the project changes from initial proposed design. The doc might also keep track of all the changes.

.github/ISSUE_TEMPLATE/feature_proposal.md
crazy4pi314: I was just trying to emulate the format of other issue templates we have, see : https://github.com/unitaryfund/mitiq/blob/master/.github/ISSUE_TEMPLATE/bug_report.md 

It need not be required for us, I assume the volume of proposals we will get will not likely run into duplicates :)

.github/ISSUE_TEMPLATE/feature_proposal.md
crazy4pi314: I guess my thought here is not all need to be checked to be valid, its just more so we can gauge quickly how much work is involved. I take the point that breaking change might be not familiar outside of SW dev.

.github/ISSUE_TEMPLATE/feature_proposal.md
crazy4pi314: My intent is that it is broader than just purpose, in that it also asks about pros and cons of the approach outlined in the previous section and if there are other possible approaches to the problem the proposal. The purpose should have really be already described in the description of the feature. Maybe `##Discussion`` would be a better heading?

.github/ISSUE_TEMPLATE/feature_proposal.md
crazy4pi314: Exactly as you suggest, its asking about are there other ways to fulfill the purpose of the proposed feature, either with other tools, approaches, or just even different ways in Mitiq to go about implementing it. Sometimes there is only one clear way, but it can be helpful when discussing a new feature to check and see if any other approaches meet the user need without major changes.

.github/ISSUE_TEMPLATE/feature_proposal.md
crazy4pi314: Reordering is fine with me, and Context really does I think describe what I wanted here (which may not be a helpful thing i get). Basically the question here is are there other things we should consider like other papers or packages that implement part of the feature being proposed? Or is this similar to another technique that we may also want to consider implementing as it would be easy after adding this feature? Have they written up pseudo code for it already and have a link to that? It's the context that this proposal should be considered with. 

Again, maybe we want to know something different from the author and that's totally cool.

nathanshammah (commented):
This looks to me like a quite subtle difference from "Feature request". My understanding is that someone would not expect others to implement a feature request, but just that that is someone's need, and that they may be asking for help or be proposing to address it themselves – e.g., in a related pull request. 

I understood that we wanted like a standard code design document for major feature proposals. This template could be used for that and could be included inside the source code in a specific folder, i.e. not generating an issue in Github and related conversation. Something like a MEP, Mitiq Enhancement Proposal, e.g., [NEPs](https://github.com/numpy/numpy/blob/1245f9da021d65789a3332f0e046ca30d518e9c8/doc/neps/nep-0000.rst#id2). 

It is also fine to me also to add this template, I just think the difference between request and proposal is not so important to me. 

Also, if one where to go this way, I'd suggest: 
- Adding some information about all three options in the documentation, e.g., in the readme file and in the "Contributing to Mitiq" documentation section. 
- The template for the feature request should have mitiq--> Mitiq in the "about" box.

crazy4pi314:
@nathanshammah I think the design doc in the repo makes more sense here, I will close this PR and make a new one to reflect this as I think its closer to the iteration workflow folks like in google docs and we can also then include notebooks with sample code and latex notation.

PR #520, opened 2021-01-29
Clifford Data Regression Methods 

L-P-B:
Description
-----------
Implementing CDR error mitigation methods first presented in Czarnik et al. (2020) https://arxiv.org/abs/2005.10189 and Lowe et al. (2020) https://arxiv.org/abs/2011.01157.

If you have not finished all tasks in the checklist, you can also open a [Draft Pull Request](https://github.blog/2019-02-14-introducing-draft-pull-requests/) to let the others know this on-going work and keep this checklist in the PR description.

Checklist
-----------
Please make sure you have finished the following tasks before requesting a review of the pull request (PR). For more information, please refer to the [Contributor's Guide](../blob/master/CONTRIBUTING.md):

- [ ] Contributions to mitiq should follow the [pep8 style](https://www.python.org/dev/peps/pep-0008/). You can enforce it easily with [`black`](https://black.readthedocs.io/en/stable/index.html) style and with [`flake8`](http://flake8.pycqa.org) conventions.
- [ ] Please add tests to cover your changes, if applicable, and make sure that all new and existing tests pass locally.
- [ ] If the behavior of the code has changed or new feature has been added, please also [update the documentation](../blob/master/docs/CONTRIBUTING_DOCS.md).
- [ ] Functions and classes have useful [Google-style](https://sphinxcontrib-napoleon.readthedocs.io/en/latest/example_google.html) docstrings and [type hints](https://www.python.org/dev/peps/pep-0484/) in the signature of the objects.
- [ ] (Bug fix) The associated issue is referenced above using [auto-close keywords](https://help.github.com/en/articles/closing-issues-using-keywords). If the PR fixes an issue, use the keyword fix/fixes/fixed followed by the issue id. For example, if the PR fixes issue 1184, type "Fixes #1184" (without quotes).
- [ ] The [changelog](https://github.com/unitaryfund/mitiq/blob/master/CHANGELOG.md) is updated, including author and PR number (@username, gh-xxx).

PR #554, opened 2021-02-16
Increase Qiskit version in dev_requirements

andreamari:
Description
-----------
Pin Qiskit to a higher version (`~0.23.0`).  This is probably a good thing to do in itself but the motivation came from #520 where
a recent qiskit feature is used in a test file. 

Checklist
-----------


andreamari:
FYI @nathanshammah, @crazy4pi314, @rmlarose :  I tried to increase the version of qiskit to 0.2.1, this required to increase the version of NumPy to 1.2, BUT this is not compatible with `tensorflow-quantum` that we currently use in doctest.

Long story short, I will close this PR since this is not so urgent. However this is maybe something to keep in mind.

rmlarose:
What about dropping the TFQ example altogether? To me the ratio of value created to problems created by this TFQ example is quite low. Or maybe try replacing it with qsim? AFAIK the TFQ simulator is qsim and qsim is much more lightweight.

andreamari:
> What about dropping the TFQ example altogether? To me the ratio of value created to problems created by this TFQ example is quite low. Or maybe try replacing it with qsim? AFAIK the TFQ simulator is qsim and qsim is much more lightweight.

We can easily discuss and decide this at the Mitiq meeting. I add the label.

andreamari:
According to the discussion of the last meeting, I removed the TensorFlow requirements and I temporarily skipped documentation tests of corresponding executor.
 
I also opened a new issue to keep track of the skipped tests #559.

.github/workflows/ci-master.yml
```diff
@@ -7,13 +7,13 @@ on:
...
 
-    # runs just the main & benchmarking tests on py36/37/38 and mac/windows
+    # runs just the main & benchmarking tests on py37/38 and mac/windows
     test:
       runs-on: ${{ matrix.os }}
       strategy:
         matrix:
           os: [macos-latest, windows-latest]
-          python-version: [3.6, 3.7, 3.8]
+          python-version: [3.7, 3.8]
```
rmlarose, line 16: Why drop 3.6?

rmlarose (changes requested):
Is there a reason on waiting to delete the TFQ example in the docs? I'm confused why it's still there.

.github/workflows/ci-master.yml
andreamari: Because the more recent NumPy (which is required by the more recent Qiskit) requires Python >=3.7.
See e.g. left column of https://pypi.org/project/numpy/

rmlarose: Got it, thanks. Should we update setup.py with [python_requires](https://packaging.python.org/guides/distributing-packages-using-setuptools/#python-requires)?

andreamari:
> Is there a reason on waiting to delete the TFQ example in the docs? I'm confused why it's still there.

Because I think that in the last meeting there wasn't a consensus on deleting it. So I made #559 with different options.

rmlarose:
> > Is there a reason on waiting to delete the TFQ example in the docs? I'm confused why it's still there.
> 
> Because I think that in the last meeting there wasn't a consensus on deleting it. So I made #559 with different options.

Right now there's just untested code hanging around in the documentation which isn't really useful for anyone. I would suggest removing it in this PR, then taking an appropriate action in #559 - wait until its supported, add back with qsim, do nothing, etc.

andreamari:
A summary of the last changes and other attempts:

- Added `python_requires='>=3.7'` in `setup.py`
- The TF executor is for the moment removed as suggested by @rmlarose. Initially I wasn't sure about this, but now I am quite convinced that this is the right thing to do. It is not just a matter of non-tested docs, it is also a matter of compatibility. If Mitiq now requires some packages which are not compatible with TF-Quantum, the user would need to force `pip` to get both libraries in the same environment. This issue is not tracked in #559.
- I tried to also add python 3.9 to the CI (this required a small bump of matplotlib). All tests passed, however they took a very long time. The reason is that Python 3.9 is not cached in GihHub actions and so it must be installed from scratch and takes a really long time. 
- So I changed back as before: 3.8 as a maximum Python version.

This is now ready for re-review. 

@nathanshammah feel free to have a look if you are interested.

rmlarose (commented):
Thanks @andreamari. I'm actually no longer sure about adding python_requries to setup, since this is a development requirement. In principle one could use Mitiq without Qiskit on Python 3.6. Or do we not want to support this anymore? Thoughts @crazy4pi314 @willzeng?

willzeng:
My vote is to keep the dev and usage requirements separate. Ideally the usage has minimal dependency footprint and we won't want to add qiskit to that if we don't have to.

andreamari:
> Thanks @andreamari. I'm actually no longer sure about adding python_requries to setup, since this is a development requirement. In principle one could use Mitiq without Qiskit on Python 3.6. Or do we not want to support this anymore? Thoughts @crazy4pi314 @willzeng?

Python 3.6 is incompatible with recent version(s) of NumPy which is actually core requirement (it's in `requirements.txt` and not in `dev_requirements.txt`). 

Maybe it is still possible to support 3.6 with some alternative settings (e.g. by just removing `numpy` in `requirements.txt ? ). But I am not sure this would be a good solution.

rmlarose:
Ok, I'm fine to drop 3.6 if there are no concerns from others. I just wanted to check with everyone and not have this slip through unnoticed.

rmlarose (approved):

crazy4pi314:
Yeah I'm ok with it, its simpler to just drop 3.6 for now.

PR #568, opened 2021-02-26
[WIP] Add measurement support to executors

rmlarose:
Opening for comments and discussion.

Initial implementation of the Executor outlined [here](https://docs.google.com/document/d/1Y_CaqfhtMzvfr5X46r3dtrYKZcqH16kRhtecv3wYOJI/edit?usp=sharing). 

As discussed @ Mitiq meeting:

* How to deal with multiple observables? i.e., when you have to use two or more incompatible measurements to compute an expectation? 
  - Option 1: Don't support this. Rely on multiple calls to `mitiq.execute_with_[pec|zne]`. 
      - Note: Multiple calls only required if you want to store measurements. A single call is possible when returning a float by computing the linear combination inside the `circuit_to_expectation` function.
  - Option 2: Add support for observables and store measurements for the product of {circuit} x {observable set}.
  - Option 3: Other suggestions welcome :)
* What's the best interface for accepting `circuit_to_expectation`, `circuit_to_measurement`, and/or `measurement_to_expectation` user-defined functions? As currently is? Abstract base class? Other?

Beyond discussed @ Mitiq meeting:

* I would propose merging the (properly implemented) `Collecter` capabilities into the `Executor` class, assuming the design doesn't significantly change from the current design. Namely, add `Executor.run` method to accept a batch of circuits, detect duplicates, increase shots in duplicates if supported, and handle max batch size & max # shots.

### DIFF
--- a/mitiq/executor.py
+++ b/mitiq/executor.py
@@ -1,0 +1,150 @@
+# Copyright (C) 2021 Unitary Fund
+#
+# This program is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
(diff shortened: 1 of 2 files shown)


rmlarose:
Outdated after new design for execute functions - see #623. New PR coming soon.

PR #574, opened 2021-03-03
Density-matrix simulation with depolarizing noise executor

### LINKED ISSUE #269: Add a qiskit executor example for exact density matrix simulation
This is similar to the other executor examples in the docs. It isn't clear if qiskit has direct support for this.

Aaron-Robertson:
Description
-----------
Resolves #269: add a qiskit executor example for exact density matrix simulation 

Checklist
-----------

### DIFF
--- a/docs/source/guide/guide-executors.rst
+++ b/docs/source/guide/guide-executors.rst
@@ -414,8 +414,8 @@ in this example can be found `here <https://quantumcomputing.stackexchange.com/a
 Qiskit: Density-matrix Simulation with Depolarizing Noise
 -----------------------------------------------------------
 
-Analogous to the Cirq implementation, this executor can be used for noisy depolarizing
-simulation.
+This executor can be used to simulate a circuit with depolarizing noise and to return the exact
+expectation value of an observable (without the shot noise typical of a real experiment).
 
 .. testcode::
 
@@ -479,11 +479,6 @@ simulation.
 
     qc = QuantumCircuit(1)
     for _ in range(100):
-        qc.u1(0, 0)
-    assert 0.1 < execute(qc, np.diag([1, 0]), 0.02) < 1.0
-
-    qc = QuantumCircuit(1)
-    for _ in range(100):
         qc.x(0)
 
     assert execute(qc, np.diag([0, 1]), 0.0) == 0.0
@@ -493,7 +488,8 @@ simulation.
 Qiskit: Density-matrix Simulation with Depolarizing Noise and Sampling
 ------------------------------------------------------------------------
 
-You can also include both noise models and finite sampling in your executor.
+This executor can be used to simulate a circuit with depolarizing noise. The expectation value is
+estimated with a finite number of measurements and so it is affected by statistical noise.
 
 .. testcode::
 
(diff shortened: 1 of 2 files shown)


docs/source/guide/guide-executors.rst
```diff
@@ -413,12 +413,93 @@ in this example can be found `here <https://quantumcomputing.stackexchange.com/a
...
+        if len(circ.clbits) > 0:
+            raise ValueError("This executor only works on programs with no classical bits.")
+
+        circ = copy.deepcopy(circ)
+        # we need to modify the circuit to measure obs in its eigenbasis
+        # we do this by appending a unitary operation
+        eigvals, U = np.linalg.eigh(obs) # obtains a U s.t. obs = U diag(eigvals) U^dag
+        circ.unitary(np.linalg.inv(U), qubits=range(circ.num_qubits))
+        circ.snapshot_density_matrix('final')
+        circ.measure_all()
```
andreamari, line 455: Hi @Aaron-Robertson, thank you very much for this contribution!
I wonder if we should remove 
```suggestion
```
I wonder if one should remove this block of code. This was used to map the observable to the computational basis of the qubits. However I think we don't have to apply U if we directly use the density matrix.

Aaron-Robertson: Hey @andreamari, right you are! And you're welcome. I'm happy to be involved and appreciate the opportunity for feedback.

I can make the change on this executor as soon as I have a chance (though I'll leave the `circ.snapshot_density_matrix('final')`, which I snuck in there).

What about the sampled case? Better to invoke the density matrix snapshot directly I think, but should it be a separate issue?

Aaron-Robertson: I suppose they are examples, and nothing wrong with showing more than one way to do a thing. I could just add a note for the distinction.

andreamari: The qiskit noisy executor with a finite number of shots is already in the docs and I think it is fine. In that case I would keep the change of basis (i.e. the unitary U) since that executor, even if simulated, should mimic the hardware execution where the density matrix is not available.

docs/source/guide/guide-executors.rst
```diff
@@ -413,12 +413,87 @@ in this example can be found `here <https://quantumcomputing.stackexchange.com/a
...
+
+    assert execute(qc, np.diag([0, 1]), 0.0) == 0.0
+    assert np.isclose(execute(qc, np.diag([0, 1]), 0.5), 0.5)
+    assert np.isclose(execute(qc, np.diag([0, 1]), 0.001), 0.0476039)
 
 Qiskit: Density-matrix Simulation with Depolarizing Noise and Sampling
 ------------------------------------------------------------------------
 
-This executor can be used for noisy depolarizing simulation.
+You can also include both noise models and finite sampling in your executor.
```
andreamari, line 496:
```suggestion
This executor can be used to simulate a circuit with depolarizing noise. The expectation value is estimated with a finite number of measurements and so it is affected by statistical noise.
```
Just to be as explicit as possible.

andreamari (commented):
LGTM. Just a few minor comments.

rmlarose (approved):
LGTM, thanks @Aaron-Robertson.

crazy4pi314:
@all-contributors please add @Aaron-Robertson for example, test, bug

PR #463, opened 2020-12-04
Replace Qiskit utils with Qiskit executors from docs 

pchung39:
Description
-----------
This pull request is for #379 . I've transferred the Qiskit executors referenced in the back-end executors documentation  ([Back-End Executors](https://mitiq.readthedocs.io/en/stable/guide/guide-executors.html)) into `/mitiq_qiskit/qiskit_utils.py`. There were two existing function in `qiskit_utils.py` named `random_one_qubit_identity_circuit` and `run_with_noise` that I left untouched in the file. 

I also included a few tests for the added executors. They are pretty straight-forward but should cover basic functionality! Let me know if there are any additional use cases that you feel should be included.

Thanks so much! 

Checklist
-----------
Please make sure you have finished the following tasks before requesting a review of the pull request (PR). For more information, please refer to the [Contributor's Guide](../blob/master/CONTRIBUTING.md):

- [x] Contributions to mitiq should follow the [pep8 style](https://www.python.org/dev/peps/pep-0008/). You can enforce it easily with [`black`](https://black.readthedocs.io/en/stable/index.html) style and with [`flake8`](http://flake8.pycqa.org) conventions.
- [x] Please add tests to cover your changes, if applicable, and make sure that all new and existing tests pass locally.
- [x] If the behavior of the code has changed or new feature has been added, please also [update the documentation](../blob/master/docs/CONTRIBUTING_DOCS.md).
- [x] Functions and classes have useful [Google-style](https://sphinxcontrib-napoleon.readthedocs.io/en/latest/example_google.html) docstrings and [type hints](https://www.python.org/dev/peps/pep-0484/) in the signature of the objects.
- [x] (Bug fix) The associated issue is referenced above using [auto-close keywords](https://help.github.com/en/articles/closing-issues-using-keywords). If the PR fixes an issue, use the keyword fix/fixes/fixed followed by the issue id. For example, if the PR fixes issue 1184, type "Fixes #1184" (without quotes).
- [x] The [changelog](https://github.com/unitaryfund/mitiq/blob/master/CHANGELOG.md) is updated, including author and PR number (@username, gh-xxx).

PR #584, opened 2021-03-08
Replace Qiskit utils with Qiskit executors from docs (wrapup)

### LINKED ISSUE #379: Replace Qiskit utils with Qiskit executors from docs
The code in [qiskit_utils.py](https://github.com/unitaryfund/mitiq/blob/master/mitiq/mitiq_qiskit/qiskit_utils.py) is stale, not used for anything but testing, and not general enough to be useful as a utility. In my opinion, the [Qiskit executors](https://github.com/unitaryfund/mitiq/blob/master/docs/source/guide/guide-executors.rst#qiskit-executors) in the docs would be much better to include as a utility.

Aaron-Robertson:
Description
-----------
Resolves #379: Replace Qiskit utils with Qiskit executors from docs (as wrapup)

From @pchung39:

> This pull request is for #379 . I've transferred the Qiskit executors referenced in the back-end executors documentation (Back-End Executors) into /mitiq_qiskit/qiskit_utils.py. There were two existing function in qiskit_utils.py named random_one_qubit_identity_circuit and run_with_noise that I left untouched in the file.
> 
> I also included a few tests for the added executors. They are pretty straight-forward but should cover basic functionality! Let me know if there are any additional use cases that you feel should be included.
> 
> Thanks so much!

Addresses the review on source pull request (#463) and add the Qiskit noisy density matrix simulation without sampling.

Additionally makes executors more general by accepting as arg a Qiskit noise model and adds an initialized depolarizing noise method to recreate the depolarizing examples.

Checklist
-----------


Aaron-Robertson:
@rmlarose Doesn't look like I have access to request a reviewer, but this one is ready when you get a chance (to either take a look or shoot out a couple requests).

We could look into more generic approaches to cut down on redundant code—though potentially better as future enhancement.

mitiq/mitiq_qiskit/qiskit_utils.py
```diff
@@ -87,3 +89,182 @@ def run_with_noise(
...
+        # we want all gates to be actually applied,
+        # so we skip any circuit optimization
+        optimization_level=0,
+        shots=shots,
+    )
+    results = job.result()
+    counts = results.get_counts()
+    expectation = 0
+    # classical bits are included in bitstrings with a space
+    # this is what breaks if you have them
```
rmlarose, line 152: I need to test this locally, I'm not sure this comment applies anymore. Even if it does, it would be better to remove the measurements from the circuit and run the rest of the function if possible.

mitiq/mitiq_qiskit/qiskit_utils.py
```diff
@@ -87,3 +89,182 @@ def run_with_noise(
...
+        obs: The observable to measure as a NumPy array.
+        noise: The depolarizing noise strength as a float, i.e. 0.001 is 0.1%.
+
+    Returns:
+        The expectation value of obs as a float.
+    """
+    if len(circ.clbits) > 0:
+        raise ValueError(
+            "This executor only works on programs with no classical bits."
+        )
```
rmlarose, line 175: Similarly here.

mitiq/mitiq_qiskit/qiskit_utils.py
```diff
@@ -87,3 +89,182 @@ def run_with_noise(
...
+               i.e. 0.001 is 0.1%.
+        shots: The number of measurements.
+
+    Returns:
+        The expectation value of obs as a float.
+    """
+    if len(circ.clbits) > 0:
+        raise ValueError(
+            "This executor only works on programs with no classical bits."
+        )
```
rmlarose, line 229: And here.

mitiq/mitiq_qiskit/tests/test_qiskit_utils.py
```diff
@@ -0,0 +1,187 @@
+# Copyright (C) 2020 Unitary Fund
```
rmlarose, line 1:
```suggestion
# Copyright (C) 2021 Unitary Fund
```

mitiq/mitiq_qiskit/tests/test_qiskit_utils.py
```diff
@@ -0,0 +1,187 @@
...
+from qiskit import QuantumCircuit
+
+from mitiq.mitiq_qiskit.qiskit_utils import (
+    execute,
+    execute_with_shots,
+    execute_with_depolarizing_noise,
+    execute_with_shots_and_depolarizing_noise,
+)
+
+OBSERVABLE = np.array([[1, 0], [0, 0]])
```
rmlarose, line 28:
```suggestion
ONE_QUBIT_OBSERVABLE = np.array([[1, 0], [0, 0]])
```

mitiq/mitiq_qiskit/tests/test_qiskit_utils.py
```diff
@@ -0,0 +1,187 @@
...
+TWO_QUBIT_OBSERVABLE = np.diag([1, 0, 0, 0])
+SHOTS = 1000
+
+
+def test_execute():
+    """ Tests the Qiskit waveform simulation executor returns
+    appropriate expectation value given an observable
+    """
```
rmlarose, line 36: Style nit for consistency. Remove initial space and end with a period.

rmlarose:
```suggestion
    """Tests the Qiskit waveform simulation executor returns
    appropriate expectation value given an observable.
    """
```

mitiq/mitiq_qiskit/tests/test_qiskit_utils.py
rmlarose: Perhaps even better to call these `ONE_QUBIT_GS_PROJECTOR` and `TWO_QUBIT_GS_PROJECTOR` to make the tests more readable. Ignore if they're too long.

mitiq/mitiq_qiskit/tests/test_qiskit_utils.py
rmlarose: Here and elsewhere on unit test docstrings.

mitiq/mitiq_qiskit/tests/test_qiskit_utils.py
```diff
@@ -0,0 +1,187 @@
...
+    assert 1.0 == expected_value
+
+    second_circ = QuantumCircuit(1)
+    second_circ.x(0)
+    expected_value = execute(second_circ, obs=OBSERVABLE)
+    assert 0.0 == expected_value
+
+
+def test_execute_with_shots():
+    """ Tests the Qiskit waveform sampling simulation executor returns
```
rmlarose, line 49:
```suggestion
    """ Tests the Qiskit wavefunction sampling simulation executor returns
```

mitiq/mitiq_qiskit/tests/test_qiskit_utils.py
```diff
@@ -0,0 +1,187 @@
...
+        execute_with_shots(circ=circ, obs=OBSERVABLE, shots=SHOTS)
+
+
+def test_execute_with_depolarizing_noise_single_qubit():
+    """ Tests the noisy sampling executor across increasing levels
+    of single qubit gate noise
+    """
+
+    single_qubit_circ = QuantumCircuit(1)
+    single_qubit_circ.z(0)
```
rmlarose, line 87: This is to make sure there is noise, right?

mitiq/mitiq_qiskit/tests/test_qiskit_utils.py
```diff
@@ -0,0 +1,187 @@
...
+
+    noiseless_exp_value = 1.0
+
+    for noise in [0.01, 0.1, 0.2, 1.0]:
```
rmlarose, line 91: I think it's fine to just test one noise value.

mitiq/mitiq_qiskit/tests/test_qiskit_utils.py
```diff
@@ -0,0 +1,187 @@
...
+    """ Tests the noisy sampling executor across increasing levels of
+    two qubit gate noise
+    """
+
+    two_qubit_circ = QuantumCircuit(2)
+    two_qubit_circ.cx(0, 1)
+
+    noiseless_exp_value = 1.0
+
+    for noise in [0.01, 0.1, 0.2, 1.0]:
```
rmlarose, line 110: Also here, fine to just test one value.

rmlarose (changes requested):
Looking really good @Aaron-Robertson. Mostly style comments, and I want to see about removing the no classical bits requirements which I need to test locally.

Also, the `guide-executors.rst` needs to be edited. Could you remove all the functions in the Qiskit section that you're now defining in `qiskit_utils.py`? You can leave the "Qiskit executors" header and just say "See `mitiq_qiskit.qiskit_utils` for example executors you can use on Qiskit circuits" or something similar.

rmlarose:
One more thought: Right now we restrict to depolarizing noise, but it's really easy to generalize this. What about accepting a noise model as an argument to the noisy executors? E.g.

```python
def execute_with_noise(
    circ: QuantumCircuit, obs: np.ndarray, noise_model: Something, noise: float
) -> float:
```

where `Something` is whatever the noise model types are in Qiskit, or string specifiers if there are no such objects. It could default to depolarizing noise.

andreamari:
> One more thought: Right now we restrict to depolarizing noise, but it's really easy to generalize this. What about accepting a noise model as an argument to the noisy executors? E.g.
> 
> ```python
> def execute_with_noise(
>     circ: QuantumCircuit, obs: np.ndarray, noise_model: Something, noise: float
> ) -> float:
> ```
> 
> where `Something` is whatever the noise model types are in Qiskit, or string specifiers if there are no such objects. It could default to depolarizing noise.

If we add this function, I would remove the `noise` argument since it would be inside `noise_model`.

andreamari (changes requested):
There is a function in qiskit_utils, called `run_with_noise` which could be confused with executors but it is actually used only in tests. I would propose:

- to rename it to `_run_with_noise`
- Maybe to change its body by simply calling one of the new executors. (it is just a particular case of them).

rmlarose:
> There is a function in qiskit_utils, called `run_with_noise` which could be confused with executors but it is actually used only in tests. I would propose:
> 
> * to rename it to `_run_with_noise`
> * Maybe to change its body by simply calling one of the new executors. (it is just a particular case of them).

Good call Andrea. This function was only ever used for tests (and so should have been in the test file), so I would say it's even better to remove this altogether and replace its usage in unit tests with one of the new functions.

Aaron-Robertson:
Looks like the random_one_qubit_identity_circuit 

> > There is a function in qiskit_utils, called `run_with_noise` which could be confused with executors but it is actually used only in tests. I would propose:
> > 
> > * to rename it to `_run_with_noise`
> > * Maybe to change its body by simply calling one of the new executors. (it is just a particular case of them).
> 
> Good call Andrea. This function was only ever used for tests (and so should have been in the test file), so I would say it's even better to remove this altogether and replace its usage in unit tests with one of the new functions.

Agree all around and thanks both @rmlarose and @andreamari for the feedback!

What about the `random_one_qubit_identity_circuit`? That one is also only currently used in tests and shares a name with a pyquil equivalent that lives in `test_zne_mitiq_pyquil.py`.

I'll move it into `test_zne_mitiq_qiskit.py` if there's no dissent, but should we plan for a more general approach?

Aaron-Robertson:
> > One more thought: Right now we restrict to depolarizing noise, but it's really easy to generalize this. What about accepting a noise model as an argument to the noisy executors? E.g.
> > ```python
> > def execute_with_noise(
> >     circ: QuantumCircuit, obs: np.ndarray, noise_model: Something, noise: float
> > ) -> float:
> > ```
> > 
> > 
> > where `Something` is whatever the noise model types are in Qiskit, or string specifiers if there are no such objects. It could default to depolarizing noise.
> 
> If we add this function, I would remove the `noise` argument since it would be inside `noise_model`.

Hmm, I agree with the concept but I'm not sure of end-user cases just yet. I assume end-user would generally have a noise model built/customized, so it won't be a problem to pass into the executor?

Maybe we want to check if the noise model `is_ideal()`[[1]](https://qiskit.org/documentation/_modules/qiskit/providers/aer/noise/noise_model.html) and if it is raise an error.

Or, keep the check and if there's no noise use sensible default. One possibility for default could be `from_backend()`[[2]](https://qiskit.org/documentation/apidoc/aer_noise.html)[[3]](https://qiskit.org/documentation/stubs/qiskit.providers.aer.noise.NoiseModel.html#qiskit.providers.aer.noise.NoiseModel.from_backend)?

Otherwise and maybe for a future issue, we could make an initialization method as a general approach, but that could get complicated with the variety of quantum error types/params[[4]](https://qiskit.org/documentation/tutorials/simulators/3_building_noise_models.html).

For now, I'm including a one off depolarizing initializer for testing, which could also be a (probably temporary) default:
```
def initialize_depolarizing_noise(noise: float) -> NoiseModel:
    # initialize a qiskit noise model
    noise_model = NoiseModel()

    # we assume the same depolarizing error for each
    # gate of the standard IBM basis
    noise_model.add_all_qubit_quantum_error(
        depolarizing_error(noise, 1), ["u1", "u2", "u3"]
    )
    noise_model.add_all_qubit_quantum_error(
        depolarizing_error(noise, 2), ["cx"]
    )
    return noise_model
```

mitiq/mitiq_qiskit/qiskit_utils.py
Aaron-Robertson: The bitstrings still return with a space, but easy enough to just do `expectation += eigvals[int(bitstring[0:circ.num_qubits], 2)] * count / shots`?

rmlarose: If this approach works I would much prefer it to throwing an error whenever there are bits in the circuit!

rmlarose:
> What about the `random_one_qubit_identity_circuit`? That one is also only currently used in tests and shares a name with a pyquil equivalent that lives in `test_zne_mitiq_pyquil.py`.
> 
> I'll move it into `test_zne_mitiq_qiskit.py` if there's no dissent, but should we plan for a more general approach?

+1 for moving into tests (for now). We need to have a general discussion about where benchmarking circuits live. My preference is for them to be in `mitiq.benchmarks`, otherwise there is a lot of duplication in `mitiq_qiskit`, `mitiq_pyquil`, etc. This can be handled separately from this PR though.

mitiq/mitiq_qiskit/tests/test_qiskit_utils.py
Aaron-Robertson: Yep that's my interpretation, though I didn't add the line so not 100% certain. Tests fail without the gates.

Aaron-Robertson:
As far as latest pipeline failure, haven't been able to get both `black` and `flake8` happy with the spacing on `:` – `eigvals[int(bitstring[0:circ.num_qubits], 2)] * count / shots (flake8)` vs `eigvals[int(bitstring[0 : circ.num_qubits], 2)] * count / shots (black)`. One of them take priority?

mitiq/mitiq_qiskit/tests/test_qiskit_utils.py
rmlarose: I think that's correct because the noise model is defined on gates, so if there are no gates there is no noise, hence the failing test. 

Might be good to add a comment summarizing this for posterity.

Aaron-Robertson:
> Also, the `guide-executors.rst` needs to be edited. Could you remove all the functions in the Qiskit section that you're now defining in `qiskit_utils.py`? You can leave the "Qiskit executors" header and just say "See `mitiq_qiskit.qiskit_utils` for example executors you can use on Qiskit circuits" or something similar.

Should we replace the body with the tests showing how they get used? I suppose they are pretty self-explanatory in the code, but I worry that the readme would be too text-filled and sparse on example without anything.

mitiq/mitiq_qiskit/tests/test_zne_mitiq_qiskit.py
```diff
@@ -55,10 +82,64 @@ def measure(circuit, qid) -> QuantumCircuit:
...
+# TODO: Troubleshoot the problems in this executor.
+#       Causes test failures and time increase to 40+ mins
+# def qiskit_executor(qp: QPROGRAM, shots: int = 500) -> float:
+#   expectation = execute_with_shots_and_noise(
+#         qp,
+#         shots=shots,
+#         obs=ONE_QUBIT_GS_PROJECTOR,
+#         noise_model=initialized_depolarizing_noise(BASE_NOISE)
+#     )
+#   return expectation
```
Aaron-Robertson, line 140:
```suggestion
def qiskit_executor(qp: QPROGRAM, shots: int = 500) -> float:
    expectation = execute_with_shots_and_noise(
        qp,
        shots=shots,
        obs=ONE_QUBIT_GS_PROJECTOR,
        noise_model=initialized_depolarizing_noise(BASE_NOISE),
        seed=1,
    )
    return expectation
```
@rmlarose could you take a look at this executor locally and provide some feedback? I'm not totally sure where I'm going wrong. I included `seed_simulator=1` in `execute_with_shots_and_noise` and that passes all tests, should we add it as an arg like in the previous `run_with_noise`?

That said, the test run still took ~40 minutes. Still troubleshooting that now, but I assume I'm missing something that could be obvious.

Aaron-Robertson: Oh, just saw that @andreamari is the main reviewer here, so tagging in as well. I see that the time increase is coming from the projection with so many shots (qiskit_utils line 167-168):
```
    # we need to modify the circuit to measure obs in its eigenbasis
    # we do this by appending a unitary operation
    # obtains a U s.t. obs = U diag(eigvals) U^dag
    eigvals, U = np.linalg.eigh(obs)
    circ.unitary(np.linalg.inv(U), qubits=range(circ.num_qubits))
```
I'm unsure what the solution should be, but I have a test notebook if anyone wants it.

Thanks!

andreamari:
@Aaron-Robertson, thanks for your changes! I think we are converging. I write here my comments:

1. I see that you removed the previous executors with depolarizing noise and kept only the generic executor based on the `noise_model` argument. I initially expected that the old depolarizing executors were still available.  But maybe your solution makes sense. This would be ok for me but I ask @rmlarose to double check this choice. 

2. If we use only executors with a generic `noise_model`, we should remove the adjective _depolarizing_ in the corresponding sub-section in `guide-executors.rst`.

3. In the docs, it would be helpful to replace the pure text executor names with a link (sphinx cross-reference). E.g. something like this:
`mitiq_qiskit.qiskit_utils.execute` -----> :func:`mitiq.mitiq_qiskit.qiskit_utils.execute`

4. About the option of writing code snippets showing how to use the qiskit executors (question asked above by @Aaron-Robertson). I would vote for not doing it to keep the docs compact (considering also that this section is supposed to change soon).

5. Finally, about the problem of slow qiskit executors. Thanks for finding this out @Aaron-Robertson! Could you please open a new issue about this fact by just copying what you have already written in the comments? This issue is quite important, but is unrelated to the aim of this PR so we don't need to fix it here.

Aaron-Robertson:
Created #594 for the `execute_with_shots_and_noise ` issue, so this one is ready for final review!

andreamari (approved):

nathanshammah (approved):
Thanks @Aaron-Robertson. It looks fine to me. I have made some minor suggestions in https://github.com/Aaron-Robertson/mitiq/pull/1 for style in linking the functions with apidoc and for avoiding commenting the test, but just skipping it with pytest.

You could accept those suggestions in the branch and add those commits to this PR, or we could assess those minor edits in a future PR.

Note that as per my comment in #594, the test seem to run fine right now.

nathanshammah:
Thanks @Aaron-Robertson!

PR #590, opened 2021-03-13
Add braket support via rudimentary translator

rmlarose:
Still a bit WIP and needs more tests probably. 

The braket gateset is a subset of the Cirq gateset (except for `CY` and maybe a few others), so the `from_braket` strategy is mostly pure translation. The `to_braket` strategy attempts to translate then decomposes one-qubit and two-qubit ops if no translation is found.

Tips on remaining TODO's are welcome.


rmlarose:
Discussed at Mitiq meeting: I will follow up with the tester of this, if good we will merge to master so this doesn't go stale.

nathanshammah (commented):
Thanks @rmlarose, looks good to me once it's ready. 
Some comments:
- 1. Installation worked fine for me. Note that braket-sdk installs also their `simulator` and `schemas`. Maybe in the future that can be avoided like for cirq-core.  
- 2. Tested the conversions with a braket [getting started](https://amazon-braket-sdk-python.readthedocs.io/en/latest/getting-started.html#getting-started-using-an-amazon-braket-notebook), doing a 2pi-rotation bk/cirq/bk for the bell state circuit and it worked fine: 
```
from mitiq.mitiq_braket.conversions import to_braket, from_braket
from braket.circuits import Circuit
# create the circuit
bell = Circuit().h(0).cnot(0, 1) 
cirq_bell = from_braket(bell)
converted_bell = to_braket(cirq_bell)
bell == converted_bell
```
- 3. Right now I could find only a PennyLane/Braket [plugin](https://github.com/aws/amazon-braket-pennylane-plugin-python), and the BK->PL->Cirq option, even when available from pennylane-cirq, seems too convoluted to be ideal. In the future it may reduce tech debt to outsource as many conversions as possible, I guess.  
- 4. Maybe in the BK/PL plugin [translation](https://github.com/aws/amazon-braket-pennylane-plugin-python/blob/main/src/braket/pennylane_plugin/translation.py) there is some helpful hint for the TODOs of mitiq conversions.
- 5. I think that a general "conversions" example notebook in `examples` could be very handy, happy to write it up if it sounds like a good idea, in another PR.

rmlarose:
> 1. I think that a general "conversions" example notebook in the `examples` could be very handy, happy to write it up if it sounds like a good idea, in another PR, for the multiple circuits we support.

I don't think we should advertise this at all, actually. The reason we convert circuits is entirely an implementation detail, and the purpose of Mitiq is to implement error mitigation techniques, not convert circuits.

The thing to advertise is that Mitiq works the same for any supported circuit type. So I would say we should add braket examples to the docs after this PR, not docs on conversions.

rmlarose:
Documenting reasons (discussed @ Mitiq meeting) behind merging this now:

> I will follow up with the tester of this, if good we will merge to master so this doesn't go stale.

1. Response of tester was that there were no errors with the circuit conversions, but they weren't trying super exotic circuits - mostly just Pauli rotations and XX gates.

> Thanks @rmlarose, looks good to me once it's ready.

2. Decided to leave remaining TODO's as an issue for unitaryHACK. (New issue forthcoming.)

PR #579, opened 2021-03-03
fixes #577 and #578

crazy4pi314:
Description
-----------
Hopefully this fixes the greating bot and the release PyPI workflow, but needs to test with some new accounts.

Checklist
-----------
Please make sure you have finished the following tasks before requesting a review of the pull request (PR). For more information, please refer to the [Contributor's Guide](../blob/master/CONTRIBUTING.md):

- [x] Contributions to mitiq should follow the [pep8 style](https://www.python.org/dev/peps/pep-0008/). You can enforce it easily with [`black`](https://black.readthedocs.io/en/stable/index.html) style and with [`flake8`](http://flake8.pycqa.org) conventions.
- [x] Please add tests to cover your changes, if applicable, and make sure that all new and existing tests pass locally.
- [x] If the behavior of the code has changed or new feature has been added, please also [update the documentation](../blob/master/docs/CONTRIBUTING_DOCS.md).
- [x] Functions and classes have useful [Google-style](https://sphinxcontrib-napoleon.readthedocs.io/en/latest/example_google.html) docstrings and [type hints](https://www.python.org/dev/peps/pep-0484/) in the signature of the objects.
- [x] (Bug fix) The associated issue is referenced above using [auto-close keywords](https://help.github.com/en/articles/closing-issues-using-keywords). If the PR fixes an issue, use the keyword fix/fixes/fixed followed by the issue id. For example, if the PR fixes issue 1184, type "Fixes #1184" (without quotes).
- [ ] The [changelog](https://github.com/unitaryfund/mitiq/blob/master/CHANGELOG.md) is updated, including author and PR number (@username, gh-xxx).

PR #597, opened 2021-03-16
Supresses Test PyPI nightly upload on forks

### LINKED ISSUE #581: Suppress TestPyPI action from forks
Pre-Report Checklist
--------------------

- [x] I checked to make sure that this bug has not already been reported

Issue Description
-----------------

The TestPyPI-nightly action tries to update Mitiq also from forks, such as mine, `nathanshammah/mitiq`.  

How to Reproduce
----------------

See https://github.com/nathanshammah/mitiq/actions

Proposed solution
------------------

I see two possible options:
- Add these release actions to the gitignore, if possible for actions. 
- Insert a condition for the triggering event (runs "on") in the `.yml` files related to release actions (gh-release.yml, greeting.yml, pre-release.yml, publish-pypi.yml, publish-testpypi.yml), to run only from specific branches (unitaryfund/*).

purva-thakre:
Description
-----------
Fixes #581

When actions are enabled on a fork, Test PyPI tries to deploy from this fork instead of just the master. I have added a conditional statement for `deploy` to run in `publish-pypi.yml` and `publish-testpypi.yml` only when current branch is `master`. This is based on discussion at [this link](https://docs.github.com/en/actions/reference/environment-variables#determining-when-to-use-default-environment-variables-or-contexts). 

I am unsure if this worked or not because running the `deploy` workflow in a fork with the added condition still leads to same errors as before. This seems to be a WIP for the main repo [here](https://github.com/unitaryfund/mitiq/pull/579). So far, disabling the workflow in a fork seems to have solved issue #581. 

Feel free to close this PR if this ought to be worked on after the new PyPI workflow is released.

### DIFF
--- a/.github/workflows/gh-release.yml
+++ b/.github/workflows/gh-release.yml
@@ -10,6 +10,7 @@ jobs:
   #generate_artifacts:
   # TODO: make docs pdf and tarball
   create_release:
+    if: github.repository_owner == 'unitaryfund'
     runs-on: ubuntu-latest
     if: github.repository_owner == 'unitaryfund'
     steps:
--- a/.github/workflows/gh-release.yml
+++ b/.github/workflows/gh-release.yml
@@ -12,7 +12,6 @@ jobs:
   create_release:
     if: github.repository_owner == 'unitaryfund'
     runs-on: ubuntu-latest
-    if: github.repository_owner == 'unitaryfund'
     steps:
       - name: Check out Mitiq
         uses: actions/checkout@v2
--- a/.github/workflows/gh-release.yml
+++ b/.github/workflows/gh-release.yml
@@ -1,7 +1,7 @@
 on:
   push:
     branches: # specify branch to Mitiq master not forks
-      - master
+      - release
     # Sequence of patterns matched against refs/tags
     tags:
       - 'v*' # Push events to matching v*, i.e. v1.0, v20.15.10
--- a/.github/workflows/gh-release.yml
+++ b/.github/workflows/gh-release.yml
@@ -1,5 +1,7 @@
 on:
   push:
+    branches: # specify branch to Mitiq master not forks
+      - master
     # Sequence of patterns matched against refs/tags
     tags:
       - 'v*' # Push events to matching v*, i.e. v1.0, v20.15.10
--- a/.github/workflows/gh-release.yml
+++ b/.github/workflows/gh-release.yml
@@ -1,6 +1,6 @@
 on:
   push:
-    branches: # specify branch to Mitiq master not forks
+    branches: # specify branch to Mitiq release, not running from forks
       - release
     # Sequence of patterns matched against refs/tags
     tags:
--- a/.github/workflows/gh-release.yml
+++ b/.github/workflows/gh-release.yml
@@ -11,6 +11,7 @@ jobs:
   # TODO: make docs pdf and tarball
   create_release:
     runs-on: ubuntu-latest
+    if: github.repository_owner == 'unitaryfund'
     steps:
       - name: Check out Mitiq
         uses: actions/checkout@v2
--- a/.github/workflows/publish-pypi.yml
+++ b/.github/workflows/publish-pypi.yml
@@ -9,7 +9,7 @@ on:
 
 jobs:
   deploy:
-if: github.repository_owner == 'unitaryfund'
+    if: github.repository_owner == 'unitaryfund'
     runs-on: ubuntu-latest
     steps:
     - name: Check out mitiq
--- a/.github/workflows/publish-pypi.yml
+++ b/.github/workflows/publish-pypi.yml
@@ -14,8 +14,6 @@ jobs:
     steps:
     - name: Check out mitiq
       uses: actions/checkout@v2
-      with:
-        ref: 'release'
     - name: Set up Python
       uses: actions/setup-python@v2
       with:
--- a/.github/workflows/publish-pypi.yml
+++ b/.github/workflows/publish-pypi.yml
@@ -1,7 +1,7 @@
 # This workflow will upload a Python Package using Twine when a release is created
 # For more information see: https://help.github.com/en/actions/language-and-framework-guides/using-python-with-github-actions#publishing-to-package-registries
 
-name: Upload Python Package to Test PyPI
+name: Upload Release Package to PyPI
 
 on:
   release:
--- a/.github/workflows/publish-pypi.yml
+++ b/.github/workflows/publish-pypi.yml
@@ -11,7 +11,8 @@ jobs:
   deploy:
 
     runs-on: ubuntu-latest
-    if: github.repository_owner == 'unitaryfund'
+    # if: github.repository_owner == 'unitaryfund'
+    # above condition is made a comment until needed to specify main repo even when master has been specified for release branch. 
     steps:
     - name: Check out mitiq
       uses: actions/checkout@v2
--- a/.github/workflows/publish-pypi.yml
+++ b/.github/workflows/publish-pypi.yml
@@ -14,6 +14,8 @@ jobs:
     steps:
     - name: Check out mitiq
       uses: actions/checkout@v2
+      with:
+        ref: 'release'
     - name: Set up Python
       uses: actions/setup-python@v2
       with:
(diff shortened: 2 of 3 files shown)


nathanshammah (commented):
Thanks @purva-thakre. I think this is in the right direction, however my limited understanding of "yaml-git-github actions' is that `refs/heads/master` would run from master branches of forks too. Indeed if I locally do from my fork
```bash
>>> cat .git/HEAD
```
it prints
```bash
refs/heads/master
```
My understanding is that we'd need something to also exactly ping `unitaryfund/mitiq` as org/repo location, possibly jointly with this check on the master branch.

purva-thakre:
@nathanshammah I'll spend some time figuring out `Git refs`. Until then I am changing this to a draft.

nathanshammah:
thanks!

purva-thakre:
Sorry the commits are a mess for this. I was using `git pull` in order to update my branches which lead to new changes being merged ! I have figured out `rebase` is a better option in order to update branches.

purva-thakre:
@nathanshammah New commit specifies the owner of the repo as `unitaryfund`. Other option is to specify the repo as `unitaryfund/mitiq`. At [this link](https://docs.github.com/en/actions/reference/context-and-expression-syntax-for-github-actions#github-context), there's two ways to specify the path. 
- `github.repository` is used to specify the exact repository + owner
-  `github.repository_owner` directly specifies the owner. Since this workflow seems to run on a fork's `master` as well, I felt a conditional statement of owner of `master` would be enough.  

If both fail to work, then I could try `ignore` for all branches except `master`. For more info, see [this link](https://docs.github.com/en/actions/reference/workflow-syntax-for-github-actions#excluding-branches-and-tags).

nathanshammah (changes requested):
@purva-thakre thanks.
On the condition: I would have thought there was a way to check the condition at the "on" stage, so that it does not start at all otherwise. Is it possible to fix that?

Git picked up changes to files that do not seem to belong to this PR. Can you please remove them?

purva-thakre:
Sure, I'll move the condition to the `on` stage. 
> Git picked up changes to files that do not seem to belong to this PR. Can you please remove them?

Yes, I have been trying to figure out the best way to remove changes that do not belong to this PR.

purva-thakre:
@nathanshammah I have added a branch condition from which `release` branch is built. Since `publish-pypi` uses `release` branch as a reference, I did not feel this needed an additional condition for `master` branch to be specified. 

For the file that runs on a schedule, it seems[ I cannot add a specific branch](https://github.community/t/scheduled-builds-of-non-default-branch/16306/9) at the `on` stage. So, I have added a reference to the `release` branch in this file as well. 

The previous conditional statement about `repo owner` has been commented out for now, just in case these new changes fail to work.

nathanshammah (changes requested):
Hi @purva-thakre, my understanding is that fixing the branch to master would make the gh release be potentially done also on remote master branches from forks.

Also, my understanding is that the daily `TestPyPI`, once fully enabled, should run on master, not on the release branch, which is used just for the PyPI releases. 

Given these facts, and your findings that it seems to be not possible to put a condition `on` for the owner of the repository for the action to be triggered, my suggestion is to just revert to your previous proposal, with the if statement on the repository owner. This would make all the involved actions run on potentially also on forked branches, but then things should not fail in those cases.

purva-thakre:
@nathanshammah I propose having both `master` condition at `on` stage and an `if` condition specifying repo owner in `gh-release`. I was unsure of `master` at the `on` stage because there's no way to specify a non-fork `master` besides directly referencing the name of the repo, `head` ref or a repo owner. It might be good to have `release` branch created from Unitary Fund's `master` only. 

I thought `TestPyPi` was supposed to run on `release` branch based on [this comment](https://github.com/unitaryfund/mitiq/blob/3b3d8865eaa6e533a0c66c31c4d51dc13414bdf7/.github/workflows/publish-testpypi.yml#L1) made at the top of the file. Is this just a remnant of a copy and paste from `publish-pypi` ?

nathanshammah:
@purva-thakre your solution looks fine to me, about having `on: master` and the `if: owner`. I suggest deleting comments on lines 15 etc. if not implemented.

Good catch on the `release` comment. I think it is a remnant from PyPI. We'd use the release branch just for the official releases, and every night push to TestPyPI the current master. Pinging @crazy4pi314 just to be sure.

purva-thakre:
@nathanshammah @crazy4pi314  

Here's a list of the changes made : 
- In `gh-release.yml`, the workflow is set to run only on the `master` branch in addition to other conditions that were added before this PR. Since there's an issue of the workflow running on a fork's master, an additional condition of repo owner is added before jobs that will run. 
- As `publish-pypi.yml` is conditioned to run only on the `release` branch, there's no need to add an extra condition as long as `release` branch was created from Unitary Fund's `master`. 
- There's a condition added in `publish-testpypi.yml` for repo owner. Because `schedule` is supposed to be triggered after [`commit on default branch`](https://docs.github.com/en/actions/reference/events-that-trigger-workflows#schedule), the added condition makes sure this workflow does not run on a fork's master. A reference to the `release` branch condition has been removed as well.

.github/workflows/gh-release.yml
```diff
@@ -1,5 +1,7 @@
 on:
   push:
+    branches: # specify branch to Mitiq master not forks
+      - master
```
crazy4pi314, line 4: This should be the 'release' branch if I understand correctly

.github/workflows/gh-release.yml
```diff
@@ -9,6 +11,7 @@ jobs:
   # TODO: make docs pdf and tarball
   create_release:
```
crazy4pi314, line 12:
```suggestion
  create_release:
    if: github.repository_owner == 'unitaryfund'
```

.github/workflows/gh-release.yml
```diff
@@ -9,6 +11,7 @@ jobs:
...
     runs-on: ubuntu-latest
+    if: github.repository_owner == 'unitaryfund'
```
crazy4pi314, line 14:
```suggestion

```

crazy4pi314 (changes requested):
I think the repo owner solution should work, and should disable anything related to deployment to avoid those workflows from failing. There were a few references to release swapped for master which should be reverted as final releases are made from `release`

crazy4pi314:
I can't seem to make a comment on a part of a file that wasn't changed, but I think the `.github/workflows/publish-pypi.yml` file has the wrong `name:` field at the top. Currently it says `name: Upload Python Package to Test PyPI` when it should be `name: Upload Release Package to PyPI`

.github/workflows/gh-release.yml
purva-thakre: I guess I was confused about how `release` branch is created. I thought this workflow was going to create a `release` branch :sweat_smile: !

purva-thakre:
@nathanshammah @crazy4pi314 Just wanted to check if either of you had a chance to take a look at the new changes ?

.github/workflows/publish-pypi.yml
```diff
@@ -1,22 +1,19 @@
...
 
 jobs:
   deploy:
-
+    if: github.repository_owner == 'unitaryfund'
     runs-on: ubuntu-latest
-
     steps:
     - name: Check out mitiq
       uses: actions/checkout@v2
```
crazy4pi314, line 16: This has to be here I think as it instructs the release process to use the latest commit on the release branch.
```suggestion
      uses: actions/checkout@v2
      with:
        ref: 'release'
```

.github/workflows/publish-testpypi.yml
```diff
@@ -6,12 +6,12 @@ name: Nightly Upload to Test PyPI
 on:
   schedule:
     - cron: '0 0 * * *'
+
```
crazy4pi314, line 9: Don't need an extra line here I think

```suggestion
```

crazy4pi314 (approved):
LGTM! Thanks for sorting this out @purva-thakre 💖

.github/workflows/gh-release.yml
```diff
@@ -1,5 +1,7 @@
 on:
   push:
+    branches: # specify branch to Mitiq master not forks
```
nathanshammah, line 3:
```suggestion
    branches: # specify branch to Mitiq release, not running from forks
```

nathanshammah (approved):
Thanks @purva-thakre. Added a small comment. LGTM, thanks for working on it thoroughly, it ended up being a bit tricky in the end. Let's hope it works.

PR #601, opened 2021-03-17
Clifford Data Regression Methods: Training

L-P-B:
Description
-----------
Second try at pull request 520:

Implementing CDR error mitigation methods first presented in Czarnik et al. (2020) https://arxiv.org/abs/2005.10189 and Lowe et al. (2020) https://arxiv.org/abs/2011.01157.

If you have not finished all tasks in the checklist, you can also open a [Draft Pull Request](https://github.blog/2019-02-14-introducing-draft-pull-requests/) to let the others know this on-going work and keep this checklist in the PR description.

Checklist
-----------

- [ ] Contributions to mitiq should follow the [pep8 style](https://www.python.org/dev/peps/pep-0008/). You can enforce it easily with [`black`](https://black.readthedocs.io/en/stable/index.html) style and with [`flake8`](http://flake8.pycqa.org) conventions.
- [ ] Please add tests to cover your changes, if applicable, and make sure that all new and existing tests pass locally.


mitiq/cdr/clifford_training_data.py
```diff
@@ -0,0 +1,516 @@
...
+    **additional_options: dict,
+) -> Tuple[List[Circuit], List[List[float]], List[List[float]]]:
+    """Function to return a list of near-Clifford circuits to create the
+    training data.
+
+    Args:
+        circuit: A circuit of interest.
+        num_training_circuits: Number of circuits in the returned training set,
+                               assumes already compiled into gate set
+                               (Rz, Rx, Z, X, CNOT)
```
rmlarose, line 28: This is misplaced, right? The `circuit` is assumed to be compiled into this gate set, no?

mitiq/cdr/clifford_training_data.py
```diff
@@ -0,0 +1,516 @@
...
+        fraction_non_clifford: The (approximate) fraction of non-Clifford
+                               gates in each returned circuit.
+        method_select: option to define the way in which the non-Clifford
+                              gates to replace with Cliffords are chosen can
+                              take strings 'random' or 'probabilistic'.
```
rmlarose, line 33:
```suggestion
                              gates to replace with Cliffords are chosen.
                              Options are 'random' or 'probabilistic'.
```

rmlarose: "random" and "probabilistic" are synonyms. Can we pick better names here? Somewhere in the docstring they should be defined, also.

mitiq/cdr/clifford_training_data.py
```diff
@@ -0,0 +1,516 @@
...
+        method_replace: str = option to define the way the chosen non-Clifford
+                              gates are replace with a Clifford gate can take
+                              strings 'random', 'probabilistic' or 'closest'.
```
rmlarose, line 36: Same comment as above re "random" and "probabilistic"

mitiq/cdr/clifford_training_data.py
```diff
@@ -0,0 +1,516 @@
...
+        random_state: Seed for sampling.
+        additional_options: dictionary with the following keys and values:
```
rmlarose, line 38: It's not a dict but rather a sequence of `key=value` pairs.

mitiq/cdr/clifford_training_data.py
```diff
@@ -0,0 +1,516 @@
...
+            'sigma_select': float -  postitive variable definined width of
```
rmlarose, line 39:
```suggestion
            'sigma_select': float -  postitive variable which defines the width of
```

mitiq/cdr/clifford_training_data.py
```diff
@@ -0,0 +1,516 @@
...
+                                     probability distribution used in choosing
+                                     which non-Cliffords to replace, only has
+                                     an impact if
+                                     method_select = 'probabilistic'.
+            'sigma_replace': float - positive variable definined width of
```
rmlarose, line 44:
```suggestion
            'sigma_replace': float - positive variable which defines the width of
```

mitiq/cdr/clifford_training_data.py
```diff
@@ -0,0 +1,516 @@
...
+                                     probability distribution used in choosing
+                                     which Clifford to replace the
+                                     non-Clifford with, only has an impact if
+                                     method_replace = 'probabilistic'.
+    Returns:
+        List[circ.Circuits]: list of near-Clifford circuits constructed from
```
rmlarose, line 50: circ -> cirq

rmlarose: Or just use `List[Circuit]`. Circuits should not be plural

mitiq/cdr/clifford_training_data.py
```diff
@@ -0,0 +1,516 @@
...
+    # find all the non-Clifford gates:
+    # all_cliff = np.column_stack((not_rz_circ_data, rz_cliff))
+    angles_original_list = []
+    angles_replaced_list = []
+    # setting the seed:
+    if isinstance(random_state, int):
+        random_state = np.random.RandomState(random_state)
+    else:
+        random_state = randint(10 ** (8))
+        random_state = np.random.RandomState(random_state)
```
rmlarose, line 67: The function signature says `random_state` can be a `RandomState` but this isn't checked here. If you input a `RandomState` it will not actually use that but use a different one.

mitiq/cdr/clifford_training_data.py
```diff
@@ -0,0 +1,516 @@
...
+        random_state = random_states[n]
+        random_state = np.random.RandomState(random_state)
+        # Convert data arry into cirq circuit and append it to the storage
+        #  array:
+        if additional_options:
+            (
+                projected_circuit,
+                angles_original,
+                angles_replaced,
+            ) = _map_to_near_clifford(
```
rmlarose, line 82: Great!! Happy to see this function was written :)

mitiq/cdr/clifford_training_data.py
```diff
@@ -0,0 +1,516 @@
...
+    random_state: np.random.RandomState,
+    method_select: str = "random",
+    method_replace: str = "closest",
+    **additional_options: dict,
+) -> Tuple[Circuit, List[float], List[float]]:
+    """Function to take the information in some circuit of interest and
+        return a near-Clifford circuit as constructed according to
+        some user defined methods.
+    Args:
+        circuit: cirq.Circuit object of origional circuit of interest.
```
rmlarose, line 125: Can just say "Circuit of interest" or "Circuit to map to a near-clifford circuit." No need to specify it is a cirq.Circuit because this is in the type annotation.

Should list all assumptions about the circuit here, though. Such as: we assume it only has single-qubit + CNOT gates, right?

mitiq/cdr/clifford_training_data.py
```diff
@@ -0,0 +1,516 @@
...
+    count = 0
+    for o, op in enumerate(operations):
+        if (o + 1) in pos_selected:
+            qubit = op.qubits[0]
+            parameter = rz_non_cliff_selected[count]
+            operation = cirq.ops.rz(parameter)
+            projected_circuit.append(operation(qubit))
+            count += 1
+        else:
+            projected_circuit.append(op)
```
rmlarose, line 219: I thought the point of masking was to be fast with vectorized operations, but here we end up iterating through the operations anyhow. 

Did you consider replacing the operations in a vectorized manner as well? E.g., map `operations -> new_operations` where `new_operations` has some of the non-Clifford ops replaced by Clifford ops? Then you could just `return cirq.Circuit(new_operations)`.

mitiq/cdr/clifford_training_data.py
```diff
@@ -0,0 +1,516 @@
...
+        ang: rotation angle in the Rz gate.
+    Returns:
+        bool: True / False for Clifford or not.
+    """
+    ang = ang % (2 * np.pi)
+    closest_clifford_angle = _closest_clifford(ang)
+    if abs(closest_clifford_angle - ang) < tol:
+        return True
+    else:
+        return False
```
rmlarose, line 392: Remove the else and just return False.

mitiq/cdr/clifford_training_data.py
```diff
@@ -0,0 +1,516 @@
...
+
+    Args:
+        ang: angle in Rz gate.
+
+    Returns:
+        Clifford angle: closest clifford angle.
+    """
+    random_index = random_state.randint(0, 3)
+    clifford_angle = CLIFFORD_ANGLES[random_index]
+    return clifford_angle
```
rmlarose, line 443: `ang` is not used? What is this function supposed to do?

mitiq/cdr/clifford_training_data.py
```diff
@@ -0,0 +1,516 @@
...
+
+
+# vectorize so function can take array:
+_random_clifford = np.vectorize(_random_clifford)
+
+
+def _angle_to_probabilities(angle: float, sigma: float) -> float:
+    """Function to return probability disribtuion based on distance from
```
rmlarose, line 451: typo distribution

mitiq/cdr/clifford_training_data.py
```diff
@@ -0,0 +1,516 @@
...
+        Prob_project = exp(-(dist/sigma)^2) where dist = sum(dists) is the
+        sum of distances from each Clifford gate.
+    """
+    angle = angle % (2 * np.pi)
+    S = np.array([[1, 0.0], [0.0, 1j]])
+    Rz = np.array([[1, 0.0], [0.0, np.exp(angle * 1j)]])
+    dists = []
+    for i in range(4):
+        if i == 0:
+            i = 4
```
rmlarose, line 467: why start at 0 then?

mitiq/cdr/clifford_training_data.py
```diff
@@ -0,0 +1,516 @@
...
+        10 ** (8), size=2
+    )
+    random_state_select = np.random.RandomState(random_state_select)
+    random_state_replace = np.random.RandomState(random_state_replace)
+    # get the operations from the circuit and find the non-cliff angles:
+    operations = np.array(list(circuit.all_operations()))
+    positions = np.linspace(1, len(operations), len(operations))
+    gates = _get_gates(operations)
+    mask = np.array(
+        [isinstance(i, cirq.ops.common_gates.ZPowGate) for i in gates]
```
rmlarose, line 164:
```suggestion
        [isinstance(gate, cirq.ops.common_gates.ZPowGate) for gate in gates]
```

mitiq/cdr/clifford_training_data.py
```diff
@@ -0,0 +1,516 @@
...
+    if (
+        "sigma_select" not in additional_options
+        and "sigma_replace" not in additional_options
+    ):
+        raise Exception(
+            "additional options must be dicitonary with \
+                        keys containing one or both of "
+            '"sigma_select" and "sigma_replace" both \
+                        equal to some positive float'
+        )
```
rmlarose, line 184: Put this at the start of the function or add some spacing to improve readability.

mitiq/cdr/clifford_training_data.py
```diff
@@ -0,0 +1,516 @@
...
+    # set the seed for sampling, for replacement and selection:
+    (random_state_select, random_state_replace) = random_state.randint(
+        10 ** (8), size=2
+    )
+    random_state_select = np.random.RandomState(random_state_select)
+    random_state_replace = np.random.RandomState(random_state_replace)
+    # get the operations from the circuit and find the non-cliff angles:
+    operations = np.array(list(circuit.all_operations()))
+    positions = np.linspace(1, len(operations), len(operations))
+    gates = _get_gates(operations)
```
rmlarose, line 162:
```suggestion
    gates = np.array([op.gate for op in circuit.all_operations()])
```

rmlarose: And remove `operations` above, I don't think you use this anywhere except to get the gates. If so you can remove `_get_gates` below.

mitiq/cdr/clifford_training_data.py
```diff
@@ -0,0 +1,516 @@
...
+    mask = np.array(
```
rmlarose, line 163: Something like `zgates_mask` would be more descriptive.

mitiq/cdr/clifford_training_data.py
```diff
@@ -0,0 +1,516 @@
...
+        [isinstance(i, cirq.ops.common_gates.ZPowGate) for i in gates]
+    )
+    r_z_gates = operations[mask]
+    r_z_positions = positions[mask]
+    angles = _get_arguments(r_z_gates)
+    mask_non_cliff = ~_is_clifford_angle(angles)
+    rz_non_cliff = angles[mask_non_cliff]
+    pos_non_cliff = r_z_positions[mask_non_cliff]
+    rz_non_cliff_copy = rz_non_cliff.copy()
```
rmlarose, line 172: why is this needed?

mitiq/cdr/clifford_training_data.py
```diff
@@ -0,0 +1,516 @@
...
+    for o, op in enumerate(operations):
+        if (o + 1) in pos_selected:
+            qubit = op.qubits[0]
+            parameter = rz_non_cliff_selected[count]
+            operation = cirq.ops.rz(parameter)
+            projected_circuit.append(operation(qubit))
+            count += 1
+        else:
+            projected_circuit.append(op)
+    return projected_circuit, angles_original, angles_replaced
```
rmlarose, line 220: Is it just for debugging to return the original and replaced angles? If so it should be removed.

mitiq/cdr/clifford_training_data.py
```diff
@@ -0,0 +1,516 @@
+import cirq
```
rmlarose, line 1: Add license

rmlarose (changes requested):
Initial comments on just the first two functions.

General suggestion which may simplify our lives: Did you consider defining `is_clifford` and similar functions for operations instead of angles? This would remove the need to extract gates from ops then extract angles from gates etc. and simplify the code a bit.

There is a fair amount of repeated work every time `_map_to_near_clifford` is called in `generate_training_circuits` - namely, the operations to replace are the same but recomputed every time. I know I suggested this approach, but still it should be said. Since this is a helper function I think it should input the recomputed information. You can ignore this for now if you want and we can open an enhancement issue for it later, up to you.

mitiq/cdr/clifford_training_data.py
L-P-B: what do you think about 'uniform' and 'guassian'?

rmlarose: Sounds good to me!

mitiq/cdr/clifford_training_data.py
L-P-B: Yeah thats true thanks for spotting. Are you happy with me just taking an integer argument here for the seed?

mitiq/cdr/clifford_training_data.py
L-P-B: it is not, i removed it, thanks for spotting

mitiq/cdr/clifford_training_data.py
L-P-B: I have re-written this section. I am still looping through some angles to build the operations correctly. I think vectorizing is still works like looping, it is just hidden making the code a little neater.

mitiq/cdr/clifford_training_data.py
L-P-B: Because I want the order to correspond to the order of the tuple CLIFFORD_ANGLES. I could reorder that list as well I suppose, but that does not appear immediately more elegant to me...

mitiq/cdr/clifford_training_data.py
L-P-B: this function returns random angles as long as the array of angles input, hence the argument ang. But its functionality is a little hidden. Thats true.

nathanshammah (commented):
@L-P-B can you add `clifford_training_data` to the API guide, just adding the automodule to [this file](https://github.com/unitaryfund/mitiq/blob/master/docs/source/apidoc.rst)?

nathanshammah:
Tests pass locally from my fork of this branch.

mitiq/cdr/clifford_training_data.py
```diff
@@ -0,0 +1,488 @@
...
+
+
+def _select(
+    rz_non_cliff: np.ndarray,
+    fraction_non_clifford: float,
+    method_select: str,
+    sigma_select: float,
+    random_state: np.random.RandomState,
+) -> np.ndarray:
+    """Function to select the non-Clifford gates to be replace for a given set
```
nathanshammah, line 234:
```suggestion
    """Function to select the non-Clifford gates to be replaced for a given set
```

mitiq/cdr/clifford_training_data.py
```diff
@@ -0,0 +1,488 @@
...
+                       non-Clifford gates to replace: 'uniform' or
+                       'gaussian'.
+        method_replace: string defining method used to replace selected
+                        non-Clifford gates: 'closest', 'uniform' or
+                        'gaussian'.
+        random_state: Seed for sampling.
+        sigma_select: width of probability distribution used in selection
+                      of non-Clifford gates to replace, only has effect if
+                      method_select = 'gaussian'
+        sigma_replace: width of probability distribution used in replacement
```
nathanshammah, line 158: The docstring seems different from the args in the signature, they need to be updated?

mitiq/cdr/clifford_training_data.py
```diff
@@ -0,0 +1,488 @@
...
+    method_select: str,
+    sigma_select: float,
+    random_state: np.random.RandomState,
+) -> np.ndarray:
+    """Function to select the non-Clifford gates to be replace for a given set
+    of  non-Clifford gates.
+
+    Args:
+        rz_non_cliff: array of non-Clifford angles for a circuit of interest.
+        fration_non_clifford: fraction of non-Clifford gates to change.
```
nathanshammah, line 239:
```suggestion
        rz_non_clifford: array of non-Clifford angles for a circuit of interest.
        fration_non_clifford: fraction of non-Clifford gates to change.
```

nathanshammah: Here and elsewhere I would try to be consistent and keep the same kind of abbreviations. Also, it sounds more pythonic to have full word spellings.

mitiq/cdr/clifford_training_data.py
```diff
@@ -0,0 +1,488 @@
...
+        method_select: string specifying method to use to select Clifford gates
+                       to replace can be 'uniform' or 'probabilistic.
+        sigma_select: width of probability distribution used in selection
+                      of non-Clifford gates to replace, only has effect if
+                      method_select = 'gaussian'.
```
nathanshammah, line 244:
```suggestion
        method: {'uniform', 'probabilistic'} method to use to select Clifford gates
                       to replace.
        sigma: width of probability distribution used in selection
                      of non-Clifford gates to replace, only has effect if
                      method_select == 'gaussian'.
```

mitiq/cdr/clifford_training_data.py
L-P-B: Yes, thanks for spotting.

nathanshammah:
@L-P-B please see https://github.com/L-P-B/mitiq/pull/2 with fixes that should fix the tests now failing and improve the style of the docs.

nathanshammah (changes requested):
There is a test failing in the docs due to something in a docstring I guess?
https://github.com/unitaryfund/mitiq/pull/601/checks?check_run_id=2254832392

Also another validation test that should be fixed with black, I guess.

mitiq/cdr/clifford_training_data.py
rmlarose: Yeah this is fine.

mitiq/cdr/clifford_training_data.py
```diff
@@ -0,0 +1,495 @@
...
+            which non-Cliffords to replace, only has
+            an impact if method_select = 'gaussian'.
+            'sigma_replace': float - positive variable which defines the width
+            of probability distribution used in
+            choosing which Clifford to replace the
+            non-Clifford with, only has an impact if
+            method_replace = 'gaussian'.
+
+    Returns:
+        List[circ.Circuit]: list of near-Clifford circuits constructed from
```
rmlarose, line 65:
```suggestion
        List[cirq.Circuit]: list of near-Clifford circuits constructed from
```

mitiq/cdr/test_clifford_training_data.py
```diff
@@ -0,0 +1,395 @@
...
+from cirq.experiments import (
+    random_rotations_between_grid_interaction_layers_circuit,
+)
+from qiskit import compiler, QuantumCircuit
+from mitiq.mitiq_qiskit.conversions import to_qiskit, from_qiskit
+
+
+"""Tests for training circuits generation for Clifford data regression.
+"""
+CLIFFORD_ANGLES = (0.0, np.pi / 2, np.pi, (3 / 2) * (np.pi))
```
rmlarose, line 42: import this from clifford_training_data

andreamari:
@L-P-B, I think there is a spelling problem:  `np.random.ranint` ---> `np.random.randint` . Originated by 75c7b6b

mitiq/cdr/clifford_training_data.py
```diff
@@ -0,0 +1,443 @@
...
+        kwargs: Available keyword arguments are:
+            sigma_select (float): Width of the Gaussian distribution used for
+                ``method_select='gaussian'``.
+            sigma_replace (float): Width of the Gaussian distribution used for
+                ``method_replace='gaussian'``.
+
+    .. [Czarnik2020] : Piotr Czarnik, Andrew Arramsmith, Parick Coles,
+        Lukasz Cincio,
+        "Error mitigation with Clifford quantum circuit data,"
+        (https://arxiv.org/abs/2005.10189).
```
rmlarose, line 72:
```suggestion
    .. [Czarnik2020] : Piotr Czarnik, Andrew Arramsmith, Parick Coles,
        Lukasz Cincio, "Error mitigation with Clifford quantum circuit 
        data," (https://arxiv.org/abs/2005.10189).
```

mitiq/cdr/clifford_training_data.py
```diff
@@ -0,0 +1,442 @@
...
+            'closest'.
+        random_state: Seed for sampling.
+        kwargs: Available keyword arguments are:
+            sigma_select (float): Width of the Gaussian distribution used for
+                ``method_select='gaussian'``.
+            sigma_replace (float): Width of the Gaussian distribution used for
+                ``method_replace='gaussian'``.
+
+    .. [Czarnik2020] : Piotr Czarnik, Andrew Arramsmith, Parick Coles,
+        Lukasz Cincio, "Error mitigation with Clifford quantum circuit
```
rmlarose, line 70:
```suggestion
        Lukasz Cincio, "Error mitigation with Clifford quantum circuit
```

mitiq/cdr/clifford_training_data.py
```diff
@@ -0,0 +1,442 @@
...
+            replaced by Clifford gates. Options are 'uniform' or 'gaussian'.
+        method_replace: Method by which selected non-Clifford gates are
+            replaced by Clifford gates. Options are 'uniform', 'gaussian' or
+            'closest'.
+        random_state: Seed for sampling.
+        kwargs: Available keyword arguments are:
+            sigma_select (float): Width of the Gaussian distribution used for
+                ``method_select='gaussian'``.
+            sigma_replace (float): Width of the Gaussian distribution used for
+                ``method_replace='gaussian'``.
```
rmlarose, line 67:
```suggestion
        kwargs: Available keyword arguments are:
            - sigma_select (float): Width of the Gaussian distribution used for
            ``method_select='gaussian'``.
            - sigma_replace (float): Width of the Gaussian distribution used for
            ``method_replace='gaussian'``.
```

mitiq/cdr/clifford_training_data.py
```diff
@@ -0,0 +1,442 @@
...
+            replaced by Clifford gates. Options are 'uniform' or 'gaussian'.
+        method_replace: Method by which selected non-Clifford gates are
+            replaced by Clifford gates. Options are 'uniform', 'gaussian' or
+            'closest'.
+        random_state: Seed for sampling.
+        kwargs: Available keyword arguments are:
+            - sigma_select (float): Width of the Gaussian distribution used for
+            ``method_select='gaussian'``.
+            - sigma_replace (float): Width of the Gaussian distribution used for
+            ``method_replace='gaussian'``.
```
rmlarose, line 67:
```suggestion
            - sigma_replace (float): Width of the Gaussian distribution used
            for ``method_replace='gaussian'``.
```

mitiq/cdr/clifford_training_data.py
```diff
@@ -0,0 +1,442 @@
...
+        sigma: width of probability distribution used in selection
+                      of non-Clifford gates to replace, only has effect if
+                      method_select = 'gaussian'.
+        random_state: Random state for sampling.
+    """
+    if random_state is None:
+        random_state = np.random
+
+    num_non_cliff = len(non_clifford_ops)
+    num_to_replace = int(fraction_non_clifford * num_non_cliff)
```
andreamari, line 220:
```suggestion
    num_to_replace = int(round(fraction_non_clifford * num_non_cliff))
```
slightly better approximation

mitiq/cdr/clifford_training_data.py
```diff
@@ -0,0 +1,442 @@
...
+        )
+
+    # Select (indices of) non-Clifford operations to replace.
+    selected_indices = random_state.choice(
+        range(num_non_cliff),
+        num_non_cliff - num_to_replace,
+        replace=False,
+        p=distribution,
+    )
+    return [int(i) for i in sorted(selected_indices)]
```
andreamari, line 244: Aren't `selected_indices` already integers? Why this conversion?

mitiq/cdr/clifford_training_data.py
```diff
@@ -0,0 +1,442 @@
...
+    Args:
+        angle: angle to form probability distribution.
+
+    Returns:
+        discrete value of probability distribution calculated from
+        exp(-(dist/sigma)^2) where dist = sum(dists) is the
+        sum of distances from each Clifford gate.
+    """
+    dists = _make_probability_distribution(angle, sigma)
+    return sum(dists)
```
andreamari, line 400: These two functions don't return actual distrubutions or probabilities. It took me a while to understand why `sum(dists)` is not 1. 
The functions return something similar to a closeness measure: "how close is this angle to a clifford angle"? 
So maybe something like `_angle_to_proximities` and `_angle_to_proximity` could be better names for the two functions. I would not mention probabilities.

Also, I am not convinced why the second function returns `sum(dists)` .   Probably the results would be almost the same, but isn't  `max(dists)` a more natural measure of "proximity" ?

mitiq/cdr/test_clifford_training_data.py
```diff
@@ -0,0 +1,254 @@
...
+        ops, 0.0, method=method, random_state=np.random.RandomState(1)
+    )
+    assert np.allclose(indices, np.array(list(range(len(ops)))))
+
+
+@pytest.mark.parametrize("method", ("uniform", "gaussian"))
+def test_select_some(method):
+    n = 10  # Number to select.
+    q = cirq.GridQubit(1, 1)
+    ops = [cirq.ops.rz(a).on(q) for a in np.random.randn(n)]
```
andreamari, line 83:
```suggestion
    q = cirq.GridQubit(1, 1)
    rnd_state = np.random.RandomState(1)
    ops = [cirq.ops.rz(a).on(q) for a in rnd_state.randn(n)]
```
If we have bad luck this test could fail. In this way it becomes deterministic.

mitiq/cdr/test_clifford_training_data.py
```diff
@@ -0,0 +1,254 @@
...
+        method_replace=method_replace,
+        **kwargs,
+    )
+    assert len(train_circuits) == num_train
+
+    for train_circuit in train_circuits:
+        assert set(train_circuit.all_qubits()) == set(circuit.all_qubits())
+        assert count_non_cliffords(train_circuit) == int(
+            fraction_non_clifford * count_non_cliffords(circuit)
+        )
```
andreamari, line 177:
```suggestion
        assert count_non_cliffords(train_circuit) == int(
            round(fraction_non_clifford * count_non_cliffords(circuit))
        )
```
If round is used in the code of `generate_training_circuits`, this may be necessary.

mitiq/cdr/test_clifford_training_data.py
```diff
@@ -0,0 +1,254 @@
...
+        non_clifford_ops,
+        fraction_non_clifford,
+        method=method,
+        sigma=0.5,
+        random_state=np.random.RandomState(1),
+    )
+    assert all(isinstance(index, int) for index in indices)
+    assert len(indices) == int(
+        (1.0 - fraction_non_clifford) * len(non_clifford_ops)
+    )
```
andreamari, line 198:
```suggestion
    assert len(indices) == int(
        round((1.0 - fraction_non_clifford) * len(non_clifford_ops))
    )
```

andreamari (changes requested):
Thanks @L-P-B. Overall LGTM,  I only left some minor suggestions.

I agree with the TODOs saying that we could generalize many functions to operations instead of angles. But the current structure works for me.

andreamari (approved):

andreamari:
Thanks @L-P-B!   LGTM!   :champagne:

rmlarose (approved):

nathanshammah:
Great!

PR #605, opened 2021-03-18
Adds local identity insertion scaling

### LINKED ISSUE #335: RFC for Identity insertion noise scaling
Insert n identity gates in a circuit which scales certain errors by elapsing time. Suggested in recent discussions by alpha users.

purva-thakre:
Description
-----------
Fixes #335 partially - Adds local identity scaled functions. Documentation is also updated with new examples.

The test/code for these functions is similar to code found in `ZNE/folding.py`.  All the changes are made to account for only 1 gate being inserted in the circuit instead of 2 gates. The main difference between the methods can be seen when the scaling factor is not a whole number (say scale_factor=2.5) otherwise they all produce equivalent circuits (scale_factor=3). 

Other functions (global folding etc.) will be added after this PR is accepted.

### DIFF
--- a/mitiq/zne/scaling/identity_insertion.py
+++ b/mitiq/zne/scaling/identity_insertion.py
@@ -101,14 +101,14 @@ def _create_scale_mask(
         [2, 2, 1, 0]
     """
 
-    if not 0.0 <= scale_factor:
+    if not 1.0 <= scale_factor:
         raise ValueError(
-            f"Requires scale_factor >= 0.0 but scale_factor = {scale_factor}."
+            f"Requires scale_factor >= 1.0 but scale_factor = {scale_factor}."
         )
 
     # Find the maximum odd integer smaller or equal to scale_factor
     num_uniform_folds = int(scale_factor - 1.0)
-    odd_integer_scale_factor = num_uniform_folds - 1.0
+    odd_integer_scale_factor = np.abs(num_uniform_folds - 1)
 
     # Uniformly scaling all gates to reach odd_integer_scale_factor
     num_inserts_mask = []
@@ -119,7 +119,7 @@ def _create_scale_mask(
             num_inserts_mask.append(num_uniform_folds)
 
     # If the scale_factor is an odd integer, we are done.
-    if np.isclose(odd_integer_scale_factor, scale_factor):
+    if np.isclose(odd_integer_scale_factor, scale_factor, atol=0.01):
         return num_inserts_mask
 
     # Express scaling order through a list of indices
--- a/mitiq/zne/scaling/identity_insertion.py
+++ b/mitiq/zne/scaling/identity_insertion.py
@@ -1,0 +1,16 @@
+# Copyright (C) 2021 Unitary Fund
+#
+# This program is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program.  If not, see <https://www.gnu.org/licenses/>.
+
+"""Functions for identity scaling on supported circuits."""
--- a/mitiq/zne/scaling/identity_insertion.py
+++ b/mitiq/zne/scaling/identity_insertion.py
@@ -14,3 +14,46 @@
 # along with this program.  If not, see <https://www.gnu.org/licenses/>.
 
 """Functions for identity scaling on supported circuits."""
+from copy import deepcopy
+from typing import (
+    Any,
+    Dict,
+    FrozenSet,
+    List,
+    Optional,
+    Tuple,
+)
+
+import numpy as np
+from cirq import Circuit, InsertStrategy, inverse, ops, has_unitary
+
+from mitiq._typing import QPROGRAM
+from mitiq.conversions import noise_scaling_converter
+
+from mitiq.zne.scaling.folding import ( _string_keys_to_cirq_gates,
+_cirq_gates_to_string_keys, _is_measurement, _pop_measurements, _append_measurements,
+_squash_moments)
+
+class UnscalableCircuitError(Exception):
+    pass
+
+def _check_scalable(circuit: Circuit) -> None:
+    """Raises an error if the input circuit cannot be scaled.
+    Args:
+        circuit: Checks whether this circuit is able to be scaled.
+    Raises:
+        UnfoldableCircuitError:
+            * If the circuit has intermediate measurements.
+            * If the circuit has non-unitary channels which are not terminal
+              measurements.
+    """
+    if not circuit.are_all_measurements_terminal():
+        raise UnscalableCircuitError(
+            "Circuit contains intermediate measurements and cannot be scaled."
+        )
+
+    if not has_unitary(circuit):
+        raise UnscalableCircuitError(
+            "Circuit contains non-unitary channels which are not terminal "
+            "measurements and cannot be scaled."
+        )
--- a/mitiq/zne/scaling/identity_insertion.py
+++ b/mitiq/zne/scaling/identity_insertion.py
@@ -106,20 +106,20 @@ def _create_scale_mask(
             f"Requires scale_factor >= 1.0 but scale_factor = {scale_factor}."
         )
 
-    # Find the maximum odd integer smaller or equal to scale_factor
-    num_uniform_folds = int(scale_factor - 1.0)
-    odd_integer_scale_factor = np.abs(num_uniform_folds - 1)
+    # Find the maximum integer smaller or equal to scale_factor
+    num_uniform_inserts = int(scale_factor - 1.0)
+    integer_scale_factor = num_uniform_inserts + 1
 
-    # Uniformly scaling all gates to reach odd_integer_scale_factor
+    # Uniformly scaling all gates to reach integer_scale_factor
     num_inserts_mask = []
     for w in weight_mask:
         if np.isclose(w, 0.0):
             num_inserts_mask.append(0)
         else:
-            num_inserts_mask.append(num_uniform_folds)
+            num_inserts_mask.append(num_uniform_inserts)
 
-    # If the scale_factor is an odd integer, we are done.
-    if np.isclose(odd_integer_scale_factor, scale_factor, atol=0.01):
+    # If the scale_factor is an integer, we are done.
+    if np.isclose(integer_scale_factor, scale_factor, atol=0.01):
         return num_inserts_mask
 
     # Express scaling order through a list of indices
--- a/mitiq/zne/scaling/identity_insertion.py
+++ b/mitiq/zne/scaling/identity_insertion.py
@@ -139,7 +139,7 @@ def _create_scale_mask(
 
     # Fold gates until the input scale_factor is better approximated
     input_circuit_weight = sum(weight_mask)
-    output_circuit_weight = odd_integer_scale_factor * input_circuit_weight
+    output_circuit_weight = integer_scale_factor * input_circuit_weight
     approx_error = np.abs(
         output_circuit_weight - scale_factor * input_circuit_weight
     )
--- a/mitiq/zne/scaling/identity_insertion.py
+++ b/mitiq/zne/scaling/identity_insertion.py
@@ -15,32 +15,30 @@
 
 """Functions for identity scaling on supported circuits."""
 from copy import deepcopy
-from typing import (
-    Any,
-    Dict,
-    FrozenSet,
-    List,
-    Optional,
-    Tuple,
-)
+from typing import Any, List, Optional
 
 import numpy as np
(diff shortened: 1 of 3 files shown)


purva-thakre:
@andreamari @rmlarose This PR is ready for your review (in case it escaped your notice).

mitiq/zne/scaling/identity_insertion.py
```diff
@@ -0,0 +1,425 @@
...
+            new_output_circuit_weight - scale_factor * input_circuit_weight
+        )
+        # Fold the candidate gate only if it helps improving the approximation
+        if new_approx_error >= approx_error:
+            break
+        approx_error = new_approx_error
+        output_circuit_weight = new_output_circuit_weight
+        num_inserts_mask[j] += 1
+
+    return num_inserts_mask
```
rmlarose, line 162: These are all identical to the corresponding functions in https://github.com/unitaryfund/mitiq/blob/master/mitiq/zne/scaling/folding.py except with "fold" -> "scale", right? If so they should be imported, and we can decide if its good to rename.

mitiq/zne/scaling/identity_insertion.py
```diff
@@ -0,0 +1,425 @@
...
+    else:
+        mask_index = 0
+        for moment in circ_copy:
+            scaled_moment = Circuit(moment)
+            for op in moment:
+                num_inserts = num_inserts_mask[mask_index]
+                scaled_moment.append(
+                    num_inserts
+                    * [GateOperation(IdentityGate(len(op.qubits)), op.qubits)]
+                )
```
rmlarose, line 220: This function is identical to [`_apply_fold_mask`](https://github.com/unitaryfund/mitiq/blob/master/mitiq/zne/scaling/folding.py#L491) except for these lines which determine which gates are inserted. Here, it is identity, and in unitary folding it is an operation in the circuit.

I would propose generalizing `_apply_fold_mask` to the following:

```python
def _apply_fold_mask(
    circuit: Circuit,
    num_folds_mask: List[int],
    squash_moments: Optional[bool] = True,
    gates_to_insert: Optional[Sequence[cirq.Gate]] = None,
) -> Circuit:
    """...

    Args:
        ...
        gates_to_insert: Pool of gates to select from and insert into the circuit when applying the fold mask.
            When inserting a gate, we choose `G` uniformly at random from `gates_to_insert` and insert
            G^dag G into the circuit. If `None`, `G` is the preceding operation in the circuit.
    """
```

and using it here with `gates_to_insert = cirq.I`.

In addition to reducing duplication, this would increase the flexibility of folding/scaling and would allow us to test more methods. Thoughts @andreamari? 

Some considerations:

- Gates might need to be specified in a frontend-agnostic manner (e.g., `"Identity"` or `"I"` instead of `cirq.I`.)
- Might want to allow for a probability distribution (not just uniform as I wrote above) for selecting from `gates_to_insert`.
- For the inserted gate, we probably want to match the number of qubits of the preceding gate in the circuit.

rmlarose (changes requested):
Thanks @purva-thakre! This is looking pretty good and will be a valuable contribution. I want to reduce the code duplication (see comments) and, to this end, proposed a way to generalize [`_apply_fold_mask`](https://github.com/unitaryfund/mitiq/blob/master/mitiq/zne/scaling/folding.py#L491) which will give us identity insertion scaling essentially for free, plus other methods.

Let me know what you think. I haven't looked at the tests yet.

mitiq/zne/scaling/identity_insertion.py
purva-thakre: Not really. The formula at lines 110-111 is different from the folding module.

mitiq/zne/scaling/identity_insertion.py
purva-thakre: In unitary folding, there's 2 operations being inserted which is why I created this separate function. I could try to figure out if there's a way to change just those two lines when `apply_fold_mask` is used for identity scaling. I think there might be a way to define a decorator function. 

> For the inserted gate, we probably want to match the number of qubits of the preceding gate in the circuit.

I think it does so from what I remember. Some of the tests are for gates acting on more than 1 qubit and the gates inserted also act on the same number of qubits. I will change the docstrings wherever possible. 

> Gates might need to be specified in a frontend-agnostic manner 

Ok. Seems reasonable. I will make the changes. 

> Might want to allow for a probability distribution (not just uniform as I wrote above) for selecting from gates_to_insert

I think this could be added to above decorator function.

purva-thakre:
> I want to reduce the code duplication

Instead of introducing a separate module for identity scaling, I could add conditional statements in functions for unitary folding for when a user chooses to insert identity gates in the circuit.

andreamari:
Thanks @purva-thakre , this looks great! 
I still hadn't looked into the details, but I think we can first discuss the code duplication problem.

As noted by @purva-thakre, there is an important difference between unitary folding and identity insertion. In one case we append **2 gates** in the other case we append **1 identity gate**. 

I think that because of this, the proposal of @rmlarose to allow for generic gates (although nice in itself) wouldn't directly solve our problem.

My suggestion would be:
- keep the new functions even if they contain similar code
- make a new issue called "remove code duplication between unitary folding and identity insertion". This is less trivial than it seems and so it could require a new PR. For example, it may be better to re-define the convention of the folding mask where
the elements are not the **number of folds** but the **number of added gates**. With this new convention, unitary folding would always have even elements in the mask while identity insertion can have odd elements too.

We can discuss this on the Friday meeting if we are all there. Otherwise we can comment here.

rmlarose:
> As noted by @purva-thakre, there is an important difference between unitary folding and identity insertion. In one case we append **2 gates** in the other case we append **1 identity gate**.
> 
> I think that because of this, the proposal of @rmlarose to allow for generic gates (although nice in itself) wouldn't directly solve our problem.

Can't we just divide the mask by 2 when inserting identity? Alternatively, is it really that bad to insert two identities instead of one?

purva-thakre:
> Can't we just divide the mask by 2 when inserting identity?

I tried this but couldn't do it because `scale_factor=1` would insert 1 identity gate whereas we want the circuit to be returned unchanged similar to [unitary folding](https://github.com/unitaryfund/mitiq/blob/2b1b61c3db1ff092db225c03342a9ac8d1e070b5/mitiq/zne/scaling/folding.py#L437). In order to be consistent, I had to introduce [absolute value for the scale factor calculation](https://github.com/unitaryfund/mitiq/pull/605/files#diff-955bebdf27ed25b3e37bad94d918bfc5d182a5c047b956829a1c77752296b741R111). 

> Alternatively, is it really that bad to insert two identities instead of one?

I leave this to @andreamari. We discussed [this before](https://github.com/unitaryfund/mitiq/issues/335#issuecomment-759315142).

mitiq/zne/scaling/identity_insertion.py
```diff
@@ -0,0 +1,425 @@
...
+    num_inserts_mask = []
+    for w in weight_mask:
+        if np.isclose(w, 0.0):
+            num_inserts_mask.append(0)
+        else:
+            num_inserts_mask.append(num_uniform_folds)
+
+    # If the scale_factor is an odd integer, we are done.
+    if np.isclose(odd_integer_scale_factor, scale_factor, atol=0.01):
+        return num_inserts_mask
```
andreamari, line 123: I think this  should be corrected in this way. More or less, I didn't test.

```suggestion
    # Find the maximum integer smaller or equal to scale_factor
    num_uniform_inserts = int(scale_factor - 1.0)
    integer_scale_factor = num_uniform_inserts + 1

    # Uniformly scaling all gates to reach odd_integer_scale_factor
    num_inserts_mask = []
    for w in weight_mask:
        if np.isclose(w, 0.0):
            num_inserts_mask.append(0)
        else:
            num_inserts_mask.append(num_uniform_inserts)

    # If the scale_factor is an integer, we are done.
    if np.isclose(integer_scale_factor, scale_factor, atol=0.01):
        return num_inserts_mask
```

mitiq/zne/scaling/identity_insertion.py
```diff
@@ -0,0 +1,425 @@
...
+            "The option 'scaling_method' is not valid."
+            "It must be 'at_random', or 'from_left', or 'from_right'."
+        )
+
+    # Fold gates until the input scale_factor is better approximated
+    input_circuit_weight = sum(weight_mask)
+    output_circuit_weight = odd_integer_scale_factor * input_circuit_weight
+    approx_error = np.abs(
+        output_circuit_weight - scale_factor * input_circuit_weight
+    )
```
andreamari, line 145: This is based on the previous suggested renaming.

```suggestion
    # Fold gates until the input scale_factor is better approximated
    input_circuit_weight = sum(weight_mask)
    output_circuit_weight = integer_scale_factor * input_circuit_weight
    approx_error = np.abs(
        output_circuit_weight - scale_factor * input_circuit_weight
    )
```

andreamari:
813449199). So, there's a possibility the equation for identity scaling might be incorrect.
> 
Indeed, apart from names of variables, I think there is a problem in 1 equation. See my code suggestion.

Hopefully this should fix also this problem:
> Looking at the list below, I do think there needs to be a change as to how `scale_factor` is used. There's no way to insert 1 identity gate after each operation in the circuit uniformly. Right now, `scale_factor` inserts gates as follows:

mitiq/zne/scaling/identity_insertion.py
purva-thakre: @andreamari Yes, I tried using `num_uniform_inserts + 1` but there was an incompatibility with `scale_factor`. 
For identity scaling, `scale_factor=1` would insert 1 identity gate whereas `scale_factor=0` would keep the circuit unchanged. This was different from how `scale_factor=1` is used in unitary folding. 

Can I use `insert_factor` instead of `scale_factor` ? See above discussion.

andreamari: > @andreamari Yes, I tried using `num_uniform_inserts + 1` but there was an incompatibility with `scale_factor`.
> For identity scaling, `scale_factor=1` would insert 1 identity gate whereas `scale_factor=0` would keep the circuit unchanged. This was different from how `scale_factor=1` is used in unitary folding.

Are you sure?   If `scale_factor = 1`, `num_uniform_inserts` is 0 and `integer_scale_factor` is 1.  So `num_inserts_mask` should be `[0, 0, .... 0]`.  Am I wrong? 


> Can I use `insert_factor` instead of `scale_factor` ? See above discussion.
The optimal solution would be to use `scale_factor` for consistency with other folding functions. But if this is really necessary we can discuss about it.

rmlarose:
> My suggestion would be:
> 
> * keep the new functions even if they contain similar code
> * make a new issue called "remove code duplication between unitary folding and identity insertion". This is less trivial than it seems and so it could require a new PR. For example, it may be better to re-define the convention of the folding mask where
>   the elements are not the **number of folds** but the **number of added gates**. With this new convention, unitary folding would always have even elements in the mask while identity insertion can have odd elements too.

Discussed @ Mitiq meeting and I agree with this suggestion. So once Andrea's other comments are addressed here this LGTM.

crazy4pi314:
Hi @purva-thakre! I know you are probably pretty busy, but is there anything I can do to help out on getting this pull closed?

purva-thakre:
> Hi @purva-thakre! I know you are probably pretty busy, but is there anything I can do to help out on getting this pull closed?

@crazy4pi314 Not really ! I know what I need to do. I haven't had the time to make above requested fixes. I will do so soon.

purva-thakre:
I do want to complete this PR. 

At the moment, I am trying to fix the import path on my local computer. When trying to install a new software, I inadvertently changed the import path for my virtual environments. I am thinking of starting from scratch with all the installations which will take a couple more days.

crazy4pi314:
@purva-thakre if you want help troubleshooting this lmk! I think your plan of just blowing away and re-installing is a good one though. What OS are you on?

purva-thakre:
> What OS are you on?

@crazy4pi314 I am on Linux Mint. I did figure out that the only the import path for virtual environments is messed up. If I use `pip list`, it does show `mitiq` installed in my environment. But when I try to import a module from `mitiq`, I get the error similar to `mitiq : command not found`. I tried to create a new virtual environment with a fresh install of everything in my old environment and it runs into same issues. 

I think one possible issue could be the version of Python I switched to before all these issues popped up. Looking at the installed programs for my environment, [they have been installed for an older version of Python](https://stackoverflow.com/a/56619949/10241324). 

Here's some solutions I plan to try :

- Try switching to older version of python
- Create a new environment and specify python and pip version
- Add path to module via [PYTHONPATH](https://stackoverflow.com/questions/18247333/pythonpath-on-linux) (this one I am not too familiar with)

Let me know if you think of some other options.

crazy4pi314:
Glad you got it @purva-thakre! Let me know if you run into any other issues <3

mitiq/zne/scaling/identity_insertion.py
```diff
@@ -0,0 +1,425 @@
...
+"""Functions for identity scaling on supported circuits."""
+from copy import deepcopy
+from typing import Any, List, Optional
+
+import numpy as np
+from cirq import GateOperation, Circuit, has_unitary
+from cirq.ops import IdentityGate
+
+from mitiq._typing import QPROGRAM
+from mitiq.conversions import noise_scaling_converter
```
purva-thakre, line 25: mitiq.interface not mitiq.conversions

purva-thakre:
Closing this to create a new PR later.

Issue #594, opened 2021-03-15
Change of basis Qiskit executor is slow


Aaron-Robertson:
Pre-Report Checklist
--------------------

- [ ] I am running the latest version of mitiq
- [ ] I checked to make sure that this bug has not already been reported

Issue Description
-----------------
Resolve todos in `test_zne_mitiq_qiskit.py`, with focus on executor bug found in the `qiskit_executor()` below.

How to Reproduce
----------------
Uncomment this executor in `test_zne_mitiq_qiskit.py` and run `make test-qiskit`.
```
def qiskit_executor(qp: QPROGRAM, shots: int = 500) -> float:
    # initialize a qiskit noise model
    noise_model = NoiseModel()

    # we assume a depolarizing error for each gate of the standard IBM basis
    # set (u1, u2, u3)
    noise_model.add_all_qubit_quantum_error(
        depolarizing_error(BASE_NOISE, 1), ["u1", "u2", "u3"]
    )
    expectation = execute_with_shots_and_noise(
        qp,
        shots=shots,
        obs=ONE_QUBIT_GS_PROJECTOR,
        noise_model=noise_model,
        seed=1,
    )
    return expectation
```
Test run takes >40 minutes (on `test_run_factory_with_number_of_shots` and `test_mitigate_executor_with_shot_list` primarily). As of 03/11, time increase occurs in execution after change of basis with many shots.

| Shots  | %timeit eigvals, U  | %timeit circ.unitary  | %timeit circ.measure_all() | %timeit qiskit.execute | %timeit qiskit.execute (no change of basis)
| --------------- | --------------- | --------------- | --------------- | --------------- | --------------- |
| 10 ** 4 | 23.2 µs ± 6.91 µs | 286 µs ± 91.4 µs | 5.76 ms ± 3.84 ms | 1min 3s ± 8.62 s | 47.1 ms ± 15 ms |
| 10 ** 5 | 13.8 µs ± 1.66 µs | 179 µs ± 64.8 µs | 7.45 ms ± 3.67 ms | 12min 55s ± 1min | 176 ms ± 5.95 ms |
| 10 ** 6 | 11.7 µs ± 292 ns | 118 µs ± 4.86 µs | 3.92 ms ± 2.21 ms | [?] | 1.84 s ± 398 ms |

[...]

PR #614, opened 2021-03-31
Install only the qiskit packages we need

### LINKED ISSUE #396: Depend only on the Qiskit packages we really need
Issue Description
-----------------

Right now, we have `qiskit` in the `requirements.txt` file, as it is an optional third-party integration. This is a "meta" package that just installs other Qiskit libraries. For example, this is what is installed when you do `pip install qiskit` as of today:

```
qiskit                        0.16.2
qiskit-aer                    0.4.1
qiskit-aqua                   0.6.5
qiskit-ibmq-provider          0.5.0
qiskit-ignis                  0.2.0
qiskit-terra                  0.12.0
```

This is likely more than we need to run Mitiq, and additionally makes `import mitiq` quite slow. The goal of this issue is to only install the Qiskit packages we really need in `requirements.txt`.

purva-thakre:
Description
-----------

Fixes #396

Qiskit packages are specified in `dev_requirements.txt` with `aer` and `ibmq-provider`. 

`terra` is not listed here because it will be installed as dependency for `aer`. But since it is needed to build circuits, a comment is made about this in `dev_requirements.txt`. 

Checklist
-----------

Check off the following once complete (or if not applicable). The PR will be reviewed once this
checklist is complete and all tests are passing.
- [ ] I updated the [changelog](https://github.com/unitaryfund/mitiq/blob/master/CHANGELOG.md) including author and PR number (@username, gh-xxx)


purva-thakre:
The check for `docs` fails due issue with documentation page for `scipy`.  

`https://docs.scipy.org/` does not appear to be functional for the moment.

rmlarose (changes requested):
Thanks a lot @purva-thakre for revisiting this, I'm in support of a minimal Qiskit installation.

I'm guessing the changes here result in a different version of Qiskit being installed which is causing tests to fail. Can you check which version of Qiskit is installed to confirm (or refute) this? And then either

1. Install the current version of Qiskit to fix tests.
2. Make necessary updates for whatever Qiskit version is installed here to fix the tests.

Slight preference to (1), especially if the Qiskit version installed here is older than the current one.

purva-thakre:
@rmlarose Seemed to be an issue with my virtual environment. When I tested this locally, all of the tests passed because I guess the checks were being run with my previously versions of qiskit. The version of `qiskit-terra` that gets installed here is a newer version (`0.17.0` and not `0.16.4`). 

One issue I can see with a minimal installation is `pip` would always try to install latest version of the dependencies. What's your preference ?

purva-thakre:
BTW I don't know what the issue with checks being run for this PR. I have tested this locally with `make` tests run in the workflows.

rmlarose:
> Which one of these do you prefer ? I don't really have a preference about either method.
> 
> If there's a preference to bump up to higher versions then like you suggested there will need to be changes for how `qargs` are used + `QasmSimulator` being deprecated to `AerSimulator` etc.
> 
> The requested code review is for older versions of `qiskit` packages.

I guess we need the set of dependencies such that `qiskit_version>=0.24.0` for https://github.com/unitaryfund/mitiq/issues/594#issuecomment-813121317.

mitiq/_version.py
```diff
@@ -1,23 +1 @@
...
-
-"""Reads in version information.
-Note: This file will be overwritten by the packaging process."""
-import os
-
-directory_of_this_file = os.path.dirname(os.path.abspath(__file__))
-
-with open(f"{directory_of_this_file}/../VERSION.txt", "r") as f:
-    __version__ = f.read().strip()
+__version__ = "0.7.0dev"
```
rmlarose, line 1: I think this was committed by accident.

dev_requirements.txt
```diff
@@ -1,5 +1,7 @@
 # Third-party integration.
-qiskit~=0.24.0
+qiskit-terra ~= 0.16.4
```
rmlarose, line 2: nit: Remove whitespace on either side of `~=`.

dev_requirements.txt
```diff
@@ -18,3 +20,16 @@ myst-parser~=0.13.3
 pybtex~=0.22.2
 pydata-sphinx-theme~=0.3.1
 notebook~=6.1.5
+
+# Packages needed for individual qiskit install
```
rmlarose, line 24: Is this to say: If we only install terra, aer, and ibmq-provider, these are the packages that are missing compared to when we install qiskit? Is it a minimal set? Are they all necessary?

purva-thakre: Yes. I compared the packages installed after a full install to the ones installed individually. Since individual install lead to some tests failing, I just added the packages listed here. I did not test all of them individually to check absence of which ones led to errors.

purva-thakre:
> I guess we need the set of dependencies such that `qiskit_version>=0.24.0` for [#594 (comment)](https://github.com/unitaryfund/mitiq/issues/594#issuecomment-813121317).

@rmlarose The aer and terra versions specified in this PR are already set to have `qiskit_version=0.24.1`. If you instead change to `qiskit_version =0.25.0` then there will need to be changes as stated above. For more information, see the[ first two lines of version history table](https://qiskit.org/documentation/release_notes.html#version-history). 

Do you prefer `qiskit_version` be bumped up ?

dev_requirements.txt
rmlarose: Thanks @purva-thakre. Can you try this in your environment to find a minimal set?

```bash
pip install pipdeptree
pipdeptree --warn silence | grep -E '^\w+'
```

https://pypi.org/project/pipdeptree/

purva-thakre: @rmlarose Thanks for pointing me towards `pipdeptree`. It seems I did not need the packages that were listed previously. The tests were failing because `terra` was not installed before `aer`. There's an installation order and `pip dependency resolver` has issues with switching to a higher version of dependency for `terra` when a lower one was installed for `aer`. 

image

For this reason, minimal `qiskit` install by specifying `aer` only might not be a good idea. I have specified terra, aer and ibmq-provider exactly.

rmlarose: This is good to know, thanks!

dev_requirements.txt
```diff
@@ -9,7 +11,7 @@ flake8~=3.7.9
 black~=19.10b0
 
 # Documentation and examples.
-sphinx~=2.4.4
+sphinx~=3.5.3
```
rmlarose, line 14: I think this is fine, just pinging @nathanshammah and @crazy4pi314 as an fyi.

rmlarose (approved):
LGTM, thanks @purva-thakre! Made a slight edit to the PR name for clarity.

PR #471, opened 2020-12-08
Adds new Bayesian Inference based factory, BayesFactory

elmandouh:
Fixes #418 

Description
-----------
Adds a new factory based that uses Bayesian inference to extrapolate to the zero-noise-limit.

Currently `BayesFactory` extrapolates to the zne limit by estimating the model parameters of the likelihood function 
given by E(lambda) = a + b * e^{-c*lambda) using [PyMC3](https://github.com/pymc-devs/pymc3). This should be extended to accommodate any likelihood function and priors. In addition, the full probability distribution should be outputted. This assumes that the prior for a measurement is a normal distribution. A more fitting prior would be the Binomial distribution, however that requires modifying `execute_with_zne`. 

Checklist
-----------
Please make sure you have finished the following tasks before requesting a review of the pull request (PR). For more information, please refer to the [Contributor's Guide](../blob/master/CONTRIBUTING.md):

- [x] Contributions to mitiq should follow the [pep8 style](https://www.python.org/dev/peps/pep-0008/). You can enforce it easily with [`black`](https://black.readthedocs.io/en/stable/index.html) style and with [`flake8`](http://flake8.pycqa.org) conventions.
- [x] Please add tests to cover your changes, if applicable, and make sure that all new and existing tests pass locally.
- [x] If the behavior of the code has changed or new feature has been added, please also [update the documentation](../blob/master/docs/CONTRIBUTING_DOCS.md).
- [x] Functions and classes have useful [Google-style](https://sphinxcontrib-napoleon.readthedocs.io/en/latest/example_google.html) docstrings and [type hints](https://www.python.org/dev/peps/pep-0484/) in the signature of the objects.
- [x] (Bug fix) The associated issue is referenced above using [auto-close keywords](https://help.github.com/en/articles/closing-issues-using-keywords). If the PR fixes an issue, use the keyword fix/fixes/fixed followed by the issue id. For example, if the PR fixes issue 1184, type "Fixes #1184" (without quotes).
- [x] The [changelog](https://github.com/unitaryfund/mitiq/blob/master/CHANGELOG.md) is updated, including author and PR number (@username, gh-xxx).

PR #625, opened 2021-04-06
Revert "Adds new Bayesian Inference based factory, BayesFactory"

nathanshammah:
Reverts unitaryfund/mitiq#471, as it is making the CI on master fail. The CI issue is in Windows. @crazy4pi314 pointed out that it may be due to a non-Python dependency of pymc3 missing there.


rmlarose (approved):
Docs failure is unrelated and due to a recent sphinx bug I believe. See https://github.com/sphinx-doc/sphinx/issues?q=object+has+no+attribute.

`pip install docutils==0.16.0 sphinx` instead of `pip install sphinx` is a fix apparently. cc @purva-thakre

PR #622, opened 2021-04-02
Update information about formatting in RTD

purva-thakre:
Description
-----------

Fixes  #593
- [x] put code in a code block
- [x] subsection making the existing text more evident
- [x] include a link in PR template not just to Black RTD docs, but to MItiq's docs, to show the way black is configured in the makefile

PR #626, opened 2021-04-06
Pins docutils version to avoid issues with newer version

purva-thakre:
Description
-----------

Check for `docs` fails when newer version of `docutils` is installed. This PR pins older version. See [this comment ](https://github.com/unitaryfund/mitiq/pull/622#issuecomment-813500688)for more information.


nathanshammah (approved):
Thanks @purva-thakre!

Issue #320, opened 2020-08-19
Move qiskit_utils and pyquil_utils to benchmarks


rmlarose:
As far as I can tell, the contents of these files are all utilities for benchmarking. I don't think they would ever be used otherwise. Thoughts?

rmlarose:
Also these files don't have unit tests (?) and pyquil_utils.py sets a global seed "for testing"...

willzeng:
Nice catch. Some cleanup of these files would be good.

Some reasoning about their location structurally is that they are for distinctly testing the different front/backends for mitiq. Benchmarks is for testing mitiq in its native representation (Cirq). Having these are separate modules makes it easier to extend or drop support for different ones. For example if someone didn't want to install the whole pipeline to test every platform, then they could pick and choose.

rmlarose:
Closed in favor of #589

PR #632, opened 2021-04-08
Move all benchmark circuit generation to `mitiq.benchmarks`, adding option for conversions

### LINKED ISSUE #589: Move all benchmark circuits to mitiq.benchmarks
Discussed at Mitiq meeting. All benchmark circuits should be defined in `mitiq.benchmarks` and have the option to output as a specific supported type (i.e., `QPROGRAM`). This is preferable to (re)defining them in `mitiq_qiskit`, `mitiq_pyquil`, etc.

Aaron-Robertson:
Description
-----------

Resolves #589.

Removes benchmarking circuits for testing and moves conversion to `mitiq.benchmarks.generate_rb_circuits()` with supported type outputs.


Aaron-Robertson:
@rmlarose I don't think this one's ready for a full review, but could you take a look and give some conceptual/broad stroke feedback?

It's a working solution based on testing. I tried to take into account the comments I saw on the original issue, and the added unit tests are sort of dummy tests for now (maybe even until the generalized executors).

rmlarose:
> @rmlarose I don't think this one's ready for a full review, but could you take a look and give some conceptual/broad stroke feedback?

Thanks @Aaron-Robertson, looks generally good. 

- I would just add `return_type` to `generate_rb_circuits` instead of creating a new function. I think that's preferable unless you had a good reason for another function?
- I wouldn't include `random_one_qubit_identity_circuit` as its own function. If this was a utility for testing somewhere it can be defined there, but it doesn't need to be in the public API.
  - Actually, I would be in favor of removing this function altogether, even from tests. It's simpler and fewer lines of code to just call `generate_rb_circuits` with `num_qubits=1`. 
- Conversions can be slow, so I:
  - Would conservatively test with `return_type` as `QPROGRAM \ Cirq`, e.g. a few circuits with a small `num_cliffords`.
  - Only execute when `return_type` is Cirq. We don't need to test executing the other circuits here, just that they are returned as the right type.

Aaron-Robertson:
Sounds good and can do! With conversion built in, I agree it makes sense to remove `random_one_qubit_identity_circuit` post proof of concept.

I included the separate function for `converted_rb_circuits` to try to answer a concern I saw on #320 that 
> Benchmarks is for testing mitiq in its native representation

But I'm happy to combine them.

Aaron-Robertson:
@rmlarose This one's ready for review if you have a chance to review/request reviews.

mitiq/benchmarks/randomized_benchmarking.py
```diff
@@ -62,6 +79,19 @@ def generate_rb_circuits(
...
+                _random_two_q_clifford(
+                    *qubits,  # type: ignore
+                    num_cliffords,
+                    cfd_matrices,
+                    cliffords,
+                ),
+                return_type,
+            )
+            for _ in range(trials)
+        ]
```
rmlarose, line 94: I think it's (marginally) better to do something like below for readability / fewer LOC.

```python
circuits = [
        _random_two_q_clifford(
            *qubits,  # type: ignore
            num_cliffords,
            cfd_matrices,
            cliffords,
        )
        for _ in range(trials)
    ]

if return_type is None or return_type == "cirq":
    return circuits

return [convert_from_mitiq(circuit, return_type) for circuit in circuits]
```

mitiq/mitiq_pyquil/tests/test_zne_mitiq_pyquil.py
```diff
@@ -30,28 +30,10 @@
     generate_qcs_executor,
     ground_state_expectation,
 )
-from mitiq.mitiq_pyquil.conversions import to_pyquil
 from mitiq.benchmarks.randomized_benchmarking import generate_rb_circuits
 
 TEST_DEPTH = 30
 
-
-def random_one_qubit_identity_circuit(num_cliffords: int) -> pyquil.Program:
```
rmlarose, line 39: So much better without this everywhere 🙏

mitiq/benchmarks/randomized_benchmarking.py
```diff
@@ -37,6 +42,8 @@ def generate_rb_circuits(
         num_cliffords: The number of Clifford group elements in the
             random circuits. This is proportional to the depth per circuit.
         trials: The number of random circuits at each num_cfd.
+        return_type: String specifier for Mitiq-supported circuit type
+            conversion.
```
rmlarose, line 46:
```suggestion
        return_type: String which specifies the type of the returned circuits. See the keys of ``mitiq.QPROGRAM`` for options. If ``None``, the returned circuits are ``cirq.Circuit``s.
```

rmlarose (changes requested):
Two small suggestions then LGTM.

mitiq/benchmarks/randomized_benchmarking.py
Aaron-Robertson: It's much better 😅  Can define the `random_single_q`s as `circuits` too then there's only one conditional return and one call to convert 💯

rmlarose (approved):

rmlarose:
Thanks @Aaron-Robertson!

Issue #623, opened 2021-04-05
Add `Observable`s and conceptually generalize executors


rmlarose:
Implementation plan for [this design document](https://docs.google.com/document/d/1NKOrOAiQVetbsuZXRAw8hbpIFBoY5Go4EuBYBpP2G30/edit?usp=sharing).

## Steps

- [ ] Add a `PauliString` class satisfying the key features in the design doc.
- [ ] Add an `Observable` class which represents a linear combination of `PauliString`s. 
- [ ] Change signature of `execute_with_[zne|pec]` to (optionally) input an `Observable`.
  - At this point, only execute functions which return floats are supported, but `Observable`s can be used.
- [ ] Rename `mitiq.Collector` to `mitiq.Executor` and support execute functions with the following signatures:
  - `QPROGRAM` -> `MeasurementResult`
  - `QPROGRAM` -> `Sequence[Wavefunction]`
  - `QPROGRAM` -> `DensityMatrix`
  - `QPROGRAM` -> `float`
  - Batched versions of each of the above, e.g., `Sequence[MeasurementResult]`.
- [ ] Use `mitiq.Executor` in `execute_with_[zne|pec]`. 
- [ ] Allow `mitiq.Executor` to adjust the shots of the execute function (if the function supports this).

rmlarose:
#552 is part of this for documenting the new executors.

PR #633, opened 2021-04-08
Add `PauliString`

rmlarose:
Description
-----------

Part of #623. This is a thin wrapper for `cirq.PauliString` with appropriate I/O for us.

Checklist
-----------

Check off the following once complete (or if not applicable). The PR will be reviewed once this
checklist is complete and all tests are passing.

- [ ] I updated the [changelog](https://github.com/unitaryfund/mitiq/blob/master/CHANGELOG.md) including author and PR number (@username, gh-xxx)

Tips
----

  autoformatter.

### DIFF
--- a/mitiq/observable/pauli.py
+++ b/mitiq/observable/pauli.py
@@ -1,0 +1,128 @@
+# Copyright (C) 2021 Unitary Fund
+#
+# This program is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program.  If not, see <https://www.gnu.org/licenses/>.
+
+from typing import Optional, Sequence
+
+import numpy as np
+import cirq
+
+from mitiq import QPROGRAM
+from mitiq.conversions import atomic_converter
+
+
+class PauliString:
+    _string_to_gate_map = {"I": cirq.I, "X": cirq.X, "Y": cirq.Y, "Z": cirq.Z}
+
+    def __init__(
+        self,
+        *,
+        spec: str,
+        coeff: complex = 1.0,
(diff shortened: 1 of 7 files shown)


mitiq/observable/pauli.py
```diff
@@ -0,0 +1,128 @@
...
+                )
+            measured = (
+                circuit + pauli.to_z_basis_ops() + cirq.measure(*pauli.qubits)
+            )
+
+            # Transform circuit back to original qubits.
+            reverse_qubit_map = dict(zip(qubit_map.values(), qubit_map.keys()))
+            return measured.transform_qubits(lambda q: reverse_qubit_map[q])
+
+        return _measure_in(circuit, self._pauli)
```
andreamari, line 102: This function `measure_in` is nice but I think it measures the Pauli operator without considering the coefficient.  E.g. it acts in the same way when measuring XZ or 7*XZ. 
I am not sure what is the best solution, e.g.:
- we add a note in the docstring of `measure_in` saying that the coefficient is neglected.
- we could force `PauliString` to always have coefficient 1 and store the coefficients in a different place/object.

andreamari (approved):
Hi @rmlarose, sorry for the long delay, I forgot about it.
LGTM apart from the comment about `measure_in`.

A general comment: do we want the user to use this class?
If PauliString and other similar future objects will only be used internally by Mitiq, maybe we don't need to support circuit conversions.

mitiq/observable/pauli.py
rmlarose: The coefficient shouldn't go with the circuit but rather the results from executing the circuit. In other words, the circuits you would run to measure XZ or 7*XZ are the same, so they should be the same here. Perhaps there is a better name than `measure_in`? Or I can add a note to the docstring.

rmlarose:
> A general comment: do we want the user to use this class?
> If PauliString and other similar future objects will only be used internally by Mitiq, maybe we don't need to support circuit conversions.

I think the workflow should be identical for both `Observable` and `PauliString` and so this class should be public.

mitiq/observable/pauli.py
rmlarose: Verbose, but what this method really does is `add_single_qubit_basis_rotations_to_measure_in(circuit: QPROGRAM) -> QPROGRAM`. Or we could steal the Cirq name of `to_z_basis_ops`. Thoughts?

andreamari: With `to_z_basis_ops` one would expect only the final rotation, not the full circuit. So `measure_in` is fine.

PR #646, opened 2021-04-19
Test Bayes on windows

rmlarose:
Description
-----------

Checklist
-----------

Tips
----

  autoformatter.


crazy4pi314:
A summary of the issue here as described in the community call today:
- PyMC3 is has dependencies on Windows that cannot be managed through pip. You basically need MinGW to build/support the package. This comes from the linear algebra libraries they are using, you can see in the docs for one of them, install is [only supported through conda](https://aesara.readthedocs.io/en/latest/install_windows.html).
- Anaconda is a very popular distro of python, and is in fact the [preferred way ](http://qutip.org/docs/latest/installation.html#installing-with-conda) to do a QuTiP installation (for me the only way as the external installs of C++ compiler don't work). I use conda almost exclusively as I have worked on many projects that need that ability to manage requirements outside of Python.
- I would like to know more about the objections to using conda, what user feedback we have gotten etc. I feel the easiest way to have a uniform experience for our users is to use conda as the package management, and as we grow I feel like this won't be the only thing that this will be useful for. 
- Other proposals were to make it an optional dependency and then have error messages if a user tries to use it without it installed, they get a link to the install instructions. Also test partitioning would be needed to ensure if a user runs the tests locally without it, it doesn't fail

andreamari:
I would like to add my view on this topic. It looks that the are two different decisions:

**first decision**: `pip` vs `conda`. I don't have an opinion on this. I agree with the points of @crazy4pi314 in favor of `conda`, but I also wonder why many quantum projects (Cirq, Qiskit, PyQuil, PennyLane, etc.) decided to use `pip`. 

**second decision**:  should PyMC3 be a base requirement of Mitiq? This link looks a bit scary to me:  https://stackoverflow.com/questions/48334627/pyro-vs-pymc-what-are-the-difference-between-these-probabilistic-programming-fr .  The point is not just Theano vs PyTorch vs TensorFlow. But do we want to Mitiq to be unusable without Theano or PyTorch or TensorFlow ?

rmlarose:
> * Other proposals were to make it an optional dependency and then have error messages if a user tries to use it without it installed, they get a link to the install instructions. Also test partitioning would be needed to ensure if a user runs the tests locally without it, it doesn't fail

Decided @ MItiq meeting to do this!

PR #647, opened 2021-04-20
Improves PEC sampling performance via fewer conversions

### LINKED ISSUE #507: PEC with Qiskit is slow
Origin: https://github.com/unitaryfund/mitiq/pull/500#issuecomment-764640997

As mentioned there I believe it is because of conversions but am not sure. Tests with Qiskit like [this one](https://github.com/unitaryfund/mitiq/blob/master/mitiq/pec/tests/test_pec.py#L234) are markedly slower than with pyquil/cirq.

ckissane:
Description
-----------

Fixes #507
https://github.com/unitaryfund/mitiq/issues/507

With over 25x Speed Improvements for qiskit
And minor ~2x speed imrpovements for pyquil

Checklist
-----------

Tips
----

  autoformatter.


mitiq/mitiq_qiskit/conversions.py
```diff
@@ -46,6 +46,22 @@ def _remove_barriers(circuit: qiskit.QuantumCircuit) -> qiskit.QuantumCircuit:
...
+    Args:
+        qasm: QASM to remove barriers from.
+
+    Note:
+        According to the OpenQASM 2.X language specification
+        (https://arxiv.org/pdf/1707.03429v2.pdf), "Statements are separated by
+        semicolons. Whitespace is ignored. The language is case sensitive.
+        Comments begin with a pair of forward slashes and end with a new line."
+    """
+    # The inside regex removes comments, the outside regex removes barriers
```
rmlarose, line 61: Why remove comments?

rmlarose (commented):
Nice!! Thanks for this.

> With over 25x Speed Improvements for qiskit
> And minor ~2x speed imrpovements for pyquil

I believe it, but how did you measure this? It's mostly from the original conversion in `execute_with_pec`, right? How much savings comes from this vs using `_remove_qasm_barriers`? 

LGTM but I'll wait for @andreamari to take a look. Specifically, do you think we should keep `_remove_barriers` if we introduce `_remove_qasm_barriers` here? At minimum I think `_remove_barriers` should call `_remove_qasm_barriers`.

andreamari:
Thanks @ckissane! Very nice fix.

Could you also add a test (or two) to check that the new function `_remove_qasm_barriers` works as expected?
The place for this would be in: `mitiq/tests/test_conversions.py` .

I think we can remove the old  `_remove_barriers` function since now it is not used anymore. Unless you need it
for testing the new function  :-)

andreamari (changes requested):
Thanks @ckissane ! (see previous comment)

mitiq/pec/pec.py
```diff
@@ -120,10 +121,10 @@ def execute_with_pec(
 
     sampled_circuits = []
     signs = []
-
+    converted_circut, _ = convert_to_mitiq(circuit)
     for _ in range(num_samples):
         sampled_circuit, sign, _ = sample_circuit(
-            circuit, representations, random_state
+            converted_circut, representations, random_state
```
Aaron-Robertson, line 127: Just a nit comment here that `converted_circut` should be `converted_circuit` :sweat_smile:

Nice find on this repeated conversion!

mitiq/mitiq_qiskit/conversions.py
HaoTy: Removing the comments is to remove all the weird things a comment can contain, for example, a semicolon, which will interfere with our matching of the barrier statements. Qiskit's qasm parser won't care about comments anyway.

HaoTy (changes requested):
The first non-comment statement must be QASM's version number, ending with a semicolon, so a starting barrier is not an edge case. The current implementation, however, can fail on corner cases like when there is a statement `include "barrier.inc";`. Although it will almost never happen, I still suggest we change back to the previous implementation.

Edit: I believe I missed the case where a gate definition's closing curly bracket may appear between the semicolon and the barrier, so the regex still needs some tweaks to cover all corner cases.

briancylui:
Thanks for the feedback.  Addressed some PR comments in the latest commits.

@HaoTy and @ckissane: Thank you for the very nice fix - could you please report the runtime comparison of the corresponding test?  I tried to reproduce the claimed speed-up below, but unfortunately without much success.  Specifically, a report of the runtimes of the command 
```
pytest --durations=0 mitiq/pec/tests/test_pec.py -k 'test_qiskit_noiseless_decomposition_multiqubit' -vv`
```
at the root directory after running `pip install pytest` and commenting out [this test-skipping line](https://github.com/unitaryfund/mitiq/blob/c4794c8b270205cdd2f9627b1952d18712c73aee/mitiq/pec/tests/test_pec.py#L202) will be helpful.

To start with, I report the following local runtimes (in seconds) for the test `mitiq/pec/tests/test_pec.py::test_qiskit_noiseless_decomposition_multiqubit[num_qubits]` (`num_qubits` can be 1 or 2):
| Runtime (s) for test at branch/commit  | 1-qubit  | 2-qubit  |
|---|---|---|
| `master`  | 126.79  | 146.30  |
| `a816fe4` (`faster-PEC-qiskit` before my changes)  | 131.31  | 166.63  |
|  `f5f8599` (`faster-PEC-qiskit` after my changes, latest) | 125.86  | 150.24  |

According to these metrics, the current change doesn't seem to provide the claimed speed-up.  I believe the claimed speed-up was attained at some intermediate stages according to screenshots on the Mitiq channel of the Unitary Fund Discord server, which could have been masked by accident throughout the code clean-up process.

@ckissane and @HaoTy:  Could you please verify the runtime comparison (the exact figures will vary from machine to machine) and point out where in your code change is supposed to provide the speed-up?

ckissane:
_remove_qasm_barriers should now be very robust and the regex construction is much clearer

ckissane:
Should be fixed now

andreamari:
Thanks @ckissane !

I think you probably forgot about this comment:

> Could you also add a test (or two) to check that the new function `_remove_qasm_barriers` works as expected?
> The place for this would be in: `mitiq/tests/test_conversions.py` .
> 

For example one could test that `_remove_qasm_barriers` works as expected with a tricky input. E.g. a QASM with comments,
imports, etc..

If you are not familiar with [pytest](https://docs.pytest.org/en/6.2.x/getting-started.html), no problem. We can help.

ckissane:
> Thanks @ckissane !
> 
> I think you probably forgot about this comment:
> 
> > Could you also add a test (or two) to check that the new function `_remove_qasm_barriers` works as expected?
> > The place for this would be in: `mitiq/tests/test_conversions.py` .
> 
> For example one could test that `_remove_qasm_barriers` works as expected with a tricky input. E.g. a QASM with comments,
> imports, etc..
> 
> If you are not familiar with [pytest](https://docs.pytest.org/en/6.2.x/getting-started.html), no problem. We can help.

Ok I can add one that does tricky cases, I already have a test case I used when making it, I just need to commit it as a test.

ckissane:
> The first non-comment statement must be QASM's version number, ending with a semicolon, so a starting barrier is not an edge case. The current implementation, however, can fail on corner cases like when there is a statement `include "barrier.inc";`. Although it will almost never happen, I still suggest we change back to the previous implementation.
> 
> Edit: I believe I missed the case where a gate definition's closing curly bracket may appear between the semicolon and the barrier, so the regex still needs some tweaks to cover all corner cases.

This is all fixed now

andreamari:
Thanks @ckissane! LGTM it just needs some style fixing. Probably just trailing whitespaces: https://github.com/unitaryfund/mitiq/runs/2450153599

ckissane:
@andreamari that actually looks like a bug in flake8 considering trailing spaces matter in a multiline string :P.

I removed them anyway because it shouldn't effect the test.

andreamari (approved):
Thanks @ckissane, @HaoTy and @briancylui for this PR!
This is great and helpful contribution!

ckissane:
@andreamari merge?

rmlarose (approved):

crazy4pi314:
@all-contributors please add @HaoTy for code, bug

crazy4pi314:
@all-contributors please add @briancylui for code, bug

PR #648, opened 2021-04-21
Folding fixes and additions

### LINKED ISSUE #527: Add functionality to fold all gates of a certain type
A common literature technique is to "square CNOTs": see Fig. 2 caption of https://arxiv.org/abs/1803.03326 and Fig. 1 of https://arxiv.org/abs/2011.01157. This is not easily possible with unitary folding, even when folding by fidelity. 

I propose an interface like

```python
def fold_all(circuit: QPROGRAM, num_folds: int, gates_to_fold: List[something]) -> QPROGRAM:
```

which would do G (G^\dagger G)^{num_folds} for each gate G in `gates_to_fold` appearing in the `circuit`. 

With this, squaring CNOTs would be like `fold_all(circuit, num_folds=1, gates_to_fold=[CNOT])`. 

Alternatively, we could add `gates_to_fold` as an optional argument of each (local) folding method. Thoughts?

### LINKED ISSUE #624: Make it so `fold_left` and `fold_right` match on odd scale factors
We want to change `fold_gates_from_left` and `fold_gates_from_right` so that they match on odd scale factors. Namely, we want it so that

```python
fold_gates_from_left(circuit, 2k + 1) == fold_gates_from_right(circuit, 2k + 1)
```

for all integers `k`. This is to match how they are defined in https://arxiv.org/abs/2005.10921.

## Current behavior

Currently the circuit is folded up to a scale factor of 3, then this circuit is folded again treating all gates as though they were in the original circuit for scale factors beyond 3.

```python
import cirq
from mitiq.zne import scaling

circuit = cirq.Circuit(cirq.H(cirq.LineQubit(0)), cirq.CNOT(*cirq.LineQubit.range(2)))
# 0: ───H───@───
#           │
# 1: ───────X───

# They are the same for scale_factor = 3.
lscaled = scaling.fold_gates_from_left(circuit, 3)
rscaled = scaling.fold_gates_from_right(circuit, 3)
print(lscaled)
# 0: ───H───H───H───@───@───@───
#                   │   │   │
# 1: ───────────────X───X───X───

# They are different for scale_factor = 5, but we want them to be the same.
scaling.fold_gates_from_left(circuit, 5)
# 0: ───H───H───H───H───H───H───H───@───@───@───
#                                   │   │   │
# 1: ───────────────────────────────X───X───X───

scaling.fold_gates_from_right(circuit, 5)
# 0: ───H───H───H───@───@───@───@───@───@───@───
#                   │   │   │   │   │   │   │
# 1: ───────────────X───X───X───X───X───X───X───
```

For the last case (scale factor = 5), the returned circuit should be (for both left/right scaling)
```
0: ───H───H───H───H───H───@───@───@───@───@───
                          │   │   │   │   │
1: ───────────────────────X───X───X───X───X───
```

This would be a breaking change, so some scaling tests with scale factors > 3 may need to be modified.

rmlarose:
Description
-----------

Closes #527 and fixes #624.

New strategy for `fold_left` and `fold_right`, instead of a recursive call, is to:

- Determine the fraction of gates that needs to be folded.
- Fold these gates in the same way as before, keeping track of added virtual moments.
- Fold all moments an integer number of times to reach the scale factor, if needed, excluding virtual moments.

The last step calls for a `fold_all` type function as in #527, so I added that to close that issue as well.

Note: This is technically a breaking change although it shouldn't really affect anyone.

Checklist
-----------

Tips
----

  autoformatter.

### DIFF
--- a/mitiq/zne/scaling/folding.py
+++ b/mitiq/zne/scaling/folding.py
@@ -15,6 +15,7 @@
 
 """Functions for local and global unitary folding on supported circuits."""
 from copy import deepcopy
+import math
 from typing import (
     Any,
     Callable,
@@ -299,7 +300,7 @@ def _fold_all(
             folded.append(op, strategy=InsertStrategy.EARLIEST)
             if op.gate not in to_exclude:
                 folded.append(
-                    [op, inverse(op)] * num_folds,
+                    [inverse(op), op] * num_folds,
                     strategy=InsertStrategy.EARLIEST,
                 )
 
@@ -453,16 +454,11 @@ def fold_gates_from_left(
     if fidelities and not all(0.0 < f <= 1.0 for f in fidelities.values()):
         raise ValueError("Fidelities should be in the interval (0, 1].")
 
-    # Copy the circuit and remove measurements
+    # Copy the circuit and remove measurements.
     folded = deepcopy(circuit)
     measurements = _pop_measurements(folded)
 
-    # Determine the stopping condition for folding
-    scale_factor, integer_num_folds = (
-        scale_factor % 3.0 + float(scale_factor > 3.0),
-        int(scale_factor // 3.0),
-    )
-
+    # Determine the stopping condition.
     ngates = len(list(folded.all_operations()))
     weights: Optional[Dict[str, float]]
     if fidelities:
@@ -472,14 +468,15 @@ def fold_gates_from_left(
         weights = None
         total_weight = ngates
 
-    stop = total_weight * (scale_factor - 1.0) / 2.0
+    fraction_to_fold = (scale_factor - 1.0) / 2.0
+    fraction_to_fold, integer_num_folds = math.modf(fraction_to_fold)
+    stop = total_weight * fraction_to_fold
 
-    # Fold gates from left to reach scale_factor % 3.
+    # Do the fractional folding (i.e., fold some but not all gates).
     tot = 0.0
     moment_shift = 0
     virtual_moments: List[int] = []
     for (moment_index, moment) in enumerate(circuit):
-        # TODO: This is ugly. Refactor so the stopping condition isn't duplicated.
         if tot >= stop:
             break
 
@@ -503,16 +500,17 @@ def fold_gates_from_left(
             if tot >= stop:
                 break
 
-    # Fold all gates to reach scale_factor // 3.
-    # TODO: This needs to account for fidelities, in particular perfect fidelity gates.
+    # Do the integer folding (i.e., fold all gates an equal number of times).
+    if not weights:
+        weights = dict()
+    exclude = frozenset(k for k, v in weights.items() if v == 0.0)
+
     if integer_num_folds > 0:
-        folded_moments = range(len(folded))
-        _fold_moments(
+        folded = _fold_all(
             folded,
-            moment_indices=[
-                i for i in folded_moments if i not in virtual_moments
-            ],
-            num_folds=integer_num_folds,
+            num_folds=int(integer_num_folds),
+            exclude=exclude,
+            skip_moments=frozenset(virtual_moments),
         )
 
     _append_measurements(folded, measurements)
--- a/mitiq/zne/scaling/folding.py
+++ b/mitiq/zne/scaling/folding.py
@@ -20,6 +20,7 @@ from typing import (
     Callable,
     Collection,
     Dict,
+    FrozenSet,
     Iterable,
     List,
     Optional,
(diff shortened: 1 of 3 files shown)


mitiq/zne/scaling/folding.py
```diff
@@ -289,6 +367,45 @@ def _get_num_to_fold(scale_factor: float, ngates: int) -> int:
...
+    """
+    _check_foldable(circuit)
+
+    folded = deepcopy(circuit)
+    measurements = _pop_measurements(folded)
+
+    folded = _fold_all(folded, num_folds, exclude)
+
+    _append_measurements(folded, measurements)
+    return folded
```
andreamari, line 406: This function is not properly a noise scaling function since it does not take a `scale_factor` as an input. So it should not be directly used as an argument of `execute_with_zne` .  Maybe we could:

- Add a note to warn the user about this problem.
or
- Change the function to 
```
@noise_scaling_converter
def uniform_scaling(
    circuit: QPROGRAM,
    scale_factor: int,
    exclude: FrozenSet[Any] = frozenset(),
) -> QPROGRAM:
    """Returns a circuit with all gates folded locally. This uniform
    folding method can effectively scale the noise by odd integers.
    Args:
        circuit: Circuit to fold.
        scale_factor: The noise scale factor. It must be an odd integer.
        exclude: Do not fold these gates. Supported gate keys are listed in
            the following table.::
                Gate key    | Gate
                -------------------------
                "H"         | Hadamard
                "X"         | Pauli X
                "Y"         | Pauli Y
                "Z"         | Pauli Z
                "I"         | Identity
                "CNOT"      | CNOT
                "CZ"        | CZ gate
                "TOFFOLI"   | Toffoli gate
                "single"    | All single qubit gates
                "double"    | All two-qubit gates
                "triple"    | All three-qubit gates
    """
    _check_foldable(circuit)

    folded = deepcopy(circuit)

    measurements = _pop_measurements(folded)
    
    if scale_factor is not np.close to an odd integer:
           raise error

    num_folds = int((scale_factor - 1) / 2)
    folded = _fold_all(folded, num_folds, exclude)

    _append_measurements(folded, measurements)
    return folded

File: mitiq/zne/scaling/folding.py [hunk 2]
```diff
@@ -289,6 +367,45 @@ def _get_num_to_fold(scale_factor: float, ngates: int) -> int:
 
 
 # Local folding functions
+@noise_scaling_converter
+def fold_all(
+    circuit: QPROGRAM,
+    num_folds: int = 1,
+    exclude: FrozenSet[Any] = frozenset(),
```
andreamari, line 374:
```suggestion
    exclude: FrozenSet[str] = frozenset(),
```

andreamari (commented):
Thanks @rmlarose!  

I am actually also working on #624 using a different approach which should clean up our local folding methods (including `fold_gates_at_random` which suffers from the same recursive problems of `fold_gates_from_left/right`). A draft is here #649.

My idea is to first define function which just produces a list of integers (a kind of mask) telling
how many times each gate of a circuit should be folded.  In this way it is very easy to define
all folding methods in an abstract but intuitive way without using circuits at all.

After this, we only need a unique function which folds the gates of a circuit according to a folding mask (list of integers). 

But I think it is not a big problem if we now apply your changes and later we consider my changes.

rmlarose:
> I am actually also working on #624 using a different approach which should clean up our local folding methods (including `fold_gates_at_random` which suffers from the same recursive problems of `fold_gates_from_left/right`). A draft is here #649.

Oh sweet. Let's keep this open as an alternate then. Except I would like to include `fold_all`, let's see if I can convince you.

mitiq/zne/scaling/folding.py
rmlarose: Ok I think you convinced me. It's a bit of a bummer we can't use `num_folds` here as it's so natural to think about and I've gotten this feedback multiple times. But as far as extrapolate to 0 is concerned we really need 0 to correspond to no noise, but `num_folds=0` is the base noise not zero noise.

mitiq/zne/scaling/folding.py
rmlarose: I would keep this `Any` because Cirq gates are also allowed.

rmlarose:
Back to you @andreamari. Up to you how to handle this with #649.

andreamari (approved):
LGTM! I think we can marge this. The new PR #649 may introduce additional changes as a second step on top of this.

PR #649, opened 2021-04-21
Update code of local folding functions

### LINKED ISSUE #624: Make it so `fold_left` and `fold_right` match on odd scale factors

### LINKED ISSUE #650: Make folding at_random and from_right and from_left consistent with odd scale factors
This is issue is equivalent to #624 but related to the function `fold_gates_at_random`.

For a description of the problem see #624.

There is already [WIP] PR (#649 ) related to local folding functions and could also fix this issue.

andreamari:
Description
-----------

Draft PR to:

- Simplify the code of folding functions. 

- fixes #650 

Checklist
-----------

Tips
----

  autoformatter.


andreamari:
@rmlarose , this is now ready for review.

Main change 1: re-written local folding functions with via folding masks. This also fixes #650.
Main change 2: removed many helper functions which now are not used anymore.

Minor note 1: I made `squash_moments` private such that now all public functions in zne.folding are valid scaling methods which can be safely imported and used for ZNE. If you don't like this, I can revert it to public. (Maybe utils.py file in `mitiq/zne/scaling` could be useful in the future).

Minor note 2: you don't need to review `fold_global`. Even if it is marked as green in the diff, it was actually unchanged.

mitiq/zne/scaling/folding.py
```diff
@@ -124,7 +116,7 @@ def _check_foldable(circuit: Circuit) -> None:
         )
 
 
-def squash_moments(circuit: Circuit) -> Circuit:
+def _squash_moments(circuit: Circuit) -> Circuit:
```
rmlarose, line 119: No change needed, but we really need to make submodule structure consistent. Here, `squash_moments` is never imported in `zne.scaling.__init__.py`, so it is private and the underscore is extraneous. In other places though (`zne/inference.py`) the underscore is needed because the module structure is different.

mitiq/zne/scaling/folding.py
```diff
@@ -421,6 +289,262 @@ def fold_all(
...
+            docstring of local folding function for mode details.
+
+    Returns: The list of weights associated to all the gates.
+    """
+    if fidelities and not all(0.0 < f <= 1.0 for f in fidelities.values()):
+        raise ValueError("Fidelities should be in the interval (0, 1].")
+    weights = None
+    if fidelities:
+        # Round to avoid ugly numbers like 0.09999999999999998 instead of 0.1
+        # https://docs.python.org/2/tutorial/floatingpoint.html
```
rmlarose, line 367: Since we're python 3+ I would remove or update this link. Seems more or less self-explanatory without a reference though.

mitiq/zne/scaling/folding.py
```diff
@@ -421,6 +289,262 @@ def fold_all(
...
+    num_folds_mask = []
+    for w in weight_mask:
+        if np.isclose(w, 0.0):
+            num_folds_mask.append(0)
+        else:
+            num_folds_mask.append(num_uniform_folds)
+
+    # If the scale_factor is an odd integer, we are done.
+    if np.isclose(odd_integer_scale_factor, scale_factor):
+
```
rmlarose, line 441:
```suggestion
```

mitiq/zne/scaling/folding.py
```diff
@@ -421,6 +289,262 @@ def fold_all(
...
+    weight_mask: List[float],
+    scale_factor: float,
+    folding_method: str,
+    seed: Optional[int] = None,
+) -> List[float]:
+    r"""Returns a list of integers determining how many times each gate a
+    circuit should be folded to realize the desired input scale_factor.
+
+    More precicely, the j_th element of the returned list is associated
+    to the j_th gate G_j of a circuit that we want to scale and determines
```
rmlarose, line 388: "jth gate of a circuit" could be stated more clearly - a circuit is a 2d object. (What is the jth element of a matrix?)

mitiq/zne/scaling/folding.py
```diff
@@ -421,6 +289,262 @@ def fold_all(
...
+
+def _apply_fold_mask(
+    circuit: Circuit,
+    num_folds_mask: List[int],
+    squash_moments: Optional[bool] = True,
+) -> Circuit:
+    r"""Applies local unitary folding to the gates of the input circuit
+    according to the input num_folds_mask.
+
+    More precicely, G_j^\dag G_j is applied after the j_th gate G_j of
```
rmlarose, line 497: Same (minor) comment above about the linear indexing a circuit.

mitiq/zne/scaling/folding.py
```diff
@@ -421,6 +289,262 @@ def fold_all(
...
+            placed as early as possible in the circuit. If False, new moments
+            are created for folded gates. This option only applies to QPROGRAM
+            types which have a "moment" or "time" structure. Default is True.
+
+    Returns: The folded quantum circuit.
+    """
+    if squash_moments is False:
+        pass
+    else:
+        squash_moments = True
```
rmlarose, line 514: huh?

mitiq/zne/scaling/folding.py
```diff
@@ -421,6 +289,262 @@ def fold_all(
...
+            f" The number of gates is {num_gates}"
+            f" but len(num_folds_mask) is {len(num_folds_mask)}."
+        )
+
+    folded_circuit = circ_copy[:0]
+    if squash_moments:
+        for op, num_folds in zip(circ_copy.all_operations(), num_folds_mask):
+            folded_circuit.append([op] + num_folds * [inverse(op), op])
+    else:
+        index = 0
```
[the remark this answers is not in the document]
rmlarose: for clarity

mitiq/zne/scaling/tests/test_folding.py
```diff
@@ -931,6 +618,14 @@ def test_fold_left_and_fold_right_match_on_odd_scale_factors():
             require_measurement_equality=True,
         )
 
+    for s in (3, 7, 15):
+        assert _equal(
+            fold_gates_from_left(circuit, s),
+            fold_gates_at_random(circuit, s),
```
rmlarose, line 624: Why don't you test `from_right` here too?

mitiq/zne/scaling/tests/test_folding.py
```diff
@@ -1848,3 +1482,255 @@ def test_folding_keeps_measurement_order_with_qiskit():
 
     folded = fold_gates_at_random(circuit, scale_factor=1.0)
     assert folded == circuit
+
+
+def test_create_weight_mask_with_fidelities():
+    """Check the correct weight mask is created."""
```
rmlarose, line 1488: Not that descriptive of a comment, can be inferred from test name. Add more description or remove.

mitiq/zne/scaling/tests/test_folding.py
```diff
@@ -1848,3 +1482,255 @@ def test_folding_keeps_measurement_order_with_qiskit():
...
+@pytest.mark.parametrize(
+    "weight_mask",
+    ([0.1, 0.2, 0.3, 0.0], [0.3, 0.5, 0.7, 0.0], [1.0, 1.0, 1.0, 0.0]),
+)
+@pytest.mark.parametrize("scale_factor", (1, 3, 5, 7, 9, 11))
+@pytest.mark.parametrize("method", ("at_random", "from_left", "from_right"))
+def test_create_fold_mask_with_odd_scale_factors(
+    weight_mask, scale_factor, method,
+):
+    """Check _create_fold_mask works as expected"""
```
rmlarose, line 1520: Not that descriptive of a comment, can be inferred from test name. Add more description or remove.

mitiq/zne/scaling/tests/test_folding.py
```diff
@@ -1848,3 +1482,255 @@ def test_folding_keeps_measurement_order_with_qiskit():
...
+    fold_mask = _create_fold_mask(weight_mask, scale_factor, method)
+    num_folds = int((scale_factor - 1) / 2)
+    assert fold_mask == [num_folds, num_folds, num_folds, 0]
+
+
+def test_create_fold_mask_with_real_scale_factors_from_left():
+    """Check _create_fold_mask works as expected"""
```
rmlarose, line 1527: Not that descriptive of a comment, can be inferred from test name. Add more description or remove.

mitiq/zne/scaling/tests/test_folding.py
```diff
@@ -1848,3 +1482,255 @@ def test_folding_keeps_measurement_order_with_qiskit():
...
+    fold_mask = _create_fold_mask(
+        weight_mask=[1.0, 1.0, 1.0, 1.0],
+        scale_factor=3.9,
+        folding_method="from_left",
+    )
+    assert fold_mask == [2, 2, 1, 1]
+
+
+def test_create_fold_mask_with_real_scale_factors_from_right():
+    """Check _create_fold_mask works as expected"""
```
rmlarose, line 1559: Not that descriptive of a comment, can be inferred from test name. Add more description or remove.

rmlarose: ditto for remaining tests

rmlarose (changes requested):
This is very nice, thanks @andreamari. I'm happy to see 300 fewer lines of code here. See inline comments before merging.

Can be merged without remaining comments, but food for thought and perhaps future work:

- Should the masks be two-dimensional instead of one-dimensional to match circuit structure? It's easy enough to `np.reshape` between these, 2d seems a little more natural to me.
- Should `_apply_fold_mask` be user-facing? (Involves above comment.) This is quite general and perhaps useful.
- Should we recommend that custom folding methods be written using `_apply_fold_mask`? 
- I'm not the biggest fan of `folding_method` being a string in `_create_fold_mask`, but I guess it's ok.

mitiq/zne/scaling/folding.py
andreamari: I clarified this in the docstring(s) of all the 3 functions related to masks.
This is the addition:
 ```
    The gate ordering is equal to the one used in the `all_operations()`
    method of the :class:`cirq.Circuit` class: gates from earlier moments
    come first and gates within the same moment follow the order in which
    they were given to the moment's constructor.
```

This comment is similar to the docstring of `all_operations`:

https://github.com/quantumlib/Cirq/blob/0ab45cf24e53a0de3b60404ce5f713dc92491cc3/cirq/circuits/circuit.py#L868-L875

File: mitiq/zne/scaling/folding.py [hunk 13]
```diff
@@ -421,6 +289,262 @@ def fold_all(
...
+            num_folds_mask.append(num_uniform_folds)
+
+    # If the scale_factor is an odd integer, we are done.
+    if np.isclose(odd_integer_scale_factor, scale_factor):
+
+        return num_folds_mask
+
+    # If necessary, fold a subset of gates to approximate the scale_factor
+    input_circuit_weight = sum(weight_mask)
+    output_circuit_weight = odd_integer_scale_factor * input_circuit_weight
```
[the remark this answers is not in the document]
andreamari: I agree! Done in this way: 946ba78

mitiq/zne/scaling/folding.py
```diff
@@ -421,6 +289,262 @@ def fold_all(
...
+            new_output_circuit_weight - scale_factor * input_circuit_weight
+        )
+        # Fold the candidate gate only if it helps improving the approximation
+        if new_approx_error < approx_error:
+            approx_error = new_approx_error
+            output_circuit_weight = new_output_circuit_weight
+            num_folds_mask[j] += 1
+        # TODO: decide if removing next else case to get better approximations
+        else:
+            break
```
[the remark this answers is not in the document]
andreamari: I removed the TODO (946ba78). It was just an idea to improve the scale factor approximation but I think it is better to keep the behavior of the function as simple and predictable as possible to avoid confusing the user.

mitiq/zne/scaling/folding.py
andreamari: e063dfa

mitiq/zne/scaling/folding.py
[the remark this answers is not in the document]
andreamari: renamed to `mask_index`.

mitiq/zne/scaling/tests/test_folding.py
andreamari: I think it is redundant. 
First we test that `from_left == from_right` (right above),  then we test `from_left == at_random` .  So if both tests pass, it should automatically imply that `from_right == at_random` .

mitiq/zne/scaling/folding.py
andreamari: it was a very ugly way of setting `squash_moments = True` when `squash_moments = None`.  Now I removed this and used the default option of the `kwargs.get() ` method instead.  ( 9bb5e16)

andreamari:
About your general comments:

> - Should the masks be two-dimensional instead of one-dimensional to match circuit structure? It's easy enough to `np.reshape` between these, 2d seems a little more natural to me.
I see your point but I am not sure about it, I think we can discuss about a 2D mask in the future. My fear is that it could make
things more complicated. I see that Cirq use moments (which is a nice thing to handle idle pauses), but it is also true that in theoretical descriptions circuits are naturally written as ordered sequences of operations. Also, for users working with other circuit types (qiskit, pyquil) the 1D mask is perhaps more natural.  But happy to discuss and change my mind on this.

> - Should `_apply_fold_mask` be user-facing? (Involves above comment.) This is quite general and perhaps useful.
I think it is possible. However, before doing this, it would be good to clearly separate "standard" scaling functions (where scale_factor is an argument) from other anomalous functions like this one. Maybe using different files. 

> - Should we recommend that custom folding methods be written using `_apply_fold_mask`?
It can be a suggestion but not a requirement. Sometimes it is not appropriate e.g. `fold_global` and sometimes a direct approach is probably simpler (e.g. `fold_all`) . 
- I'm not the biggest fan of `folding_method` being a string in `_create_fold_mask`, but I guess it's ok.
As long as it is private I think it is probably fine. If we need something more flexible in the future, we could use a list of indices instead of a string to identify a generic folding order.

mitiq/zne/scaling/tests/test_folding.py
rmlarose: Oh, missed it above because it didn't show up in the diff, oops!

rmlarose (approved):

PR #654, opened 2021-04-28
Fix custom factory example in docs

andreamari:
I noticed that the custom factory example in the docs passes current tests but is actually not very appropriate.

The problem is that the `extrapolate` function returns a float, however a `Factory` expects also additional data in the case `full_output=True`.
This can create problems, for example when calling the `reduce()` method.
These problems are not captured by the current tests since they simply call the static function.

The proposed solution of this PR is to first call the existing `extrapolate` function of `LinearFactory` and, after this, clip the result.

Whoever has time, can have a look. Not urgent.


rmlarose (approved):
LGTM

PR #672, opened 2021-05-12
Remove classical register transformations in Qiskit conversions

### LINKED ISSUE #561: Only need to transform qubit registers in Qiskit conversions, not classical registers
Originally classical registers were transformed, but this didn't preserve the order of measurements (#551). The measurement order was fixed in #557 by removing all measurements then adding them back in the original order. It is therefore no longer needed to transform the classical registers (but this is still done). Rather, one can:

* Remove all measurements and classical registers, storing the classical registers.
* Do any operations on the Qiskit circuit, e.g. fold it.
* Add the classical registers back to the modified circuit, then add measurements in the correct order.

This simplifies the `_transform_registers` function in `mitiq_qiskit.conversions`.

Aaron-Robertson:
Description
-----------
Resolves #561 

Simplifies the basic `to_qiskit` conversion and the `transform_registers` post process.


Aaron-Robertson:
@rmlarose Is this approach what you had in mind for #561? I removed/readded the registers without transformation, but there are a few cases where I'm not sure that the result is intended. If it is intentional I can change the tests to match.

rmlarose:
> @rmlarose Is this approach what you had in mind for #561? I removed/readded the registers without transformation, but there are a few cases where I'm not sure that the result is intended. If it is intentional I can change the tests to match.

Yes, more or less. What tests are failing / what results are different now?

I think you can also remove the classical bit/register options from `_map_bits`.

Aaron-Robertson:
Alright made a bit more progress here. It's failing on tests with measurement, and I think it's because the converted qubits have different name(s)? `q1_0` vs `q_0`:
```
q, c = QuantumRegister(2), ClassicalRegister(2)
circuit = QuantumCircuit(q, c)
circuit.x(q[0])
circuit.measure(range(2), range(2))
print(circuit)
converted = from_qiskit(circuit)
converted_back = to_qiskit(from_qiskit(circuit))
print(converted_back)

      ┌───┐┌─┐
q1_0: ┤ X ├┤M├
      └┬─┬┘└╥┘
q1_1: ─┤M├──╫─
       └╥┘  ║ 
c1: 2/══╩═══╩═
        1   0 
          ┌───┐┌─┐
     q_0: ┤ X ├┤M├
          └┬─┬┘└╥┘
     q_1: ─┤M├──╫─
           └╥┘  ║ 
m_c1_1: 1/══╩═══╬═
            0   ║ 
                ║ 
m_c1_0: 1/══════╩═
                0 
```

rmlarose:
> I think it's because the converted qubits have different name(s)? `q1_0` vs `q_0`

They need to have the same names to be equal. It looks like you're skipping the part which does the renaming? All the (qu)bits are different in the circuits below. 

> ```
>       ┌───┐┌─┐
> q1_0: ┤ X ├┤M├
>       └┬─┬┘└╥┘
> q1_1: ─┤M├──╫─
>        └╥┘  ║ 
> c1: 2/══╩═══╩═
>         1   0 
>           ┌───┐┌─┐
>      q_0: ┤ X ├┤M├
>           └┬─┬┘└╥┘
>      q_1: ─┤M├──╫─
>            └╥┘  ║ 
> m_c1_1: 1/══╩═══╬═
>             0   ║ 
>                 ║ 
> m_c1_0: 1/══════╩═
>                 0
> ```

Aaron-Robertson:
Should have a chance to wrap this one up this weekend, but not really seeing why the renaming does not occur. @rmlarose 
 if you have a couple minutes to take a look I'd appreciate it. I'm somewhat confused about the differences in testing scaled vs converted measurements that I see in these tests/results:
```
from qiskit import QuantumRegister, ClassicalRegister, QuantumCircuit
from mitiq.mitiq_qiskit.conversions import from_qiskit, to_qiskit

q, c = QuantumRegister(2), ClassicalRegister(2)
circuit = QuantumCircuit(q, c)
circuit.x(q[0])
circuit.measure(range(2), range(2))
print(circuit)
converted = from_qiskit(circuit)

converted_back = to_qiskit(from_qiskit(circuit))
print(converted_back)
      ┌───┐┌─┐
q0_0: ┤ X ├┤M├
      └┬─┬┘└╥┘
q0_1: ─┤M├──╫─
       └╥┘  ║ 
c0: 2/══╩═══╩═
        1   0 
          ┌───┐┌─┐
     q_0: ┤ X ├┤M├
          └┬─┬┘└╥┘
     q_1: ─┤M├──╫─
           └╥┘  ║ 
m_c0_1: 1/══╩═══╬═
            0   ║ 
                ║ 
m_c0_0: 1/══════╩═
                0 
```
and
```
from qiskit import QuantumRegister, ClassicalRegister, QuantumCircuit

q, c = QuantumRegister(2), ClassicalRegister(2)
circuit = QuantumCircuit(q, c)
circuit.h(q[0])
circuit.measure(q, c)

print(circuit)

folded = fold_gates_at_random(circuit, scale_factor=1.)

print(folded)
      ┌───┐┌─┐
q1_0: ┤ H ├┤M├
      └┬─┬┘└╥┘
q1_1: ─┤M├──╫─
       └╥┘  ║ 
c1: 2/══╩═══╩═
        1   0 
      ┌───┐┌─┐
q1_0: ┤ H ├┤M├
      └┬─┬┘└╥┘
q1_1: ─┤M├──╫─
       └╥┘  ║ 
c1: 2/══╩═══╩═
        1   0 
```

rmlarose:
> I'm somewhat confused about the differences in converted vs scaled measurements that I see in these tests/results:

This (and the two examples you show) is because there is [additional logic](https://github.com/unitaryfund/mitiq/blob/master/mitiq/conversions.py#L188-L205) when converting circuits for noise scaling. 

If you just call `mitiq_qiskit.[to|from]_qiskit`, this additional logic doesn't happen, so you see in the first example there are two classical registers with different names in `converted_back`.

This was screwing up things when running ZNE with Qiskit, because the structure of measurement (counts) dictionaries in Qiskit depends on the classical register structure. So we had to add the additional logic to preserve the register structure to make it usable.

So this is expected behavior. But I'm not sure I'm answering your question - does this help?

mitiq/mitiq_qiskit/conversions.py
```diff
@@ -269,42 +251,19 @@ def to_qasm(circuit: cirq.Circuit) -> QASMType:
...
-def to_qiskit(
-    circuit: cirq.Circuit,
-    qregs: Optional[List[qiskit.QuantumRegister]] = None,
-    cregs: Optional[List[qiskit.ClassicalRegister]] = None,
-    add_cregs_if_cannot_transform: bool = True,
-) -> qiskit.QuantumCircuit:
-    """Returns a Qiskit circuit equivalent to the input Mitiq circuit.
+def to_qiskit(circuit: cirq.Circuit,) -> qiskit.QuantumCircuit:
+    """Returns a Qiskit circuit equivalent to the input Mitiq circuit. Note
+    that the output circuit register names do not match the input.
```
rmlarose, line 256:
```suggestion
    that the output circuit registers may not match the input circuit registers.
```

mitiq/mitiq_qiskit/conversions.py
```diff
@@ -269,42 +251,19 @@ def to_qasm(circuit: cirq.Circuit) -> QASMType:
...
 
 
-def to_qiskit(
-    circuit: cirq.Circuit,
-    qregs: Optional[List[qiskit.QuantumRegister]] = None,
-    cregs: Optional[List[qiskit.ClassicalRegister]] = None,
-    add_cregs_if_cannot_transform: bool = True,
-) -> qiskit.QuantumCircuit:
-    """Returns a Qiskit circuit equivalent to the input Mitiq circuit.
+def to_qiskit(circuit: cirq.Circuit,) -> qiskit.QuantumCircuit:
```
rmlarose, line 254:
```suggestion
def to_qiskit(circuit: cirq.Circuit) -> qiskit.QuantumCircuit:
```

mitiq/mitiq_qiskit/conversions.py
```diff
@@ -93,27 +93,25 @@ def _map_bit_index(
 
 
 def _map_bits(
-    bits: List[Union[qiskit.circuit.Qubit, qiskit.circuit.Clbit]],
-    registers: List[Union[qiskit.QuantumRegister, qiskit.ClassicalRegister]],
+    bits: List[qiskit.circuit.Qubit],
```
rmlarose, line 96:
```suggestion
    qubits: List[qiskit.circuit.Qubit],
```

rmlarose: I think we should also rename `_map_bits` to `_map_qubits`.

mitiq/mitiq_qiskit/conversions.py
```diff
@@ -269,42 +251,19 @@ def to_qasm(circuit: cirq.Circuit) -> QASMType:
     return circuit.to_qasm()
 
 
-def to_qiskit(
-    circuit: cirq.Circuit,
-    qregs: Optional[List[qiskit.QuantumRegister]] = None,
-    cregs: Optional[List[qiskit.ClassicalRegister]] = None,
-    add_cregs_if_cannot_transform: bool = True,
```
rmlarose, line 276: When we discussed this @ Mitiq meeting, I totally forgot these were also arguments to `to_qiskit`. I think it's fine to remove them, since we will always transform registers in noise scaling functions via the `noise_scaling_converter`. But pinging @andreamari for a second opinion on removing these arguments. Sorry for the confusion @Aaron-Robertson, mea culpa.

rmlarose (commented):
Thanks @Aaron-Robertson!! This is looking good, just a few minor comments to address then good to go.

mitiq/mitiq_qiskit/conversions.py
andreamari: @rmlarose I don't see big problems at the moment. PEC is quite delicate with respect to the names of registers, but if all tests pass, it's fine.

Aaron-Robertson:
@rmlarose I think this one's ready to go!

mitiq/mitiq_qiskit/conversions.py
rmlarose: Sounds good, it's fine to reintroduce if needed.

rmlarose (approved):

PR #677, opened 2021-05-14
CDR part two - Clifford data regression functions to fit training data

L-P-B:
Description
-----------
In this pull request I include function to increase the noise in the training circuits and circuits of interest. Furthermore the functions to execute those circuits and collect the data into a training set are included. Additionally, the functions that use that data to complete a regression are included. 

The only missing part (left for the next pull request) is the `execute_with_cdr` (and maybe `execute_cdr_with_training_data`) function which will tie all the pieces in this request together. 

This is a work in progress. 

For a detailed look at the methods implemented see: Czarnik et al (https://arxiv.org/abs/2005.10189) and Lowe et al (https://arxiv.org/abs/2011.01157).

Checklist
-----------

Tips
----


mitiq/cdr/data_regression.py
```diff
@@ -0,0 +1,294 @@
...
+    if intercept:
+        a = list(args[0][: noise_levels + 1])
+        return f_intercept(x, a)
+    else:
+        a = list(args[0][:noise_levels])
+        return f(x, a)
+
+
+# TODO: Discuss if this format for converting shots makes sense in general.
+def dictionary_to_state_counts(counts: Dict, Q: int) -> Dict:
```
rmlarose, line 269: Q -> nqubits

mitiq/cdr/cdr_execution.py
```diff
@@ -0,0 +1,208 @@
...
+
+def execute_with_CDR(
+    circuit: Circuit,
+    executor: Callable[[Circuit], dict],
+    simulator: Callable[[Circuit], Union[dict, np.ndarray]],
+    observable: Union[List[np.ndarray], np.ndarray],
+    num_training_circuits: int,
+    fraction_non_clifford: float,
+    scale_noise: Optional[Callable[[Circuit, float], Circuit]] = None,
+    noise_scaling_factors: Optional[Union[float, List[float]]] = None,
```
rmlarose, line 46: How common is it to have just a float? Can we just force this to always be a list?

rmlarose, line 46: Use `scale_factors` to be consistent with ZNE.

mitiq/cdr/data_regression.py
```diff
@@ -0,0 +1,294 @@
...
+        all_training_circuits: list of all training circuits at all noise
+                               levels.
+        executor: user defined executor function.
+        simulator: user defined simulator function either returns counts or
+                   a statevector.
+        noise_levels: number of noise levels to be used in the regression.
+    """
+    number_of_training_circuits = int(
+        len(all_training_circuits) / noise_levels
+    )
```
rmlarose, line 70: This is extremely confusing. If I understand right, the logic is:

- Run all circuits with `executor`.
- Run circuits with scale factor 1 with `simulator`.

If so, just get rid of this helper function altogether. You probably want to store the circuits in a way that's amenable to what you want to do. How about a 2d structure with dimensions # of training circuits x # of scale factors? Then you just pass in the first column to `simulator` and the all circuits to `executor`. 

```python
circuits =  2d data structure described above
executor_data = [executor(c) for c in circuits.flatten()]
simulator_data = [simulator(c) for c in 1st column of circuits]
```

It's much shorter and much more readable.

This should take into account batching as well, but let's go one step at a time.

mitiq/cdr/data_regression.py
```diff
@@ -0,0 +1,294 @@
...
+            training_circuit_simulated_result = simulator(training_circuit)
+            training_circuits_simulated_data.append(
+                training_circuit_simulated_result
+            )
+    return (training_circuits_raw_data, training_circuits_simulated_data)
+
+
+# TODO: this function below is not really necessary, but is quite useful and
+# follows the structure of the above format.
+def execute_circuit_of_interest(
```
rmlarose, line 87: Again I don't think you really want or need a helper function for this. The way it's called by `execute_with_cdr`, you just need

```python
scaled_circuits = [scale_noise(circuit, s) for s in scale_factors]
results_dict_circuit_of_interest = [executor(c) for c in scaled_circuits]
```

mitiq/cdr/cdr_execution.py
```diff
@@ -0,0 +1,208 @@
...
+        executor,
+        simulator,
+        noise_levels,
+    )
+    # run the circuit of interest in the desired backend:
+    results_dict_circuit_of_interest = execute_circuit_of_interest(
+        circuits, executor
+    )
+
+    if isinstance(observable, list):
```
rmlarose, line 174: Either (i) require it to always be a list or (ii) make it a list if it's not already one. We don't want to repeat this entire codeblock verbatim.

mitiq/cdr/cdr_execution.py
```diff
@@ -0,0 +1,208 @@
...
+
+from data_regression import (
+    scale_noise_in_circuits,
+    execute_training_circuits,
+    execute_circuit_of_interest,
+    construct_training_data_floats,
+    construct_circuit_data_floats,
+    find_optimal_parameters,
+    wrapper_fit_func,
+    calculate_observable,
```
rmlarose, line 34: This isn't used.

mitiq/cdr/data_regression.py
```diff
@@ -0,0 +1,294 @@
...
+            Y_data[count] = training_obs_sim
+    return (X_data, Y_data)
+
+
+# TODO: again this function below is not really necessary, but is quite useful
+#  and follows the structure of the above format.
+
+
+def construct_circuit_data_floats(
+    circuit_data: List[Dict], observable: np.ndarray, noise_levels: int = 1,
```
rmlarose, line 190: noise_levels isn't used.

mitiq/cdr/data_regression.py
```diff
@@ -0,0 +1,294 @@
...
+    return sum(a * x for a, x in zip(params, X_data))
+
+
+def wrapper_fit_func(x, noise_levels, *args, intercept: bool = True):
+    if intercept:
+        a = list(args[0][: noise_levels + 1])
+        return f_intercept(x, a)
+    else:
+        a = list(args[0][:noise_levels])
+        return f(x, a)
```
rmlarose, line 265: Can't you just have one function which always has an intercept that could be zero?

rmlarose (changes requested):
See inline comments. I think we can & should remove a lot of lines of code here. In my opinion, the only functions that should be in `data_regression.py` are `find_optimal_parameters` and *one* linear fit function.

mitiq/cdr/cdr_execution.py
L-P-B: Yes I think thats fine.

mitiq/cdr/data_regression.py
L-P-B: Do you mean setting the intercept to zero? Or allowing it to fit and expecting it to be zero? 

Having the extra freedom to fit an intercept changes the fit. We found for vnCDR including a constant makes the ansatz less natural to motivate. Whereas for CDR it is more natural. In experiments on real data I have normally found including a constant to be better for both vnCDR and CDR. 

Personally I am not tied keeping this option if you think it is messy and arbitrary. In the future it would be ideal for a user to define their own functions to use for the regression anyway.

nathanshammah:
@L-P-B, as a suggestion, opening some issues for the upcoming parts of CDR could help have a comprehensive picture.  Or even a "master" CDR issue detailing the relation among the various PRs (and children issues). 

Some of the tasks of this PR could have been broken down in different PRs, possibly all addressing a specific part of an open issue. Besides the formal structure, this can help in review, as less changes need to be looked into. (There are also pros in having one big PR for coherence, but this can slow down the review stage. A way to have both coherence and sliced down PRs could be to work on a specific remote branch with all changes and then open specific branches and hence PRs for target features.)

mitiq/cdr/cdr_execution.py
rmlarose: Great. That's consistent with setting scale factors in `Factory`s which always take a list, and makes the function easier for you to implement.

mitiq/cdr/data_regression.py
rmlarose: Gotcha. I'm fine leaving the two functions then. 

> In the future it would be ideal for a user to define their own functions to use for the regression anyway.

If this is the case, the function should really be an argument to `execute_with_cdr`. I suggest:

1. Add `regression: Callable[..., float] = linear_fit` to `execute_with_cdr`. (`linear_fit` is currently called `f`).
2. Do the [fitting with scipy](https://github.com/unitaryfund/mitiq/pull/677/files#diff-30fe2c40c2db4727e319f1930148eff86562fab3dde123ff71ebee6d97ad4b14R228-R235) inside `execute_with_cdr`, or keep the `find_optimal_paramaters` helper function but generalize it to any regression (right now it's specific to `wrapper_fit_function`).
3. Remove `wrapper_fit_function` - after (1) and (2) you don't need it anymore.

What do you think? Does this sound viable?

L-P-B: I think this is a very good suggestion.

mitiq/cdr/data_regression.py
L-P-B: for this 2d structure what would you recommend. Is a lists of lists okay?

mitiq/cdr/data_regression.py
L-P-B: After thinking about this a little more I think it might be tricky to generalise the fitting process. It might be better to have the user write the regression functions and how to calculate the optimal parameters. I propose combining the `find_optimal_parameters` function and the `linear_fit` function. I think we can still call this `linear_fit` but just involve the fitting with scipy inside that function.

L-P-B: actually, just realised this won't work because we also need access to the function form to complete the mitigation. So we would need two user defined functions. The functional form to fit and the function to calculate the optimal parameters of this function from the training data.

mitiq/cdr/data_regression.py
```diff
@@ -0,0 +1,294 @@
...
+
+
+def f_intercept(X_data: np.ndarray, params: List) -> float:
+    """ Function used to map noisy to exact expectation values. Fitted using
+    data from the training circuits.
+    Args:
+        X_data: array of noisy observables are various noise levels.
+        *params: parameters of function.
+    """
+    return sum(a * x for a, x in zip(params, X_data)) + params[-1]
```
andreamari, line 246: If the last element of `params` is the intercept, I think this should be:
```suggestion
    return sum(a * x for a, x in zip(params[:-1], X_data)) + params[-1]
```
or maybe:
```suggestion
    return np.append(x_data, 1.0).dot(params)
```

mitiq/cdr/data_regression.py
```diff
@@ -0,0 +1,294 @@
...
+    """ Function used to map noisy to exact expectation values. Fitted using
+    data from the training circuits. No intercept.
+    Args:
+        X_data: array of noisy observables are various noise levels.
+        *params: parameters of function.
+    """
+    return sum(a * x for a, x in zip(params, X_data))
+
+
+def wrapper_fit_func(x, noise_levels, *args, intercept: bool = True):
```
andreamari, line 259: Can we deduce `noise_levels` from `x`? An can we deduce `intercept` comparing the `len(x)` and `len(args)` ?
```
def wrapper_fit_func(x, *args, intercept: bool = True): 
    if len(x) == len(args):   
        return x.dot(args)
    else:
        return x.dot(args[:-1]) + args[-1]
```

 In this case we can also remove `f` and` f_intercept `

mitiq/cdr/data_regression.py
andreamari, line 269: If we normalize the counts, this should probably be renamed as: `dictionary_to_probabilities` or `dictionary_to_state_probs`

mitiq/cdr/data_regression.py
```diff
@@ -0,0 +1,294 @@
...
+        if key not in counts:
+            counts[key] = 0
+        for j in range(len(counts)):
+            if list(basis.values())[i] == list(counts.keys())[j]:
+                counts_order[i] = list(counts.values())[j]
+    state = {
+        list(basis.values())[i]: (counts_order[i] / sum(counts_order))
+        for i in range(2 ** Q)
+    }
+    return state
```
andreamari, line 294: Here we could use `collecions.Counter`, which should do exactly what we need:
```
>>> from collections import Counter
>>> z = ['blue', 'red', 'blue', 'yellow', 'blue', 'red']
>>> Counter(z)
Counter({'blue': 3, 'red': 2, 'yellow': 1})
```

andreamari (commented):
Thanks @L-P-B! I left some comments. Note: I didn't review the test for the moment.

mitiq/cdr/cdr_execution.py
```diff
@@ -79,10 +80,13 @@ def execute_with_CDR(
                                subsituted in the training circuits. The higher
                                this fraction the more costly the simulations,
                                but more successful the mitigation.
+        ansatz: the function to map noisy to exact data. Takes array of noisy
+                and data and parameters returning a float.
+        num_parameters: the number of paramters the ansatz takes.
```
rmlarose, line 85: Let's do `ansatz` -> `fit` and `num_parameters` -> `num_fit_parameters`.

Also, include the needed signature of the `fit` and an example of calling it. It's clear to you and me, but not to someone reading this for the first time.

mitiq/cdr/cdr_execution.py
```diff
@@ -0,0 +1,202 @@
...
+)
+
+
+def execute_with_CDR(
+    circuit: Circuit,
+    executor: Callable[[Circuit], dict],
+    simulator: Callable[[Circuit], Union[dict, np.ndarray]],
+    observables: List[np.ndarray],
+    num_training_circuits: int,
+    fraction_non_clifford: float,
```
rmlarose, line 42: These should both have default values.

rmlarose:
Tests are quite slow here - `pytest mitiq/cdr` takes around 3-4 minutes on my computer. This needs to be reduced before merging.

mitiq/cdr/data_regression.py
andreamari: Actually I think you just want to add missing keys and normalize the values. So this should work:

```python
def dictionary_to_probabilities(
    counts: Dict[str, int], nqubits: int,
) -> Dict[str, float]:
    """Expresses the result of the simulation in the form of a dictionary
    whose values are the modulus squared of the components of the final state.
    The return probabilities are normalized by the number of counts.
    Args:
        counts: Dictionary of counts with binary keys identifying the state.
        nqubits: Number of qubits in the system.
    Returns:
        state: Dictionary whose keys are the base elements of the nqubits qubit
        system and whose values are the modulus of the corresponding squared
        amplitudes.
    """
    total_counts = sum(counts.values())
    # Initialize probabilities to 0.0
    state = {bin(j): 0.0 for j in range(2 ** nqubits)}
    # Update with normalized counts
    for key, value in counts.items():
        state[key] = value / total_counts
    return state
```

andreamari: By the way, can't we just return the array of probabilities instead of a dictionary?
Do we use the keys?

mitiq/cdr/cdr_execution.py
```diff
@@ -0,0 +1,202 @@
...
+                and data and parameters returning a float.
+        num_parameters: the number of paramters the ansatz takes.
+        scale_noise: optional argument containing a user defined function on
+                     how to increase the noise. If this argument is given then
+                     the mitigation method will be vnCDR.
+        scale_factors: factors by which to scale the noise, should not
+                               include 1 as this is just the original circuit.
+        kwargs: Available keyword arguments are:
+
+            TRAINING SET CONSTRUCTION OPTIONS:
```
andreamari, line 93: The indentation of` TRAINING SET CONSTRUCTION OPTIONS:`  is very shifted to the right. Is this intentional?

mitiq/cdr/cdr_execution.py
```diff
@@ -0,0 +1,202 @@
...
+             corresponding to the many raw expectation values (at different
+             noise levels) and the associated mitigated values.
+
+    .. [Czarnik2020] : Piotr Czarnik, Andrew Arramsmith, Patrick Coles,
+        Lukasz Cincio, "Error mitigation with Clifford quantum circuit
+        data," (https://arxiv.org/abs/2005.10189).
+    .. [Lowe2020] : Angus Lowe, Max Hunter Gordon, Piotr Czarnik,
+        Andrew Arramsmith, Patrick Coles, Lukasz Cincio,
+        "Unified approach to data-driven error mitigation,"
+        (https://arxiv.org/abs/2011.01157)."""
```
andreamari, line 121:
```suggestion
        (https://arxiv.org/abs/2011.01157).
        """
```

mitiq/cdr/cdr_execution.py
```diff
@@ -0,0 +1,202 @@
...
+    # Extracting kwargs:
+    method_select = kwargs.get("method_select", "uniform")
+    method_replace = kwargs.get("method_replace", "closest")
+    random_state = kwargs.get("random_state", None)
+    training_set_generation_kwargs_keys = ["sigma_select", "sigma_replace"]
+    kwargs_for_training_set_generation = {
+        key: kwargs.get(key) for key in training_set_generation_kwargs_keys
+    }
```
andreamari, line 129:
```suggestion
    kwargs_for_training_set_generation = {
        "sigma_select": kwargs.get("sigma_select"),
        "sigma_replace": kwargs.get("sigma_replace"),
    }
```
Unless I am missing something, I think this can be simplified.

mitiq/cdr/cdr_execution.py
```diff
@@ -0,0 +1,202 @@
...
+                               this fraction the more costly the simulations,
+                               but more successful the mitigation.
+        ansatz: the function to map noisy to exact data. Takes array of noisy
+                and data and parameters returning a float.
+        num_parameters: the number of paramters the ansatz takes.
+        scale_noise: optional argument containing a user defined function on
+                     how to increase the noise. If this argument is given then
+                     the mitigation method will be vnCDR.
+        scale_factors: factors by which to scale the noise, should not
+                               include 1 as this is just the original circuit.
```
andreamari, line 90:
```suggestion
        scale_factors: factors by which to scale the noise, should not
                               include 1 as this is just the original circuit.
```
It would be much better if scale_factors can include 1 because this is the convention that we use everywhere in Mitiq. If necessary, you can always internally remove all the 1 elements (usually the just first).

mitiq/cdr/cdr_execution.py
```diff
@@ -0,0 +1,202 @@
...
+            if i == 0:
+                training_circuits_simulated_data.append(simulator(circuit))
+    results_dict_training_circuits = [
+        training_circuits_raw_data,
+        training_circuits_simulated_data,
+    ]
+    # circuit of interest execution:
+    results_dict_circuit_of_interest = []
+    for circuit in circuits:
+        circuit_raw_result = executor(circuit[0])
```
andreamari, line 179: I am confused by the execution of `circuit[0]`. Is `circuit` actually a list of `circuits`?

mitiq/cdr/cdr_execution.py
```diff
@@ -0,0 +1,202 @@
...
+    for i, training_circuits in enumerate(training_circuits_list):
+        for j, circuit in enumerate(training_circuits):
+            training_circuits_raw_data[i].append(executor(circuit))
+            # runs the circuits with no increased noise in the simulator:
+            if i == 0:
+                training_circuits_simulated_data.append(simulator(circuit))
+    results_dict_training_circuits = [
+        training_circuits_raw_data,
+        training_circuits_simulated_data,
+    ]
```
andreamari, line 175: In `training_circuits_list` the circuits are ordered with increasing noise levels.
If so, isn't more natural placing the simulated data (zero noise) in the first position instead of last position? I.e.:
```python
results_dict_training_circuits = [
  training_circuits_simulated_data,
  training_circuits_raw_data,
]
```

mitiq/cdr/cdr_execution.py
andreamari: @rmlarose actually I am afraid `fit` gives the idea of the action of fitting instead of "the function to be fitted".
So I am not sure about this renaming.

mitiq/cdr/data_regression.py
L-P-B: hey, yes I think your right, we can just return the probabilities.

mitiq/cdr/cdr_execution.py
L-P-B: whats the conclusion on this?

mitiq/cdr/cdr_execution.py
```diff
@@ -0,0 +1,202 @@
...
+    arrays representing the diagonal of the observables to be measured. Note
+    these observables MUST be diagonal in z-basis measurements corresponding to
+    the circuit of interest.
+
+    Returns list of raw observables (at many noise levels) and mitigated
+    observables.
+
+    This function returns the mitigated observable/s.
+    Args:
+        circuit: circuit of interest compiled in the correct basis.
```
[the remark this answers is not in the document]
L-P-B: why should this be capitalized?

mitiq/cdr/cdr_execution.py
L-P-B: yes thats right.

mitiq/cdr/cdr_execution.py
rmlarose: Since we use `ansatz` for mitiq_curve_fit, I'm ok with this. Usually ansatz refers to circuit structure though, and `fit` or `fit_function` is more accurate imo.

andreamari: `fit_function` is good for me

mitiq/cdr/cdr_execution.py
```diff
@@ -0,0 +1,179 @@
...
+from mitiq.cdr.data_regression import (
+    construct_training_data_floats,
+    construct_circuit_data_floats,
+    linear_fit_function,
+)
+from mitiq.zne.scaling import fold_gates_at_random
+
+
+# TODO: Allow for any QPROGRAM, not just a cirq.Circuit.
+def execute_with_CDR(
```
andreamari, line 35:
```suggestion
def execute_with_cdr(
```

For style convention and for consistency with other Mitiq `execute_with_something` functions.

mitiq/cdr/cdr_execution.py
```diff
@@ -0,0 +1,179 @@
...
+                                        ['uniform', 'gaussian', 'closest']
+            - sigma_select (float): Width of the Gaussian distribution used for
+                                    ``method_select='gaussian'``.
+            - sigma_replace (float): Width of the Gaussian distribution used
+                                     for ``method_replace='gaussian'``.
+            - random_state (int): Seed for sampling.
+
+    Returns: The tuple (raw_expectations, mitigated_expectations)
+             corresponding to the many raw expectation values (at different
+             noise levels) and the associated mitigated values.
```
andreamari, line 113:
```suggestion
    Returns: The tuple (cdr_values, cdr_raw_data) corresponding
             to the mitigated expectation values and to the raw expectation values
             (at different noise levels).
             
  For consistency with `execute_with_pec`.  I also think that it is better to return the important result first and then all additional data. In the future we could expand cdr_raw_data to contain additional information (e.g. the optimal parameters of the fit_function, maybe the circuits, etc...).
```

mitiq/cdr/cdr_execution.py
```diff
@@ -0,0 +1,179 @@
...
+        fitted_params, _ = curve_fit(
+            lambda x, *params: ansatz(x, params),
+            train_data[0].T,
+            train_data[1],
+            p0=np.zeros(num_parameters),
+        )
+        mitigated_observables.append(ansatz(circuit_data, fitted_params))
+        raw_observables.append(circuit_data)
+
+    return raw_observables, mitigated_observables
```
andreamari, line 179: If you agree with the previous comment these should be switched (here and in tests):
```suggestion
    return mitigated_observables, return raw_observables
```

mitiq/cdr/cdr_execution.py
L-P-B: done

rmlarose (approved):

PR #687, opened 2021-05-19
mention Braket in the README

andreamari:
add Braket to the list of supported libraries


nathanshammah (approved):
Thanks @andreamari. Maybe we can also add something in the `guide-getting-started` in docs, that shows up here:
https://mitiq.readthedocs.io/en/stable/guide/guide-getting-started.html?highlight=pyquil#general-workflow-front-ends-backends-and-executors

rmlarose (approved):

Issue #374, opened 2020-09-25
Manually solve for QPR decomposition given noisy gateset


karalekas:
Following the trend of #370, we would like to be able to compute the quasi-probability distribution (QPR) of an "ideal gate" in a more general fashion, without assuming there exists some well-behaved/simple noise model to describe our QPU. In this case, we might be given some description of a "noisy implementable gateset" (e.g. as superoperators or process matrices). Using this description, we can solve the linear problem described in [Temme2017](https://arxiv.org/abs/1612.02058) to find the decomposition with minimum overhead (defined as 1-norm of the QPR coefficients).

karalekas:
Punting on this for now, as it is not blocking for an initial end-to-end demonstration of PEC.

nathanshammah:
I'm happy to look into this. I am coordinating with @andreamari.

PR #694, opened 2021-05-26
Preparing for optimal representations: Helper functions for channel manipulation.

andreamari:
Description
-----------

Renamed `mitiq/pec/utils.py` to` mitiq/pec/channels.py` and add several functions to manipulate channel matrices (Kraus, super-operators, Choi states, etc...).

Once we have these tools, I can make a new PR focused on optimal quasi-probability representations (#374).

As noticed by @rmlarose, at some point we may import some of these functions directly from Cirq (when they will become available). Meanwhile we can use our own functions. They are just matrix manipulations, so it's not a big deal I think.

Checklist
-----------


mitiq/pec/channels.py
```diff
@@ -0,0 +1,212 @@
...
+    CNOT,
+    LineQubit,
+    DensityMatrixSimulator,
+    DepolarizingChannel,
+    AmplitudeDampingChannel,
+    channel,
+)
+
+
+def _max_ent_state_circuit(num_qubits: int) -> Circuit:
```
andreamari, line 40: This function existed before this PR in a different file. No need to review.

mitiq/pec/channels.py
```diff
@@ -0,0 +1,212 @@
...
+    simulator = DensityMatrixSimulator()
+    num_qubits = len(circuit.all_qubits())
+    # Copy and remove all operations
+    full_circ = deepcopy(circuit)[0:0]
+    full_circ += _max_ent_state_circuit(2 * num_qubits)
+    full_circ += circuit
+    return simulator.simulate(full_circ).final_density_matrix  # type: ignore
+
+
+def _operation_to_choi(operation_tree: OP_TREE) -> np.ndarray:
```
andreamari, line 93: This function existed before this PR in a different file. No need to review.

mitiq/pec/channels.py
```diff
@@ -0,0 +1,219 @@
...
+) -> List[np.ndarray]:
+    """Given the a local channel (defined by its kraus operators),
+    this functions returns the Kraus operators corresponding to tensor
+    product of "num_channels" copies of the input channel.
+    """
+    tensored_kraus = []
+    for kraus_tuple in product(local_kraus, repeat=num_channels):
+        kraus_product = np.eye(1)
+        for kraus in kraus_tuple:
+            kraus_product = np.kron(kraus_product, kraus)
```
rmlarose, line 120: Do you think this portion should be factored out as a helper function? (Do you think it could be useful elsewhere?)

```python
def tensor(*args):
    val = args[0]
    for term in args[1:]:
        val = np.kron(val, term)
    return val
```

mitiq/pec/channels.py
```diff
@@ -0,0 +1,219 @@
...
+    |i><j| --> |i,j>.
+    """
+    return density_matrix.flatten()
+
+
+def vector_to_matrix(vector: np.ndarray) -> np.array:
+    """Reshapes a vector to a density matrix, according to the rule:
+    |i,j> --> |i><j|.
+    """
+    dim = int(np.round(np.sqrt(vector.size)))
```
rmlarose, line 169: Might be good to check if `vector.size` is a square number

mitiq/pec/channels.py
```diff
@@ -0,0 +1,219 @@
...
+    return _circuit_to_choi(circuit)
+
+
+def _tensor_product_kraus(
+    local_kraus: List[np.ndarray], num_channels: int
+) -> List[np.ndarray]:
+    """Given the a local channel (defined by its kraus operators),
+    this functions returns the Kraus operators corresponding to tensor
+    product of "num_channels" copies of the input channel.
+    """
```
rmlarose, line 115: I think the docstring / arguments could be clearer here. Maybe `nqubits` or `ncopies` instead of `num_channels`? Also, there's no requirement that `local_kraus` has to be local really, and local isn't defined. 

I guess what you're doing is:

Given Kraus operators K_1, ..., K_m, returns {K_{i_1} \otimes \cdots \otimes K_{i_m}} for all m^`ncopies` combinations (i_1, ..., i_m) of indices (1, ..., m).

mitiq/pec/channels.py
```diff
@@ -0,0 +1,219 @@
...
+
+    S = \\sum_j K_j \\otimes K_j^*,
+
+    where K_j are the superoperators.
+    The mapping is based on the following isomorphism:
+
+    A|i><j|B  <=>  (A \\otimes B^T) |i>|j>.
+    """
+    terms_to_sum = [np.kron(k, k.conj()) for k in kraus_ops]
+    return sum(terms_to_sum)
```
rmlarose, line 188:
```suggestion
    return sum([np.kron(k, k.conj()) for k in kraus_ops])
```

rmlarose: Do you need to take the transpose? `k.conj()` -> `k.conj().T`?

mitiq/pec/tests/test_channels.py
```diff
@@ -0,0 +1,188 @@
...
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program.  If not, see <https://www.gnu.org/licenses/>.
+
+"""Tests related to mitiq.pec.utils functions."""
```
rmlarose, line 16: Update.

rmlarose (changes requested):
See inline comments.

How long do these tests take? Might be good to trim down loops over nqubits / dim if possible.

mitiq/pec/channels.py
andreamari: Good idea! I called it `tensor_product` and I removed the less general function `_tensor_product_kraus`.

mitiq/pec/channels.py
andreamari: This happens in two functions and I expect to be a recurrent situation in the future, so I made a new function `_safe_sqrt` for handling this situation.

mitiq/pec/channels.py
andreamari: > Given Kraus operators K_1, ..., K_m, returns {K_{i_1} \otimes \cdots \otimes K_{i_m}} for all m^`ncopies` combinations (i_1, ..., i_m) of indices (1, ..., m).
Yes, this is correct. The resulting kraus operators correspond to the tensor product of `ncopies` of the input channel.

Anyway, I removed this function since (now that we have `tensor_product()`), it can be directly written with a single line:
`[tensor_product(*tup) for tup in product(local_kraus, repeat=num_qubits)]`

andreamari:
> How long do these tests take? Might be good to trim down loops over nqubits / dim if possible.

This is not a problem, tests are almost instantaneous:

```
0.14s call     mitiq/pec/tests/test_channels.py::test_kraus_to_super
0.03s call     mitiq/pec/tests/test_channels.py::test_kraus_functions[global_depolarizing_kraus]
0.01s call     mitiq/pec/tests/test_channels.py::test_kraus_functions[local_depolarizing_kraus]
0.01s call     mitiq/pec/tests/test_channels.py::test_kraus_to_choi
0.01s call     mitiq/pec/tests/test_channels.py::test_super_to_choi
0.01s call     mitiq/pec/tests/test_channels.py::test_circuit_to_choi
```

rmlarose:
> > How long do these tests take? Might be good to trim down loops over nqubits / dim if possible.
> 
> This is not a problem, tests are almost instantaneous:
> 
> ```
> 0.14s call     mitiq/pec/tests/test_channels.py::test_kraus_to_super
> 0.03s call     mitiq/pec/tests/test_channels.py::test_kraus_functions[global_depolarizing_kraus]
> 0.01s call     mitiq/pec/tests/test_channels.py::test_kraus_functions[local_depolarizing_kraus]
> 0.01s call     mitiq/pec/tests/test_channels.py::test_kraus_to_choi
> 0.01s call     mitiq/pec/tests/test_channels.py::test_super_to_choi
> 0.01s call     mitiq/pec/tests/test_channels.py::test_circuit_to_choi
> ```

Sweet!

mitiq/pec/channels.py
```diff
@@ -106,20 +106,16 @@ def _operation_to_choi(operation_tree: OP_TREE) -> np.ndarray:
     return _circuit_to_choi(circuit)
 
 
-def _tensor_product_kraus(
-    local_kraus: List[np.ndarray], num_channels: int
-) -> List[np.ndarray]:
-    """Given the a local channel (defined by its kraus operators),
-    this functions returns the Kraus operators corresponding to tensor
-    product of "num_channels" copies of the input channel.
+def tensor_product(*args: np.ndarray) -> np.ndarray:
```
rmlarose, line 109: Interested to see if we want to use this elsewhere. If so, probably move to utils. For now here makes sense.

mitiq/pec/channels.py
```diff
@@ -139,53 +135,72 @@ def local_depolarizing_kraus(
...
-    |i,j> --> |i><j|.
+def _safe_sqrt(perfect_square: int) -> int:
+    """Takes the square root of the input integer and
+    raises an error if the input is not a perfect square."""
+    square_root = int(np.round(np.sqrt(perfect_square)))
+    if square_root ** 2 != perfect_square:
+        raise ValueError(
+            "The expected dimension of the input matrix (or array) must be a"
+            " square number but dimension is {perfect_square}."
+        )
```
rmlarose, line 174: nit: This assumes `perfect_square` is the dimension of a matrix / array but the function is for any int. Error message could originate from calling function.

mitiq/pec/channels.py
```diff
@@ -106,20 +106,16 @@ def _operation_to_choi(operation_tree: OP_TREE) -> np.ndarray:
...
 
 
-def _tensor_product_kraus(
-    local_kraus: List[np.ndarray], num_channels: int
-) -> List[np.ndarray]:
-    """Given the a local channel (defined by its kraus operators),
-    this functions returns the Kraus operators corresponding to tensor
-    product of "num_channels" copies of the input channel.
+def tensor_product(*args: np.ndarray) -> np.ndarray:
+    """Returns the Kronecker product of the input arguments (NumPy objects).
```
rmlarose, line 110:
```suggestion
    """Returns the Kronecker product of the input arguments (NumPy arrays).
```

rmlarose: I guess really it doesn't have to be arrays or numpy types. e.g., `np.kron` for scalars is multiplication. `np.kron(a, b)` says `a` and `b` should be "array like".

mitiq/pec/tests/test_channels.py
```diff
@@ -0,0 +1,202 @@
...
+
+def test_kraus_to_choi():
+    for dim in (2, 4, 8, 16):
+        rand_kraus_ops = [np.random.rand(dim, dim) for _ in range(7)]
+        super_op = kraus_to_super(rand_kraus_ops)
+        expected_choi = super_to_choi(super_op)
+        assert np.allclose(kraus_to_choi(rand_kraus_ops), expected_choi)
+
+
+def test_tensor_product():
```
rmlarose, line 199: Would be good to have tests for 1 and 0 elements. 1 element just returns the element, and 0 elements throws an IndexError, which I think is appropriate behavior.

andreamari: I added a TypeError if no arguments are given. This is the type of error that e.g. `np.kron()` or `sum()` would give.

mitiq/pec/channels.py
andreamari: changed with `product of the input array-like arguments.`

mitiq/pec/channels.py
andreamari: I added a custom error string which should have fixed this problem.

PR #700, opened 2021-05-29
Warning for short programs [unitaryHACK]

### LINKED ISSUE #275: Warn users when their programs are too short
As discussed in the engineering meeting, some programs are too short for ZNE via unitary folding to work properly (the example was a program with a single gate). While this example program is silly, and as @andreamari noted it is implicit in the protocol that there need to be enough gates for it to work, I still think it is worth warning the user. Someone also mentioned using a different default factory (linear instead of Richardson) could improve low-depth behavior, but that might be a discussion for another issue.

Yash-10:
Description
-----------
- Fixes #275 
- Added a warning which checks for the inequality of scale factors of the `factory` and the ratio of depths of the final and the initial circuits.
- *This might probably need some polishing* to ensure it works correctly. If somebody could provide insights on the warning handling condition, it would be really great! Thanks!

Checklist
-----------

*For adding warnings, other checklist options might not be necessary?*


andreamari:
@Yash-10, thanks for your PR!

I think there can be a problem with this approach. Even if `len(output_circuit) / len(input_circuit)` is often a good measure of
the scale factor this is not always the case for two reasons: (i) `len()` measures the number of moments but sometimes the number of gate is the relevant quantity. (ii) Most importantly, in some cases (e.g. in folding by fidelity), gates should be counted with different weights. For example, often people are interested in folding only CNOT gates and so, in these cases, `output_cnots / input_cnots` would be the good quantity to consider.

Given this variety of situations, probably the most pragmatic solution is to just raise a warning of the input circuit is too short, independently from the values of the scale factors.  E.g. if  `len(qp) < 10` we may raise `"The input circuit is very short. This may reduce the accuracy of noise scaling." `

I know that this is still not optimal but, at least, I think it is a good compromise which is easy to maintain and to improve in the future.

**Important:** the function `too_few_gates_warning` should contain the word `test` otherwise is not tested by pytest. It should  be renamed  `test_short_circuit_warning` or something similar.

Yash-10:
@andreamari Thank you so much for your guidance and insights. I have tried to modify the code as per your suggestions. Could you have a look?

Thanks!

andreamari:
Hi @Yash-10, thanks! I think we are almost there. There are only a couple of style issues:
```
./mitiq/zne/inference.py:524:80: E501 line too long (97 > 79 characters)
./mitiq/zne/inference.py:731:80: E501 line too long (97 > 79 characters)
./mitiq/zne/tests/test_inference.py:38:1: F401 'mitiq.zne.scaling.folding.fold_gates_at_random' imported but unused
```
Could you fix them? A way to check the style is to run `make check-style` (or just `flake8`) in your local mitiq folder.

Yash-10:
@andreamari Thanks again, I just fixed the style errors. For the coverage, it seems that Adaptive Factories are not tested. I would need to include them in the tests too.

andreamari:
> Hi @Yash-10, thanks! I think we are almost there. There are only a couple of style issues:
> 
> ```
> ./mitiq/zne/inference.py:524:80: E501 line too long (97 > 79 characters)
> ./mitiq/zne/inference.py:731:80: E501 line too long (97 > 79 characters)
> ./mitiq/zne/tests/test_inference.py:38:1: F401 'mitiq.zne.scaling.folding.fold_gates_at_random' imported but unused
> ```
> 
> Could you fix them? A way to check the style is to run `make check-style` (or just `flake8`) in your local mitiq folder.

Sorry, I missed your last commits in which you fixed the style. LGTM, but let's wait for the runnig CI tests

andreamari:
> @andreamari Thanks again, I just fixed the style errors. For the coverage, it seems that Adaptive Factories are not tested. I would need to include them in the tests too.

Exactly, testing also `AdaExpFactory` would be great!

Yash-10:
So sorry for so many commits! Hopefully by the addition of `AdaExpFactory`, the coverage would be fine.

andreamari (approved):
Great! Thanks for testing also adaptive factory! 
This is the last day of unitaryHACK, so congratulations @Yash-10 for your last-minute PR!
:champagne:

Yash-10:
Thanks @andreamari !

crazy4pi314:
@all-contributors please add @Yash-10 for tests and code

PR #708, opened 2021-06-05
Mitiq tutorial example - VQE [unitaryHACK]

Misty-W:
Description
-----------
Simple VQE tutorial example in pyQuil with depolarizing noise, mitigated with ZNE.

Checklist
-----------

Tips
----


crazy4pi314:
Hi @Misty-W ! It looks like some of the style tests are failing, if you run `make check-style` you can have the tools fix those and then commit :)

Misty-W:
hi @crazy4pi314, I believe the style issues are fixed now.

crazy4pi314:
Running the tests again now! They don't run automatically for new contributors as per a GitHub wide security policy.

andreamari:
Hi @Misty-W, thanks for sharing your draft!

This PR contains python file but, in principle, examples should be added in a `.myst` / `.md` format such that they are rendered
in the documentation (https://mitiq.readthedocs.io/en/latest/contributing_docs.html#adding-code-examples). 
If you are interested, you can look at other examples files and at the contribution guide-lines at this link:
https://mitiq.readthedocs.io/en/latest/contributing_docs.html#adding-code-examples.

Other general comments:

- Examples in `.myst` / `.md` should contain many markdown comments that allow even a non-expert reader to easily understand all the steps of the example.

- The code contained in each example is automatically executed in CI (to build the docs). Therefore, ideally, it should be as light as possible.

nathanshammah (changes requested):
Thanks @Misty-W. As a first review (not complete):
- Question: is this still a draft PR? If not we can update the title. 
 The example should be included in the table of contents at `docs/source/examples.myst`
- I understood that examples were to be `.myst` files or `.ipynb`. 
- In any case, if I try on a local fork of this branch to include the `*.py` file in the `examples.myst` table of contents, I get the error
```
Warning, treated as error:
/Users/nathanshammah/github/mitiq/docs/source/examples/examples.myst:3:toctree contains reference to nonexisting document 'examples/var_classifier_mitiq/variational_classifier_mitiq'
make[1]: *** [html] Error 2
make: *** [docs] Error 2
```

About the style and content of the notebook:
- You can remove `_mitiq` from the folder and title. 
- Just as a suggestion: I propose to create a `data/` folder possibly also for other examples, putting there the `.txt` data file, and move the example further up in the tree, in the same folder as the other examples. 
- Please include a `mitiq.about()` command as in other examples.

Misty-W:
@andreamari thanks for the feedback! I need to take some time to familiarize myself with the .myst / .md format, but hopefully I'll have another version ready soon.

Also, can you clarify a bit what is meant by "the code should be as light as possible"?

Misty-W:
@nathanshammah thanks for the review! For now the PR is still a draft, and I realize I need to become familiar with the .myst / .md format. Hope to have another version out to you soon!

andreamari:
> Also, can you clarify a bit what is meant by "the code should be as light as possible"?
I meant that, ideally, the time required to execute the code should not be too long. We don't have strict rule but, just to give an order of magnitude, I would put ~5 minutes as an upper bound.

WrathfulSpatula:
@Misty-W, for your remaining `codecov/project` test failure, I don't think an example requires unit test coverage, unless the team disagrees. Hence, I'd suggest that all you need to do to fix the test is [tell codecov to ignore your new files](https://docs.codecov.io/docs/ignoring-paths).

However, I'm new to the Unitary Fund team, and I'm not sure where `codecov` is controlled, here, but it should be in the hidden/configuration files of the repository you have checked out, itself. You commit the configuration change in your local repository and push, and the pull request tests will automatically reflect the configuration change. I'd guess it's `.coveragerc`, but I would expect `codecov.yml` from the documentation I linked above, and I'm just not familiar with the tool, yet. Any help from the core team is appreciated, on this.

By the way, **merge `master` first** into this open pull request, and potentially save yourself a major headache, if the coverage drop is just from not being up-to-date with the trunk. You have no code conflicts, so `git` can just "fast-forward."

(@crazy4pi314 also tells me `pragma: no cover` works for making `codecov` ignore specific single lines, but I think your new files should be categorically excluded, so the "right" way likely isn't the pragma.)

Misty-W:
After working with the variational classifier example a bit more, I realized a simpler problem would be much more manageable both in terms of runtime and tuning the noise mitigation. To that end, I switched to a simple VQE example. 

I still need to experiment with the noise scaling and extrapolation to get the VQE example to actually converge on the noisy simulated device, but at least it runs more efficiently than the classifier example. I'll be digging further into the Mitiq documentation to figure that out, but if you have any suggestions please let me know. 

I also put it in the .md format as requested :)

crazy4pi314:
Hi @Misty-W! Is there anything you are blocked on, if so let us know so we can help!

Misty-W:
hi @crazy4pi314, thanks for checking in! 

I think I fixed the doc failing issue in my latest commit with the addition of the Jupytext header content, but it looks like the tests haven't run yet. Is there something I can run to check it myself? The previewer in VS Code didn't catch the missing header in the previous version.  

More concerning is that the actual VQE example in Python still isn't converging with ZNE. Even the expectation value calculated with the initial parameters is not improved with ZNE, and sometimes it's slightly worse. I've tried the linear, Richardson, and polynomial extrapolation methods and adjusted the scale factors. I've pasted some of them below. Any suggestions would be appreciated!
 
    linear_fac = zne.inference.LinearFactory(scale_factors = [1.0, 3.0]) 
    richardson_fac = zne.inference.RichardsonFactory(scale_factors = [1.0, 2.0, 4.0]) 
    poly_fac = zne.inference.PolyFactory(scale_factors = [1.0, 2.0, 3.0, 4.0], order=2)

crazy4pi314:
> More concerning is that the actual VQE example in Python still isn't converging with ZNE.

@rmlarose @andreamari @nathanshammah does anyone see any obvious reasons why the samples here for ZNE wouldn't converge?

docs/source/examples/vqe-demo.md
```diff
@@ -0,0 +1,236 @@
...
+    return expectation.real
+```
+
++++
+
+Calculates the parity of elements at indexes in marked_qubits
+   
+```{code-cell} 
+def parity_even_p(state, marked_qubits):
+"""
```
rmlarose, line 89:
```suggestion
    """
```

rmlarose (commented):
> > More concerning is that the actual VQE example in Python still isn't converging with ZNE.
> 
> @rmlarose @andreamari @nathanshammah does anyone see any obvious reasons why the samples here for ZNE wouldn't converge?

By "converge" I'm assuming you mean converge to the minimum energy. i.e., "doesn't converge" means that the optimizer fails.

Can you make a heatmap of the noisy & mitigated energy landscapes? My guess is that the mitigated landscape is not smooth and causes the optimizer to fail. 

You can try to increase the noise level or change the noise model. The ansatz has two gates so every folding method is the same. I don't know what noise model `"2q-noisy-qvm"` uses but `zne.inference.ExpFactory([1, 3, 5, 7, 9])` should work well with depolarizing noise.

nathanshammah:
The CI is failing, although the error message is not very helpful to understand what's wrong:
```
Warning, treated as error:
Execution Failed: /home/runner/work/mitiq/mitiq/docs/source/examples/vqe-demo.md
make[1]: *** [Makefile:20: html] Error 2
make[1]: Leaving directory '/home/runner/work/mitiq/mitiq/docs'
make: *** [Makefile:34: docs] Error 2
Error: Process completed with exit code 2.
```
@Misty-W was the documentation building correctly locally?
This can be tested from the Terminal with 
```
make docs-clean
make doctest
```

Misty-W:
Hi @rmlarose, thanks for the suggestions! Indeed, the optimizer is failing on both the noisy and mitigated runs. I tried ```zne.inference.ExpFactory([1, 3, 5, 7, 9])``` and encountered the following error:

> ExtrapolationError: The extrapolation fit failed to converge. The problem may be solved by switching to a more stable extrapolation model such as `LinearFactory`.

I generated heatmaps of the noisy and mitigated (linear factory) energy landscapes. They look almost identical. 

Noisy_Energy_Landscape
Mitigated_Energy_Landscape_LinearFactory

rmlarose:
Thanks a lot @Misty-W! I suspect the heatmaps are so similar because the ansatz only has two (single-qubit) gates. (Noise is applied after each gate, so if there are only two gates there is not much noise in the computation.) You could test this hypothesis by computing the same landscape without noise (use `"2q-qvm"` instead of `"2q-noisy-qvm"`) and comparing. 

Also, the `ExtrapolationError` from the exponential fit will probably go away with higher noise and / or more gates.

Misty-W:
hi @nathanshammah 

- `make docs-clean`  I tried the Windows version described [here](https://stackoverflow.com/questions/23723364/windows-7-make-is-not-recognized-as-an-internal-or-external-command-operabl), which is ```nmake docsclean```, but it results in the error 

> fatal error U1073: don't know how to make 'docs-clean'
     
   I’m guessing I don’t have the right makefile, but I’m not sure where to look for it.

- `nmake doctest`  results in a similar error:

> fatal error U1073: don't know how to make 'doctest'

I was able to run `python -m doctest -v vqe-demo.md`, which according to the [Python 3 docs](https://docs.python.org/3/library/doctest.html#:~:text=There%20is%20also%20a%20command%20line%20shortcut%20for,the%20command%20line%3A%20python%20-m%20doctest%20-v%20example.txt) will infer that it must be run with `testfile()` , not `testmod()` since it is not a `.py` file. However, the result was my file "had no tests"

> 1 items had no tests:
>     vqe-demo.md
> 0 tests in 1 items.
> 0 passed and 0 failed.
> Test passed.

Misty-W:
hi @rmlarose,

As expected, the energy landscape looks similar with and without noise for two single-qubit gates. 
Noisy_Energy_Landscape_2gatesNo_Noise_E_Landscape_2gates

After that I tried the exponential fit on four gates, but it still failed.  

My main concern about increasing the number of gates has been (and still is) runtime. When running the optimization on four gates with `LinearFactory`, 100 iterations took 24 minutes, and the optimizer did not converge to the minimum energy within the 100 iterations.

I also ran parameter sweeps to generate heatmaps with four gates, and it took 2.5 hours to run the noise-mitigated sweep with `LinearFactory`, versus 15 minutes without the noise mitigation. I didn't time the noise-mitigated sweep on with the two gate ansatz, but it took somewhere around 1 hour. 

No_Noise_E_Landscape_4gates Noisy_Energy_Landscape_4gates Mitigated_E_Landscape_4gates_LinearFac

 I haven't had a chance to try increasing the noise level yet; I'll look into that next.

rmlarose:
> My main concern about increasing the number of gates has been (and still is) runtime

About the runtime: It shouldn't increase in any noticeable way if you have 2 parameterized gates + 2 non-parameterized gates on two qubits. 

Were you doing 4 parameterized gates? If so the time to do a grid search will go up a lot. But you can add in as many non-parameterized gates as you want to make the results noisier while still doing a grid search over 2 parameters (still on two qubits).

Misty-W:
@rmlarose yes, I was using 4 parameterized gates. I switched to 2 parameterized gates + 2 non-parameterized gates, still on 2 qubits

`program = Program(RX(params[0], 0), H(0), RX(params[1], 0), CNOT(0,1))`

and re-did the grid search. I ran it twice and both times I encountered a timeout error: 

> TimeoutError: Timeout on client tcp://127.0.0.1:5555, method name quil_to_native_quil, class info: <rpcq._client.Client object at 0x000001ABF52406A0>


I also tried out Quil's parametric compilation (as a test run I used the example [here](https://pyquil-docs.rigetti.com/en/stable/migration3-declare.html?highlight=qc.run)). I ran into trouble trying to use the Quil `memory_map` with `execute_with_zne` 

> RPCError: Unhandled error in host program:
> Reference to unknown memory region theta.

    def ansatz():
        program = Program()
        theta = program.declare('theta', memory_type='REAL')
        ro = program.declare('ro', memory_type='BIT', memory_size=1)
        program += RY(theta, 0) 
        program += MEASURE(0, ro[0])
        return program

    def expectation_fn(theta, program):
        program.wrap_in_numshots_loop(1000) 
        executable = qc.compile(program)
        bitstrings = qc.run(executable, memory_map= {'theta': [theta]})
        return np.mean(bitstrings[:,0])

    thetas = np.linspace(0, 2*np.pi, 21)
    program = ansatz()
    results_zne = []
    for theta in thetas:
        results_zne.append(execute_with_zne(program, partial(expectation_fn, theta)))

crazy4pi314:
Ok so I haven't solved the converging issue, but I have now updated the VS Code docker configuration so that if you are on Windows and have docker installed you can with a couple of clicks have a linux box gets spun up with the appropriate dockers for the QVM/quilc and port forwarded.

Misty-W:
thank you @crazy4pi314 ! I tried building locally and I get the following error:

> Configuration error:
> There is a programmable error in your configuration file:
> 
> Traceback (most recent call last):
>   File "C:\Users\misty\AppData\Local\Packages\PythonSoftwareFoundation.Python.3.9_qbz5n2kfra8p0\LocalCache\local-packages\Python39\site-packages\sphinx\config.py", line 327, in eval_config_file
>     execfile_(filename, namespace)
>   File "C:\Users\misty\AppData\Local\Packages\PythonSoftwareFoundation.Python.3.9_qbz5n2kfra8p0\LocalCache\local-packages\Python39\site-packages\sphinx\util\pycompat.py", line 88, in execfile_
>     exec(code, _globals)
>   File "C:\Users\misty\Documents\GitHub\mitiq\docs\source\conf.py", line 16, in <module>
>     import pybtex.style.formatting
> ModuleNotFoundError: No module named 'pybtex'

Let me see what the container does...

crazy4pi314:
> >     import pybtex.style.formatting
> > ModuleNotFoundError: No module named 'pybtex'
> 

It seems like here the local dev requirements are not installed/installed in the right place?

Anyway, I have tested the container now a few times, and once it is up (your editor should looks just like it would if it was opening the files locally) you can from the terminal in VS Code run things like make docs etc

Misty-W:
@crazy4pi314 I tried running `make docs` in the Mitiq dev container. It builds some of the docs successfully, only to error out on mine. I got the warning messages:

> WARNING: Execution Failed: /workspace/docs/source/examples/vqe-demo.myst
> WARNING: Couldn't find cache key for notebook file source/examples/vqe-demo.myst. Outputs will not be inserted.
>   Last execution failed with traceback saved in /workspace/docs/build/html/reports/vqe-demo.log

Looking into the log, I see

> QVMNotRunning: No QVM server running at http://127.0.0.1:5000

purva-thakre:
@Misty-W Have you checked if your QVM server is running ? [See `NOTE:`](https://mitiq.readthedocs.io/en/stable/contributing.html#running-local-tests)

Misty-W:
hi @purva-thakre, yes, I typed in the terminal
    
and I can see their images running in Docker

Misty-W:
@crazy4pi314 @rmlarose I committed the .py files with pyquil noise models I mentioned on Friday in case you want to take a look. Looks like the custom (noisy) gate is not applied whether it is inside or outside the executor, when compared with the result from the ideal gate.

andreamari:
Hi @Misty-W, I don't have a complete solution to the problems of this PR but I played with your notebook and this is what I found:

1. The fact that mitigated and unmitigated results are very similar is due to the particular noise model of the QVM. For short circuits errors are dominated by measurement errors which are not mitigated by unitary folding. For this reason you get a negligible mitigation.
2. Mitigation by unitary folding + ZNE starts to give interesting results for longer circuits. For long circuits one should also be careful and to avoid compiling a long circuit into a short circuit (the compiler seems to be quite smart in reducing the length of the circuits). I managed to get something in a test notebook that you can find here if you are interested (https://github.com/unitaryfund/mitiq/blob/test-pyquil-example/docs/source/examples/vqe-pyquil.myst).
3. Unfortunately, in the regime of long circuits discussed in the previous point, the execution time is very slow. Especially if you are not just interested in plotting the landscape but also in solving the optimization problem associated to the VQE. This is fine for a local test  and for potential applications but is quite bad if we are going to merge it in the docs since it would slow down CI a lot. Maybe there is a way to speedup the mentioned test notebook https://github.com/unitaryfund/mitiq/blob/test-pyquil-example/docs/source/examples/vqe-pyquil.myst but I am not sure how.
4. Maybe, more realistically, the simplest solution is to manually build a simple noise model (depolarizing or  dephasing or  amplitude damping, etc.) for a short circuit. This is the same approach that we use for other frontends. I think that the way of building a simple noise model in PyQuil is actually not so simple but in principle it is doable (https://pyquil-docs.rigetti.com/en/stable/noise.html). I can help if necessary.

If we decide for option 3, we should find a way to keep the execution time between 2 to 5 minutes. More or less.

If we decide for option 4, the example would look a bit "artificial" but at least it should be much faster to execute.

Misty-W:
Hi @andreamari, thanks for the suggestions! I tried option 4, and it seems a little better but still not as consistent as I'd like. I put it in a new notebook and added to the PR, if you want to take a look. Next I'll try adding a few more gates with the custom noise model (but not too many because of the runtime).

andreamari:
@Misty-W I tried to execute this notebook: https://github.com/unitaryfund/mitiq/blob/d47771126b2c52ab210207f247cea87d620e0f35/docs/source/examples/vqe-demo_custom-gate.myst
with the following settings:
`prob = 0.8`
`scale_factors = [1.0, 3.0]`
`init_angle = 3.0`  

Everything seems to work as expected and it's quite fast.

Setting `init_angle = 3.0` is a bit like cheating since the minimum is `pi`. However, given the simple shape of the landscape, it is clear that any optimizer with sufficiently many steps will reach the minimum. Setting `init_angle = 3.0` , we simply avoid a long building time of the docs.

P.s. The default method BFGS  (`method=None`) is usually better than `method="Nelder-Mead"`. But if you prefer Nelder-Mead for some reason, it's ok.

P.s. Some cells are not formated according [PEP 8](https://www.python.org/dev/peps/pep-0008/). A simple way of improving the style is to `pip install nb_black` and temporarily add this cell at the beginning of the noteook:
```
# Improve the style of code cells
%load_ext nb_black
```
After executing all the notebook, all code will be formatted automatically. If you do this, remember to remove that cell before pushing to GitHub since  `np_black` is not present in the Mitiq dev environment.

Misty-W:
@andreamari I tried the settings you suggested and they look good to me. I updated the PR with them and also with the reformatting via `%load_ext nb_black`.

andreamari:
> @andreamari I tried the settings you suggested and they look good to me. I updated the PR with them and also with the reformatting via `%load_ext nb_black`.

Great! The content LGTM. Maybe we can improve the presentation a little bit. A few ideas:

- Is it possible to import the function `expectation_from_sampling`  from PyQuil? If not it would be nice to add a comment like `This function is a modified version of: link-to-the-pyquil-function.`
- Is it possible to import the function `parity_even_p`  from PyQuil?
- We can move `noise_sim` next to the definition of the noiseless `program`.
- What about renaming `noise_sim` to `add_noise_to_circuit`? Adding a docstring would also help. This is a non-trivial step.
- Docstrings have a different style compared to what is used in Mitiq. See e.g. https://github.com/unitaryfund/mitiq/blob/master/mitiq/pec/pec.py .  Note also that listing all the `Args` and `Returns` is optional for a notebook. Sometimes a single sentence is enough to explain what a functions does, unless it is not clear what is the meaning of the arguments.
- Annotations are also optional for a notebook. But if used, I would use them everywhere. 
- What about using these more explicit names: `qc` ---> `backend`, `ro` ---> `readout_qubit` ?

Misty-W:
@andreamari, thanks for the feedback! Example is looking cleaner now... 

> Is it possible to import the function expectation_from_sampling from PyQuil? If not, it would be nice to add a comment like This function is a modified version of: link-to-the-pyquil-function.

Unfortunately `expectation_from_sampling` had to be modified because `define_noisy_gate` has to come after applying the `MEASURE` gate but before `qc.run` is called, and both are inside `expectation_from_sampling`. I wish I could think of a better way- the example would look nicer if I could import `expectation_from_sampling` directly. Anyway, I cleaned up the docstring and added a description in the markdown text with a link to the source. I did the same for `executor`, since much of it comes from `expectation` in grove.pyvqe.vqe

> Is it possible to import the function parity_even_p from PyQuil?

Yes, done.

> We can move noise_sim next to the definition of the noiseless program.

Done

> What about renaming noise_sim to add_noise_to_circuit? Adding a docstring would also help since this is a non-trivial part of the code.

Agree, done

> Docstrings have a different style compared to what is used in Mitiq. See e.g. https://github.com/unitaryfund/mitiq/blob/master/mitiq/pec/pec.py . Note also that listing all the Args and Returns is optional for a notebook. Sometimes a single sentence is enough to explain what a functions does, unless it is not clear what is the meaning of the arguments.

I made some updates based on what I saw in the example link. Further improvement suggestions welcome!

> Annotations are also optional for a notebook. But if used, I would use them everywhere.

After looking over the annotations in my file, I decided they weren't adding much useful information and removed them.

> What about using these more explicit names: qc ---> backend, ro ---> readout_qubit ?

Agree, done

Misty-W:
@crazy4pi314, I see a conflict with the examples list- how shall we resolve?

Misty-W:
@nathanshammah and @rmlarose, your feedback is welcome as well!

crazy4pi314:
> @crazy4pi314, I see a conflict with the examples list- how shall we resolve?

Got it! new build is kicked off, though nothing should really be a problem from that change.

Misty-W:
@andreamari I reverted the change where I imported `parity_even_p` from Grove (not pyQuil) because I think that's what caused the docs test to fail. I'm guessing the failure occurred because the Mitiq dev environment doesn't have Grove. Earlier the checks appeared to pass with it in, but later I saw two checks were run in those instances.

andreamari:
> @andreamari I reverted the change where I imported `parity_even_p` from Grove (not pyQuil) because I think that's what caused the docs test to fail. I'm guessing the failure occurred because the Mitiq dev environment doesn't have Grove. Earlier the checks appeared to pass with it in, but later I saw two checks were run in those instances.

@Misty-W You are right I didn't realize that the function it was in Grove and not in PyQuil. So I am ok with the current solution.

- A final suggestion from my side: can we add 2 or 3 sections in the notebook? I.e. markdown cells with `## Section title.` They usually help the overall readability (see e.g. other Mitiq examples).

I like the example and tests are passing so I approve this PR! I won't merge it to allow other reviewers to have a look. 

Thanks @Misty-W for the great work!

andreamari (approved):

docs/source/examples/examples.myst
```diff
@@ -7,5 +7,6 @@ maxcut-demo.myst
 Understanding probabilistic error cancellation (PEC) <pec-tutorial.myst>
 Zero noise extrapolation on IBMQ backends<ibmq-backends.myst>
 Zero noise extrapolation on PyQuil parametric programs <pyquil_demo.myst>
+vqe-demo_custom-gate.myst
```
nathanshammah, line 10: If you add a title next to the file name it should show up nicely like the other ones, I think. 
"Variational Quantum Eigensolver improved with Mitiq". 

Also, we'd better find a common choice for all files to either have hyphens or underscore. Can you pick one for this example? For example 
```
Variational Quantum Eigensolver improved with Zero Noise Extrapolation <vqe-demo-custom-gate>
```
I'd drop the 'with Mitiq', as this is already the Mitiq library documentation so it does not add much information.

docs/source/examples/vqe-demo_custom-gate.myst
```diff
@@ -0,0 +1,282 @@
...
+    format_name: myst
+    format_version: 0.13
+    jupytext_version: 1.10.3
+kernelspec:
+  display_name: Python 3 (ipykernel)
+  language: python
+  name: python3
+---
+
+# Variational Quantum Eigensolver improved with Mitiq
```
nathanshammah, line 14: I'd drop the "with Mitiq" part.
```suggestion
# Variational Quantum Eigensolver improved with Zero Noise Extrapolation
```
the modification is a suggestion.

docs/source/examples/vqe-demo_custom-gate.myst
```diff
@@ -0,0 +1,282 @@
...
+from typing import List, Union
+from collections import Counter
+from matplotlib import pyplot as plt
+from scipy import optimize
+import mitiq
+from mitiq import zne
+from mitiq.zne.scaling.folding import fold_gates_from_right
+```
+
+Use the get_qc command to initialize the simulated backend
```
nathanshammah, line 39:
```suggestion
Use the `get_qc` command to initialize the simulated backend 
```
it would actually be even nicer to use Intersphinx and see if it works on the Examples' part of the documentation, to automatically link to the relevant part of the PyQuil docs. Here is an example of its use 
https://mitiq.readthedocs.io/en/stable/guide/guide-executors.html#qiskit-density-matrix-simulation-with-noise-and-sampling for the `execute_with_shots_and_noise()` function. [Link to github raw file](https://github.com/unitaryfund/mitiq/blob/51a10d1defffdd050e881fa5070c2f5484a4d6b9/docs/source/guide/guide-executors.rst#qiskit-density-matrix-simulation-with-noise-and-sampling).

docs/source/examples/vqe-demo_custom-gate.myst
```diff
@@ -0,0 +1,282 @@
...
+kernelspec:
+  display_name: Python 3 (ipykernel)
+  language: python
+  name: python3
+---
+
+# Variational Quantum Eigensolver improved with Mitiq
+In this example we investigate how Mitiq can help reduce errors
+and improve convergence of a simple VQE problem executed
+on a simulated noisy backend.
```
nathanshammah, line 17: I would add a bit more details here in the initial introduction. For example you can cite ZNE (with a link to the part in the docs where the specific technique is examplained). 

You could also spend a few words, on why VQE was chosen and what it is about, possibly with a reference. 
Maybe we can try to format the reference in the same style of the user guide, so that hopefully it shows up at https://mitiq.readthedocs.io/en/latest/zz_bibliography.html

docs/source/examples/vqe-demo_custom-gate.myst
```diff
@@ -0,0 +1,282 @@
...
+
+
+The pyQuil VQE example shown here is adapted from
+https://grove-docs.readthedocs.io/en/latest/vqe.html#, 
+and the VQE function in Grove.
```
nathanshammah, line 22:
```suggestion
and the VQE function in Grove.
```
Is a verb missing? Like "and the VQE function is taken from Grove"?

docs/source/examples/vqe-demo_custom-gate.myst
```diff
@@ -0,0 +1,282 @@
...
+
+Simulate depolarizing noise on the static gates
+
+```{code-cell} ipython3
+def add_noise_to_circuit(quil_prog):
+    """Define pyQuil gates with a custom noise model via Kraus operators:
+        1. Generate Kraus operators at given survival probability
+        2. Append Kraus operators to the gate matrices
+        3. Add custom gates to circuit
+    """
```
nathanshammah, line 66: Can you complete the docstring using Google style for input and output? More info and links [here in the documentation](https://mitiq.readthedocs.io/en/latest/contributing.html#style-guidelines).

docs/source/examples/vqe-demo_custom-gate.myst
```diff
@@ -0,0 +1,282 @@
...
+    d = 4 ** num_qubits
+    d_sq = d ** 2
+    kraus_list = [(1 - prob) / d] * d
+    kraus_list[0] += prob
+    kraus_ops = pauli_kraus_map(kraus_list)
+    k_list = [(1 - prob) / d_sq] * d_sq
+    k_list[0] += prob
+    k_ops = pauli_kraus_map(k_list)
+    T_gate = np.array([[1, 0], [0, np.exp(1j * np.pi / 4)]])
+    CN_gate = np.block(
```
nathanshammah, line 78:
```suggestion
    CN_gate = np.block(
```
This is a CNOT, right? Can you be  more explicit in the function name?

docs/source/examples/vqe-demo_custom-gate.myst
```diff
@@ -0,0 +1,282 @@
...
+    S_gate = np.array([[1, 0], [0, 1j]])
+    Z_gate = np.array([[1, 0], [0, -1]])
+    quil_prog.define_noisy_gate("T", [0], append_kraus_to_gate(kraus_ops, T_gate))
+    quil_prog.define_noisy_gate("CNOT", [1, 0], append_kraus_to_gate(k_ops, CN_gate))
+    quil_prog.define_noisy_gate("S", [0], append_kraus_to_gate(kraus_ops, S_gate))
+    quil_prog.define_noisy_gate("Z", [0], append_kraus_to_gate(kraus_ops, Z_gate))
+    return quil_prog
+```
+
+ Hamiltonian in this example is just sigma_z on the zeroth qubit
```
nathanshammah, line 90:
```suggestion
 Hamiltonian in this example is just `sigma_z` on the zeroth qubit
```
also here, could try to see if Inter-Sphinx works and link to the relevant function.

docs/source/examples/vqe-demo_custom-gate.myst
```diff
@@ -0,0 +1,282 @@
...
+    executor,
+    init_angle,
+    args=(backend, readout_qubit, samples, hamiltonian, program),
+    method="Nelder-Mead",
+    options={"xatol": 1.0e-3, "fatol": 1.0e-2},
+)
+print(res)
+```
+
+Now introduce noise mitigation technique Zero Noise Extrapolation
```
nathanshammah, line 215:
```suggestion
Now we introduce as quantum error  mitigation technique, Zero Noise Extrapolation.
```

docs/source/examples/vqe-demo_custom-gate.myst
```diff
@@ -0,0 +1,282 @@
...
+
+Create new objective function to calculate expecation value w/ noise mitigation
```
nathanshammah, line 217:
```suggestion
We first create a new objective function to calculate the expectation value with noise mitigation
```
In general, the text could be set more like a single discourse.

docs/source/examples/vqe-demo_custom-gate.myst
```diff
@@ -0,0 +1,282 @@
...
+
+```{code-cell} ipython3
+def mitigated_expectation(
+    thetas, backend, readout_qubit, samples, pauli_sum, executable: Program, factory
+) -> float:
```
nathanshammah, line 222: Can you add docstrings in google style format?

docs/source/examples/vqe-demo_custom-gate.myst
```diff
@@ -0,0 +1,282 @@
...
+    mitigated_exp = zne.execute_with_zne(
+        executable,
+        lambda p: executor(thetas, backend, readout_qubit, samples, pauli_sum, p),
+        factory,
+        fold_gates_from_right,
+    )
+    return mitigated_exp
+```
+
+Here we use a linear technique for the Zero Noise Extrapolation
```
nathanshammah, line 232: You could add a link to the relevant part in the Users Guide where linear factories are discussed, e.g., https://mitiq.readthedocs.io/en/latest/guide/guide-zne.html#classical-fitting-and-extrapolation-factory-objects

docs/source/examples/vqe-demo_custom-gate.myst
```diff
@@ -0,0 +1,282 @@
...
+
+```{code-cell} ipython3
+fac = mitiq.zne.inference.LinearFactory(scale_factors=[1.0, 3.0])
+```
+
+Scan over the parameter theta and plot energy expecation with noise mitigation
```
nathanshammah, line 238:
```suggestion
Scan over the parameter `theta` and plot the energy expectation with error mitigation
```

docs/source/examples/vqe-demo_custom-gate.myst
```diff
@@ -0,0 +1,282 @@
...
+)
+print(res_zne)
+```
+
+## References
+[1] [VQE tutorial in PyQuil.]
+https://grove-docs.readthedocs.io/en/latest/vqe.html#
+
+[2] Rigetti Computing (2018) Grove (Version 1.7.0) 
+[Source code].https://github.com/rigetti/grove
```
nathanshammah, line 274: Are [1] and [2] referenced earlier in the example? In case, I missed it, but they should be. I am not sure that this way the links will work, but I may be wrong. I would have guessed this format worked:

[VQE tutorial in PyQuil.](https://grove-docs.readthedocs.io/en/latest/vqe.html) 
Rigetti Computing (2018) Grove (Version 1.7.0) 
[Source code.](https://github.com/rigetti/grove)

docs/source/examples/vqe-demo_custom-gate.myst
```diff
@@ -0,0 +1,282 @@
...
+
++++
+
+display information about Mitiq, packages, and Python version/platform
```
nathanshammah, line 278:
```suggestion
This final blog displays information about Mitiq, installed packages, and Python version/platform
```

docs/source/examples/vqe-demo_custom-gate.myst
```diff
@@ -0,0 +1,282 @@
...
+
+```{code-cell} ipython3
+res_zne = optimize.minimize(
+    mitigated_expectation,
+    init_angle,
+    args=(backend, readout_qubit, samples, hamiltonian, program, fac),
+    method="Nelder-Mead",
+    options={"xatol": 1.0e-3, "fatol": 1.0e-2},
+)
+print(res_zne)
```
nathanshammah, line 266: After this block, please conclude the example with an analysis of results and a short conclusion.

nathanshammah (changes requested):
Thank you @Misty-W! I left some comments. Overall I would try to be a bit more descriptive in the text part, especially in the initial motivation and introduction and in the final part, addressing the results of the mitigation.

Misty-W:
Awesome, thanks @andreamari!

docs/source/examples/examples.myst
Misty-W: All good points, updated accordingly.

docs/source/examples/vqe-demo_custom-gate.myst
Misty-W: Agree, added.

docs/source/examples/vqe-demo_custom-gate.myst
Misty-W: Good catch! I added the references to [1] and [2] and updated the hyperlinks to the format shown above.

docs/source/examples/vqe-demo_custom-gate.myst
Misty-W: I realize the introduction could've been clearer. I reworded it, so hopefully it makes more sense now.

docs/source/examples/vqe-demo_custom-gate.myst
Misty-W: I updated the introduction with a few sentences on ZNE and another small paragraph on VQE.

docs/source/examples/vqe-demo_custom-gate.myst
Misty-W: I reworded this section, so hopefully it's clearer now.

docs/source/examples/vqe-demo_custom-gate.myst
Misty-W: Good catch! I fixed the references in the example and also reformatted the reference section at the end to hide the link addresses.

docs/source/examples/vqe-pyquil-demo.myst
```diff
@@ -0,0 +1,312 @@
...
+Define example ansatz, consisting of a rotation by angle `theta` and a layer of static gates: 
+
+```{code-cell} ipython3
+program = Program()
+theta = program.declare("theta", memory_type="REAL")
+program += RX(theta, 0)
+program += T(0)
+program += CNOT(1, 0)
+program += S(0)
+program += Z(0)
```
rmlarose, line 73: This ansatz is a little strange  - the only gate that does anything (without noise) is RX(theta, 0).

docs/source/examples/vqe-pyquil-demo.myst
```diff
@@ -0,0 +1,312 @@
...
+    meas_basis_change = Program()
+    marked_qubits = []
+    for index, gate in term:
+        marked_qubits.append(index)
+        if gate == "X":
+            meas_basis_change.inst(RY(-np.pi / 2, index))
+        elif gate == "Y":
+            meas_basis_change.inst(RX(np.pi / 2, index))
+        program += meas_basis_change
+    readout_qubit = program.declare("ro", "BIT", max(marked_qubits) + 1)
```
rmlarose, line 127: This is a bit overkill - to measure Z0 you just need a measurement on the zeroth qubit.

Let's keep it though - this logic is now incorporated in `mitiq.Observable` and it would be good to update this example to use that sometime in the future.

docs/source/examples/vqe-pyquil-demo.myst
```diff
@@ -0,0 +1,312 @@
...
+    expectation = 0.0
+    pauli_sum = PauliSum([pauli_sum])
+    for j, term in enumerate(pauli_sum.terms):
+        qubits_to_measure = []
+        for index, gate in term:
+            qubits_to_measure.append(index)
+            meas_outcome = expectation_from_sampling(
+                thetas, noisy, qubits_to_measure, backend, samples
+            )
+            expectation += term.coefficient * meas_outcome
```
rmlarose, line 166: Again a bit overkill but ok!

docs/source/examples/vqe-pyquil-demo.myst
```diff
@@ -0,0 +1,312 @@
...
+
+    
+    exp_val = 0
+    for bitstring, count in freq.items():
+        bitstring_int = int("".join([str(x) for x in bitstring[::-1]]), 2)
+        if parity_even_p(bitstring_int, marked_qubits):
+            exp_val += float(count) / samples
+        else:
+            exp_val -= float(count) / samples
+    return exp_val
```
rmlarose, line 194: Another FYI, no change requested - this logic is in `mitiq.Observable` now. https://github.com/unitaryfund/mitiq/blob/5ee3da4ba39a732060f700ae269f9ab0b2ca96a9/mitiq/observable/observable.py#L103-L105

rmlarose (approved):
LGTM, thanks Misty!

In the final plot:

image

you may want consider adding the noiseless curve (cosine) and add a legend.

docs/source/examples/vqe-pyquil-demo.myst
Misty-W: I know...I used it to add noise via static gates, and I wanted the ideal energy expectation to have a clear minimum.

docs/source/examples/vqe-demo_custom-gate.myst
Misty-W: I added some explanations around the results and a short conclusion section at the end.

docs/source/examples/vqe-demo_custom-gate.myst
Misty-W: Yes it is, agree it makes sense to make it explicit. It's updated in the example now.

docs/source/examples/var_classifier_mitiq/variational_classifier_mitiq.py
```diff
@@ -0,0 +1,365 @@
...
+import numpy as np
+import matplotlib.pyplot as plt
+
+# initialize quantum device
+qpu = get_qc("2q-noisy-qvm")
+shots = 1
+# create function to calculate expectation value
+executor_fn = generate_qcs_executor(qpu, ground_state_expectation, shots)
+
+# initialize quanutm circuit
```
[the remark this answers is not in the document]
Misty-W: I'm going to mark this as resolved since the file referenced in this comment is no longer part of the PR.

docs/source/examples/vqe-demo_custom-gate.myst
Misty-W: Done!

docs/source/examples/vqe-demo_custom-gate.myst
Misty-W: Done!

Misty-W:
@nathanshammah, thanks for the detailed review! I believe I have resolved all of your comments.

Misty-W:
@rmlarose thanks for your comments! I added your suggestion for the ideal landscape and legend to the example.

nathanshammah (approved):
Thanks @Misty-W, looks great to me after the last points by Ryan are addressed. 

You can also add your name to the top of the example if you'd like to, for authorship and/or with contact email.

nathanshammah:
🎉

PR #714, opened 2021-06-08
Clearer return type for extrapolate and remove deprecated method

### LINKED ISSUE #651: Return type of RichardsonFactory
Pre-Report Checklist
--------------------

- [X] I am running the latest version of mitiq
- [X] I checked to make sure that this bug has not already been reported

Issue Description
-----------------

The return type of `extrapolate` in `RichardsonFactory` is a non-obvious `Union`, which seems to be a bad solution for a couple of reasons:
1. It's confusing to the user.
2. It might make it harder to write modular code, as some functions might depend on one or the other type of the output, which might cause confusion or ugly bugs
3. It might not work well with static type checking (though I haven't checked it properly!)

One solution that I can think of is having two functions, e.g. `extrapolate` and `extrapolate_with_full_output`? 
Any other ideas @dexter2206 ?

### Code Snippet

```python
Union[
        float,
        Tuple[
            float,
            Union[float, None],
            List[float],
            Union[np.ndarray, None],
            Callable[[float], float],
        ],
```

rmlarose:
Description
-----------

Closes #651. In that issue we decided to change the type, but really the issue wasn't about the return type itself, but rather how it was documented/written. As such, I made this clearer.

Also removes a deprecated method which I found in doing the above.

Checklist
-----------

Tips
----

### DIFF
--- a/mitiq/zne/inference.py
+++ b/mitiq/zne/inference.py
@@ -38,6 +38,18 @@ from mitiq import QPROGRAM
 from mitiq.collector import Collector
 
 
+ExtrapolationResult = Union[
+    float,  # The zero-noise value.
+    Tuple[
+        float,  # The zero-noise value.
+        Optional[float],  # The (estimated) error on the zero-noise value.
+        List[float],  # Optimal parameters found during fitting.
+        Optional[np.ndarray],  # Covariance of fitting parameters.
+        Callable[[float], float],  # Function that was fit.
+    ],
+]
+
+
 class ExtrapolationError(Exception):
     """Error raised by :class:`.Factory` objects when
     the extrapolation fit fails.
@@ -302,16 +314,6 @@ class Factory(ABC):
         """
         raise NotImplementedError
 
-    def iterate(
-        self, noise_to_expval: Callable[..., float], max_iterations: int = 100
-    ) -> "Factory":
-        warnings.warn(
-            "The `iterate` method is deprecated in v0.3.0 and will be removed "
-            "in v0.4.0. Use `run_classical` instead.",
-            DeprecationWarning,
-        )
-        return self.run_classical(noise_to_expval)
-
     def push(
         self, instack_val: Dict[str, float], outstack_val: float
     ) -> "Factory":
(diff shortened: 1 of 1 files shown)


mitiq/zne/inference.py
```diff
@@ -38,6 +38,18 @@
...
+ExtrapolationResult = Union[
+    float,  # The zero-noise value.
+    Tuple[
+        float,  # The zero-noise value.
+        Optional[float],  # The (estimated) error on the zero-noise value.
+        List[float],  # Optimal parameters found during fitting.
+        Optional[np.ndarray],  # Covariance of fitting parameters.
+        Callable[[float], float],  # Function that was fit.
+    ],
+]
```
nathanshammah, line 50: Will it show up in docs?

andreamari: I think the docs would be unchanged by this PR. The return type shown in the docs would remain explicit: `Union[float Tuple[ ... ]]`.

andreamari (approved):
Thanks @rmlarose, I am not sure this PR actually closes the issue, since I think most of the user-side problems for writing a custom function remain. But for sure it doesn't harm and at least it simplifies the source code.

What about if we merge this but keep the issue open (removing the PR-issue link)?

mitiq/zne/inference.py
nathanshammah: yeah, I meant this will not show up in docs, unlike the proposal in #651.

rmlarose:
I thought the issue was about the source code, I didn't realize it shows up in docs also. The docs look like:

image

Is this really unclear? The type is explained directly below it.

I'm pushing against a different return type because

```python
zne_value = Factory.extrapolate(...)
```

is intuitive and a very common statement in Mitiq. It's not horrible to make this

```python
zne_value = Factory.extrapolate(...).zne_value
```

but I'm not sure the desire to have a different type signature in the API docs is good enough motivation for this change.

Can we change the API doc some other way?

nathanshammah (approved):
LGTM. I still think that a class may be handy in general. With a returned float either for ZNE, CDR, etc. Can be looked at again in the future.

PR #742, opened 2021-06-21
Update energy landscape example in docs

### LINKED ISSUE #734: Documentation shows a lot of warnings about programs being too short
Look at https://mitiq.readthedocs.io/en/stable/examples/simple_landscape.html for example.

image

andreamari:
This PR updates the energy landscape example in the docs.
Apart from minor changes to improve visibility of the plot, the main reason of this PR is to remove the repeated warnings due to a short circuit.
Since this example is the only place in which the warning appears in the docs, I think this PR fixes #734.

### DIFF
--- a/docs/source/examples/simple_landscape.myst
+++ b/docs/source/examples/simple_landscape.myst
@@ -18,19 +18,19 @@ kernelspec:
 This tutorial shows an example in which the energy landscape for a two-qubit variational 
 circuit is explored with and without error mitigation.
 
-
 ```{code-cell} ipython3
 import matplotlib.pyplot as plt
 import numpy as np
-from cirq import Circuit, H, rx, CNOT, DensityMatrixSimulator, LineQubit, depolarize
-import mitiq
+from cirq import Circuit, X, rx, CNOT, DensityMatrixSimulator, LineQubit, depolarize
+
 from mitiq.zne import mitigate_executor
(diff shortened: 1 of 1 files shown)


docs/source/examples/simple_landscape.myst
```diff
@@ -45,7 +45,7 @@ def variational_circuit(gamma: float) -> Circuit:
         
     q0, q1 = LineQubit.range(2)
         
-    return Circuit([CNOT(q0, q1), rx(gamma)(q1), CNOT(q0, q1)])
+    return Circuit([X(q1), CNOT(q0, q1), rx(gamma)(q1), CNOT(q0, q1), X(q1)])
```
rmlarose, line 48: This made me laugh :)

Can we make it so the gates are less-obviously doing nothing? Or, even better, that they actually are doing something? I think it's fine to completely change the ansatz here because it's a contrived example.

rmlarose (changes requested):

docs/source/examples/simple_landscape.myst
andreamari: I added a less trivial circuit. As a byproduct, the shape of the landscape is now less boring :-)
image

rmlarose: Nice, thanks!

rmlarose (approved):

PR #747, opened 2021-06-22
Patch release 0.9.1

andreamari:
This is a PR from `patch-release-0.9.1` to `release` updating 0.9.0 to 0.9.1.
It contains only 4 commits which are cherry picked from master. 

I think we have more options:

1. Merge this into `release`. In principle this is nice but it may create problems when releasing 0.10.0. I am not sure.
2. Keeping this as an independent branch (`patch-release-0.9.1`). Make a `v0.9.1` tag on this branch.
3. Other suggestions?


rmlarose:
Thanks Andrea. Don't we want to include #744 for this? You'll need to update the version to 0.9.1 also.

Unless I'm missing something, it should be merged into release.

rmlarose (approved):

PR #754, opened 2021-06-24
Update simple_landscape.myst

DSamuel1:
Added noiseless landscape to the simple landscape tutorial and used a different factory and noise scaling for better results.

Description
-----------
I updated the simple landscape tutorial in the mitiq docs so that the noiseless landscape is also visible to compare to. I also used a different factory and added noise scaling to the mitigated landscape for better results.

Checklist
-----------

Tips
----

### DIFF
--- a/docs/source/examples/simple_landscape.myst
+++ b/docs/source/examples/simple_landscape.myst
@@ -4,7 +4,7 @@ jupytext:
     extension: .myst
     format_name: myst
     format_version: 0.13
-    jupytext_version: 1.10.2
+    jupytext_version: 1.11.3
 kernelspec:
   display_name: Python 3
   language: python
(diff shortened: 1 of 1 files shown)


andreamari (commented):
Thanks @DSamuel1 for this PR! 
I like the idea of introducing and plotting also the noiseless landscape. This is a very nice improvement. 

However, there are other changes which I am not sure about:

- the `FakeNodesFactory` is a quite unconventional inference method (it known to be applicable only in the limit of many scale factors) and its application in this setting seems unmotivated. 
- the noise scaling method `fold_all` is compatible only with odd integer `scale_factors`. Moreover for odd integer scale factors it behaves as the default folding method of `mitigate_executor`. So, for simplicity, I would not introduce it in this example.

In conclusion my suggestion would be to **add the plot of the noiseless landscape** without changing the error mitigation part of the existing example. This would also require restoring the noise level back to `p=0.04` and some of the text comments.

Other suggestions:

- You could replace `plt.scatter` with `plt.plot` for the noiseless plot, to better highlight that it is a theoretical curve.
- You could add the noiseless landscape line also in the plot of the unmitigated landscape in the same way as you did for the last plot.
- If you feel more comfortable with restarting from scratch from in new fresh branch feel free to do it. It may be easier than reverting some of the changes related to the error mitigation method.

DSamuel1:
Thanks for the comments. As you suggested, I'm making the changes in another branch and am accordingly closing this PR.

Issue #755, opened 2021-06-25
Behavior of `zne.scaling.fold_all` for scale factors that are not odd integers


andreamari:
Users may be tempted to use this function:
https://github.com/unitaryfund/mitiq/blob/2cb68d64a0f2ebfc98f6c3fbb03cb9c67f516cdf/mitiq/zne/scaling/folding.py#L240

with arbitrary scale factors. However this function, by construction, makes sense only if the scale factor is an odd integer. 
We may sill keep the type of `scale_factor` as `float` for consistency with other methods, but we should check if `scale_factor` is approximately equal to an odd integer. If not, we can raise a `ValueError`.

asuzukosi:
Hey, i'll like to work on this issue

asuzukosi:
@andreamari I've made the suggested changes and created a pull request, you can review it now.

rmlarose:
Should we round it to the nearest odd integer instead of raising an error @andreamari?

andreamari:
> Should we round it to the nearest odd integer instead of raising an error @andreamari?

Actually, the rounding that you suggest is what the function is already doing.

Maybe we can only raise a warning (instead of an error) if the `scale_factor` is not `np.isclose` to its odd integer approximation.

My only concern is that the user may not be aware of the such a crude rounding approximation. E.g. `scale_factor = 2` is rounded to `1`.

rmlarose:
I'm ok with a warning but I don't think its necessary as long as the function is well-documented and clearly states it will round the scale factor.

PR #756, opened 2021-06-25
feat: add odd number check #755

asuzukosi:
Added Odd number check to make sure that scale factor is an odd number before proceeding with the rest of the function.

### DIFF
--- a/mitiq/zne/scaling/folding.py
+++ b/mitiq/zne/scaling/folding.py
@@ -277,6 +277,12 @@ def fold_all(
         raise ValueError(
             f"Requires scale_factor >= 1 but scale_factor = {scale_factor}."
         )
+        
+    if scale_factor % 2 == 0:
+        raise ValueError(
+            f"Requires scale_factor to be an odd number but {scale_factor} is even."
+        )
+        
     _check_foldable(circuit)
 
     folded = deepcopy(circuit)


mitiq/zne/scaling/folding.py
```diff
@@ -277,6 +277,12 @@ def fold_all(
         raise ValueError(
             f"Requires scale_factor >= 1 but scale_factor = {scale_factor}."
         )
+        
+    if scale_factor % 2 == 0:
```
crazy4pi314, line 281: The `scale_factor` parameter seems to be a float, will the mod operator be sufficient here?

crazy4pi314 (commented):
Thanks for picking this up!

crazy4pi314:
@rmlarose @andreamari Is this the desired behavior when the scale factor is odd?

rmlarose:
We don't want to raise an error (apologies to @asuzukosi) and the desired behavior for non odd integer scale factors is still up for discussion (https://github.com/unitaryfund/mitiq/issues/755#issuecomment-871243276). I leave the final decision to @andreamari.

andreamari:
I agree with the comment by @crazy4pi314 . We cannot use the `%` operation since `scale_factor` is a float.

Anyway, based on the last comments in the discussion of #755, I am closing this PR.

Thanks anyway to @asuzukosi for your attempt!

PR #695, opened 2021-05-26
Reduced type annotation errors and updated the CI [unitaryHack] #489

TripleR47:
Description
-----------
* Partially Fixes #489 
* Type annotations corrected in `benchmarks/utils.py`, `zne/zne.py`, `_deprecations.py`, `mitiq_cirq/cirq_utils.py`, `mitiq_braket/conversions.py`, `cdr/clifford_training_data.py`, `mitiq_qiskit/conversions.py`, `conversions.py`, `zne/scaling/folding.py`, `collector.py`, `zne/inference.py`, `mitiq_braket/conversions.py`, `zne/scaling/parameter.py`, `benchmarks/randomized_benchmarking.py`, `pec/types/types.py`, `observable/pauli.py`, `pec/pec.py`, `benchmarks/maxcut.py`
* Added type check in the CI
* Created mypy configuration file `mypy.ini`

Checklist
-----------

Check off the following once complete (or if not applicable) after opening the PR. The PR will be reviewed once this checklist is complete and all tests are passing.

- [x] I added unit tests for new code.
- [x] I used [type hints](https://www.python.org/dev/peps/pep-0484/) in function signatures.
- [x] I used [Google-style](https://sphinxcontrib-napoleon.readthedocs.io/en/latest/example_google.html) docstrings for functions.
- [x] I [updated the documentation](../blob/master/docs/CONTRIBUTING_DOCS.md) where relevant.

If some items remain, you can mark this a [draft pull request](https://github.blog/2019-02-14-introducing-draft-pull-requests/).

Tips
----

- If the validation check fails:

    1. Run `make check-style` (from the root directory of the repository) and fix any [flake8](http://flake8.pycqa.org) errors.

    2. Run `make format` to format your code with the [black](https://black.readthedocs.io/en/stable/index.html) autoformatter.

  For more information, check the [Mitiq style guidelines](https://mitiq.readthedocs.io/en/stable/contributing.html#style-guidelines).
  
- Write "Fixes #XYZ" in the description if this PR fixes Issue #XYZ.

PR #764, opened 2021-06-28
Fix mypy errors and add type check to CI

### LINKED ISSUE #489: Add type check to CI
We had a PR (#326) to add mypy to CI, but while this PR fixed mypy errors it never actually added a job to CI. Do we want to add this?

rmlarose:
Description
-----------

Started from #695 and fixed tests, then fixed the remaining mypy errors. I was probably to judicial with `# type: ignore`, particularly for `extrapolate` methods in `zne.infernece`, but on the whole it will be better to get this in asap so future PRs are careful with typing. (But please review carefully.)

Closes #489. 

Checklist
-----------

Tips
----


mitiq/zne/zne.py
```diff
@@ -28,7 +28,7 @@ def execute_with_zne(
     factory: Optional[Factory] = None,
     scale_noise: Callable[[QPROGRAM, float], QPROGRAM] = fold_gates_at_random,
     num_to_average: int = 1,
-) -> float:
+) -> Optional[float]:
```
andreamari, line 31: The fact that `execute_with_zne` returns an `Optional[float]` instead of a `float` is a bit confusing (this propagates also to zne decorators). But I guess this is the only way to make mypy happy. 

Would something like this be useful to fix the problem:

```python
zne_limit = factory.run(...).reduce()
if zne_limit is None:
    # I actually don't know how to generate this error but it could help mypy
    raise RuntimeError("Zero-noise extrapolation failed.")
return zne_limit
```

Otherwise I am fine with `Optional[float]`

andreamari (approved):
Thanks @rmlarose and @TripleR47. 
It is great to have type checking passing!

I think it is better having many `#type ignore` than not testing at all.  
In the future we could give a bounty for every `#type ignore` which is removed :-)

I left a minor comment about `execute_with_zne` returning an `Optional[float].`

mitiq/zne/zne.py
rmlarose: I don't know why it was optional actually. mypy is fine with just float. Fixed.

mitiq/zne/inference.py
```diff
@@ -962,7 +962,7 @@ def new_curve(scale_factor: float) -> float:
                 FakeNodesFactory._map_to_fake_nodes([scale_factor], a, b)[0]
             )
 
-        return zne_limit, zne_error, opt_params, params_cov, zne_curve
+        return zne_limit, zne_error, opt_params, params_cov, new_curve
```
andreamari, line 965: Thanks

andreamari (approved):
LGTM, thanks!

andreamari:
The building of the docs seems blocked but it should be unrelated to the last commit.

rmlarose:
Yeah I think its GitHub, I just built locally fine.

TripleR47:
Thanks, @rmlarose @andreamari @nathanshammah. I was unable to continue contributing to the PR due to some unforeseen circumstances. It's good to see that the issue is fixed. 🙂

PR #769, opened 2021-06-29
Ensure short circuit warning is multi-platform

andreamari:
Description
-----------

Fix bug related to short circuit warning + Braket circuits. 
Fixes #767

Issue #767, opened 2021-06-29
Cannot run `execute_with_zne` with braket circuits


rmlarose:
[Checking the length of the circuit in `BatchedFactory.run`](https://github.com/unitaryfund/mitiq/blob/master/mitiq/zne/inference.py#L524-L528) bugs out because braket circuits do not have `__len__` defined.

## MWE

(Cirq is only in the picture here because I don't know how to simulate circuits in braket.)

```python
from braket.circuits import Circuit, Instruction, gates
import cirq

from mitiq import zne
from mitiq.interface import accept_any_qprogram_as_input


circuit = Circuit([Instruction(gates.H(), 0)] * 10)


@accept_any_qprogram_as_input
def execute(circuit, noise_level: float = 0.1) -> float:
    noisy = circuit.with_noise(cirq.depolarize(p=noise_level))
    return cirq.DensityMatrixSimulator().simulate(noisy).final_density_matrix[0, 0].real

zne.execute_with_zne(circuit, execute)
```

Raises

```python
TypeError: object of type 'Circuit' has no len()
```

## Proposed fix

Convert the circuit to Cirq and check `len(list(circuit.all_operations()))` in place of https://github.com/unitaryfund/mitiq/blob/master/mitiq/zne/inference.py#L524-L528.

PR #750, opened 2021-06-23
Add CDR example to docs + small change to `cdr.calculate_observable`

rmlarose:
Description
-----------

Closes #693 with an example showing how to use the CDR API.

I also changed the dictionary/counter key type of measurements to be integers, since this is what Cirq returns by default and I think it's more natural.

Additional docs/examples could be added on generating training circuits and doing the regression.

Checklist
-----------

Check off the following once complete (or if not applicable) after opening the PR. The PR will be reviewed once this checklist is complete and all tests are passing.

- [ ] I added unit tests for new code.
- [ ] I used [type hints](https://www.python.org/dev/peps/pep-0484/) in function signatures.
- [ ] I used [Google-style](https://sphinxcontrib-napoleon.readthedocs.io/en/latest/example_google.html) docstrings for functions.
- [ ] I [updated the documentation](../blob/master/docs/CONTRIBUTING_DOCS.md) where relevant.

If some items remain, you can mark this a [draft pull request](https://github.blog/2019-02-14-introducing-draft-pull-requests/).

Tips
----

- If the validation check fails:

    1. Run `make check-style` (from the root directory of the repository) and fix any [flake8](http://flake8.pycqa.org) errors.

    2. Run `make format` to format your code with the [black](https://black.readthedocs.io/en/stable/index.html) autoformatter.

  For more information, check the [Mitiq style guidelines](https://mitiq.readthedocs.io/en/stable/contributing.html#style-guidelines).
  
- Write "Fixes #XYZ" in the description if this PR fixes Issue #XYZ.

PR #770, opened 2021-06-29
Patch release 0.9.2

andreamari:
PR cherry picking #769 from master to fix the Braket error mentioned in this issue #767 .

Do we want to add something else?

I will update version and changelog as a final commit when we are ready, to avoid multiple updates.


rmlarose:
> Do we want to add something else?

Maybe add #750 also so there is documentation on CDR.

andreamari:
> Maybe add #750 also so there is documentation on CDR.

Done.

> If we could also resolve #755 for this that would be good, but not essential.

I would keep this out of the release. Probably everything will reduce to just a warning addition and I think this is not urgent.

rmlarose (approved):
LGTM

PR #779, opened 2021-07-01
Test execute_with_pec with all supported programs

### LINKED ISSUE #768: Make sure that error mitigation techniques are tested for all SUPPORTED_PROGRAM_TYPES
For example, at the time moment, zero-noise extrapolation is never tested with Braket circuits (see  #767).

A good way to automatically test all (current and **future**) supported circuit types is to loop over `mitiq._types.SUPPORTED_PROGRAM_TYPES` when writing important tests.  For example this was done here:
https://github.com/unitaryfund/mitiq/blob/ff6d3e540d7f14430a3dfcf4f7d1f1a5bdc50109/mitiq/cdr/tests/test_cdr.py#L36

andreamari:
Tests `execute_with_pec` with all `SUPPORTED_PROGRAM_TYPES` ,
in the same way as it is already done for `execute_with_zne` and `execute_with_cdr`.

Fixes #768. 

Additional minor change: I restored a previously skipped PEC test by transforming a warning into an error.

### DIFF
--- a/mitiq/pec/tests/test_pec.py
+++ b/mitiq/pec/tests/test_pec.py
@@ -14,7 +14,7 @@
 # along with this program.  If not, see <https://www.gnu.org/licenses/>.
 
 """Unit tests for PEC."""
-
+import warnings
 from typing import List, Optional
 from functools import partial
 import pytest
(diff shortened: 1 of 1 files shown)


rmlarose (approved):
I'm pretty relieved PEC works with braket 😅

PR #811, opened 2021-07-08
Fix multiplication order when adding NoisyOperations

andreamari:
Description
-----------

If we add two noisy operations ([A] +[B]) we get the circuit -[A]-[B]- . The corresponding superoperator should be `S_AB = S_B @ S_A`, while in the current implementation we get  `S_AB = S_A @ S_B`.


nathanshammah (approved):
LGTM.

rmlarose (approved):

PR #826, opened 2021-07-11
Initial support for executors which return measurement results: part 1/2

rmlarose:
Description
-----------

Part of #623. Here all measurement results must be `cirq.Result`s. Part 2/2 will add and handle conversions to/from any supported library (Qiskit, pyQuil, braket).

Note: This is technically a breaking change since `-> np.ndarray` executors are no longer recognized as a batched. (We want executors which return density matrices which are `np.ndarray`s.)

Checklist
-----------

Tips
----

### DIFF
--- a/mitiq/collector.py
+++ b/mitiq/collector.py
@@ -21,12 +21,21 @@ import inspect
 from typing import Any, Callable, Iterable, List, Sequence, Tuple, Union
 
 import numpy as np
+from cirq import Result as MeasurementResult  # TODO: Generalize frontend.
 
 from mitiq import QPROGRAM
-from mitiq.interface import (
-    convert_from_mitiq,
-    convert_to_mitiq,
-)
+from mitiq.interface import convert_from_mitiq, convert_to_mitiq
+
+# An `executor` function inputs a quantum program and outputs an object from
+# which expectation values can be computed. Explicitly, this object can be one
+# of the following types:
+QuantumResult = Union[
+    float,  # The expectation value itself.
+    MeasurementResult,  # Sampled bitstrings.
+    # TODO: Support the following:
+    # np.ndarray,  # Density matrix.
+    # Sequence[np.ndarray],  # Wavefunctions sampled via quantum trajectories.
+]
 
 
 def generate_collected_executor(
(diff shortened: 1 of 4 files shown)


mitiq/pec/pec.py
```diff
@@ -134,10 +134,13 @@ def execute_with_pec(
...
     )
-    exp_values = collected_executor(sampled_circuits)
+    results = collected_executor(sampled_circuits)
 
     # Evaluate unbiased estimators [Temme2017] [Endo2018] [Takagi2020]
-    unbiased_estimators = [norm * s * val for s, val in zip(signs, exp_values)]
+    unbiased_estimators = [
+        norm * s * val  # type: ignore[operator]
+        for s, val in zip(signs, results)
+    ]
```
andreamari, line 143: I see that you renamed `exp_values` to `results`. This is good for me, but this doesn't mean that PEC now works with an executor that returns a `MeasurementResult`. Right? Just to double-check if I am following.

mitiq/collector.py
```diff
@@ -21,20 +21,29 @@
...
+# An `executor` function inputs a quantum program and outputs an object from
+# which expectation values can be computed. Explicitly, this object can be one
+# of the following types:
+QuantumResult = Union[
+    float,  # The expectation value itself.
+    MeasurementResult,  # Sampled bitstrings.
+    # TODO: Support the following:
+    # np.ndarray,  # Density matrix.
+    # Sequence[np.ndarray],  # Wavefunctions sampled via quantum trajectories.
+]
```
andreamari, line 38: Initially we planned to use a universal Python dictionary to return raw counts, instead of supporting many independent platform dependent objects. 

However, I actually like what you are proposing (it is easier for the user), as long as we are aware that this will cost  more work from our side (conversions).

andreamari (approved):

mitiq/pec/pec.py
rmlarose: Yes that's right. mypy complained about multiplying measurements by floats here (its awesome we have mypy now, I never would have caught this otherwise). PEC does not work with executors that return measurements, but nothing does for now and won't until observables are fully supported so that we can take measurements -> expectation values.

mitiq/collector.py
rmlarose: Probably both a `dict[int, int]` and `Counter[int]` will be recognized as `QuantumResult`s eventually. It's very easy to convert these to/from `cirq.Result`, `qiskit.Whatever`, etc., much easier than converting circuits.

PR #832, opened 2021-07-14
Fix warnings in docs log output during build

andreamari:
If you look at the current log output in CI there are some warnings about short circuits. This should fix the problem.


nathanshammah (approved):
Thanks @andreamari. How come some ZNE values worsened?

andreamari:
> Thanks @andreamari. How come some ZNE values worsened?

Because I added two gates, and so the overall noise increased. However the mitigated errors are still better when compared to the unmitigated error.

rmlarose (approved):

PR #835, opened 2021-07-19
Make ZNE (more) usable with PyQuil

### LINKED ISSUE #834: Support pyQuil parametric compilation
Parametric compilation saves time by compiling the program ansatz once, instead of compiling each time the parameters are updated. In pyQuil 2.28.1, a `memory_map` argument is used to substitute in values for previously-declared Quil variables in the pre-compiled executable. It would be helpful to be able to leverage the runtime improvement from parametric compilation when using Mitiq. Right now using a pyQuil memory map with Mitiq results in an error (details below).

The following code snippet (no mitigation) runs through and gives the expected result:

    from pyquil import get_qc, Program
    from pyquil.gates import RY, MEASURE
    import mitiq
    from mitiq import zne
    from mitiq.zne import execute_with_zne
    import numpy as np
    from scipy import optimize
    from functools import partial
    # Initialize the quantum device
    qc = get_qc("2q-noisy-qvm")

    # Set up quantum circuit
    def ansatz():
        program = Program()
        theta = program.declare('theta', memory_type='REAL')
        ro = program.declare('ro', memory_type='BIT', memory_size=1)
        program += RY(theta, 0)
        program += MEASURE(0, ro[0])
        return program

    print(ansatz())

    def expectation_fn(theta, executable):
        bitstrings = qc.run(executable, memory_map= {'theta': [theta]})
        return np.mean(bitstrings[:,0])

    def expectation(thetas, program):
        program.wrap_in_numshots_loop(1000) 
        executable = qc.compile(program)
        bitstrings = qc.run(executable, memory_map={'theta': thetas})
        result = np.mean(bitstrings[:, 0])
        return -result

    thetas = np.linspace(0, 2*np.pi, 21)
    program = ansatz()
    results = []
    program.wrap_in_numshots_loop(1000) 
    quil_prog = qc.compiler.quil_to_native_quil(program, protoquil=True)

    for theta in thetas:
        results.append(expectation_fn(theta, quil_prog))`

Adding the code snippet below results in an error:

    program = ansatz()
    results_zne = []
    for theta in thetas:
        results_zne.append(execute_with_zne(quil_prog, partial(expectation_fn, theta)))

The error message is 

> QVMError: At line 28: Bad memory region name "theta" in MOVE. This is probably due to either:
    * a missing DECLARE for this memory,
    * a misspelling of the memory reference, or
    * a misspelling of the DECLAREd memory.

On the surface it looks like a problem with Quil, but the error only occurs with the addition of `execute_w_zne`. I suspect the error arises because Mitiq does not support `memory_map`, hence I filed the issue against Mitiq.

rmlarose:
Description
-----------

Fixes #834 + makes measurements in scaled pyquil programs the same + keeps metadata (shots) in scaled programs.

`generate_qcs_executor` was originally meant to do this but did so in an extremely rigid and not good way, so I removed it in favor of the above changes.

Checklist
-----------

Tips
----


andreamari (approved):
Thanks for the quick fix @rmlarose ! 

Maybe in the future we could think of a more systematic way of preserving metadata at the conversion level such that everything could work also with PEC and CDR. 

E.g. all conversion function(s) "Cirq->Something", in addition to the `cirq_circuit` to converted, could take as an (optional) argument a native `template_circuit` from which all metadata can be extracted. E.g. something like:

```
draft_native = to_something(cirq_circuit)
good_native = deepcopy(template_circuit) 
good_native = good_native[:0] # remove instructions (this step may be front-end dependent)
good_native += draft_native # (this step may be front-end dependent)
```
Just some random thoughts. Not sure if it makes sense.

rmlarose:
It's not a bad idea. We steered away from it at the start but because of all all the platform-specific metadata a Mitiq Circuit object may make sense, especially if we have to deal with this metadata for each technique - here it's only ZNE, I feel pretty confident PEC or CDR on PyQuil would need similar fixes. We need more testers & tests.

PR #836, opened 2021-07-23
[WIP] Add support for pennylane circuits

rmlarose:
Description
-----------

Checklist
-----------

Tips
----

### DIFF
--- a/dev_requirements.txt
+++ b/dev_requirements.txt
@@ -4,6 +4,7 @@ qiskit-aer~=0.7.6
 qiskit-ibmq-provider~=0.12.2
 pyquil~=2.28.0
 amazon-braket-sdk~=1.5.10
+PennyLane-qiskit~=0.16.0
 
 # Unit tests, coverage, and formatting/style.
 pytest-xdist[psutil]~=2.2.1
--- a/mitiq/interface/mitiq_pennylane/conversions.py
+++ b/mitiq/interface/mitiq_pennylane/conversions.py
@@ -1,0 +1,29 @@
+# Copyright (C) 2021 Unitary Fund
+#
+# This program is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program.  If not, see <https://www.gnu.org/licenses/>.
+
+"""Functions to convert between Mitiq's internal circuit representation and
+Pennylane's circuit representation.
+"""
+from cirq import Circuit
+from cirq.contrib.qasm_import import circuit_from_qasm
+from pennylane_qiskit import AerDevice, load_qasm
+
+
+def from_pennylane(device: AerDevice) -> Circuit:
+    return circuit_from_qasm(device._circuit.qasm())
+
+
+def to_pennylane(circuit: Circuit):
+    return load_qasm(circuit.to_qasm())
--- a/mitiq/interface/mitiq_pennylane/__init__.py
+++ b/mitiq/interface/mitiq_pennylane/__init__.py
@@ -1,0 +1,18 @@
+# Copyright (C) 2021 Unitary Fund
+#
+# This program is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program.  If not, see <https://www.gnu.org/licenses/>.
+
+from mitiq.interface.mitiq_pennylane.conversions import (
+    from_pennylane, to_pennylane
+)


mitiq/interface/mitiq_pennylane/tests/test_conversions_pennylane.py
```diff
@@ -0,0 +1,69 @@
...
+    )
+
+    converted = from_pennylane(to_pennylane(circuit))
+
+    print(circuit)
+    print(converted)
+
+    assert cirq.testing.assert_allclose_up_to_global_phase(
+        cirq.unitary(converted), cirq.unitary(circuit), atol=1e-7,
+    )
```
rmlarose, line 69: Round-trip conversions not preserving qubit order. Output of this test:

```text
      ┌──┐
0: ─────×────────
        │
1: ────×┼────H───
       ││
2: ────×┼────@───
        │    │
3: ─────×────@───
      └──┘
            ┌─────────┐
q_0: ───×────H────────────M('c_0')───
        │
q_1: ───×────@────────────M('c_1')───
             │
q_2: ───×────┼M('c_2')───────────────
        │    │
q_3: ───×────@────────────M('c_3')───
            └─────────┘
```

rmlarose: Hey @trbromley is it possible to reorder the wires of a quantum tape? (Long story motivating this question below.)

1. Input Cirq `circuit`:

```
      ┌──┐
0: ─────×────────
        │
1: ────×┼────H───
       ││
2: ────×┼────@───
        │    │
3: ─────×────@───
      └──┘
```

2. Resulting `tape` from `tape = to_pennylane(circuit)`:

```
 1: ──╭SWAP───H──┤  
 2: ──╰SWAP──╭C──┤  
 0: ──╭SWAP──│───┤  
 3: ──╰SWAP──╰Z──┤  
```

  - Note this is the same as above with permuted qubit order. Namely (0, 1, 2, 3) -> (1, 2, 0, 3).

3. `print(tape.to_openqasm(rotations=True))`

```
OPENQASM 2.0;
include "qelib1.inc";
qreg q[4];
creg c[4];
swap q[0],q[1];
h q[0];
swap q[2],q[3];
cz q[1],q[3];
measure q[0] -> c[0];
measure q[1] -> c[1];
measure q[2] -> c[2];
measure q[3] -> c[3];
```

  - Note the indices of the qubits in the QASM program do not correspond to the indices of the wires in the tape.
  - (Aside from main point about qubit order, but there are also added measurements that are not in the tape.)

4. When converting back to Cirq from this QASM, the qubit order is the same as in the QASM program, so is permuted:

```
            ┌─────────┐
q_0: ───×────H────────────M('c_0')───
        │
q_1: ───×────@────────────M('c_1')───
             │
q_2: ───×────┼M('c_2')───────────────
        │    │
q_3: ───×────@────────────M('c_3')───
            └─────────┘
```

This can be fixed by reordering the wires in step 2, hence the question.

trbromley (commented):
Thanks @rmlarose and sorry for being late to look at this!

The changes look great and thanks for making a start on the tests. My main question is whether we want `rotations=True` when converting to QASM from a tape. The positive is that we don't need to worry about the observable being measured, but a potential negative is that we might end up folding the rotation gates.

The current approach also loses the PL [return type](https://pennylane.readthedocs.io/en/stable/introduction/measurements.html) (e.g., measure expval, var, probs). The only way I can think to support that is with `rotations=False` and adding additional arguments to `to_pennylane` to persist the output measurement, but I can understand why it's preferable to have just the circuit as an input in `to_pennylane`.

It's also nontrivial to support custom wire labels (e.g., `["alice", "bob"]` rather than integers). :thinking: 

The simplest solution is probably to restrict the PL circuits that can be mitigated. For example, we could raise an error if the input tape has custom wire labels or non-expval return types. If this makes sense, I'll be happy to make a PR from a fork.

mitiq/interface/mitiq_pennylane/conversions.py
```diff
@@ -0,0 +1,48 @@
...
+from pennylane import from_qasm as pennylane_from_qasm
+
+
+def from_pennylane(tape: QuantumTape) -> Circuit:
+    """Returns a Cirq circuit equivalent to the input QuantumTape.
+
+    Args:
+        tape: Pennylane QuantumTape to convert to a Cirq circuit.
+    """
+    return cirq_from_qasm(tape.to_openqasm(rotations=True))
```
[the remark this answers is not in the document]
rmlarose: It's a good point, thanks for bringing this up! Without regard to folding and just thinking about `from_pennylane`, it should include the rotations, otherwise the Cirq circuit - which only supports computational basis measurements - will not be equivalent.

However this makes it so that the Pennylane -> Cirq -> Pennylane sequence can start with non-computational basis measurements and end with computational basis measurements, which is a little strange. With hindsight we've realized conversions typically have problems like this, and have handled it in a "first do sloppy conversion, then clean it up" way. (Two examples of cleaning it up are [adding back PyQuil declarations](https://github.com/unitaryfund/mitiq/blob/9140ec2080651fe621f65d61d1ab2fb13e4ca680/mitiq/interface/conversions.py#L230-L265) and [keeping the same register structure in Qiskit](https://github.com/unitaryfund/mitiq/blob/9140ec2080651fe621f65d61d1ab2fb13e4ca680/mitiq/interface/conversions.py#L267-L282).) This is a con of the design in my opinion, but the pro is that as long as you have `to_whatever` and `from_whatever`, even if it's (necessarily) a little sloppy, it makes it very easy to add support for `whatever`.

Anyway, you asked about folding the rotations: Since all other platforms measure in the Z basis we do fold rotations by default. There are options to not fold single-qubit gates, or certain types of gates, so whether or not rotations are there doesn't matter too much from the perspective of folding.

rmlarose:
> However, it looks like an issue in tape.to_openqasm, since the wires in the QASM script are matched to the order specified by > tape.wires = <Wires = [1, 2, 0, 3]>. A quick hack fix is to do:
> ```python
> tape.to_openqasm(wires=sorted(tape.wires))
> ```
> Overall this looks like a bug/unexpected behaviour and I'll discuss with the team.

Sounds good, thanks! This is currently the main blocker for me. I'm fine with this suggested fix and the requirement that the tape's wires must be sortable. (Which isn't the case with wires labeled `[0, "a"]`, for example.)

mitiq/interface/mitiq_pennylane/tests/test_conversions_pennylane.py
rmlarose: Sounds good, thanks! This is currently the main blocker for me. I'm fine with this suggested fix and the requirement that the tape's wires must be sortable. (Which isn't the case with wires labeled [0, "a"], for example.)

mitiq/interface/mitiq_pennylane/conversions.py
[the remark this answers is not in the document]
andreamari: What about having the `rotations` argument in the function `from_pennylane`?

It seems that if we consider `from_pennylane` as a stand-alone function, the natural choice would be `rotations=True` .

On the other hand, if we also want to restore the full metadata of the tape (return type, observable, etc.) somewhere along the Mitiq worflow, it is probably easier if one can set `rotations=False` to decouple the bare circuit from the rest.

dev_requirements.txt
```diff
@@ -4,6 +4,7 @@ qiskit-aer~=0.8.2
 qiskit-ibmq-provider~=0.16.0
 pyquil~=2.28.0
 amazon-braket-sdk~=1.5.10
+pennylane-qiskit~=0.16.0
```
trbromley, line 7: Thanks, I've posted an issue to share with the team regarding the chained dependency on all of qiskit: https://github.com/PennyLaneAI/pennylane-qiskit/issues/149

mitiq/interface/mitiq_pennylane/tests/test_conversions_pennylane.py
trbromley: Thanks for adding in the fix with pre-sorting the wires! Regarding the [follow up comment](https://github.com/PennyLaneAI/pennylane/issues/1500#issuecomment-893763852) on the PL-issue, is that a blocker for this PR? :thinking:

trbromley (commented):
Thanks @rmlarose! This is looking great!

To confirm, the remaining item left to do is to add a check for the return type of the quantum tape in `from_pennylane()` to ensure that only returning expectation values (for now) is supported? I can help with this, just want to make sure you're on board.

I'd also like to do a few manual trial runs of mitigating PL tapes using this new functionality, to test for any edge cases I can think of. Is there anything remaining to be done in the frontend to get the PL-tape compatible with, for example, `execute_with_zne` or the higher level folding functionality?

dev_requirements.txt
rmlarose: Thanks!

rmlarose:
Thanks @trbromley. I just pushed code adding Pennylane tapes to `mitiq.QPROGRAM` and the top-level to/from conversions. It's still a bit rough around the edges, and needs more tests.

> Thanks for adding in the fix with pre-sorting the wires! Regarding the follow up comment on the PL-issue, is that a blocker for this PR? :thinking:

There are some failing tests with PEC, so I would consider that blocking, yes.

> To confirm, the remaining item left to do is to add a check for the return type of the quantum tape in `from_pennylane()` to ensure that only returning expectation values (for now) is supported? I can help with this, just want to make sure you're on board.

I'm actually not sure what you mean by this, sorry. Can you elaborate?

> Is there anything remaining to be done in the frontend to get the PL-tape compatible with, for example, execute_with_zne or the higher level folding functionality?

Probably - An end-to-end example will tell. With the latest code we can kinda sorta fold gates in Pennylane tapes:

```python
import numpy as np

import pennylane as qml
from mitiq import zne

with qml.tape.QuantumTape() as tape:
    qml.RX(np.pi / 2, wires=[0])
    qml.CNOT(wires=[0, 1])
    qml.expval(qml.PauliZ(wires=[0]))
    qml.expval(qml.PauliZ(wires=[1]))

print(tape.draw())

scaled_tape = zne.scaling.fold_all(tape, scale_factor=3)
print(scaled_tape.draw())
```

```
 0: ──RX(1.57)──╭C──┤ ⟨Z⟩ 
 1: ────────────╰X──┤ ⟨Z⟩ 


 0: ──RX(1.57)──RX(-1.57)──RX(1.57)──╭C──╭C──╭C──┤  
 1: ─────────────────────────────────╰X──╰X──╰X──┤  
```

Note the `⟨Z⟩`s are removed in the `scaled_tape`. This would probably cause issues in an end-to-end example.

mitiq/interface/mitiq_pennylane/conversions.py
```diff
@@ -0,0 +1,60 @@
...
+        tape: Pennylane QuantumTape to convert to a Cirq circuit.
+    """
+    try:
+        wires = sorted(tape.wires)
+    except TypeError:
+        raise UnsupportedQuantumTapeError(
+            f"The wires of the tape must be sortable, but could not sort "
+            f"{tape.wires}."
+        )
+
```
trbromley, line 45:
```suggestion

    if not all(m.return_type is Expectation for m in tape.measurements):
        raise UnsupportedQuantumTapeError("Only expectation value measurements are supported.")
```

mitiq/interface/mitiq_pennylane/conversions.py
```diff
@@ -0,0 +1,60 @@
...
+# along with this program.  If not, see <https://www.gnu.org/licenses/>.
+
+"""Functions to convert between Mitiq's internal circuit representation and
+Pennylane's circuit representation.
+"""
+
+from cirq import Circuit
+from mitiq.interface.mitiq_qiskit import from_qasm as cirq_from_qasm, to_qasm
+from pennylane.wires import Wires
+from pennylane.tape import QuantumTape
```
trbromley, line 23:
```suggestion
from pennylane.tape import QuantumTape
from pennylane.operation import Expectation
```

mitiq/interface/mitiq_pennylane/tests/test_conversions_pennylane.py
```diff
@@ -0,0 +1,85 @@
...
+    assert _equal(circuit, correct, require_qubit_equality=False)
+
+
+def test_from_pennylane_unsupported_tapes():
+    with qml.tape.QuantumTape() as tape:
+        qml.CZ(wires=[0, "a"])
+
+    with pytest.raises(UnsupportedQuantumTapeError, match="could not sort"):
+        from_pennylane(tape)
+
```
trbromley, line 72:
```suggestion


def test_no_variance():
    with qml.tape.QuantumTape() as tape:
        qml.CNOT(wires=[0, 1])
        qml.var(qml.PauliZ(0))

    with pytest.raises(UnsupportedQuantumTapeError, match="Only expectation value measurements"):
        from_pennylane(tape)

```

trbromley (commented):
Thanks @rmlarose!

> >    To confirm, the remaining item left to do is to add a check for the return type of the quantum tape in from_pennylane() to ensure that only returning expectation values (for now) is supported? I can help with this, just want to make sure you're on board.
>
> I'm actually not sure what you mean by this, sorry. Can you elaborate?

The suggestions I've left in this review should fix things. In PennyLane we can have other [return types](https://pennylane.readthedocs.io/en/stable/introduction/measurements.html), for example:
```python
with qml.tape.QuantumTape() as tape:
    qml.RX(0.2, wires=0)
    qml.RY(0.6, wires=1)
    qml.CNOT(wires=[0, 1])
    qml.var(qml.PauliY(0))
```
This would give the variance of `PauliY`. I think for this PR it's sufficient to support just expectation values.

> There are some failing tests with PEC, so I would consider that blocking, yes.

This should be fixed in https://github.com/PennyLaneAI/pennylane/pull/1559 - though this would not make it to a release version of PennyLane until September :thinking:

nathanshammah:
The validate CI fails seems fixable with `black`. The `test/test` fail seems do to an `AssertionError `.

crazy4pi314:
Thanks to @trbromley @andreamari @nathanshammah @WrathfulSpatula and everyone who joined the community debug session today on discord! 💖

We narrowed down some of the failing tests here to being issues with the conversion functions `to_pennylane` and `from_pennylane` that fail only when used in CDR and PEC (ZNE seems to be ok)

Here are direct line links to those functions:
https://github.com/unitaryfund/mitiq/blob/99bcdbf31eba08cd427857f9e4cb4130300c5b50/mitiq/interface/mitiq_pennylane/conversions.py#L33-L34
https://github.com/unitaryfund/mitiq/blob/99bcdbf31eba08cd427857f9e4cb4130300c5b50/mitiq/interface/mitiq_pennylane/conversions.py#L57-L57

andreamari:
Thanks @crazy4pi314 for summarizing the discussion!

The problem with failing tests originated from default **final measurements** which are always added by `tape.to_openqasm` even if absent in the tape.

In the last commit final measurements are manually removed during the conversion and all previous tests are fixed.
This is probably not an optimal solution, but I think that supporting tapes without measurements is better than supporting tapes where all qubits must always be measured.  If you disagree with removing the measurements, we can revert this change introducing a better handling of measurement gates.

Note: there is a new failing test associated to the conversion of `cirq.IdentityGate` which, when converted to a tape, becomes a numerical matrix. Not sure if this is intentional but it seems to create problems with Mitiq since the numerical matrix is not recognized by Mitiq as a valid gate. 

This is the test:
```
    q = cirq.LineQubit(0)
    converted = from_pennylane(to_pennylane(circuit))
    assert _equal(circuit, converted, require_qubit_equality=False)
```

mitiq/interface/mitiq_pennylane/conversions.py
```diff
@@ -0,0 +1,80 @@
...
+def from_pennylane(tape: QuantumTape) -> Circuit:
+    """Returns a Mitiq circuit equivalent to the input QuantumTape.
+
+    Args:
+        tape: Pennylane QuantumTape to convert to a Mitiq circuit.
+
+    Returns:
+        Mitiq circuit representation equivalent to the input QuantumTape.
+    """
+    try:
```
WrathfulSpatula, line 45: Looking at the reverse conversion, to pennylane, we label pennylane `Wires()` with a `range(len(n))` pattern, which will necessarily pack `[0,...n-1]` for `n` wires. However, all we're requiring here is that `Wires()` are _sortable_, (potentially with gaps,) rather than densely packed from `[0,...n-1]`.

We could check, here, that the wires are sequentially packed:
```
    for i in range(len(wires)):
        if wires[i] != i:
            raise UnsupportedQuantumTapeError(
                f"The wire labels of the tape must contiguously pack 0 "
                f"to n-1, for n wires."
            )
```

Further, the pennylane documentation specifies that `Wires()` is _already_ a sorted iterable collection. So, I think we can actually count on the `Wires()` to be sorted in the first place with, without the `try` block here that attempts to sort, currently in this method.

rmlarose: I vaguely remember a test with wires named "a" (str) and 0 (int) which couldn't be sorted.

andreamari:
Hello everyone, a quick update.

There was a remaining failing test due to a problem related to conversions of identity gates.
So I updated the failing test from an exact circuit comparison to a matrix comparison:
https://github.com/unitaryfund/mitiq/blob/e82132d0089056dccee6e4651bd7e4942146fbfe/mitiq/interface/mitiq_pennylane/tests/test_conversions_pennylane.py#L112

The reason is that the Identity operation doesn't exist in PennyLane at the moment although there is a plan to add it (https://github.com/PennyLaneAI/pennylane/issues/1632).
See, also https://github.com/PennyLaneAI/pennylane-qiskit/issues/153.

Once those issues will be fixed, we can improve the test asserting gate equality instead of matrix equality.

There may be other subtle edge cases that I didn't notice. So feel free to add more tests.

A thing to decide before merging is what kind of `tape` we want to explicitly support. Do we want to raise an error each time a tape contains other things that are not gates (measurements, expvals, etc..)?  This is sub-optimal but probably safer solution.
If I am not wrong, @trbromley suggested something like this during the last PL community call.

nathanshammah:
About @andreamari's proposal:
> A thing to decide before merging is what kind of tape we want to explicitly support. Do we want to raise an error each time a tape contains other things that are not gates (measurements, expvals, etc..)? This is a sub-optimal but probably safer solution.
I am in favor of this. We can extend functionality later. We could then add an example with basic integration.

nathanshammah:
I resolved the conflict updating the requirements to include the latest braket update on master branch.

mitiq/interface/mitiq_pennylane/conversions.py
```diff
@@ -0,0 +1,87 @@
...
+"""
+
+from cirq import Circuit, Moment
+
+from mitiq.interface.mitiq_qiskit import from_qasm as cirq_from_qasm, to_qasm
+from mitiq.utils import _pop_measurements
+
+from pennylane.wires import Wires
+from pennylane.tape import QuantumTape
+from pennylane.operation import Expectation
```
trbromley, line 27:
```suggestion
```
(if we make below changes)

mitiq/interface/mitiq_pennylane/conversions.py
```diff
@@ -0,0 +1,87 @@
...
+
+from pennylane import from_qasm as pennylane_from_qasm
```
trbromley, line 29:
```suggestion
from pennylane import from_qasm as pennylane_from_qasm

from pennylane_qiskit.qiskit_device import QISKIT_OPERATION_MAP

SUPPORTED_PL = set(QISKIT_OPERATION_MAP.keys())
UNSUPPORTED = {"CRX", "CRY", "CRZ", "S", "T"}
SUPPORTED = SUPPORTED_PL - UNSUPPORTED
```
This is for decomposing unsupported operations in the tape.

mitiq/interface/mitiq_pennylane/conversions.py
```diff
@@ -0,0 +1,87 @@
...
+            raise UnsupportedQuantumTapeError(
+                "The wire labels of the tape must contiguously pack 0 "
+                "to n-1, for n wires."
+            )
+
+    if not all(m.return_type is Expectation for m in tape.measurements):
+        raise UnsupportedQuantumTapeError(
+            "Only expectation value measurements are supported."
+        )
+    output = cirq_from_qasm(tape.to_openqasm(rotations=True, wires=wires))
```
trbromley, line 64:
```suggestion
    tape = tape.expand(stop_at=lambda obj: obj.name in SUPPORTED)
    
    output = cirq_from_qasm(tape.to_openqasm(rotations=True, wires=wires))
```
This expands the tape to only supported operations.

mitiq/interface/mitiq_pennylane/tests/test_conversions_pennylane.py
```diff
@@ -0,0 +1,123 @@
...
+    assert np.allclose(cirq.unitary(circuit), cirq.unitary(converted))
+
+
+def test_non_consecutive_wires_error():
+    with qml.tape.QuantumTape() as tape:
+        qml.CNOT(wires=[0, 2])
+    with pytest.raises(
+        UnsupportedQuantumTapeError, match="contiguously pack",
+    ):
+        from_pennylane(tape)
```
trbromley, line 123:
```suggestion
        from_pennylane(tape)


def test_integration():
    n_wires = 4

    gates = [
        qml.PauliX,
        qml.PauliY,
        qml.PauliZ,
        qml.S,
        qml.T,
        qml.RX,
        qml.RY,
        qml.RZ,
        qml.Hadamard,
        qml.Rot,
        qml.CRot,
        qml.Toffoli,
        qml.SWAP,
        qml.CSWAP,
        qml.U1,
        qml.U2,
        qml.U3,
        qml.CRX,
        qml.CRY,
        qml.CRZ,
    ]

    layers = 3
    np.random.seed(1967)
    gates_per_layers = [np.random.permutation(gates) for _ in range(layers)]

    with qml.tape.QuantumTape() as tape:
        np.random.seed(1967)
        for gates in gates_per_layers:
            for gate in gates:
                params = list(np.pi * np.random.rand(gate.num_params))
                rnd_wires = np.random.choice(range(n_wires), size=gate.num_wires, replace=False)
                gate(
                    *params,
                    wires=[
                        int(w) for w in rnd_wires
                    ]  # make sure we do not address wires as 0-d arrays
                )

    base_circ = from_pennylane(tape)
    tape_recovered = to_pennylane(base_circ)
    circ_recovered = from_pennylane(tape_recovered)

    u_1 = cirq.unitary(base_circ)
    u_2 = cirq.unitary(circ_recovered)

    assert np.allclose(u_1, u_2)

```
Proposing the above integration test, based upon [this](https://github.com/PennyLaneAI/pennylane/blob/032b3016919fd1671e26c6e89e46de967ed65836/pennylane/devices/tests/test_compare_default_qubit.py#L187) test.

Having this test pass makes me much more confident that reasonable circuits can be converted back and forth.

trbromley (commented):
Thanks everyone! This looks great.

I agree with @andreamari's suggestion to restrict the input tape to not have any measurements at this stage.

I have added a few suggestions, including an integration test that should give a deeper confirmation that the conversion is working as expected.

mitiq/interface/mitiq_pennylane/conversions.py
andreamari: This was added a few commits below keeping the "try/execpt" and the suggestion by @WrathfulSpatula.

link: https://github.com/unitaryfund/mitiq/pull/836/commits/7283e56ce3bbb3a8920f92a3484f3975be55fb16

mitiq/interface/mitiq_pennylane/conversions.py
```diff
@@ -0,0 +1,87 @@
...
+    # tape.to_openqasm always introduces measurements that we remove
+    measurements = _pop_measurements(output)
+    # Remove final empty moment if present
+    if measurements and output[-1] == Moment():
+        return output[:-1]
```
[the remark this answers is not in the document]
andreamari: +1 for  `tape.to_openqasm(rotations=True, wires=wires, measure_all=False)`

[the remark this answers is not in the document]
andreamari: Applied in this commit: https://github.com/unitaryfund/mitiq/pull/836/commits/eb143d113999cca6fab79c117d7399b65e706758 . It works!

I also explicitly added `pennylane>=0.18` in `dev_requirements.txt `to ensure it works in local environments with previous  versions of PL.

mitiq/interface/mitiq_pennylane/conversions.py
andreamari: Cool!

andreamari:
Thanks for all the great suggestions, @trbromley !

Side note: I am impressed that, after directly applying all your complex GitHub suggestions, tests passed out of the box! :smiley: 

I am approving this PR, but I would wait for a double check by other Mitiq people before merging.

andreamari (approved):

dev_requirements.txt
```diff
@@ -3,6 +3,8 @@ qiskit-terra~=0.18.2
 qiskit-aer~=0.8.2
 qiskit-ibmq-provider~=0.16.0
 pyquil~=2.28.0
+pennylane-qiskit~=0.18.0
+pennylane>=0.18
```
trbromley, line 7: Do you specify anywhere in the docs a minimum version for soft dependencies?

trbromley (approved):
Thanks @andreamari for the latest updates and to everyone for contributing to this! It looks good to go and I'm excited to have this integration :rocket: 

Regarding updating the documentation (e.g. adding to [here](https://mitiq.readthedocs.io/en/stable/readme.html#supported-frontends)), is that best done in a follow up PR? I'd think so, it's nice to get this PR across the line first.

dev_requirements.txt
andreamari: @crazy4pi314, @rmlarose, @nathanshammah  what do you think?

In principle it would be nice to add info about minimum versions but there are some downsides:

- It can easily become outdated.
- In general, we actually don't know what are the minimum versions since we only test the specific versions  pinned in `dev_requirements.txt`.

crazy4pi314:
I'll merge this now, and make sure we have an issue for user docs on this! Grats everyone 🎉

andreamari:
Even if tests were passing in this PR, after merging on master tests are failing for some `pip` installation problem. I would propose clicking on the revert button above to undo the changes until we figure out how to fix the problem.

PR #840, opened 2021-07-29
Show output of Mitiq paper code blocks in example on RTD

nathanshammah:
On RTD, in the latest [build](https://mitiq.readthedocs.io/en/latest/examples/mitiq-paper/mitiq-paper-codeblocks.html) the Mitiq paper code blocks outputs are not shown, unlike for the other files, including the CDR API file. I thus enforced the `.md` extension for the paper code blocks file and copied the CDR API file metadata. 

Note that also in the old way (current master), if I was doing `make docs` locally I would see the run code blocks.


andreamari (approved):
Thanks Nathan, I merge it so we can see if it works on RTD.

Issue #813, opened 2021-07-09
Enable interactive code in the documentation


nathanshammah:
This could be done with [`jupyter-book`](https://jupyterbook.org/intro.html), as discussed in the Mitiq meeting, which extends capabilities of `my-binder`.

nathanshammah:
Quick update on the findings so far:

- 1. As expected, setting up My Binder was very straightforward, and the Mitiq runs on cloud if you click here ---> [![Binder](https://mybinder.org/badge_logo.svg)](https://mybinder.org/v2/gh/unitaryfund/mitiq/HEAD), where one can fire up a new notebook and test that Mitiq is installed:
<img width="916" alt="Screen Shot 2021-07-23 at 12 04 37" src="https://user-images.githubusercontent.com/14573436/126760729-2c5359d0-3ac2-44e2-8496-6edae47411e0.png">
 
- 2. Right now there are no jupyter notebooks in the documentation to point to, so the user does not see interactive examples. Point 1. was enabled just by copy-pasting the Mitiq repo link in the Mybinder.org website. It may be possible that by adding specifications in an appropriate `.yml` file in the MItiq Github repository, the My Binder instructions will be such that they include plugins needed to make the `.rst`, `.md`, and `.myst` examples interactive and runnable, such as `Myst-NB` and `jupytext`. I am looking into this. 
- 3. The alternative approach is to use `Jupyter Book`, which is great, and I went through the whole online documentation. I particularly like the `Thebe` options that keeps you in the webpage and make it interactive without changing it at all, and the `plotly` integrations (maybe plotly is also configurable with jupyter nb). However, there are a couple of points to understand properly and decide upon here. I found already these issues: 
- 4. Currently `Jupyter Book` is not hosted on Read The Docs (RTD), which is were we are hosting the documentation. I understand that RTD is considering integrating, but for the time this would mean not having the documentation hosted at  and changes also in continuous integration. Jupyter Book would use GitHub Pages for hosting. I actually think that this could be even better, but it is quite a big breaking change, involving also changes on internet traffic, etc. 
[...]

PR #841, opened 2021-07-29
Add My Binder

nathanshammah:
Add My Binder. Customize install to include all developmnent requirements, and allow to customize the My Binder so that also myst files show up interactively as explained in #813. To do so, include an `environment.yml` configuration file that My Binder should pick up.


nathanshammah:
I am now specifying the build on [mybinder.org](https://mybinder.org/) to tag this branch, to check that it works:

my-binder-snapshot

nathanshammah:
My Binder builds correctly, as one can check at [Binder](https://mybinder.org/v2/gh/nathanshammah/mitiq/binder) (`nathanshammah/mitiq` binder branch).

Still, even by installing all repos as in `environment.yml`, the myst, rst and md files do not render in My Binder. See:
broken-myst


This is not the case locally when one makes `make docs`. Even more, it is possible to configure My Binder to properly show such text files, as shown in the Jupytext My Binder, so I will try to copy from that environment, 
https://mybinder.org/v2/gh/mwouts/jupytext/master?filepath=demo.

nathanshammah:
Simply adding `jupytext` to the requirements in the last commit worked out fine. Now `myst`, `md` and `rst` files render correctly on My Binder, and the requirements are installed. 

Screen Shot 2021-07-29 at 19 23 18


One could even check if not adding the `environment.yml` file, and just adding  `jupytext` to the `requirements.txt` file things work, to avoid duplications. But I want to be sure that all `dev_requirements.txt` dependencies are installed, so I'll leave it as it is. I could aslo try without setitng up conda environment with conda forge, but it just works. In the next commit I will add a shield in the readme to point to the master branch. Until this branch is merged however, that button will not show up correctly the myst files, etc. 

One thing I'd like to customize is to point already to the examples folder. So that the user does not get lost by opening a jupyter notebook menu in the main mitiq folder. I will configure the startup folder to be `examples`, since the `examples.myst`, even if it shows up in notebok format cell, it does not provide clickable elements in the TOC. I wonder if a markdown-based TOC would work better. 

See the current view:
Screen Shot 2021-07-29 at 19 28 19


and the hyperlink that works in markdown:

Screen Shot 2021-07-29 at 19 29 39

nathanshammah:
Ok, the way to correctly point to the `example.myst` file is to type in mybinder.org:

**GitHub repository name or URL**: https://github.com/nathanshammah/mitiq
**Git ref (branch, tag, or commit)**: binder
**File (optional)**: /docs/source/examples/examples.myst

I will set up in the Readme

**GitHub repository name or URL**: https://github.com/unitaryfund/mitiq
**Git ref (branch, tag, or commit)**: master
**File (optional)**: /docs/source/examples/examples.myst

Note that this way the my binder will build from scratch the image if the current head is more recent that the latest commit logged in. This may slow down things, but it's probably the best to be sure that we're on track. We could also pinpoint the latest released version. I think it's ok to pin master for now. 

I am trying to modify the `examples.myst` so that below the TOC it shows links to the files. Probably the way to do it is to add another index jupyter file.

nathanshammah:
Ok, my work around was to add some links in markdown just for My Binder, which is different from the rst table of contents. At [this link](https://mybinder.org/v2/gh/nathanshammah/mitiq/binder?filepath=%2Fdocs%2Fsource%2Fexamples%2Fexamples.myst
) you can see how it will render also on master.

rmlarose:
> Ok, my workaround was to add some links in markdown just for My Binder, which is different from the rst table of contents. At [this link](https://mybinder.org/v2/gh/nathanshammah/mitiq/binder?filepath=%2Fdocs%2Fsource%2Fexamples%2Fexamples.myst) you can see how it will render also on master.

When I click links to rst files it just downloads the rst files. In the non-rst files, I had to pip install mitiq manually before I could run anything.

nathanshammah:
@rmlarose the `.rst` files get downloaded instead of linked and are not rendered interactively. 

Mitiq is installed correctly (check [here](https://hub.gke2.mybinder.org/user/nathanshammah-mitiq-0mrnp8es/tree/docs/source/examples) or [here](https://hub.gke2.mybinder.org/user/nathanshammah-mitiq-0mrnp8es/tree/)) and the `.myst` notebooks and they run fine. Actually also `.md` the CDR API Mitiq runs fine. Only the Mitiq paper notebook does not run. I wonder if the `import mitiq` that fails in the mitiq paper notebook is related to the absence (or presence) of the [`mitiq.py`](https://github.com/unitaryfund/mitiq/blob/master/docs/source/examples/mitiq.py) file put there for read the docs.

rmlarose:
> @rmlarose the `.rst` files get downloaded instead of linked and are not rendered interactively.

I think we should convert the rst docs to something that actually renders if we're going this route.

environment.yml
```diff
@@ -0,0 +1,26 @@
+name: mitiq-mybinder-environment
+channels:
+  - conda-forge
+dependencies:
```
rmlarose, line 4: Very not ideal to list out all dependencies + dev dependencies explicitly...

andreamari (commented):
**Proposal:** 
instead of using the same `examples.myst` file as a TOC for the docs and as a TOC for mybinder, maybe we could use new file e.g. `mybindertoc.myst` dedicated only to mybinder and not rendered in the html docs.

In this file we could link only those examples that run properly on mybinder. Moreover, if you now open `examples.myst` with mybinder, before the links, you see the unrelated sphinx source code which is a bit odd. This would not be present in a dedicated `mybindertoc.myst`.

nathanshammah:
Quick update to mention that I am translating the rst files in the examples' section into md/myst, with [rst-to-myst](https://rst-to-myst.readthedocs.io/en/latest/). Now troubleshooting errors.

README.md
```diff
@@ -10,10 +10,10 @@
 [![arXiv](https://img.shields.io/badge/arXiv-2009.04417-<COLOR>.svg)](https://arxiv.org/abs/2009.04417)
 [![Downloads](https://static.pepy.tech/personalized-badge/mitiq?period=total&units=international_system&left_color=black&right_color=green&left_text=Downloads)](https://pepy.tech/project/mitiq)
 [![Repository](https://img.shields.io/badge/GitHub-5C5C5C.svg?logo=github)](https://github.com/unitaryfund/mitiq)
-[![Unitary Fund](https://img.shields.io/badge/Supported%20By-Unitary%20Fund-FFFF00.svg)](https://unitary.fund)
+[![Unitary Fund](https://img.shields.io/badge/Supported%20By-Unitary%20Fund-FFFF00.svg)](http://unitary.fund)
 [![Discord Chat](https://img.shields.io/badge/dynamic/json?color=blue&label=Discord&query=approximate_presence_count&suffix=%20online.&url=https%3A%2F%2Fdiscord.com%2Fapi%2Finvites%2FJqVGmpkP96%3Fwith_counts%3Dtrue)](http://discord.unitary.fund)
-
-Mitiq is a Python toolkit for implementing error mitigation techniques on
+[![Binder](https://mybinder.org/badge_logo.svg)](https://mybinder.org/v2/gh/unitaryfund/mitiq/master?filepath=%2Fdocs%2Fsource%2Fexamples)
+Mitiq is a )Python toolkit for implementing error mitigation techniques on
```
rmlarose, line 16: Random )

docs/source/examples/examples.myst
```diff
@@ -4,12 +4,26 @@
...
+
+[Defining Hamiltonians as Linear Combinations of Pauli Strings](hamiltonians.md)
+
+[Mitigating the energy landscape of a variational circuit with Mitiq](simple_landscape.myst)
+
+[Solving MaxCut with QAOA improved with Mitiq](maxcut-demo.myst)
+
+[CLifford data regression API](cdr_api.md)
+
+[Mitiq paper codeblocks](mitiq-paper/mitiq-paper-codeblocks.md)
```
rmlarose, line 29: It's really not great that the TOC has to be duplicated. Is there any other way to do this?

docs/source/examples/examples.myst
```diff
@@ -4,12 +4,26 @@
...
+
+[Error mitigation on IBMQ backends](ibmq-backends.myst)
+
+[Defining Hamiltonians as Linear Combinations of Pauli Strings](hamiltonians.md)
+
+[Mitigating the energy landscape of a variational circuit with Mitiq](simple_landscape.myst)
+
+[Solving MaxCut with QAOA improved with Mitiq](maxcut-demo.myst)
+
+[CLifford data regression API](cdr_api.md)
```
rmlarose, line 27: typo

docs/source/examples/examples.myst
```diff
@@ -4,12 +4,26 @@
...
+hamiltonians.md
+simple_landscape.myst
 maxcut-demo.myst
 pyquil_demo.myst
 cdr_api.md
 mitiq-paper/mitiq-paper-codeblocks
 pec-tutorial.myst
 ```
+
+## Links to the examples for My Binder
```
rmlarose, line 17: Should "My Binder" be just one word? Or just "Binder"?

docs/source/examples/ibmq-backends.myst
```diff
@@ -0,0 +1,24 @@
...
+
+
+This tutorial shows an example of how to mitigate noise on IBMQ
+backends, broken down in the following steps.
+
+[comment]: <> (-   [Setup: Defining a circuit](#Setup:-Defining-a-circuit))
+[comment]: <> (-   [High level usage](#High-level-usage))
+[comment]: <> (-   [Cirq front end](#Cirq-front-end))
+[comment]: <> (-   [Low level usage](#Low-level-usage))
+
```
rmlarose, line 24: where's the rest of the tutorial?

docs/source/examples/template.myst
```diff
@@ -3,8 +3,8 @@ jupytext:
   text_representation:
     extension: .md
     format_name: myst
-    format_version: '0.8'
-    jupytext_version: '1.4.1'
+    format_version: 0.13
+    jupytext_version: 1.11.4
```
rmlarose, line 7: Not specific to this PR, but the amount of churn from different versions is pretty annoying.

rmlarose (commented):
If binder was an amazing way to host docs, I could maybe be ok with duplicating the table of contents, and maybe maybe stomach duplicating the requirements. But binder isn't a great option imo, and the bad practices needed to make it work compound this. So I'm not the biggest fan of these changes unfortunately.

My suggestion is still to simply add a link to Colab in each notebook. It's just https://colab.research.google.com/github/unitaryfund/mitiq/blob/master/ for the TOC and appending path/to/notebook.ipynb for a specific notebook (we don't have any notebooks currently, so I can't include an actual link). It's lightweight, fast, and avoids the hacks for binder mentioned above.

andreamari:
@nathanshammah,  did you try this trick (https://github.com/jupyterhub/binder/issues/51) to avoid double requirements?  
I implemented it here: https://github.com/unitaryfund/mitiq/tree/binder-req but I am not sure how to properly do a pull request to your fork.

There should be a way to avoid double requirements. After fixing this aspect and the missing part of the IBM example, the rest LGTM.

binder/requirements.txt
```diff
@@ -0,0 +1,2 @@
+-r ../dev_requirements.txt
+.
```
andreamari, line 2:
```suggestion
.[development]
```

This is much faster (pip resolves dependencies faster) and corresponds to the command that we use in GitHub for CI.

crazy4pi314:
@andreamari If you want to check out and commit to a PR you can use https://cli.github.com/ or I use the VS Code GitHub extension. From the command line you could then run `gh pr checkout 841` and push back to this PR.

docs/source/examples/template.myst
crazy4pi314: It was an attempt to regularize, they would not need changed often. Before they were all over the place and I wanted there to be consistency in the behavior/expectations from document to document.

crazy4pi314 (approved):

docs/source/examples/ibmq-backends.myst
crazy4pi314: Fixed

binder/requirements.txt
crazy4pi314: Nice!

docs/source/examples/examples.myst
crazy4pi314: The convention specified by the service is mybinder.org, will update to this.

docs/source/examples/examples.myst
crazy4pi314: File removed, resolving

crazy4pi314:
> If binder was an amazing way to host docs, I could maybe be ok with duplicating the table of contents, and maybe maybe stomach duplicating the requirements.

I disagree, it is one of the biggest projects supported by NumFOCUS and is widely used in not only the Python ecosystem but also Ruby, R, .NET, Julia, etc. It has plugins for most major documentation building engines. The reason we have to make these modifications is technical debt/bad design on our side, not theirs.

> My suggestion is still to simply add a link to Colab in each notebook. It's just https://colab.research.google.com/github/unitaryfund/mitiq/blob/master/ for the TOC and appending path/to/notebook.ipynb for a specific notebook (we don't have any notebooks currently, so I can't include an actual link). It's lightweight, fast, and avoids the hacks for binder mentioned above.

The process you have listed here is basically identical to what we are doing with mybinder.org. The changes we need here are general cleaning of regularization of versions and extensions and adding implicit requirements that should have been explicit before. This would also require us to either configure Colab to understand the myst notebook format, or duplicate ipynb files in the repo. Also Colab is Google-proprietary, where as all of the tools/backend for mybinder.org are open source and community maintained which is more consistent with our mission.

Sorry for the infodump, but I personally am excited to have an option for folks to get started with Mitiq without any installation <3

README.md
```diff
@@ -12,7 +12,7 @@
 [![Repository](https://img.shields.io/badge/GitHub-5C5C5C.svg?logo=github)](https://github.com/unitaryfund/mitiq)
 [![Unitary Fund](https://img.shields.io/badge/Supported%20By-Unitary%20Fund-FFFF00.svg)](https://unitary.fund)
 [![Discord Chat](https://img.shields.io/badge/dynamic/json?color=blue&label=Discord&query=approximate_presence_count&suffix=%20online.&url=https%3A%2F%2Fdiscord.com%2Fapi%2Finvites%2FJqVGmpkP96%3Fwith_counts%3Dtrue)](http://discord.unitary.fund)
-
+[![Binder](https://mybinder.org/badge_logo.svg)](https://mybinder.org/v2/gh/unitaryfund/mitiq/master?filepath=%2Fdocs%2Fsource%2Fexamples)
```
andreamari, line 15:
```suggestion
[![Binder](https://mybinder.org/badge_logo.svg)](https://mybinder.org/v2/gh/unitaryfund/mitiq/master?filepath=%2Fdocs%2Fsource%2Fexamples)

```
Otherwise "Mitiq is a Python toolkit for implementing..." doesn't go to a new line.

nathanshammah:
Read the docs failed:
```
Command killed due to excessive memory consumption
Command time: 861s Return: 137
```

andreamari (commented):
I just manually converted the ibm example into a myst file using the standard Jupyter `code_cell` instead of `evaluate_rst`.     In this way the example can be opened with Jupyter. (and binder).

Plus a couple of other changes:

- Added more print statements.
- Added the option `USE_REAL_HARDWARE` such that when it is set to `False`, a classical **noisy** simulator is applied. Previously the simulator was noiseless.

Please have a look at these last changes before merging.
All the rest LGTM.

dev_requirements.txt
```diff
@@ -13,6 +13,7 @@ black~=19.10b0
 mypy~=0.910
 
 # Documentation and examples.
+jupytext
```
andreamari, line 16: This is the only change that could be related to readthedocs failng.

We could try to pin it and maybe move at at the bottom of the list (I don't know if the order could be important during the install process).

nathanshammah:
Thanks @rmlarose @crazy4pi314 and @andreamari. 

@andreamari what is failing now is readthedocs,which takes too long. If we remove jupytext I am not sure the myst notebooks will look interactively.

@rmlarose: I agree with you that Binder could be faster, and agree with @crazy4pi314 on some of the shortcomings of Colab. Andrea's hack seems nice to me. Maybe if in the future we try toping Binder to a `mitiq-nightly` on `pip`, instead of the dev requirements, that will be even faster (but without extra dependencies).

@rmlarose thanks for spotting the IBM code missing, thanks @andreamari for helping put it back. 

@rmlarose I cut the unnecessary TOC now. 

I still think that we could do a better job at engaging the community by putting all the myst ("notebook") examples in a top `mitiq/examples` folder, and using the `examples.myst` TOC to collect them from `docs/source/` with `../..`. This would make the examples list more prominent.

In another PR, I am in favor of adding a Colab link with a ipynb notebook in the future and a `!pip install ` block there, so that users do what they prefer. 

Thanks all!

crazy4pi314:
> we try pinging Binder to a mitiq-nightly on pip, instead of the dev requirements, that will be even faster (but without extra dependencies)

I am not sure how that would help? By having the backends as dev requirements we have to have all of those installed to run the docs. A nightly docker image with everything updated and pre-build would speed up the mybinder.org spinup time. Not to hard to have GH actions do this.

> I am in favor of adding a Colab link with a ipynb notebook in the future and a !pip install block there

I guess I don't see the utility in this, it adds more things for us to check/maintain and managing state like that at a notebook level can be dangerous if then folks use it locally. One notebook could resolve dependencies one way, that conflicts with another notebook that has other dependencies. We should make sure that we help users and tools manage a _global_ environment (or pipenv/conda env as appropriate). That is one of the purposes of mybinder.org, it sets up an environment where everything should work without pip installing at a notebook level. Maybe Collab can do similar specification but the proposed usecase here is using pip installs which is what creates the problem.

crazy4pi314 (approved):
lgtm

nathanshammah:
@crazy4pi314 while I like Binder, to me the added value of Colab is that it's collaborative and persistent. So as a user I'm sure that whatever tinkering I do with Mitiq there I'll be able to get back to it later. Connection timeouts or page closing don't mean risking to lose everything. So the comparison is more with BinderHub, but the login and hosting are fully managed by google (with its cons but also pros in robustness). It's also way faster in my experience when starting up a new notebook and comes in an environment with a bunch of stuff pre-installed, like matplotlib, scikit-learn etc, and offers also RAM usage. It's not far-fetched to imagine a time when also Google QPUs access will be enabled easily with the Colab infrastructure. So, to me, this is no maintenance overhead as we know that doing `!pip install mitiq` should work as it's the distributed package.

nathanshammah:
There is a final consideration here to make, I guess. Given the amount of activity on development, the fact that right now the link to Binder takes one to `master`, means that very likely every time the user is clicking on the `My Binder` link, the HEAD will not be synced with the latest image. Since we have not enabled yet the branch division for new version releases, pinging the most recently released version of Mitiq seems not ideal at the moment. So my suggestion is to add a new action that is triggered on master commit merges and that fires up a Binder. This way when a user clicks on the button, they will find the Docker image ready and waiting time should go down ~tenfold.

Issue #816, opened 2021-07-10
New technique: Post-selection of measurement results


crazy4pi314:
### Problem

- Trivial to implement (see below).
    - Technically can be framed as quantum subspace expansion with a classical code, but probably not the best to frame this way in Mitiq.
- Can significantly improve performance when applicable.

```python
from typing import Dict
 
def post_select(
    measurements: Dict[str, int], hamming_weight: int
) -> Dict[str, int]:
   return {k: v for k, v in measurements.items() if sum(map(int, k)) == hamming_weight}
```

### Rough requirements

- Requires support for measurement results.

rmlarose:
Closed by #844.

PR #844, opened 2021-08-02
Set measurement result type and add post-selection

rmlarose:
Description
-----------

This is part 2/2 of #826 except we decided to not add conversions at all and assume a simple format (for now) for measurement results. Also adds post-selection in a new readout error mitigation (REM) module, closing #816. 

Checklist
-----------

Tips
----

### DIFF
--- a/mitiq/_typing.py
+++ b/mitiq/_typing.py
@@ -57,13 +57,14 @@ QPROGRAM = Union[_Circuit, _Program, _QuantumCircuit, _BKCircuit]
 
 # Supported measurement results.
 Bitstring = List[int]
+MeasurementResult = List[Bitstring]
 
 # An `executor` function inputs a quantum program and outputs an object from
 # which expectation values can be computed. Explicitly, this object can be one
 # of the following types:
 QuantumResult = Union[
     float,  # The expectation value itself.
-    List[Bitstring],  # Sampled bitstrings.
+    MeasurementResult,  # Sampled bitstrings.
     # TODO: Support the following:
     # np.ndarray,  # Density matrix.
     # Sequence[np.ndarray],  # Wavefunctions sampled via quantum trajectories.
--- a/mitiq/_typing.py
+++ b/mitiq/_typing.py
@@ -13,15 +13,22 @@
 # You should have received a copy of the GNU General Public License
 # along with this program.  If not, see <https://www.gnu.org/licenses/>.
 
-"""Defines SUPPORTED_PROGRAM_TYPES - all supported packages / circuits which
-Mitiq can interface with - and QPROGRAM - all supported packages / circuits
-which are installed in the environment Mitiq is run in.
+"""Defines input / output types for a quantum computer (simulator):
+
+  * SUPPORTED_PROGRAM_TYPES: All supported packages / circuits which Mitiq can
+       interface with,
+  * QPROGRAM: All supported packages / circuits which are installed in the
+       environment Mitiq is run in, and
+  * QuantumResult: An object returned by a quantum computer (simulator) running
+       a quantum program from which expectation values to be mitigated can be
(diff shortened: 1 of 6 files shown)


andreamari (commented):
Thanks @rmlarose. Can we separate this in two PRs. 

(PR1) dedicated to measurement type 
(PR2) dedicated to exploring the REM module.  

I would approve PR1 immediately, but I would need to think a bit more about PR2.

The reason is that #816 looks more like more an idea to explore and to prototype rather than a module that we are ready to implement with a clear design. https://github.com/unitaryfund/mitiq/projects/

rmlarose:
Sure, probably a good idea. I thought a little about `mitiq.rem` but let's discuss as a group. Practical reason for adding here was user-driven development, and that the PR is really small without it. (And that it's an easy function to write.)

rmlarose:
Actually, I think this PR would be a good to drive the discussion about `mitiq.rem`. There's not much use in PR1 by itself because you can't do anything with measurements, so if you agree I'll just leave this open as a discussion point, and we can break it up if needed after discussing.

andreamari:
> Actually, I think this PR would be a good to drive the discussion about `mitiq.rem`. There's not much use in PR1 by itself because you can't do anything with measurements, so if you agree I'll just leave this open as a discussion point, and we can break it up if needed after discussing.

Sounds good!

rmlarose:
Part of #623.

mitiq/rem/post_select.py
```diff
@@ -0,0 +1,45 @@
...
+# along with this program.  If not, see <https://www.gnu.org/licenses/>.
+
+from mitiq._typing import MeasurementResult
+
+
+def post_select(
+    measurement_result: MeasurementResult,
+    hamming_weight: int,
+    inverted: bool = False,
+) -> MeasurementResult:
```
andreamari, line 23: We could keep the name `post_select` for something more general that we could introduce in the future. E.g. one of the arguments could be a custom function to specify what should be post-selected e.g.  `select: Callable[[MeasurementResult], bool] `. 

The present funciton could be renamed to something more specific `post_select_weight` or  `select_hamming_weight` or  etc..

mitiq/_typing.py
```diff
@@ -45,4 +52,20 @@
 except ImportError:  # pragma: no cover
     _BKCircuit = _Circuit
 
+# Supported + installed quantum programs.
 QPROGRAM = Union[_Circuit, _Program, _QuantumCircuit, _BKCircuit]
+
+# Supported measurement results.
+Bitstring = List[int]
+MeasurementResult = List[Bitstring]
```
andreamari, line 60: This is a question not a suggesiton: 

What happens if an executor returns the measurement results as an `np.ndarray`. 
Do we want to explicitly support this case with  `MeasurementResult = Union[List[Bitstring], np.ndarray]` . 
The disadvantage is that `np.ndarray` is a very ambiguous type.

andreamari (changes requested):

mitiq/_typing.py
rmlarose: It's possible, but we also want to use `np.ndarray` for wavefunctions and density matrices. I'm very close to suggesting we define data classes for everything an executor can return. This breaks the (very strict) rule of "no new objects a user wouldn't already be using," but I think it's very low overhead and would make our lives much easier. In that case `MeasurementResult` could contain `np.ndarray` as you suggest, it would just have to be wrapped as `mitiq.Bitstring(measurement_array)`.

mitiq/rem/post_select.py
rmlarose: Great suggestion!!

rmlarose: Done

andreamari (approved):
Great! LGTM

PR #846, opened 2021-08-04
update references in README

andreamari:
Currently we only cite https://arxiv.org/abs/1612.02058 as a reference for both ZNE and PEC.
I propose adding a couple more to give proper credit to the other works and to give more context to the reader.


rmlarose (approved):
Thanks! Everything is chronological except the first two ZNE references, which is a little weird. But LGTM.

PR #576, opened 2021-03-03
Fix problem of notebooks without output in Read the Docs

andreamari:
This is a possible way to fix #573 . 

Description
-----------

The current problem is that when read the docs executes a notebook, mitiq is not installed and the the root folder of the mitiq repo  is not visible from the notebook. 

This PR adds a `mitiq.py ` file in `docs/source/examples` which is fake wrap of the true `mitiq` module. In this way,
the instruction `import mitiq` in each notebook works properly during the building of the docs.

This trick was suggested here: https://mg.readthedocs.io/importing-local-python-modules-from-jupyter-notebooks/sys-path-in-helper-module/path-helper.html

@crazy4pi314, I see only now that you self-assigned #576 .  What do you think?

For a preview of the result: https://mitiq.readthedocs.io/en/try-fixing-notebooks/examples/simple_landscape.html

Checklist
-----------
Please make sure you have finished the following tasks before requesting a review of the pull request (PR). For more information, please refer to the [Contributor's Guide](../blob/master/CONTRIBUTING.md):

- [ ] Contributions to mitiq should follow the [pep8 style](https://www.python.org/dev/peps/pep-0008/). You can enforce it easily with [`black`](https://black.readthedocs.io/en/stable/index.html) style and with [`flake8`](http://flake8.pycqa.org) conventions.
- [ ] Please add tests to cover your changes, if applicable, and make sure that all new and existing tests pass locally.
- [ ] If the behavior of the code has changed or new feature has been added, please also [update the documentation](../blob/master/docs/CONTRIBUTING_DOCS.md).
- [ ] Functions and classes have useful [Google-style](https://sphinxcontrib-napoleon.readthedocs.io/en/latest/example_google.html) docstrings and [type hints](https://www.python.org/dev/peps/pep-0484/) in the signature of the objects.
- [x] (Bug fix) The associated issue is referenced above using [auto-close keywords](https://help.github.com/en/articles/closing-issues-using-keywords). If the PR fixes an issue, use the keyword fix/fixes/fixed followed by the issue id. For example, if the PR fixes issue 1184, type "Fixes #1184" (without quotes).
- [x] The [changelog](https://github.com/unitaryfund/mitiq/blob/master/CHANGELOG.md) is updated, including author and PR number (@username, gh-xxx).

Issue #573, opened 2021-03-02
Readthedocs fails to reproduce output cells of some myst notebooks


andreamari:
Read The Docs fails to execute some notebooks and just reproduces input cells without the output cells. 
Se e.g. https://mitiq.readthedocs.io/en/stable/examples/simple_landscape.html or https://mitiq.readthedocs.io/en/stable/examples/maxcut-demo.html.

It works locally, in my usual environment where `mitiq` is installed. It fails if I create a new conda environment.

The problem currently happens for `maxcut-demo.myst` and `simple_landscape.myst` and seems to be related to the
line `import mitiq`.

Indeed: 
   - `test.myst` works but
   - If I add `import mitiq` in a code cell of `test.myst`, then `make docs` fails.
   - If I execute `pip install -e .`, `make docs` works.

### Error output from local log files in docs/build/html/reports/ in a new conda environment

```
Traceback (most recent call last):
  File "/home/andrea/anaconda3/envs/py38/lib/python3.9/site-packages/jupyter_cache/executors/utils.py", line 51, in single_nb_execution
    executenb(
  File "/home/andrea/anaconda3/envs/py38/lib/python3.9/site-packages/nbclient/client.py", line 1112, in execute
    return NotebookClient(nb=nb, resources=resources, km=km, **kwargs).execute()
  File "/home/andrea/anaconda3/envs/py38/lib/python3.9/site-packages/nbclient/util.py", line 74, in wrapped
    return just_run(coro(*args, **kwargs))
  File "/home/andrea/anaconda3/envs/py38/lib/python3.9/site-packages/nbclient/util.py", line 53, in just_run
    return loop.run_until_complete(coro)
  File "/home/andrea/anaconda3/envs/py38/lib/python3.9/asyncio/base_events.py", line 642, in run_until_complete
    return future.result()
  File "/home/andrea/anaconda3/envs/py38/lib/python3.9/site-packages/nbclient/client.py", line 553, in async_execute
    await self.async_execute_cell(
  File "/home/andrea/anaconda3/envs/py38/lib/python3.9/site-packages/nbclient/client.py", line 857, in async_execute_cell
    self._check_raise_for_error(cell, exec_reply)
  File "/home/andrea/anaconda3/envs/py38/lib/python3.9/site-packages/nbclient/client.py", line 760, in _check_raise_for_error
    raise CellExecutionError.from_cell_and_msg(cell, exec_reply_content)
[...]

PR #847, opened 2021-08-04
Fix invisible output in mitiq codeblocks

### LINKED ISSUE #842: Output not shown in paper code blocks
Code output is not shown in paper code blocks on [RTD](https://mitiq.readthedocs.io/en/latest/examples/mitiq-paper/mitiq-paper-codeblocks.html).

Screen Shot 2021-07-30 at 17 59 54

Fixing the extension metadata did not work, in #840.

andreamari:
This should fix #842.
It seems same problem that we had here #573 and that we fixed here #576. 

The only difference is that the Mitiq codeblocks example is in a different folder.

### DIFF
--- a/docs/source/examples/mitiq-paper/mitiq.py
+++ b/docs/source/examples/mitiq-paper/mitiq.py
@@ -1,0 +1,23 @@
+# This is a virtual mitiq package.
+# When a notebook is executed in this folder, the line "import mitiq"
+# can be executed without errors even if mitiq is not installed via pip.
+
+# See: https://mg.readthedocs.io/importing-local-python-modules-from-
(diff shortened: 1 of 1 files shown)


nathanshammah (approved):
This LGTM as an incremental solution, thanks @andreamari. Maybe there is also another general solution with RTD (maybe @astrojuanlu knows a hack) for the future.

astrojuanlu:
I guess the "general solution" is to `pip install mitiq` (that's how I always approach it). However, from here:

https://github.com/unitaryfund/mitiq/blob/809c0ca5ed96feca9e2b7392a9ced6a587f3cb09/docs/source/examples/mitiq-paper/mitiq.py#L1-L3

I understand that you have some reason to not do it?

PR #852, opened 2021-08-06
Add `Observable` 

rmlarose:
Description
-----------

Initial support for `mitiq.Observable`, a collection of `mitiq.PauliString`s.

Part of #623.

Checklist
-----------

Tips
----


mitiq/tests/test_utils.py
```diff
@@ -31,17 +31,22 @@
...
+    ops,
     depolarize,
 )
 
 from mitiq.utils import (
+    _append_measurements,
     _are_close_dict,
     _equal,
+    _is_measurement,
+    find_all_qubits_with_terminal_measurements,
```
rmlarose, line 43: Note: I moved some utility functions in `zne.scaling` because I wanted to use them for the 1st implementation of something in `Observable`. The current implementation does not use them, but it's fine / better to move these into utils.

mitiq/observable/observable.py
```diff
@@ -0,0 +1,93 @@
...
+            for (i, plist) in enumerate(plists):
+                if all(pauli.can_be_measured_with(p) for p in plist):
+                    plists[i].append(pauli)
+                    added = True
+                    break
+
+            if not added:
+                plists.append([pauli])
+
+        return set([frozenset(plist) for plist in plists])
```
andreamari, line 62: What happens if we just return `plists`?

Is `frozenset` necessary for something? 
Is the final `set()` conversion necessary? 

Note: considering that `masure_in` returns a list of circuits, it could be more natural if partition returned a list too.

mitiq/observable/tests/test_observable.py
```diff
@@ -0,0 +1,164 @@
...
+    pauli2 = PauliString(spec="IZ")
+    pauli3 = PauliString(spec="ZZ")
+    obs = Observable(pauli1, pauli2, pauli3)
+    assert obs.nterms == 3
+
+    sets = obs.partition()
+    # expected = {frozenset([pauli1, pauli2, pauli3])}
+
+    assert len(sets) == 1
+    # assert sets == expected
```
andreamari, line 60: are there problems here?

mitiq/observable/tests/test_observable.py
```diff
@@ -0,0 +1,164 @@
...
+    y = PauliString(spec="Y")
+    z = PauliString(spec="Z")
+    obs = Observable(x, y, z)
+    assert obs.nterms == 3
+
+    sets = obs.partition()
+    # expected = {frozenset([p]) for p in (x, y, z)}
+
+    assert len(sets) == 3
+    # assert sets == expected
```
andreamari, line 74: ditto

andreamari (commented):
Hi  @rmlarose,  this looks very nice! 
My general comment is that we should probably be careful when taking the `set()` of the `PauliStrings` that compose the `Observable`. For two main reasons:

1. We should be sure that there is a well-behaving notion of equality or hash for PauliStrings. E.g. how equal Paulis with different real coefficients are compared? 
2. We should be sure that removing equal PauliString is a good idea in itself. Maybe just keeping `List` of all `PauliStrings` including repetitions can simplify the whole workflow at the cost some memory inefficiency.

mitiq/observable/observable.py
```diff
@@ -0,0 +1,93 @@
...
+        self._paulis = set(paulis)
+        self._nterms = len(self._paulis)
+
+    @property
+    def nterms(self) -> int:
+        return self._nterms
+
+    def _qubits(self) -> Set[cirq.Qid]:
+        """Returns all qubits acted on by the Observable."""
+        return {q for pauli in self._paulis for q in pauli._pauli.qubits}
```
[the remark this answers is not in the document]
rmlarose: `Observable._qubits` are the `cirq.Qid`s and so should be private. The public variable is the `Observable.qubit_indices` which is platform-agnostic.

mitiq/observable/tests/test_observable.py
rmlarose: Yes :) Thanks for reminding me to investigate

mitiq/observable/observable.py
rmlarose: Agree. I will change the return type to `List[Frozenset[PauliString]]`. This is actually essential for estimating the expectation value of the observable - you get a bunch of circuits to run then you get the measurements for each circuit, and you need them to match, but set is unordered.

FrozenSet was needed (as opposed to set) because the objects in a set need to be hashable. I think we should still keep it when changing to a list, though. 

P.S. I considered defining another class for a set of `PauliString`s which can be measured at the same time via single-qubit basis rotations. This would take the place of `FrozenSet[PauliString]`. It may be a good idea, but for now I'm just sticking to `FrozenSet[PauliString]`.

mitiq/observable/observable.py
```diff
@@ -0,0 +1,93 @@
...
+
+    @property
+    def qubit_indices(self) -> List[int]:
+        return [cast(cirq.LineQubit, q).x for q in sorted(self._qubits())]
+
+    @property
+    def nqubits(self) -> int:
+        return len(self.qubit_indices)
+
+    def partition(self) -> Set[FrozenSet[PauliString]]:
```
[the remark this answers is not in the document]
rmlarose: Agree, thanks.

mitiq/observable/observable.py
```diff
@@ -0,0 +1,93 @@
...
+
+import numpy as np
+import cirq
+
+from mitiq.observable.pauli import PauliString
+
+
+class Observable:
+    def __init__(self, *paulis: PauliString) -> None:
+        self._paulis = set(paulis)
```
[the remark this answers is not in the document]
rmlarose: Thanks. It's actually just an unordered collection with potential duplicates.

```python
{PauliString(spec="Z"), PauliString(spec="Z")}
>>> {cirq.Z(cirq.LineQubit(0)), cirq.Z(cirq.LineQubit(0))}
```

I think the correct behavior should be that duplicates are combined in such a way that, e.g., `Z + Z -> 2Z`.

mitiq/observable/observable.py
rmlarose: Actually changed to `List[List[PauliString]]` due to note about duplicates. I almost want to consider something like `Observable(PauliString("Z"), PauliString("Z"))` instead of 2 * Z user error, but it should not change the PauliStrings in such a way that they become unequal.

rmlarose:
Thanks @andreamari for the very helpful review. Ready for another look.

> * We should be sure that there is a well-behaving notion of equality or hash for `PauliString`s. E.g. how equal Paulis with different real coefficients are compared?

Done, see tests.

> * We should be sure that removing equal `PauliString`s is a good idea in itself. Maybe just keeping `List` of all `PauliString`s including repetitions can simplify the whole workflow at the cost some memory inefficiency.

Changed to `List[List[PauliString]]` for now, with a todo note to combine duplicates at some point.

mitiq/observable/tests/test_observable.py
rmlarose: Outdated, see new test.

mitiq/observable/tests/test_observable.py
rmlarose: Outdated, see new test.

mitiq/observable/observable.py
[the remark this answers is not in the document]
rmlarose: Note that in the latest changes

```python
{PauliString(spec="Z"), PauliString(spec="Z")}
>>> {cirq.Z(cirq.LineQubit(0))}
```

In `Observable` we are no longer using sets though, so we are not unintentially deleting `PauliString`s.

andreamari:
> Thanks @andreamari for the very helpful review. Ready for another look.
> 
> > * We should be sure that there is a well-behaving notion of equality or hash for `PauliString`s. E.g. how equal Paulis with different real coefficients are compared?
> 
> Done, see tests.
> 
> > * We should be sure that removing equal `PauliString`s is a good idea in itself. Maybe just keeping `List` of all `PauliString`s including repetitions can simplify the whole workflow at the cost some memory inefficiency.
> 
> Changed to `List[List[PauliString]]` for now, with a todo note to combine duplicates at some point.

Great! This new approach looks safer and makes the code much simpler to understand.
We could combine duplicates at some point as you suggested. Notice however that even with the current structure of the code, if there are two equal `PauliString`s they will end up in the same partition and so they won't be physically measured twice. The number of physical measurements is the same, there is just a small classical overhead.

andreamari (approved):

PR #855, opened 2021-08-13
[Test] Binder 2

nathanshammah:
Testing hacks to make #841 work.


nathanshammah:
Ok somehow this is still failing. I will start afresh in a new PR hand-picking changes from #841.

Issue #839, opened 2021-07-26
Issues with dev install on windows fixed by unpinning numpy & scipy in requirements


rmlarose:
One of our contributors on windows got a wild lapack error from `pip install -e .` (or just `pip install -r requirements.txt`). We were able to fix it by changing `requirements.txt` to just be cirq~=0.10 (i.e., removing numpy and scipy). This could be a one-off, but may be general. It might be an indication to relax the aggressive pinning in our requirements.

OS: Windows
Python version 3.8 & 3.9

nathanshammah:
I'm generally against unpinning versions as I think it reduces reproducibility in the long run.

crazy4pi314:
IMO the philosophy on this that makes sense to me is that we only add things to the `requirements.txt` that we directly use. If we had something that returned a plot from a pandas table we would add pandas, and rely on it to bring in a correct version of matplotlib. AFAIK Lapack is a native numeric library that was likely coming in from numpy. I usually use conda oto avoid things like this (native dependencies not getting pulled in) as it can better manage dependancies not in Python.

I do agree with @nathanshammah's point on this that for reproducibility we really should pin everything. This is generally done with lock files in other languages but there isn't a great option for Python. There you would have your requirements file, and at build time the compiler would generate a lock file (if there wasn't one already). I was looking into this already for speeding up testing in CI. I found [this guide](https://realpython.com/pipenv-guide/#the-pipfilelock) helpful for the pipenv tooling, and it's pretty straight forward to do with conda (just export the env after building). Probably we will have to go one of these ways at some point, just to make things more reproducible but also easier for us to manage.

dbanty:
I currently can't install this package with Python 3.9 because the scipy that's required (1.4.1) doesn't have a wheel for 3.9 and building from source fails with the latest setuptools.

Does `mitiq` actually not support scipy > 1.4.1, or can the requirements be loosened to something like ~=1.4 so new non-major versions are supported?

crazy4pi314:
@dbanty Thanks for the feedback on this! It seems like we have some issues in version specs in the requirements files when it comes to versions of those packages for windows. I am also on windows, and am seeing if I can update the file now!

crazy4pi314:
FWIW I was able to reproduce this, as it looks like the versions we pinned for numpy/scipy were both pretty out of date.

PR #871, opened 2021-08-25
Updating pinned Scipy version

crazy4pi314:
Description
-----------
Addressing #839 and related install issues.

Checklist
-----------


andreamari:
@crazy4pi314 ,  this is not an important test and I would not consider it as a blocker.   I suggest the following options:

- Try replacing all the exponents `e9` in the variable `initial_guess` with something bigger `e10` or `e11`. 
- If the previous option doesn't work, you can mark it with the pytest option "skip" and I will think how to fix/replace this test without blocking this PR.

crazy4pi314:
> * Try replacing all the exponents `e9` in the variable `initial_guess` with something bigger `e10` or `e11`.
I tried doing this but it didn't seem to help
> * If the previous option doesn't work, you can mark it with the pytest option "skip" and I will think how to fix/replace this test without blocking this PR.
Did this, it seems that `scipy.minimize` has slightly different behavior that is preventing it from failing like it did before.  Might be a change to a default method? Taking a quick look now..

crazy4pi314:
Couldn't find a delta for the behavior for that function, we can troubleshoot this later then 👍

nathanshammah (approved):
LGTM, thank you @crazy4pi314.

PR #894, opened 2021-09-02
Fix html rendering problems of the PEC example

andreamari:
Description
-----------
The [current html rendering](https://mitiq.readthedocs.io/en/latest/examples/pec-tutorial.html) of the PEC example has some problems (not appearing in Jupyter):

- Some latex equations are bad visualized. 
- The link to sub-sections don't work in the html page.

This PR should fix both problems.
Preview the result after the changes of this PR: https://mitiq.readthedocs.io/en/fix-pec-example/examples/pec-tutorial.html


crazy4pi314:
I saw that there was some errors in the rendering with the link checker and not getting to http versions of sites, so I updated them all to the correct https links. Let's see if that helps the build get farther :)

crazy4pi314:
So after bit more investigation (sorry if I am duplicating, I just wanted to unblock other builds) it seems like some part of doi.org's handoff to IOP journals has a bad ssl certificate. Not much we can do ourselves other than manually ignore the links, or swap the doi ones for the direct journal or arxiv links. Preferences?

andreamari:
I tried with a direct link to IOP journals, but the problem remains. So it seems to be a problem of IOP journals and not of doi.org.

Since the links actually work if one clicks on them and are not critical, I propose that we simply ignore these two links when calling `linkcheck`. There is an option for this, that I used in the last commit.

crazy4pi314 (approved):
LGTM, a reasonable plan till IOP sorts out their website ssl certificate.

PR #901, opened 2021-09-03
Improve OperationRepresentation printing

andreamari:
This is not an urgent PR but I am opening it just to keep track of an old branch where I tried to fix this TODO:

https://github.com/unitaryfund/mitiq/blob/cd474fd0be42b3e0135cb2a9d94cd53ce109005e/mitiq/pec/types/types.py#L516

For single qubit gates: add parenthesis to better separate the real coefficients from the circuits and the qubit numbers. 
For multi-qubit gates: print the linear combination of circuits in a vertical way.

Single-qubit example:
```
>>> print(single_qubit_representation):
0: ───H─── = 0.500*(0: ───H───)+0.250*(0: ───H───X───)+0.250*(0: ───H───Y───)
```

Multi-qubit example:
```
>>> print(two_qubit_representation):
0: ───@───
      │
1: ───X─── =

0.500
0: ───H───@───H───
          │
1: ───H───X───H───

+0.500
0: ───Z───@───Z───
          │
1: ───Z───X───Z───
```


andreamari:
This is ready for review. Not urgent.

nathanshammah (approved):
LGTM, better than before. I would find it a bit more intuitive to include also the * next to the parenthesis.

PR #904, opened 2021-09-06
Add `Observable.expectation` with support for measurement results and density matrices; Refactor `Collector` -> `Executor`

rmlarose:
Description
-----------

Part of #623.

Checklist
-----------

Tips
----

### DIFF
--- a/mitiq/observable/observable.py
+++ b/mitiq/observable/observable.py
@@ -82,7 +82,11 @@ class Observable:
     def measure_in(self, circuit: QPROGRAM) -> List[QPROGRAM]:
         return [pset.measure_in(circuit) for pset in self._groups]
 
-    def matrix(self, qubit_indices: Optional[List[int]] = None, dtype: type = np.complex128) -> np.ndarray:
+    def matrix(
+        self,
+        qubit_indices: Optional[List[int]] = None,
+        dtype: type = np.complex128,
+    ) -> np.ndarray:
         """Returns the (potentially very large) matrix of the Observable."""
         if qubit_indices is None:
             qubit_indices = self.qubit_indices
--- a/mitiq/observable/observable.py
+++ b/mitiq/observable/observable.py
@@ -120,7 +120,11 @@ class Observable:
 
             return np.trace(density_matrix @ self.matrix())
         else:
-            raise NotImplementedError
+            raise ValueError(
+                f"Arg `execute` must be a function with annotated return type "
+                f"that is either mitiq.MeasurementResult or np.ndarray but "
+                f"was {result_type}."
+            )
 
     def _expectation_from_measurements(
         self, measurements: List[MeasurementResult]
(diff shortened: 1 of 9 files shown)


mitiq/observable/tests/test_observable.py
```diff
@@ -32,6 +34,25 @@
...
+
+    n = sum(result._measurement_shape()[1].values())
+    bitstrings = [
+        list(map(int, np.binary_repr(value[0], width=n)))
+        for value in result.data.to_numpy()
+    ]
+    return MeasurementResult(
+        bitstrings,
+        qubit_indices=tuple(int(i) for i in result.data.columns[0].split(",")),
+    )
```
andreamari, line 49: Not important for this test but, just to sketch the typical workflow that  the user should apply, we could find something simpler.

For example, I find `result.measurements` more intuitive with respect to `result.data`, since the output is very similar to the input of `MeasurementResult`.

The following should work, even in the case of multiple measurements:

```suggestion
    return MeasurementResult(
        result=np.column_stack(list(result.measurements.values())),
        qubit_indices=tuple(
            int(q) for k in result.measurements.keys() for q in k.split(",")
        ),
    )
```

mitiq/observable/observable.py
```diff
@@ -94,16 +96,21 @@ def matrix(self, dtype: type = np.complex128) -> np.ndarray:
...
+            to_run = self.measure_in(circuit)
+            results = Executor(execute).run(to_run)
+            return self._expectation_from_measurements(
+                cast(List[MeasurementResult], results)
+            )
+        elif result_type is np.ndarray:
+            density_matrix = cast(np.ndarray, execute(circuit))
+            return np.trace(density_matrix @ self.matrix())
+        else:
+            raise NotImplementedError
```
andreamari, line 113: Here we could rise a a more explicit error since this can happen often. 
For example: `"The output type of "execute" must be ... or ... but is  ....  ."`

E.g. the user may try the standard old-style executor that returns a float or may just forget annotation.

mitiq/observable/tests/test_observable.py
rmlarose: Thanks! Much simpler, went with this.

mitiq/observable/observable.py
rmlarose: Added.

rmlarose:
Thanks @andreamari. In addition to the above comments I added a partial trace to the density matrix if the observable is on fewer qubits than the ansatz circuit, and added the density matrix executor to the last test which measures, e.g., <X1> on a three-qubit state.

mitiq/observable/observable.py
rmlarose: Note that support for batched versions, e.g. `List[mitiq.MeasurementResult]`, need to be added. Meant to add a TODO comment but forgot.

andreamari (approved):

PR #936, opened 2021-09-20
Updating release process

crazy4pi314:
Description
-----------
We have had some release git history issues lately, so this PR modifies the process to make it simpler and hopefully easier. 
tl;dr when you want to make a release, update the changelog and version.txt on master, tag that commit with `v0.xx.xx` and the 
GH release will be drafted for you. After approving that, go to the [actions page](https://github.com/unitaryfund/mitiq/actions/workflows/publish-pypi.yml) and there should be a manual trigger for the pypi deployment, you just manually provide the correct version number.

Checklist
-----------


.github/workflows/gh-release.yml
```diff
@@ -1,10 +1,7 @@
 on:
   push:
-    branches: # specify branch to Mitiq release, not running from forks
-      - release
-    # Sequence of patterns matched against refs/tags
     tags:
-      - 'v*' # Push events to matching v*, i.e. v1.0, v20.15.10
+      - 'v[0-9].[0-9]+.[0-9]+' # Push events to matching sematic versioning, i.e. v1.0, v20.15.10
```
nathanshammah, line 4: great improvement.

.github/workflows/publish-pypi.yml
```diff
@@ -4,10 +4,11 @@
...
 
 on:
-  release:
-    types: [published]
   workflow_dispatch:
-    inputs: {}
+    inputs:
+      release_version:
+        description: 'The semantic version number of the release (v[0-9].[0-9]+.[0-9]+)'
+        required: true
```
nathanshammah, line 11: does `true` work if it's lower case?

nathanshammah (approved):
Looks good to me, thank you @crazy4pi314. I wonder if this time the pypi upload will work automatically or not, let's see.

docs/source/release.rst
```diff
@@ -142,6 +126,23 @@ In case the action for the automatic release on PyPI fails, the commands to rele
...
+    a new release.
+
+------------------------------------------------
+Update the new development version
+------------------------------------------------
+
+Add a new section to the ``CHANGELOG.md`` to track changes in the following
+release, meaning that if ``vX.Y.Z`` was just released, then there should be
+a section for ``vX.(Y+1).Z`` that is marked "In Development". Also, change the
+version in the ``VERSION.txt`` file from ``vX.Y.Z`` to ``vX.(Y+1).Zdev``.
```
andreamari, line 144:
```suggestion
a section for ``vX.(Y+1).0`` that is marked "In Development". Also, change the
version in the ``VERSION.txt`` file from ``vX.Y.Z`` to ``vX.(Y+1).0dev``.
```
I think the third number should go to 0. If I am wrong, please ignore this suggestion.

docs/source/release.rst
```diff
@@ -73,11 +70,15 @@ follow SemVer, so typically a release will involve changing the version from
...
+Do the release
+--------------
+
+^^^^^^^^^^^^^^^^
 Create a new tag
-----------------
+^^^^^^^^^^^^^^^^
 
-Once the PR to ``release`` is approved, tag the new commit on release
+Once the PR to ``release`` is approved, tag the new commit on master
```
andreamari, line 81: If we do a PR from `master` to `release`, I think the new commit will be on `release`.
So which commit should be tagged on `master`? The last one before merging to `release`?

.github/workflows/publish-pypi.yml
crazy4pi314: Not exactly sure I understand, but it is using a regex to check the version number so it must be a lower case v. The provided version number is always required (until I find a better way for this action to scrape it).

docs/source/release.rst
crazy4pi314: You are def right! fixing in next PR

docs/source/release.rst
crazy4pi314: I fixed this/ made it more clear, thanks!

PR #939, opened 2021-09-22
Foldability check includes a check for inverse

### LINKED ISSUE #911: Better check for fold-ability / support for parameterized circuits
Right now we say that if a circuit doesn't have a unitary, it has non-unitary gates which are not terminal measurements and therefore cannot be folded.

https://github.com/unitaryfund/mitiq/blob/31f10c252b7015a7415e450c9347f19d5b9fe400/mitiq/zne/scaling/folding.py#L69-L73

A Cirq circuit with symbols doesn't have a unitary but does have an inverse:

```python
import sympy
import cirq


theta = sympy.Symbol("theta")
q = cirq.LineQubit(0)
ansatz = cirq.Circuit(cirq.ry(theta).on(q))
print(ansatz)  # 0: ───Ry(theta)───

cirq.has_unitary(ansatz)  # False
cirq.inverse(ansatz)  # 0: ───Ry(-theta)───
```

So a better check is looking for an inverse rather than looking for a unitary.

## Current behavior

`zne.scaling.fold_*(ansatz)` raises `UnfoldableCircuitError`.

## Desired behavior

Ability to fold parameterized circuits. For the above circuit, zne.scaling.fold_*(ansatz, scale_factor=3) should return `0: ───Ry(theta)───Ry(-theta)───Ry(theta)───`.

purva-thakre:
Description
-----------
Fixes #911 

- [ ] For a Cirq circuit, before an error is raised, checks if the circuit operations have valid inverse. 
- [ ] Add tests for Cirq - When there is a valid inverse, non-valid inverse (measurements)
- [ ] Braket inverse check + tests
- [ ] Pyquil inverse check + tests
- [ ] Qiskit inverse check + tests

### DIFF
--- a/mitiq/zne/scaling/folding.py
+++ b/mitiq/zne/scaling/folding.py
@@ -67,10 +67,11 @@ def _check_foldable(circuit: Circuit) -> None:
         )
 
     if not has_unitary(circuit):
-        raise UnfoldableCircuitError(
-            "Circuit contains non-unitary channels which are not terminal "
-            "measurements and cannot be folded."
-        )
+        if inverse(circuit, None) is None:
+            raise UnfoldableCircuitError(
+                "Circuit contains non-invertible channels which are not"
+                "terminal measurements and cannot be folded."
+            )
 
 
 def _squash_moments(circuit: Circuit) -> Circuit:
--- a/mitiq/zne/scaling/tests/test_folding.py
+++ b/mitiq/zne/scaling/tests/test_folding.py
@@ -407,14 +407,14 @@ def test_fold_with_intermediate_measurements_raises_error(fold_method):
 )
 def test_fold_with_channels_raises_error(fold_method):
     """Tests local folding functions raise an error on circuits with
-    non-unitary channels (which are not measurements).
+    non-invertible channels (which are not measurements).
     """
     qbit = LineQubit(0)
     circ = Circuit(
         ops.H.on(qbit), ops.depolarize(p=0.1).on(qbit), ops.measure(qbit)
     )
     with pytest.raises(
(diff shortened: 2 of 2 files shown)


mitiq/zne/scaling/folding.py
```diff
@@ -67,10 +67,11 @@ def _check_foldable(circuit: Circuit) -> None:
         )
 
     if not has_unitary(circuit):
-        raise UnfoldableCircuitError(
-            "Circuit contains non-unitary channels which are not terminal "
-            "measurements and cannot be folded."
-        )
+        if inverse(circuit, None) is None:
```
andreamari, line 70:
```suggestion
        if inverse(circuit, default=None) is None:
```
Just to be more explicit and readable.
I didn't know `inverse` had a default option, nice!

andreamari (approved):
Thanks @purva-thakre, this LGTM!   I am approving since I think it is better to support other frontends in a different PR (see last comment of #911).

I would wait for a review of @rmlarose, for a second opinion.

rmlarose (changes requested):
Thanks @purva-thakre. `has_unitary` ignores terminal measurements, but the `inverse` check does not. If you change the added test to `ansatz_circ = Circuit(ry(theta).on(q), ops.measure(q))` this will fail, but we want to support this.

To fix this, you could ignore terminal measurements when checking for an inverse. Something like:

```python
if inverse(circuit_with_terminal_measurements_removed, default=None) is None:
```

purva-thakre:
> has_unitary ignores terminal measurements, but the inverse check does not

@rmlarose Thanks for pointing this out !! I checked non-terminal measurement but forgot to check the scenario of a terminal measurement.

PR #940, opened 2021-09-23
Fix AWS example by fixing a bug in mirror circuits

### LINKED ISSUE #937: Flaky behavior with ZNE Braket example
Sometimes the final plot looks good: (https://mitiq.readthedocs.io/en/latest/examples/braket_mirror_circuit.html)

image

Other times its nonsensical: (https://mitiq.readthedocs.io/en/stable/examples/braket_mirror_circuit.html)

image

I believe this is due to how the noisy simulation is done with braket.

## Desired behavior

Having the example (https://github.com/unitaryfund/mitiq/blob/master/docs/source/examples/braket_mirror_circuit.md) reliably produce the first plot.

andreamari:
The qubit order of the correct bitstring in mirror circuits is non-deterministic.
This PR fixes this bug and, as a consequence, fixes #937 too (Braket example).

### DIFF
--- a/mitiq/benchmarks/mirror_circuits.py
+++ b/mitiq/benchmarks/mirror_circuits.py
@@ -205,7 +205,9 @@ def generate_mirror_circuit(
     )
 
     # Compute the bitstring this circuit should sample.
-    res = cirq.Simulator().run(circuit + cirq.measure(*circuit.all_qubits()))
+    res = cirq.Simulator().run(
+        circuit + cirq.measure(*sorted(circuit.all_qubits()))
+    )
     bitstring = list(res.measurements.values())[0][0].tolist()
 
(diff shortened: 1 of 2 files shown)


rmlarose (approved):
`cirq.Result` - confusing me since 1990. Great catch, thanks @andreamari!

natestemen:
Just discovered this bit in our code and was wondering why the qubits were sorted!

It seems to me anytime you want to measure all the qubits you should sort them, according to this. Is that right? Or is there a case where it might not matter?

andreamari:
I have little memory about this PR :-) 
But it seems that qubits ordering in measurement results can be quite unpredictable. 
See e.g. https://github.com/quantumlib/Cirq/issues/4144

PR #948, opened 2021-09-26
ZNE with `Observable`s

rmlarose:
Description
-----------

Part of #623. Example usage:

```python
import functools

import cirq
from mitiq import PauliString, Observable, zne

observable = Observable(PauliString(spec="XX", coeff=-1.21), PauliString(spec="ZZ", coeff=0.7))
circuit = cirq.Circuit(cirq.H.on(cirq.LineQubit(0)), cirq.CNOT.on(*cirq.LineQubit.range(2)))

print(f"Problem: ⟨𝜓| {observable} |𝜓⟩ where |𝜓⟩ is given by", circuit, sep="\n\n")

print("\nTrue value:", observable.expectation(circuit, functools.partial(mitiq_cirq.compute_density_matrix, noise_level=(0,))))
print("Noisy value:", observable.expectation(circuit, mitiq_cirq.sample_bitstrings))

zne_value = zne.execute_with_zne(
    circuit,
    executor=mitiq_cirq.sample_bitstrings,
    observable=observable,
)
print("ZNE value:", zne_value)
```

Sample output:

```text
Problem: ⟨𝜓| (-1.21+0j)*X(0)*X(1) + (0.7+0j)*Z(0)*Z(1) |𝜓⟩ where |𝜓⟩ is given by

0: ───H───@───
          │
1: ───────X───

True value: (-0.5099999696016311+0j)
Noisy value: (-0.4831689453125001+0j)
ZNE value: (-0.5143603515624993+0j)
```

Checklist
-----------

Tips
----


mitiq/zne/inference.py
```diff
@@ -541,19 +558,27 @@ def run(
...
-                res = executor(to_run)
-            else:
-                res = executor(to_run, kwargs_list=kwargs_list)
-        else:
+        if observable is not None:
             res = [
-                executor(circ, **kwargs)  # type: ignore
-                for circ, kwargs in zip(to_run, kwargs_list)
+                observable.expectation(circuit, executor) for circuit in to_run
             ]
```
andreamari, line 564: In this way, in the case in which the `observable` argument is used, `kwargs_list` is ignored. 
If this is a temporary solution it's fine since `kwargs_list` is rarely used.

If instead we want to support it I guess we need to change `observable.expectation` to allow for *kwargs.
Alternatively,  things would be easy if we had an `observable.expectation_from_quantum_result` method which could be handy also in other contexts. Basically it would represent the concept of "trace of rho with O".

Anyway, if it's too complicated I would not block this PR.

rmlarose:
Any ideas why docs build is failing?

mitiq/zne/inference.py
```diff
@@ -739,10 +771,24 @@ def scale_factor_to_expectation_value(
         ) -> float:
             """Evaluates the quantum expectation value for a given
             scale_factor and other executor parameters."""
-            expectation_values = []
+            expectation_values: List[QuantumResult] = []
```
andreamari, line 774: If I am not wrong:
```suggestion
            expectation_values: List[float] = []
```

andreamari (commented):
Thanks @rmlarose, in general it looks good but I have some small suggestions and a doubt about the handling of `kwargs_list`.

andreamari:
> Any ideas why docs build is failing?

Its seems related with 3 notebooks:
```
WARNING: Execution Failed: /home/runner/work/mitiq/mitiq/docs/source/examples/vqe-pyquil-demo.myst
WARNING: Execution Failed: /home/runner/work/mitiq/mitiq/docs/source/examples/pyquil_demo.myst
WARNING: Execution Failed: /home/runner/work/mitiq/mitiq/docs/source/examples/maxcut-demo.myst
```
The problem is that `execute_with_zne` is called with positional arguments and the order of arguments changed because observable is now introduced in the middle of previous arguments. This is technically a breaking change.

Since `observable` is supposed to be `Optional`, we could consider pushing it at later (if not last) position.

rmlarose:
> > Any ideas why docs build is failing?

> The problem is that `execute_with_zne` is called with positional arguments and the order of arguments changed because observable is now introduced in the middle of previous arguments. This is technically a breaking change.

Ahah! Thanks. I wish there was an actual error message for the docs build besides "Execution failed"...

> Since `observable` is supposed to be `Optional`, we could consider pushing it at later (if not last) position.

Let me think about this. It conceptually goes with `executor` so makes sense to come after this, and I think arguments common to all `execute_with_technique` functions (i.e., `circuit`, `executor`, `observable`) should come before optional arguments specific to the particular technique. Fine to have a breaking change if it's an improvement and clearly communicated in the release notes.

mitiq/interface/mitiq_cirq/cirq_utils.py
```diff
@@ -32,6 +33,37 @@
 ]
 
 
+# Executors.
+def sample_bitstrings(
+    circuit: cirq.Circuit,
+    noise_model: cirq.NOISE_MODEL_LIKE = cirq.amplitude_damp,  # type: ignore
+    noise_level: Tuple[float] = (0.01,),
```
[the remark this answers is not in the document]
rmlarose: Something like `cirq.asymmetric_depolarize` has three arguments - `p_x`, `p_y`, and `p_z`. `noise_level` as a `Tuple[float]` instead of just a `float` is to allow for this.

mitiq/__init__.py
```diff
@@ -13,6 +13,7 @@
 # You should have received a copy of the GNU General Public License
 # along with this program.  If not, see <https://www.gnu.org/licenses/>.
 
+from mitiq.observable import PauliString, Observable
 from mitiq._typing import SUPPORTED_PROGRAM_TYPES, QPROGRAM
```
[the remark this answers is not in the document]
rmlarose: Our imports are a disaster - I'll be opening another issue with a suggestion for this. For now I just wanted to be able to do `from mitiq import PauliString, Observable, zne` for the example usage in https://github.com/unitaryfund/mitiq/pull/948#issue-742963243.

mitiq/zne/inference.py
```diff
@@ -305,6 +311,10 @@ def run(
             qp: Quantum circuit to scale noise in.
             executor: Function which inputs a (list of) quantum circuits and
                 outputs a (list of) expectation values.
```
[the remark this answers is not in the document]
rmlarose: Done, thanks.

mitiq/zne/inference.py
rmlarose: Done (see change below). `executor` returns `QuantumResult`, hence the original. It must return `float` if `observable` is none though, so needed to tell mypy that.

mitiq/zne/inference.py
rmlarose: This gets a little subtle, see #952. It's worth discussing that issue separately from this so as to not settle on something suboptimal to get this merged - it's really about the behavior of `Executor.run` and not so much about ZNE. I added a TODO note with a link to this issue. Let me know what you think.

andreamari (approved):

Issue #908, opened 2021-09-08
Improve documentation organization


crazy4pi314:
### Problem

The current docs organization is not ideal based on user surveys, and the dev team has gotten good feedback on what could help the make the documentation more usable. Additions to the docs are minimal in this issue (wanting to leave additional content to further issues), and we focus here on the general flow and organization of the documentation.

### Rough requirements

- Stronger focus on easy to find and clear descriptions of the project and techniques
- With #906 starting with a "quick start" section to get folks up and going
- Clear and separate documentation sections for each mitigation method

A possible high-level TOC could look something like this:
- Readme/intro page
  - Notes about where Mitiq can help, and file an issue if its not working for you 
- Quick start for each method and backend
- EM Methods
  - ZNE, PEC, CDR, etc.
  - Theory relevant to each method/relevant research
- Lower level component documentation: Factories, Folding methods, etc. that are common building blocks to multiple methods
- API docs

rmlarose:
RFC for this here: https://docs.google.com/document/d/13un5TZPknSOhmOBkrL2rsofjGfdp2jDnd-DywLpGFPc/edit?usp=sharing

andreamari:
@rmlarose added some feedback on the RFC

crazy4pi314:
Also added some comments, thanks everyone!

nathanshammah:
Closed by #1021 #1004 #1031.

PR #953, opened 2021-09-28
Add workflow figures and technique descriptions

### LINKED ISSUE #944: Add workflow figures to document techniques
It would be nice to add workflow figures to document techniques, taken or adapted from the Mitiq paper. https://arxiv.org/pdf/2009.04417.pdf. This is related to #908 as it would give more immediate information on the techniques in Mitiq.

nathanshammah:
Close #944.
- Add workflow figures and their descriptions for ZNE, PEC and CDR. 
- Added section on CDR to be able to include the figure

By the way, I did not manage to add captions. The `:caption:` directive does not work.


docs/source/guide/guide-cdr.rst
```diff
@@ -0,0 +1,27 @@
+.. cdr:
+
+************************
+Clifford Data Regression
+************************
+
+Clifford Data Regression (CDR) is a learning-based quantum error mitigation technique. It is based on the
```
rmlarose, line 7: ?

rmlarose (commented):
Which guide (1-5) does `guide-cdr.rst` fit into w.r.t. the docs reorganization RFC in https://github.com/unitaryfund/mitiq/issues/908#issuecomment-926790261? 

IMO the images should go in 1. ("How do I use this technique?"), but `guide-cdr.rst` doesn't explain how to use CDR.

nathanshammah:
> Which guide (1-5) does `guide-cdr.rst` fit into w.r.t. the docs reorganization RFC in [#908 (comment)](https://github.com/unitaryfund/mitiq/issues/908#issuecomment-926790261)?
> 
> IMO the images should go in 1. ("How do I use this technique?"), but `guide-cdr.rst` doesn't explain how to use CDR.

@rmlarose this PR is only involved with adding the workflow and technique figures. Especially for CDR, this is a minimal addition given that @L-P-B is working on the full theory. But to be able to add the figure I had to create the relative section.  The order of 1-5 is not relevant here.

rmlarose:
> But to be able to add the figure I had to create the relative section. The order of 1-5 is not relevant here.

I don't see how the documentation reorganization we discussed for two weeks is not relevant. Maybe restructure one of the technique docs according to the RFC and add one of the figures there - probably in section 4: "What happens when I use this technique?".

nathanshammah:
Thanks @andreamari. It took me a bit to work out how to link to a myst file from rst...

andreamari:
Thanks @nathanshammah, I fixed a link in the last commit and all the rest is good for me.

@rmlarose: it would be good to have this in the 0.11 stable release. Too late?

nathanshammah:
@andreamari I worked out the relative link fix, using a label and reference that works between rst and myst. It should be fine. I now just need to do the diff with your latest commit.

nathanshammah:
The test failing on my latest commit is not due to the changes in this PR but a random value in the ZNE code.

nathanshammah:
I rerun the CI and now tests pass.

PR #1004, opened 2021-10-27
PEC docs

andreamari:
Drafting and testing PEC docs according to new structure based on the recent [docs RFC](https://docs.google.com/document/d/13un5TZPknSOhmOBkrL2rsofjGfdp2jDnd-DywLpGFPc/edit?usp=sharing).

Sections ready for review:
- [x] How do I use PEC?
- [x] When should I use PEC? Advantages and disadvantages are presented without an explicit use-case example. The use-case example is a quite non-trivial addition and it may be a good idea to have a new issue + a new PR dedicated to it.
- [x] What additional options are available in PEC?
- [x] What happens when I use PEC?
- [x] What is the theory behind PEC?


nathanshammah:
Re-running CI to see if the `linkcheck` error is just a blip, as sometimes happens.

docs/source/guide/pec-1-intro.myst
```diff
@@ -0,0 +1,139 @@
...
+Once the necessary `OperationRepresentation`s have been defined, probabilistic error cancellation can be applied as follows:
+
+```{code-cell} ipython3
+from mitiq import pec
+
+pec_value = pec.execute_with_pec(circuit, executor, representations=reps)
+
+print(f"Error without PEC: {abs(ideal_value - noisy_value) :.5f}")
+print(f"Error with PEC:    {abs(ideal_value - pec_value) :.5f}")
+```
```
nathanshammah, line 139:
```suggestion
print(f"Error with PEC:    {abs(ideal_value - pec_value) :.5f}")
``` In this example, the noise has been reduced of a factor 30 with respect to the unmitigated case. 
```

nathanshammah (changes requested):
I reviewed the first file

nathanshammah:
I left a review only for the first file. By the way it seems that [roles](https://myst-parser.readthedocs.io/en/latest/syntax/syntax.html#syntax-roles) in myst are rendered as
```
{func}`func_name`
```

instead of (rst)
```
:func:`func_name`
```

docs/source/guide/pec-1-intro.myst
```diff
@@ -0,0 +1,139 @@
...
+```{code-cell} ipython3
+from mitiq import benchmarks
+
+circuit = benchmarks.generate_rb_circuits(
+  n_qubits=1, num_cliffords=2, return_type = frontend,
+)[0]
+
+print(circuit)
+```
+
```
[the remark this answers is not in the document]
andreamari: Applied here: ba945aa

docs/source/guide/pec-1-intro.myst
```diff
@@ -0,0 +1,139 @@
...
+The aim of PEC is to reduce this error.
+
++++
+
+## Represent ideal gates as linear combinations of noisy gates
+
++++
+
+Before applying PEC, we need to represent the gates of the ideal `circuit` as linear combinations 
+of noisy gates. In Mitiq, one can express a gate representation via an `OperationRepresentation`
```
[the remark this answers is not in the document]
andreamari: I agree! Applied here: ba945aa

docs/source/guide/pec-1-intro.myst
andreamari: The results are non-deterministic so I rephrased it without giving a number for the error reduction factor.
See here: ba945aa

docs/source/guide/pec-5-theory.myst
```diff
@@ -0,0 +1,156 @@
...
+kernelspec:
+  display_name: Python 3
+  language: python
+  name: python3
+---
+
+# What is the theory behind PEC?
+
+
+Probabilistic error cancellation (PEC) [[1-3]](#references) is a noise-aware error mitigation technique which is
```
nathanshammah, line 17: Here and elsewhere in the file: what about using the common References section at the end, with  `{cite}`ref1`` format?

docs/source/guide/pec-5-theory.myst
```diff
@@ -0,0 +1,156 @@
...
+All the noisy expectation values $\langle A_{\vec{\alpha}}\rangle_{\rm noisy}$ can be directly measured with 
+a noisy backend since they only require circuits composed of implementable noisy operations.
+In principle, by combining all the noisy expectation values, one could compute the ideal result $\langle A \rangle_{\rm ideal}$.
+Unfortunately this approach requires the execution of an exponentially large number of circuits and 
+which is typically unfeasible. 
+
+An important fact at the basis of PEC is that, for weak noise, only a small number of noisy expectation values actually
+contribute to the linear combination because many of the coefficients $\eta_{\vec \alpha}$ are negligible.
+For this reason, it is more efficient to estimate $\langle A \rangle_{\rm ideal}$ using an
+ ([importance-sampling](https://en.wikipedia.org/wiki/Importance_sampling)) Monte Carlo approach as described in the next section.
```
nathanshammah, line 101:
```suggestion
 [importance-sampling](https://en.wikipedia.org/wiki/Importance_sampling) Monte Carlo approach as described in the next section.
```

nathanshammah: just a suggestion. The "an" read strange without the parenthesis. Or maybe could be a(n)

docs/source/guide/pec-5-theory.myst
```diff
@@ -0,0 +1,156 @@
...
+This can be achieved as follows:
+
+$$ \mathcal{G_i} = \sum_{\alpha} \eta_{i, \alpha} \mathcal{O}_{i, \alpha} 
+= \gamma_i \sum_{\alpha} p_i(\alpha) \, {\rm sgn}(\eta_{i, \alpha})\, \mathcal{O}_{i, \alpha},$$
+
+where $p_{i}(\alpha)=|\eta_{i, \alpha}|/\gamma_i$ is a valid probability distribution with respect to $\alpha$.
+
+If for each gate $\mathcal G_i$ of the circuit we sample a value of $\alpha$ from $p_{i}(\alpha)$ and we apply the corresponding noisy operation
+$\mathcal O_{i, \alpha}$, we are effectively sampling a noisy circuit $\Phi_{\vec{\alpha}}$ from the
+global probability distribution $p(\vec{\alpha})= |\eta_{\vec{\alpha}}| / \gamma$.
```
nathanshammah, line 115: could define \vec{alpha} more explicitly, maybe. In the previous subsection.

docs/source/guide/pec-5-theory.myst
```diff
@@ -0,0 +1,156 @@
...
+
+$$\langle A \rangle_{\text{ideal}} = \gamma\, 
+\mathbb E \left\{  {\rm sgn}(\eta_{\vec{\alpha}}) \langle A_{\vec{\alpha}}\rangle_{\rm noisy} \right\}.$$
+
+By averaging a finite number $N$ of samples we obtain an unbiased estimate of $\langle A \rangle_{\text{ideal}}$.
+Assuming a bounded observable $|A|\le 1$, the number of samples $N$
+necessary to approximate $\langle A\rangle_{\text{ideal}}$ within an absolute error $\delta$,
+scales as [[5]](#references):
+
+$$ N \propto \frac{\gamma^2}{\delta^2}. $$
```
nathanshammah, line 133: Maybe could add a final comment on this scaling.

nathanshammah (commented):
Added comments for file 5 with the theory.

nathanshammah (changes requested):
Added minor comments also for file 3. 
Sometimes while reading some parts I would have liked, sometimes already previously defined:
1) reference to an equation, but equation numbers are not rendered at this stage in the docs, something to think about
2) reference to some paper (in some cases could be pinned down to a section of the NEPEC paper). 

A more general comment on the docs structure: I am more than ever convinced that pushing the workflow figures about the techniques  until file 4, in the section "What happens when I use PEC?", requires a "leap of faith" that is not realistic from beginners and general readers. The theory is quite complex, there is no way around it (it is explained very well here) and it is quite counterintuitive to keep until the end the overarching concept that can help someone get interested into and understand at a higher level the main passages, before delving into the details. If it is decided that the workflow figures need to be in section 4, for each technique, I even suggest that section 4 is moved to second place in the order of sections. Also, even from the "hands-on" approach we have chosen, section 4 to me seems more compelling than "When should I use PEC?", which could be swapped with it, to be in position 4.

docs/source/guide/pec-5-theory.myst
andreamari: Applied here: f908f64

docs/source/guide/pec-5-theory.myst
andreamari: Applied here: 40ae292

docs/source/guide/pec-5-theory.myst
andreamari: Applied here: 40ae292

andreamari:
In the section `When should I use PEC?` advantages and disadvantages are presented without an explicit use-case example. A  use-case example for PEC is a quite non-trivial addition and it may be a good idea to have a new issue + a new PR dedicated to it.
If you agree, this is **ready for review**. If you disagree, we can keep this PR open until I add a reasonable use-case.

@nathanshammah, @rmlarose, @crazy4pi314, @vprusso,   if and when you have time, feel free to take a look.

docs/source/guide/pec-1-intro.myst
```diff
@@ -0,0 +1,142 @@
...
+```{code-cell} ipython3
+frontend = "qiskit"  # Supported: "cirq", "qiskit", "pyquil", "braket", "pennylane".
+```
+
+## Define the circuit of interest
+
++++
+
+We first define the circuit of interest. For simplicity, in this example we use
+a single-qubit randomized-benchmarking circuit whose ideal execution is equivalent to the
```
rmlarose, line 35: Here "ideal" is used again without having been explicitly defined. I think it's a good idea to add definitions after the introduction.

"To apply PEC, you must know how to represent *ideal gates* as linear combinations of *noisy implementable gates*. By *ideal gates*, we mean noiseless unitary operations. By *noisy implementable gates* we mean..."

docs/source/guide/pec-1-intro.myst
```diff
@@ -0,0 +1,142 @@
...
+identity operation.
+
+```{code-cell} ipython3
+from mitiq import benchmarks
+
+circuit = benchmarks.generate_rb_circuits(
+  n_qubits=1, num_cliffords=2, return_type = frontend,
```
rmlarose, line 42: nit: remove space before/after `=` in `return = frontend`.

docs/source/guide/pec-1-intro.myst
```diff
@@ -0,0 +1,142 @@
...
+)[0]
+
+print(circuit)
+```
+In the code cell above, we used the function {func}`~mitiq.benchmarks.randomized_benchmarking.generate_rb_circuits()`, in which `num_cliffords` is the 
+number of random Clifford group elements and is proportional to the depth of the circuit.
+
+## Define a noisy executor
```
rmlarose, line 50: Remove "noisy". We should always assume people are using Mitiq to run on hardware which is inherently noisy.

docs/source/guide/pec-1-intro.myst
```diff
@@ -0,0 +1,142 @@
...
+
++++
+
+To interface generic frontends and backends with Mitiq, we must define an `executor`.
+In the simplest case, an `executor` is a Python function that takes as input a quantum
+circuit and returns the noisy expectation value of some observable.
+See the [Executors](guide-executors.myst) section for more information on 
+how to define more advanced executors.
```
rmlarose, line 58: I think we should remove all of this text and just link to the executors guide for more background. "We now define an `executor`(guide-executors.myst) to compute the ground state probability. We add a simple noise model since we are using a simulator."

docs/source/guide/pec-1-intro.myst
```diff
@@ -0,0 +1,142 @@
...
+    mitiq_circuit, _ = convert_to_mitiq(circuit)
+    noisy_circuit = mitiq_circuit.with_noise(depolarize(p=noise_level))
+    rho = DensityMatrixSimulator().simulate(noisy_circuit).final_density_matrix
+    return rho[0, 0].real
+```
+
+The `executor` can be used to evaluate noisy (unmitigated) expectation values.
+
+```{code-cell} ipython3
+# Compute the expectation value of the |0><0| observable.
```
rmlarose, line 84: nit: Use the \langle \rangle brackets (copy them from the docstring of `executor`).

docs/source/guide/pec-1-intro.myst
```diff
@@ -0,0 +1,142 @@
...
+
+**Note:** The body of this function is supposed to be written by the user and depends on the specific 
+frontend and backend. Here, for simplicity, we first convert the input circuit to the Mitiq internal
+representation (Cirq) and then simulate it.
+
+```{code-cell} ipython3
+from cirq import DensityMatrixSimulator, depolarize
+from mitiq.interface import convert_to_mitiq
+
+def executor(circuit, noise_level=0.01):
```
rmlarose, line 70: I would strongly encourage calling this function `execute` instead of `executor`. Function names should be verbs and object names (`mitiq.Executor`) should be nouns. It's confusing to call both the function and the object executor.

I know we are plagued by a bad history of this, but I think we should make this distinction and start being clearer about it.

docs/source/guide/pec-1-intro.myst
```diff
@@ -0,0 +1,142 @@
...
+
+The `executor` can be used to evaluate noisy (unmitigated) expectation values.
+
+```{code-cell} ipython3
+# Compute the expectation value of the |0><0| observable.
+noisy_value = executor(circuit)
+ideal_value = executor(circuit, noise_level=0.0)      
+print(f"Error without mitigation: {abs(ideal_value - noisy_value) :.5f}")
+```
+
```
rmlarose, line 89: Nothing up to this point has anything to do with PEC, and I think its useful to emphasize this. I suggest the following sections and header levels

## Problem setup

### Define a circuit

### Define an executor

(It might be better to get rid of the two subheaders above.)

## Apply PEC

### Represent ideal gates as...

### Run PEC

docs/source/guide/pec-1-intro.myst
```diff
@@ -0,0 +1,142 @@
...
+The aim of PEC is to reduce this error.
```
rmlarose, line 90: I want to suggest changing this to "We now show how to use PEC which should reduce this error." since this doc is about how to use PEC, not what the goal of PEC is. I might be splitting hairs too much though and I'm fine either way.

docs/source/guide/pec-1-intro.myst
```diff
@@ -0,0 +1,142 @@
...
+
++++
+
+## Represent ideal gates as linear combinations of noisy gates
+
++++
+
+Before applying PEC, we need to represent the gates of the ideal `circuit` as linear combinations
```
rmlarose, line 98: "To apply PEC, we need to represent some or all of the gates..."

docs/source/guide/pec-1-intro.myst
```diff
@@ -0,0 +1,142 @@
...
+The aim of PEC is to reduce this error.
+
++++
+
+## Represent ideal gates as linear combinations of noisy gates
+
++++
+
+Before applying PEC, we need to represent the gates of the ideal `circuit` as linear combinations 
+of noisy gates. In Mitiq, one can express a gate representation via an {class}`~.OperationRepresentation`
```
rmlarose, line 99: "noisy gates" -> "noisy implementable gates" to be consistent with the vocabulary in the introduction

rmlarose, line 99: "gate representation as an `OperationRepresentation` sounds funny. I suggest we the Cirq meanings in which "operation" is qubit-dependent and "gate" is not. The main reason being `OperationRepresentation` is in fact qubit-dependent.

Since I think we should use "operation" instead of "gate" here, I think we should be consisten and use "operation" instead of "gate" everywhere in this doc. Or, at least note somewhere we are using them interchangably, and note the difference between qubit-dependent and qubit-independent.

docs/source/guide/pec-1-intro.myst
```diff
@@ -0,0 +1,142 @@
...
+object, which depends on the specific noise model of the backend.
```
rmlarose, line 100: Remove "which depends on the..." You can instantiate an `OperationRepresentation` without regard to any noise model or backend, so I find this statement misleading.

docs/source/guide/pec-1-intro.myst
```diff
@@ -0,0 +1,142 @@
...
+
++++
+
+In this example, we assume that local depolarizing noise of some fixed `noise_level` is acting after
+the application of each gate. For this simple noise model, one can easily obtain the necessary representations as follows:
```
rmlarose, line 105: "For this simple noise model, one can easily obtain the necessary representations as follows"...

This reads like magic and doesn't answer any natural questions which arise from this statement, e.g., "What if I have another simple noise model but it isn't depolarizing noise?" or "What if I don't have a simple noise model?"

I think this section needs to be restructured and include the three possible ways we have discussed to do this:

1. Assume depolarizing noise and guess a noise level / obtain it from hardware.
2. Run an RB experiment and infer the depolarizing noise level.
3. Run GST.

There's space to mention the other analytic representations we have (e.g., amplitude damping) in the first point.

rmlarose (changes requested):
Thanks @andreamari. I reviewed the first doc. I'll get to the others after you respond to these comments to make it more manageable for both of us.

docs/source/examples/pec-tutorial.myst
```diff
@@ -10,7 +10,7 @@ kernelspec:
   language: python
   name: python3
 ---
-(label-pec-example)=
+
```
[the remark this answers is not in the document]
andreamari: Fine to delete it, to avoid overlapping content with the main docs. In the future we can add a new PEC tutorial that highlights something different with respect to the main docs.

docs/source/guide/pec-4-low-level.myst
```diff
@@ -0,0 +1,224 @@
...
+  circuit, reps, num_samples=200,
+)
+```
+
+We are now ready for the second step of PEC, i.e., the inference
+of the ideal expectation value from the noisy execution of the `sampled_circuits`.
+
++++
+
+## Step 2: Inference of the ideal expectation value from the noisy execution of the auxiliary circuits.
```
nathanshammah, line 133:
```suggestion
## Step 2: Inference of the ideal expectation value from the noisy execution of the auxiliary circuits
```

docs/source/guide/pec-3-options.myst
```diff
@@ -0,0 +1,409 @@
...
+```{code-cell} ipython3
+# Positive and normalized distribution p(alpha)=|eta_alpha|/gamma
+h_rep.distribution()
+```
+
+## Estimating expectation values with PEC
+
++++
+
+The second main step of the PEC is to make use of the previously defined {class}`.OperationRepresentation`s to estimate
```
nathanshammah, line 252:
```suggestion
The second main step of PEC is to make use of the previously defined {class}`.OperationRepresentation`s to estimate
```

docs/source/guide/pec-3-options.myst
```diff
@@ -0,0 +1,409 @@
...
+pec_data["unbiased_estimators"][:5]
+```
+
+The statistical error `pec_error` corresponds to `pec_std` / `sqrt(num_samples)`, where `pec_std` is
+the standard deviation of the unbiased samples, i.e., the square root of the mean squared deviation of
+`unbiased_estimators` from `pec_value`.
+
+```{code-cell} ipython3
+pec_data["pec_error"]
+```
```
nathanshammah, line 409: Suggestion: Comment with a short sentence or paragraoh to conclude the section.

nathanshammah (approved):
Thank you @andreamari, looks great. I left minor suggestions.

docs/source/guide/pec-1-intro.myst
```diff
@@ -0,0 +1,142 @@
...
+  display_name: Python 3
+  language: python
+  name: python3
+---
+
+# How do I use PEC?
+
+Probabilistic Error Cancellation (PEC) is an error mitigation technique for improving the estimation
+of expectation values (see [What is the theory behind PEC?](pec-5-theory.myst)). To be applied, PEC requires the 
+full knowledge of the noise model. In particular, the user must know how to represent ideal gates as linear
```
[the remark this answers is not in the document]
andreamari: Done

docs/source/guide/pec-1-intro.myst
andreamari: done

docs/source/guide/pec-1-intro.myst
andreamari: This is now explained at the beginning of the notebook

docs/source/guide/pec-1-intro.myst
andreamari: Done. I decided to merge your cases 1 and 2 into a unique case for simplicity. Please have a look to check if it is clear enough. I think we need to find the right trade-off between explaining all the possible cases and the simplicity of a "getting-started" section.

rmlarose: Looks great! Thanks for making these nontrivial changes.

docs/source/guide/pec-2-use-case.myst
```diff
@@ -0,0 +1,33 @@
...
+## Disadvantages
+
+PEC is a noise-aware technique that can be applied only if
+the **full tomographic knowledge** of the hardware gates is known. Indeed, in order to represent
+ideal gates as linear combination of noisy gates, one typically needs to know the super-operator
+matrix associated to each noisy gate.
+
+Another practical problem of PEC is that it involves the execution of many different circuits.
+This typically requires more clock time compared to the repeated execution of equal circuits.
+Batched execution of circuits can alleviate this problem.
```
rmlarose, line 33:
```suggestion
Batched execution of circuits can alleviate this problem to some extent.
```

rmlarose: "...if supported by the hardware provider."

docs/source/guide/pec-2-use-case.myst
```diff
@@ -0,0 +1,33 @@
...
+
+The main advantage of PEC is that, under the assumption of perfect gate tomography,
+it provides an unbiased estimation of expectation values.
+This means that, in the limit of many samples, the error mitigated expectation values
+converge to the ideal expectation values.
+
+
+## Disadvantages
+
+PEC is a noise-aware technique that can be applied only if
```
rmlarose, line 26: Splitting hairs on this point again, but you can apply PEC whenever you want. I would rather say the PEC result converges to the noiseless results only if full tomographic knowledge of the hardware gates is known.

docs/source/guide/pec-3-options.myst
```diff
@@ -0,0 +1,414 @@
...
+    representations=representations,
+    num_samples=5, # Number of PEC samples
+    random_state=0, # Seed for reproducibility of PEC sampling
+    full_output=True, # Return "pec_data" in addition to "pec_value"
+)
+```
+
+Similar to other error mitigation modules, `observable` is an optional argument of 
+{func}`.execute_with_pec`. If `observable` is `None` the executor must return an expectation value,
+while if `observable` is a Mitiq {class}`.Observable` the executor must return raw bitstrings.
```
rmlarose, line 326: "otherwise the executor must return a `mitiq.QuantumResult` from which the expectation value can be computed with the observable."

docs/source/guide/pec-3-options.myst
```diff
@@ -0,0 +1,414 @@
...
+    circuit_copy = circuit.copy()    
+    noise_model = initialized_depolarizing_noise(BASE_NOISE)
+    projector_on_zero = [[1, 0], [0, 0]]
+    return execute_with_noise(circuit_copy, projector_on_zero, noise_model)
+
+# Cast the execute function into an Executor class to record the execution histroy
+executor = Executor(execute)
+```
+
+### Options for estimating expectation values
```
rmlarose, line 302: I would definitely start with this section (i.e., move this section to the top), then get into the longer discussion about operation representations. That way the easiest and most relevant information is at the beginning, and you can learn more by reading more.

docs/source/guide/pec-3-options.myst
```diff
@@ -0,0 +1,414 @@
...
+
+The method {meth}`.OperationRepresentation.sample()` can be used for this scope:
+
+```{code-cell} ipython3
+noisy_op, sign, coeff = h_rep.sample()
+print(f"The sampled noisy operation is: {noisy_op}.")
+print(f"The associated coefficient is {coeff:g}, whose sign is {sign}.")
+```
+
+***Hint:** try re-executing the previous cell to get different samples.*
```
rmlarose, line 226: Hint? or Note?

rmlarose (changes requested):
Thanks for the revisions to doc 1! Added (fewer) comments to docs 2 and 3!

docs/source/guide/pec-4-low-level.myst
```diff
@@ -0,0 +1,224 @@
...
+
+## Step 1: Probabilistic generation of all the auxiliary circuits
+
++++
+
+### Define the circuit of interest and the quasi-probability representations
+
++++
+
+We define the circuit of interest as we did in Section [How do I use PEC?](pec-1-intro.myst)
```
rmlarose, line 51:
```suggestion
We define the circuit of interest as we did in [How do I use PEC?](pec-1-intro.myst)
```

docs/source/guide/pec-4-low-level.myst
```diff
@@ -0,0 +1,224 @@
...
+- Append the sampled gate to `sampled_circuit` and the corresponding sign to `gate_sign_list`;
+- When the loop is completed, return `sampled_circuit` and the associated `sampled_sign` (which is the product of all
+   the elements of `gate_sign_list`).
+
+Instead of manually applying all the previous steps, one can call the Mitiq function {func}`.sample_circuit()` to probabilistically generate an integer number `num_samples` of auxiliary circuits.
+
+```{code-cell} ipython3
+from mitiq import pec
+
+pec.sample_circuit(circuit, reps, num_samples=5)
```
rmlarose, line 110: Would be nice to see the circuits in this output, but really that's a Qiskit problem.

image

docs/source/guide/pec-4-low-level.myst
```diff
@@ -0,0 +1,224 @@
...
+    expectation_values = []
+    for circuit in circuits:
+        mitiq_circuit, _ = convert_to_mitiq(circuit)
+        noisy_circuit = mitiq_circuit.with_noise(depolarize(p=noise_level))
+        rho = DensityMatrixSimulator().simulate(noisy_circuit).final_density_matrix
+        expectation_values.append(rho[0, 0].real)
+    return expectation_values
+```
+
+We also initialize a Mitiq {class}`.Executor` object. This step is optional but recommended.
```
rmlarose, line 165:
```suggestion
We also initialize a Mitiq {class}`.Executor` object. This step is optional but recommended to store/access executed circuits and results.
```

docs/source/guide/pec-5-theory.myst
```diff
@@ -0,0 +1,141 @@
...
+The one-norm $\gamma$ of the global quasi-probability distribution is the product of those of the gates:
+
+$$
+\sum_{\vec \alpha} \eta_{\vec{\alpha}}=1,  \qquad  \gamma = \sum_{\vec{\alpha}} |\eta_{\vec \alpha}| = \Pi_{i=1}^{t} \gamma_i.
+$$
+
+All the noisy expectation values $\langle A_{\vec{\alpha}}\rangle_{\rm noisy}$ can be directly measured with 
+a noisy backend since they only require circuits composed of implementable noisy operations.
+In principle, by combining all the noisy expectation values, one could compute the ideal result $\langle A \rangle_{\rm ideal}$.
+Unfortunately this approach requires the execution of an exponentially large number of circuits and
```
rmlarose, line 95:
```suggestion
Unfortunately this approach requires executing a number of circuits which grows exponentially with the circuit depth and 
```

rmlarose (approved):
Minor comments then LGTM, thanks @andreamari.

The PEC theory section is very nice, kudos!

docs/source/guide/pec-4-low-level.myst
andreamari: I agree. Done here: a326948 .
I picked an _ad hoc_ seed such that we don't sample 5 equal circuits. In this way it is more clear how PEC circuits may look like.

PR #993, opened 2021-10-20
Bump qiskit to version 0.31 and pin it explicitly in dev requirements

andreamari:
According to the discussion of the last Mitiq community call, this PR adds `qiskit`  in  `dev_requirements.txt`
and remove individual qiskit sub-packages to avoid conflicts.

Moreover it also updates `qiskit_utils.py ` for bumping Qiskit to the current latest version.

PR #1011, opened 2021-11-03
Increase shots in test_zne.py

andreamari:
Hopefully, this will fix the flaky zne tests (https://github.com/unitaryfund/mitiq/pull/993#issuecomment-958386702). 

Note: I tested locally and it only increased the pytest duration for `test_zne.py` by 6 seconds. So timing is not an issue.

### DIFF
--- a/mitiq/zne/tests/test_zne.py
+++ b/mitiq/zne/tests/test_zne.py
@@ -296,8 +296,6 @@ def test_qiskit_execute_with_zne():
     )
     base = qiskit_executor(circuit)
     zne_value = execute_with_zne(circuit, qiskit_executor)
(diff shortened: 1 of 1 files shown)


mitiq/zne/tests/test_zne.py
```diff
@@ -296,7 +296,8 @@ def test_qiskit_execute_with_zne():
     )
     base = qiskit_executor(circuit)
     zne_value = execute_with_zne(circuit, qiskit_executor)
-
+    print(true_zne_value - zne_value)
+    print(true_zne_value - base)
```
rmlarose, line 300:
```suggestion
```

rmlarose (approved):
Thanks!

PR #1018, opened 2021-11-09
[Bug fix] Add back-conversions in execute_with_pec

andreamari:
A back-conversion of the sample circuits to the native circuit type is missing in `execute_with_pec`. 
This is a bug that makes PEC not working with non-Cirq frontends and so it may justify a patch-release. 

Note: this bug was not detected since we use `convert_to_mitiq` + Cirq simulation in most test executors.
For this reason a specific new test is added in this PR.


rmlarose (commented):
Was this also a bug in 0.10?

Looks good, let's think about the patch release strategy.

andreamari:
> Was this also a bug in 0.10?
> 
It was present since a long time. I think it was introduced in this PR: #647 .

rmlarose (approved):

PR #1020, opened 2021-11-09
Increase shots in zne tests with shot_list

### LINKED ISSUE #1015: A test for ZNE with Qiskit circuit is non-deterministically failing
https://github.com/unitaryfund/mitiq/runs/4095356632?check_suite_focus=true 

Possible reasons:
- qiskit executors in mitiq_qiskit changed.
- handling of keywords for executors in factories changed.

Note: it is possible that this is the expected behavior and for some reason it was erroneously passing in the past. In this case we should probably increase the number of shots or enforce in some way to have deterministic results.

andreamari:
Attempt to fix #1015 related to non-deterministic tests in `test_zne.py`

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -12,7 +12,7 @@ Please use `mitiq.about()` to document your dependencies and working environment
 You can open a [pull request](https://github.com/unitaryfund/mitiq/pulls) by pushing changes from a local branch, explaining the bug fix or new feature.
 
 ### Version control with git
-git is a language that helps keeping track of the changes made. Have a look at these guidelines for getting started with [git workflow](https://www.asmeurer.com/git-workflow/).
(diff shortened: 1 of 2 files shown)


nathanshammah:
Thanks @andreamari. This looks good to me. What is a bit strange is sort of a glitch in Github that does not see the change in the two files. They are visible only by clicking each commit.

andreamari:
> Thanks @andreamari. This looks good to me. What is a bit strange is sort of a glitch in Github that does not see the change in the two files. They are visible only by clicking each commit.

Only one file is changed: `mitiq/zne/tests/test_zne.py`   .  The other one was first changed and then reverted.

nathanshammah (approved):
LGTM, thank you @andreamari. It is not clear to me why `linkcheck` is now passing the CI tests, if nothing was changed there.

Issue #1023, opened 2021-11-10
Wrong circuits classically simulated in CDR


rmlarose:
In CDR, the training circuits at scale factor 1 should be simulated, but right now I believe the scaled circuits (not training circuits) are being simulated.

https://github.com/unitaryfund/mitiq/blob/e88d19f1b27cc99328b05466f2d58f07261b2ed9/mitiq/cdr/cdr.py#L161

Thanks @L-P-B for finding!

rmlarose:
Closed by #1026.

PR #1026, opened 2021-11-11
CDR: Run the training circuits on the simulator

rmlarose:
Description
-----------

For #1023 

Checklist
-----------

Tips
----

### DIFF
--- a/mitiq/cdr/cdr.py
+++ b/mitiq/cdr/cdr.py
@@ -164,7 +164,7 @@ def execute_with_cdr(
     # Do the regression.
     fitted_params, _ = curve_fit(
         lambda x, *params: fit_function(x, params),
-        noisy_results,
+        noisy_results[1:, :].T,
         ideal_results,
         p0=np.zeros(num_fit_parameters),
     )
--- a/mitiq/cdr/cdr.py
+++ b/mitiq/cdr/cdr.py
@@ -158,7 +158,7 @@ def execute_with_cdr(
     results = executor.evaluate(to_run, observable)
     noisy_results = np.array(results).reshape(all_circuits_shape)
 
-    results = simulator.evaluate(all_circuits[0], observable)
+    results = simulator.evaluate(training_circuits, observable)
     ideal_results = np.array(results)
 
     # Do the regression.
--- a/mitiq/cdr/cdr.py
+++ b/mitiq/cdr/cdr.py
@@ -83,10 +83,10 @@ def execute_with_cdr(
         num_fit_parameters: The number of parameters the fit_function takes.
         scale_noise: scale_noise: Function for scaling the noise of a quantum
             circuit.
-        scale_factors: Factors by which to scale the noise, should not
-            include 1 as this is just the original circuit. Note: When
-            scale_factors is provided, the method is known as "variable-noise
-            Clifford data regression."
+        scale_factors: Factors by which to scale the noise.
+            - When 1.0 is the only scale factor, the method is known as CDR.
+            - Note: When scale factors larger than 1.0 are provided, the method
+                is known as "variable-noise CDR."
         kwargs: Available keyword arguments are:
             - method_select (string): Specifies the method used to select the
                 non-Clifford gates to replace when constructing the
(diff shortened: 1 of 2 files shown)


rmlarose:
@L-P-B this is the fix, correct?

Tests are failing because accuracy gets worse, as expected. But the accuracy gets significantly worse in some cases. Before modifying tests I want to see if you agree on this being the fix.

mitiq/cdr/cdr.py
```diff
@@ -158,7 +158,7 @@ def execute_with_cdr(
     results = executor.evaluate(to_run, observable)
     noisy_results = np.array(results).reshape(all_circuits_shape)
 
-    results = simulator.evaluate(all_circuits[0], observable)
+    results = simulator.evaluate(training_circuits, observable)
     ideal_results = np.array(results)
 
     # Do the regression.
```
L-P-B, line 164: When doing the regression here it appears it is still being done on all the noisy circuits. 

This should just be done on the training circuits so it should be something like, the transpose is to ensure the shape is fitted in the linear_fit_function:

`noisy_results[1:,:].T,`

andreamari: Makes sense to me: the fit must be based only on the training (near-clifford) results. The results associated to the circuit of interest should be used only in the last line, where the final estimate is returned.

What I don't understand is why `curve_fit` doesn't give an error if `len(noisy_results)` is not equal to `len(ideal_results)`. Anyway, this is not important I guess.

andreamari:
Another thing that I think we should fix is the description of the `scale_factors` argument:
```
scale_factors: Factors by which to scale the noise, should not
            include 1 as this is just the original circuit. Note: When
            scale_factors is provided, the method is known as "variable-noise
            Clifford data regression."
```
I guess that, based on the code, it's actually the opposite situation. The first scale factor **must be** `1` or, at least, it is strongly recommended that the first element is `1.`

mitiq/cdr/cdr.py
L-P-B: Thats right, sorry for any confusion. 

Yes I also dont know why it does not give an error... pretty weird.

andreamari:
Maybe failures are due to the fact that, after removing the circuit of interest from the training circuits, the CDR problem is just more difficult to solve and more training circuits are necessary (now num_training_circuits=4 in tests).  
@L-P-B, what do you think?

L-P-B:
I would expect it to work with these circuits for even a small number of training circuits (provided there are more than 3). I think it might be something to do with the tests. Am I misunderstanding or should we always require the scalefactors to include 1. Right now in the tests it is passing a list: [3], maybe it should be [1,3] or (1,3). Sorry haven't had time to check this, just thinking outloud.

rmlarose:
> I would expect it to work with these circuits for even a small number of training circuits (provided there are more than 3). I think it might be something to do with the tests. Am I misunderstanding or should we always require the scalefactors to include 1. Right now in the tests it is passing a list: [3], maybe it should be [1,3] or (1,3). Sorry haven't had time to check this, just thinking outloud.

It should be `scale_factors=(1, 3)` for the test with variable noise, but even the test with `scale_factors=(1,)` is failing. I'm still tinkering with this, haven't found a change that's working yet.

mitiq/cdr/cdr.py
```diff
@@ -158,14 +159,14 @@ def execute_with_cdr(
...
     fitted_params, _ = curve_fit(
         lambda x, *params: fit_function(x, params),
-        noisy_results,
+        noisy_results[1:, :].T,
         ideal_results,
         p0=np.zeros(num_fit_parameters),
     )
-    return fit_function(noisy_results[:, 0], fitted_params)
+    return fit_function(noisy_results[0, :], fitted_params)
```
rmlarose, line 172: ahah!

rmlarose:
Thanks Andrea!! I fixed scale factors docstring. This looks ready to go to me.

andreamari (approved):

andreamari:
@L-P-B, the changes introduced in this PR my be relevant for your simulations. Apart from `noisy_results[:, 0]` --> `noisy_results[0, :]` an further change is the handling of `num_training_params` in `execute_with_cdr()`.

PR #1047, opened 2021-12-14
Add asv benchmarking framework

### LINKED ISSUE #1006: Add "`mitiq.raw.execute`"?
Without an observable, the canonical way to compute "raw results" (i.e., results without error mitigation) is `execute(circuit)`.

However, with an observable, the interface is different:

- Without error mitigation: Either `Executor.evaluate(circuit, observable)` or `observable.expectation(circuit, executor)`.
- With error mitigation: `execute_with_xxx(circuit, executor, observable)`.

Should we (i) support a common interface? and (ii) do this by adding a "raw" module which does not do any error mitigation but exists solely for the purpose of having a common interface?

## Pros

It would be really useful for documentation and examples. For example:

```python
import cirq

from mitiq import Executor, Observable, PauliString, zne, raw

circuit = cirq.Circuit(cirq.H.on(cirq.LineQubit(0)) for _ in range(10))
executor = Executor(mitiq_cirq.compute_density_matrix)
observable = Observable(PauliString("Z"))
print(f"Problem: ⟨𝜓| {observable} |𝜓⟩ where |𝜓⟩ is given by", circuit, sep="\n\n")

# Without error mitigation.
raw_result = raw.execute(circuit, executor, observable)
print("\nRaw result:", raw_result)
print("Executed circuits:", *executor.executed_circuits, sep="\n")

# With error mitigation.
zne_result = zne.execute_with_zne(circuit, executor, observable)
print("\nZNE result:", zne_result)
print("Executed circuits:", *executor.executed_circuits[1:], sep="\n")
```

```text
Problem: ⟨𝜓| Z(0) |𝜓⟩ where |𝜓⟩ is given by

0: ───H───H───H───H───H───H───H───H───H───H───

Raw result: 0.95
Executed circuits:
0: ───H───H───H───H───H───H───H───H───H───H───

ZNE result: 0.99
Executed circuits:
0: ───H───H───H───H───H───H───H───H───H───H───
0: ───H───H───H───H───H───H───H───H───H───H───H───H───H───H───H───H───H───H───H───H───
0: ───H───H───H───H───H───H───H───H───H───H───H───H───H───H───H───H───H───H───H───H───H───H───H───H───H───H───H───H───H───H───
```

## Cons

It doesn't really have a use beyond documentation. One may misinterpret `mitiq.raw` to do error mitigation.

rmlarose:
Description
-----------

Adds [ASV benchmarks](https://asv.readthedocs.io/en/stable/), currently with ZNE and PEC benchmarks. Example output:

image

To reproduce:

```bash
pip install asv
asv run
asv publish
asv preview
```

See ASV documentation for more.

Also adds `mitiq.raw` module for comparing to no error mitigation and closes #1006.

Also removes error mitigation (tests) from `mitiq.benchmarks` in favor of these new benchmarks.

Checklist
-----------

Tips
----


mitiq/raw/raw.py
```diff
@@ -0,0 +1,46 @@
...
+# You should have received a copy of the GNU General Public License
+# along with this program.  If not, see <https://www.gnu.org/licenses/>.
+
+"""Run experiments without error mitigation."""
+from typing import Callable, Optional, Union
+
+from mitiq import Executor, Observable, QPROGRAM, QuantumResult
+
+
+def execute(
```
andreamari, line 22: This name can be confused with the `execute` function typically defined by the Mitiq users. 

More explicit names could be:

`execute_with_raw`  (where `raw` as the name of the trivial error mitigation technique)
`execute_without_mitiq`
`execute_without_mitigation`

Other ideas?

andreamari (commented):
I reviewed only the `raw` module part and LGTM.

Do we want also the `raw.mitigate_executor` and `raw.raw_decorator` ?
I am not sure if they are useful. 

Using avs seems a nice idea to me but I would leave the review of that part to @vprusso .

mitiq/raw/raw.py
rmlarose: I haven't really had any good ideas here. Since the naming convention for techniques is `execute_with_xyz`, it seems natural that no technique should remove the `with_xyz` and just be `execute`.

I don't really like assuming/encouraging the name `execute` for user-defined functions. We used to reserve `execute` for `QPROGRAM -> float`, but now it's the more general `QPROGRAM -> QuantumResult`. I'd rather encourage names like `compute_density_matrix` or `sample` which are clearer. Is this sufficient to remove the confusion you're worried about?

rmlarose:
> Do we want also the `raw.mitigate_executor` and `raw.raw_decorator` ?
> I am not sure if they are useful.

I'm hesitant to say yes because it seems silly, but I guess they should be there in terms of parallelism which is the goal of this PR.

mitiq/raw/raw.py
andreamari: > I haven't really had any good ideas here. Since the naming convention for techniques is `execute_with_xyz`, it seems 
> natural that no technique should remove the `with_xyz` and just be `execute`.

Ok, I am fine with `execute`.  We can always call the function with `raw.execute` if we need to be more explicit.

> I don't really like assuming/encouraging the name `execute` for user-defined functions. We used to reserve `execute` for `QPROGRAM -> float`, but now it's the more general `QPROGRAM -> QuantumResult`. I'd rather encourage names like `compute_density_matrix` or `sample` which are clearer. 

I actually disagree with this.  The `execute`/`executor` is a key aspect of Mitiq and using a fixed name help underline this and make it easily recognizable in examples. If I had to summarize all the docs in one sentence I would say: "To use Mitiq, you must define the `execute` function. The rest are minor details." :smiley: 

Anyway this is not a blocker for this PR which LGTM!

andreamari (approved):

mitiq/raw/raw.py
rmlarose: Fair point. I guess it's a semantic argument.

rmlarose:
Ping @vprusso for thoughts on asv, but this is still WIP. Before merging I want to:

- Run a minimal version of the benchmarks in CI.
  - Trying to determine best way to do this now. I thought command line arguments to `asv` would be a solution, but it's proving nontrivial so far.
- Remove old benchmarks in https://github.com/unitaryfund/mitiq/blob/master/mitiq/benchmarks/random_circuits.py and https://github.com/unitaryfund/mitiq/blob/master/mitiq/benchmarks/maxcut.py. (Or move them into the asv benchmarks.)

benchmarks/benchmarks.py
```diff
@@ -0,0 +1,172 @@
...
+
+compute_density_matrix_noiseless = functools.partial(
+    mitiq_cirq.compute_density_matrix, noise_level=(0.0,)
+)
+benchmark_circuit_types = ("rb", "mirror")
+
+
+def get_benchmark_circuit(
+    circuit_type: str, nqubits: int, depth: int,
+) -> "cirq.Circuit":
```
vprusso, line 36: There might be a good reason for the way that this is (and if so, just ignore this comment), but usually, I only see return types quoted like this if it's in reference to returning an object in a class method whose return type is the same as the class. So, for instance:

```python
class Name:
    def get_name_obj -> "Name":
        pass
...
```

In that instance, you can just `from __future__ import annotations` to overcome the need for this quoted return type-hint. 

All this to say I'm not sure if the `cirq.Circuit` bit needs to be surrounded by quotes or not. My initial thought is it doesn't, but again, I could be missing some sound reason as to why this is the case.

vprusso:
> Ping @vprusso for thoughts on asv, but this is still WIP. Before merging I want to:

Thanks for the ping, @rmlarose. I can't say I've used `asv` for any of my past projects, but from looking at how it's used in other projects and how it's being used in `mitiq` I can't say I have an qualms with the usage so LGTM from that side of things.

mitiq/raw/raw.py
```diff
@@ -0,0 +1,46 @@
+# Copyright (C) 2020 Unitary Fund
```
vprusso, line 1: This is a tangential point to the actual code, but from my experience, files that are newly created should be stamped with copyright pertaining to the year in which they are created. Similarly, any updates to a file with an older year that is updated in a more recent year should also have the copyright value updated to reflect this. 

Note that I am not a lawyer nor do I play one on the internet. This is just a practice I've had to do previously at other institutions and it seems to be somewhat consistent. Of course, it might be a better idea to confirm with a legal professional as opposed to an in-passing comment here from me.

rmlarose: Good catch! Should be 2021, maybe even 2022 :upside_down_face:

benchmarks/benchmarks.py
rmlarose: There's actually no reason for quotes here, I forget why I did that. Will remove.

mitiq/raw/raw.py
rmlarose: Updated here and in other files.

benchmarks/benchmarks.py
rmlarose: Removed quotes.

rmlarose:
More changes, please take another look @vprusso @andreamari. Summary:

- Address review comments.
- Make sure ASV benchmarks run by testing them in CI.
- Reduce parameters to run quickly in CI.
- Remove old benchmark functions/tests in `mitiq.benchmarks`. This module now only has tools to generate benchmark circuits, not run them.
  - Note I removed a ZNE test from some test file in `mitiq.benchmarks` (https://github.com/unitaryfund/mitiq/pull/1047/commits/cd35b936c8508de1b37b6fba1c2728a1e393f4e6). This is a ZNE test so should belong in... ZNE tests. I was going to move it there, but it was superfluous with the already-existing tests which use various factories/scaling techniques with RB circuits.

mitiq/benchmarks/maxcut.py
```diff
@@ -1,189 +0,0 @@
-# Copyright (C) 2020 Unitary Fund
```
vprusso, line 1: Looks like the auto-doc builder is complaining that `maxcut.py`, `random_circuits.py`, and `utils.py` aren't present. So, if you'll be removing these files, be sure to remove the documentation usages as well.

vprusso (approved):
Modulo the minor presently failing doc building, LGTM!

andreamari (commented):
The API-docs is not updated. I think we need to:
- remove the old benchmark functions,
- add the `raw` module to the API-docs,
- (maybe) add some benchmark functions too  (e.g. `get_benchmark_circuit`) ?

rmlarose:
Thanks both for noting to update the docs - done. Ignoring the trivial coverage failure.

PR #1113, opened 2022-02-01
Remove option to squash moments in global folding

### LINKED ISSUE #653: Improve fold global function
The current `fold_global` function:

1. Has a `squash_moments` option but I guess that even if it is set to `False`, the folded blocks are always squashed anyway.
2. `fidelities` is not an option as in gate folding methods, but it could be.

Proposal for new changes:
1. Make use of a `fidelities` when applying fractional folding, in a similar way as it is now done for local folding methods.
2. Remove the `squash_moments` option and always try to **keep** the moment structure. The reason is that if one applies global folding there is not a strong motivation to squash. (we are not _inserting_ new gates in the middle of the input circuit we are just _appending_ circuits or sub_circuits circuits at the end of the input_cirucit).
3. Alternatively, keep the  `squash_moments` option but make sure that, if `squash_moments=False`, the moment structure is  preserved even for the appended circuits (virtual folded blocks). Maybe in this case  `squash_moments=False` should be default?

Between (2) and (3) I would prefer (2) for its conceptual simplicity, but (3) is fine.

purva-thakre:
Description
-----------
Fixes #653 

Checklist
-----------

Tips
----


purva-thakre:
@andreamari I think it's ok to ignore the `codecov` failure. Not sure why RTD build is failing.

andreamari:
> @andreamari I think it's ok to ignore the `codecov` failure. Not sure why RTD build is failing but is there a way to re-start it ?

Thanks @purva-thakre !
Yes, we can ignore both codecov and RTD for the moment. 

I think it is important to add test with a scale factor different from 3. 
For example:

```
Folded circuit with current fold_global function:
0: ───H──────────────

1: ───Z──────────────

2: ───S───S^-1───S───

3: ───T───T^-1───T───

Folded circuit desired result:
0: ───H─────────────────────────────────────

1: ───────Z─────────────────────────────────

2: ───────────S──────────────S^-1───S───────

3: ───────────────T───T^-1──────────────T───
```

Note that this test may fail. If so, we may need to change something in the code of `fold_global`. But for the moment, let just try and see if it works as expected.

I'll make a more detailed review on the code in the next days.

mitiq/zne/scaling/tests/test_folding.py
```diff
@@ -960,19 +960,94 @@ def test_global_fold_stretch_factor_eight_terminal_measurements():
...
-            Circuit(
-                [ops.T.on(qreg[2])],
-                [ops.TOFFOLI.on(*qreg)],
-            )
-        ),
+        inverse(circ)[0],
+        circ[-1],
+        meas,
+    )
+    assert _equal(folded, correct)
```
purva-thakre, line 967: @andreamari I have attempted a quick fix for the input circuit below but that caused an issue with an already existing test. The global folded circuit that we get back is a bit different compared to how the original global folding function would insert gates etc. 

Will spend some time trying to understand how to get the intended folded circuit back but wanted to add this comment in case you catch my mistake quickly. 

Input circuit:
0: ───H───────────────

1: ───────Z───────────

2: ───────────S───────

3: ───────────────T───

mitiq/zne/scaling/folding.py
```diff
@@ -297,12 +293,11 @@ def fold_global(
 
     if num_to_fold > 0:
         folded += Circuit(
-            [inverse(operations[-num_to_fold:])], [operations[-num_to_fold:]]
+            [inverse(base_circuit[-num_to_fold:])],
+            [base_circuit[-num_to_fold:]],
```
purva-thakre, line 297: This is currently inserting gates based on moments. Want to access operations in the moment and add them in a similar manner. For this reason, the test `test_global_fold_stretch_factor_eight_terminal_measurements` and the output circuit is not the intended globally folded circuit.

purva-thakre:
@andreamari This is ready for review. I am unsure if doctest failure is due to something I did because I googled the code and ended up a bit confused. 

image

mitiq/zne/scaling/folding.py
```diff
@@ -296,13 +292,32 @@ def fold_global(
...
+
+        for i in range(len(ops_for_partial_layers)):
+            for moment in base_circuit[moment_list[0] :]:
+                new_moment = Moment()
+                for op in moment:
+                    if op == ops_for_partial_layers[i]:
+                        new_moment = new_moment.with_operation(op)
+                circuit_partial.append(new_moment)
+
+        folded += inverse(circuit_partial) + circuit_partial
```
andreamari, line 318: I would simplify the code as in suggestion.
I see the idea of your approach, but it involves many loops which may be more difficult to read and maintain.

```suggestion
        # Create the inverse of the final partial circuit
        inverse_partial = Circuit()
        num_partial = 0
        for moment in base_circuit[::-1]:
            new_moment = Moment()
            for op in moment:
                new_moment = new_moment.with_operation(inverse(op))
                num_partial += 1
                if num_partial == num_to_fold:
                    break
            inverse_partial.append(new_moment)
            if num_partial == num_to_fold:
                break
        # Append partially folded circuit
        folded += inverse_partial + inverse(inverse_partial)
```

mitiq/zne/scaling/tests/test_folding.py
```diff
@@ -1245,6 +1319,41 @@ def test_fold_and_squash_max_stretch(fold_method):
...
+    folded_and_squashed = fold_method(
+        circuit, scale_factor=3.0, squash_moments=True
+    )
+    folded_with_squash_moments_not_specified = fold_method(
+        circuit, scale_factor=3.0
+    )  # Checks that the default is to squash moments
+
+    assert len(folded_not_squashed) == 30
+    assert len(folded_and_squashed) == 30
+    assert len(folded_with_squash_moments_not_specified) == 30
```
andreamari, line 1354: Given that `squash_moments` is not an option anymore, maybe we can remove this test?

If you want to keep it, I would avoid the use of @pytest.mark.parametrize and replace all `fold_method` variables with `fold_global`.

andreamari (changes requested):
Hi @purva-thakre, thanks! I think we are almost there, see suggestions.

mitiq/zne/scaling/folding.py
purva-thakre: Thanks ! yes, I was worried about the many loops I was using but I wanted to use as much of the original code as I could.

mitiq/zne/scaling/tests/test_folding.py
purva-thakre: Initially, I had kept it in to make sure that Cirq's insert strategies would not change the moment structure when `base_circuit` was used directly. [EARLIEST](https://quantumai.google/cirq/circuits#insertstrategies) is the default but the way I understand it, it wants to minimize the moments if there's some space in another moment. 

With the new code, the moment structure will be maintained throughout the folding process. Can you think of a case when insert strategy would cause issues here ? I don't understand them beyond what's provided in the documentation.

mitiq/zne/scaling/folding.py
```diff
@@ -296,13 +292,23 @@ def fold_global(
...
+        # Create the inverse of the final partial circuit
+        inverse_partial = Circuit()
+        num_partial = 0
+        for moment in base_circuit[::-1]:
+            new_moment = Moment()
+            for op in moment:
+                new_moment = new_moment.with_operation(inverse(op))
+                num_partial += 1
+                if num_partial == num_to_fold:
+                    break
```
purva-thakre, line 304: Must reverse the order of operations in a moment before it gets appended in the new moment. This is what's causing the current test failure. 

Note that the failing test also needs to be corrected to what it was before it was changed for this PR.

mitiq/zne/scaling/folding.py
purva-thakre: I have added an [additional check to make sure the correct operation ](https://github.com/unitaryfund/mitiq/pull/1113/files#diff-58c4c4e3cb99c8097c841f4267e46fe9702e78865350b7eb607653a332f77b52R303)is used in the inverse circuit. I don't know if there's away to reverse the order of operations in a moment because we want the last operation to be added first.

mitiq/zne/scaling/tests/test_folding.py
andreamari: @purva-thakre,  `InsertStrategy` is a Cirq option that only acts during the initial construction of the input circuit but has no effects later. In principle one could manually define the input circuit moment by moment and get the same result. 

So I propose to remove this test since the `squash_moments` option no longer exists for `fold_global`.

mitiq/zne/scaling/folding.py
andreamari: I see, good catch on the operation ordering! 

You can use the following to iterate over moment operations in reversed order:
```
for op in moment.operations[::-1]:
     new_moment =...
```
In this way you can get rid of `ops_for_partial_layers`.

andreamari (approved):
LGTM! Thanks @purva-thakre! :+1:

PR #1142, opened 2022-02-24
Add pre-executed pyquil notebooks

andreamari:
Add pre-executed PyQuil notebooks whose re-execution is skipped only on RDT.


andreamari:
@crazy4pi314, @Misty-W, @nathanshammah 

This is could be a way to get PyQuil examples rendered on readthedocs **but still tested in our GitHub CI**.  Maybe not an optimal solution but probably worth doing before releasing 0.13 (tomorrow).

What do you think?

This is a summary of what I did:

- Replaced `.myst` files with pre-executed `.ipynb` files  (only for our two PyQuil examples)
- Defined an environment variable  `SKIP_PYQUIL=True` on readthedocs admin settings
- Conditioned the execution strategy options in `conf.py` based on the value of `SKIP_PYQUIL`.

crazy4pi314:
Seems good to me for the release 👍

crazy4pi314 (approved):

Misty-W:
> Seems good to me for the release 👍

Agree- this is good for the short-term at least. We can revisit if/when we find a better solution.

nathanshammah:
@andreamari great fix!

PR #1166, opened 2022-03-11
Strickt pinning of mypy 

andreamari:
Try pinning mypy


andreamari:
I close this since the issue seems unrelated to `mypy` version.

PR #1175, opened 2022-03-22
Create file structure for digital dynamical decoupling

andreamari:
Description
-----------
This PR just creates the skeleton of the file structure for DDD as planned in the [RFC](https://docs.google.com/document/d/1cRwFCTn6kUjI1P0kNydtevxIYtE4r8Omd_iWK0Pe8qo/edit#heading=h.mckr847qq60k).

### DIFF
--- a/mitiq/ddd/__init__.py
+++ b/mitiq/ddd/__init__.py
@@ -1,0 +1,16 @@
+# Copyright (C) 2022 Unitary Fund
+#
+# This program is free software: you can redistribute it and/or modify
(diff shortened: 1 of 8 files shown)


Aaron-Robertson (commented):
LGTM @andreamari. Only question is should we include the `apidoc.rst` stubs here?

andreamari:
I would wait untill we have a MVP for DDD before changing the docs. Otherwise users may read "dynamical decoupling" and think it is implemented.

PR #1178, opened 2022-03-23
Add get_circuit_mask() for dynamical decoupling

### LINKED ISSUE #1160: Implement get_circuit_mask function for dynamical decoupling
Implement the `get_circuit_mask()` function described in the [RFC for dynamical decoupling](https://docs.google.com/document/d/1cRwFCTn6kUjI1P0kNydtevxIYtE4r8Omd_iWK0Pe8qo/edit?usp=sharing). 

This is necessary to determine slack windows (holes with idling qubits) in a circuit.

An implementation skeleton with pseudo-code is present in the [RFC](https://docs.google.com/document/d/1cRwFCTn6kUjI1P0kNydtevxIYtE4r8Omd_iWK0Pe8qo/edit?usp=sharing) but, of course,  a different approach can be used.

Aaron-Robertson:
Description
-----------

Fixes #1160. Adds `get_circuit_mask()` function to `.../ddd/insertion.py`.

License
-------


mitiq/ddd/insertion.py
```diff
@@ -0,0 +1,39 @@
...
+from mitiq import QPROGRAM
+
+
+def get_circuit_mask(circuit: QPROGRAM) -> np.ndarray:
+    """Given a circuit with n qubits and d moments returns a matrix
+    A with n rows and d columns. The matrix elements are A_{i,j} = 1 if
+    there is a gate acting on qubit i at moment j, while A_{i,j} = 0 otherwise.
+    """
+    mitiq_circuit, _ = convert_to_mitiq(circuit)
+    qubits = sorted(mitiq_circuit.all_qubits())
```
andreamari, line 29: I know I suggested sorting in the RFC, but I am actually not sure. If it works without sorting, it is probably better to remove it. Otherwise we risk unexpected re-ordering of qubits. 
Do you agree?

mitiq/ddd/tests/test_insertion.py
```diff
@@ -0,0 +1,112 @@
...
+
+def test_get_circuit_mask_one():
+    circuit1 = cirq.Circuit(
+        cirq.SWAP(q, q + 1) for q in cirq.LineQubit.range(7)
+    )
+    circuit_mask = get_circuit_mask(circuit1)
+    assert np.allclose(circuit_mask, test_mask)
+
+
+# Define test mask matrices
```
andreamari, line 45: The code below this line is from a different PR. I think we can remove it untill we merge both PRs to master.

andreamari:
Thanks @Aaron-Robertson, even if this is a Draft PR I had a preliminary look and left a few comments. Looks already very good!

mitiq/ddd/tests/test_insertion.py
Aaron-Robertson: Agreed!

mitiq/ddd/insertion.py
andreamari: Good catch! So I agree on sorting. 
I noticed that a similar sorting is implicitly applied when calling print(circuit). So the mask that we get when sorting should match the printed circuit and this is good.

andreamari:
LGTM! Could you marge master on this branch to solve the conflicts? After all test should be triggered. If the docs test still fails, we'll find a way to fix it.

mitiq/ddd/insertion.py
```diff
@@ -0,0 +1,44 @@
...
+    A with n rows and d columns. The matrix elements are A_{i,j} = 1 if
+    there is a gate acting on qubit i at moment j, while A_{i,j} = 0 otherwise.
+
+    Args:
+        circuit: Input circuit to mask with n qubits and d moments
+
+    Returns:
+        A mask matrix with n rows and d columns
+    """
+    mitiq_circuit, _ = convert_to_mitiq(circuit)
```
andreamari, line 33: I see that this function is compatible with any `QPROGRAM` but we test it only with Cirq circuits.
With other circuits there could be some problems (e.g. if a Braket gate is converted to a Cirq decomposition involving more gates the Cirq mask could be unexpected with respect to the input Braket circuit).
I see two options to proceed:

1. Explaining this potential unexpected behavior in the docstring. And maybe add a test with a different front end? 
2. Making this function private and compatible only with Cirq circuits : e.g.  _gate_circuit(circuit: cirq.Circuit): -> np.ndarray

Both options are good for me.

andreamari (commented):
@Aaron-Robertson I have a small doubt about the interoperability with different frontends. The rest looks good!

mitiq/ddd/insertion.py
Aaron-Robertson: Initially I went with (1) but looking over the RFC I think it will make the most sense to apply conversions with the decorating converter flow against `insert_ddd_sequences()`. So now it's (2).

Aaron-Robertson:
@andreamari I merged in master but the docs test still failed. Re-running it now. I don't see why it failed so let me know if you can offer any insights

andreamari (approved):
Thanks @Aaron-Robertson, LGTM!
We can ignore the GitHub pages test since it's unrelated to the content of this PR.

PR #1185, opened 2022-03-25
Idle qubits in conversion

Aaron-Robertson:
Description
-----------

Naive approach to preserve idle qubits on conversion: insert identity gate, which then gets folded, and then remove all id gates (current flow can be made to target only idle qubits, but doesn't yet). Introduces test failures and folding inefficiency because id gates are folded then removed, but otherwise works.

Checklist
---------

License
-------

Tips
----

### DIFF
--- a/mitiq/interface/mitiq_qiskit/conversions.py
+++ b/mitiq/interface/mitiq_qiskit/conversions.py
@@ -274,7 +274,6 @@ def from_qiskit(circuit: qiskit.QuantumCircuit) -> cirq.Circuit:
     Returns:
         Mitiq circuit representation equivalent to the input Qiskit circuit.
     """
-    _add_identity_to_idle(circuit)
     return from_qasm(circuit.qasm())
 
 
--- a/mitiq/interface/mitiq_qiskit/conversions.py
+++ b/mitiq/interface/mitiq_qiskit/conversions.py
@@ -104,8 +104,8 @@ def _map_qubits(
         qubits: A list of qubits to map.
         new_register_sizes: The size(s) of the new registers to map to.
             Note: These can be determined from ``new_registers``, but this
-            helper function is only called from ``_map_qubits`` where the sizes
-            are already computed.
+            helper function is only called from ``_transform_registers`` where
+            the sizes are already computed.
         new_registers: The new registers to map the ``qubits`` to.
 
     Returns:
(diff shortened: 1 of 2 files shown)


Aaron-Robertson:
@andreamari @rmlarose Let me know what you think about this approach, could certainly be made to work but I think there's a better way. Perhaps with a mask created against the input circuit, and applied post register transform.

This approach can also solve the issue (technically speaking) when applied in the `to_qiskit()` and `from_qiskit()` methods but introduces another ~50 test failures.

mitiq/interface/mitiq_qiskit/conversions.py
```diff
@@ -118,6 +118,36 @@ def _map_qubits(
...
+def _add_identity_to_idle(
+    circuit: qiskit.QuantumCircuit,
+) -> List[int]:
+    data = copy.deepcopy(circuit._data)
+    bit_indices = set()
+    idle_bit_indices = []
+    for op in data:
+        gate, qubits, cbits = op
+        bit_indices.update(set(bit.index for bit in qubits))
+    for index in range(len(circuit)):
```
andreamari, line 130:
```suggestion
    for index in range(circuit._num_qubits):
```
Am I wrong about this?

andreamari:
The implemented approach seems good to me!

The only aspect that is a bit worrying to me is the risk of eliminating identities that may have been added by the user and not by Mitiq. The final function is not able to distinguish between identities added by the user and those added by Mitiq.

It may be hard/impossible to solve this aspect. If so, I would maybe prefer not removing identities at all instead
of taking the risk of removing identity gates added by the user.

Aaron-Robertson:
> The only aspect that is a bit worrying to me is the risk of eliminating identities that may have been added by the user and not by Mitiq. The final function is not able to distinguish between identities added by the user and those added by Mitiq.

Since both `_add_identity_to_idle` and `_add_identity_to_idle` are called from within the decorated converter flow, I planned to pass the list of idle indices back to target only the idle qubits in removing identities. I'll implement it now!

andreamari (changes requested):
Great solution! 
There could be problems in the future if we need to add DDD sequences with identities (wait moments) in idle qubits, but I would not worry about it at this moment. 

Could you also add a test in which a Qiskit circuit with idle qubits is converted to Mitiq and back to Qiskit preserving idle qubits? 
And/Or a test in which a Qiskit circuit with idle qubits is folded preserving idle qubits?

andreamari:
> > Could you also add a test in which a Qiskit circuit with idle qubits is converted to Mitiq and back to Qiskit preserving idle qubits?
> > And/Or a test in which a Qiskit circuit with idle qubits is folded preserving idle qubits?
> 
> Yep, sounds good. Best to add those tests to `test_folding.py`?

I don't have a strong opinion. Wherever you prefer is good for me.

andreamari (approved):

andreamari:
Thanks @Aaron-Robertson :rocket: !

PR #1193, opened 2022-03-29
Update Cirq to 0.14

andreamari:
Simplifies the `_equal()` function to check circuit equality.
This should allow us to use the latest version of Cirq.


crazy4pi314 (approved):
LGTM! 🎉

PR #1195, opened 2022-03-30
Update Qiskit and fix conversions of qubit registers

andreamari:
Update Qiskit and fix conversions of qubit registers.

### DIFF
--- a/mitiq/interface/mitiq_qiskit/conversions.py
+++ b/mitiq/interface/mitiq_qiskit/conversions.py
@@ -247,6 +247,7 @@ def _transform_registers(
     circuit._qubit_set = set()
     circuit.qregs = []
     circuit._data = []
+    circuit._qubit_indices = {}
     circuit.add_register(*new_qregs)
 
     # Map the qubits in operations to the new qubits.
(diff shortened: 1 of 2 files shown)


mitiq/interface/mitiq_qiskit/conversions.py
```diff
@@ -247,6 +247,7 @@ def _transform_registers(
     circuit._qubit_set = set()
     circuit.qregs = []
     circuit._data = []
+    circuit._qubit_indices = {}
```
Misty-W, line 250: Just checking my understanding, before this change inconsistencies could arise in the converted circuits because the "old" qubit indices were carried through the conversion?

Misty-W (approved):

Misty-W:
@crazy4pi314, I saw you made some changes earlier, are we good to merge?

mitiq/interface/mitiq_qiskit/conversions.py
andreamari: Yes, exactly. Moreover, without resetting `_qubit_indeces` the function `_transform_registers()` sometimes failed to add some new qubits as expected.

PR #1196, opened 2022-03-30
Avoid creation of empty moments when using fold_global

andreamari:
After recent changes (#653), `fold_global` preserves the moment structure and this is good!
However, since now moment are not squashed, one can have unexpected appearance of empty moments when using the `fold_global()` function. 

E.g., with the current code we have:

```
>>> c = Circuit(X(LineQubit(0)), measure(LineQubit(0)))
>>> folded = zne.scaling.fold_global(c, 3)
>>> print(c)
0: ───X───M───
>>> print(folded)
0: ───X────────X───X───M───
```

After this PR, the function pop_measurements is updated, such that we get the expected behavior:

```
>>> c = Circuit(X(LineQubit(0)), measure(LineQubit(0)))
>>> folded = zne.scaling.fold_global(c, 3)
>>> print(c)
0: ───X───M───
>>> print(folded)
0: ───X───X───X───M───
```

### DIFF
--- a/mitiq/utils.py
+++ b/mitiq/utils.py
@@ -105,7 +105,7 @@ def _pop_measurements(
     if measurements:
         circuit.batch_remove(measurements)
         # Remove the last moment too if left empty.
-        if len(circuit[-1])==0:
+        if len(circuit[-1]) == 0:
             del circuit[-1]
     return measurements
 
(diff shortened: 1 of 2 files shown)


andreamari:
@crazy4pi314, @nathanshammah,
If we make a release tomorrow or this week, it is important to also include this change (to avoid unexpected results in the stable version). Otherwise, it's not urgent.

andreamari:
@crazy4pi314, @nathanshammah,
If we make a release tomorrow or this week, it is important to also include this change (to avoid unexpected results in the stable version). Otherwise, it's not urgent.

PR #1202, opened 2022-04-04
Add rules for dynamical decoupling 

Aaron-Robertson:
Description
-----------

Please explain the changes you made here.

Checklist
---------

License
-------

Tips
----

### DIFF
--- a/mitiq/ddd/rules/rules.py
+++ b/mitiq/ddd/rules/rules.py
@@ -16,3 +16,143 @@
 """Built-in rules determining what DDD sequence should be applied in a given
 slack window.
 """
+
+from cirq import Circuit, X, Y, I, LineQubit  # , Z
+
+
+def xx(
+    slack_length: int, num_repetitions: int = 1, spacing: int = -1
+) -> Circuit:
+    """Returns an XX digital dynamical decoupling sequence, based on inputs.
+
+    Args:
+        slack_length: Length of idle window to fill.
+        num_repetitions: How many repetitions of the dd rule to apply. Default
+            to 1.
+        spacing: How many identity spacing gates to apply between dynamical
+            decoupling gates, as a positive-value int. Defaults to evenly
+            spaced.
+    Returns:
+        An XX digital dynamical decoupling sequence, as a cirq circuit
+    """
+    default_length = 2
+    num_decoupling_gates = default_length * num_repetitions
+    slack_difference = slack_length - num_decoupling_gates
+    if spacing < 0:
+        spacing = slack_difference // num_decoupling_gates
+    uniform_spacing = (
+        1
+        if spacing == 1
+        else spacing * num_decoupling_gates // (num_decoupling_gates + 1)
+    )
+    slack_remainder = slack_difference - num_decoupling_gates * spacing
+    q = LineQubit(0)
+    slack_gates = [I(q) for _ in range(uniform_spacing)]
+    slack_fill = [I(q) for _ in range(slack_remainder)]
+    ddd_circuit = Circuit(
+        slack_fill[: slack_remainder // 2],
+        slack_gates,
+        [
+            (
+                X(q),
+                slack_gates,
+            )
+            for _ in range(num_decoupling_gates)
+        ],
+        slack_fill[slack_remainder // 2 :],
+    )
+    return ddd_circuit
+
+
+def xyxy(
+    slack_length: int, num_repetitions: int = 1, spacing: int = -1
+) -> Circuit:
+    """Returns an XYXY digital dynamical decoupling sequence, based on inputs.
+
+    Args:
+        slack_length: Length of idle window to fill.
+        num_repetitions: How many repetitions of the dd rule to apply. Default
+            to 1.
+        spacing: How many identity spacing gates to apply between dynamical
+            decoupling gates, as a positive-value int. Defaults to evenly
+            spaced.
+    Returns:
+        An XYXY digital dynamical decoupling sequence, as a cirq circuit
+    """
+    default_length = 4
+    num_decoupling_gates = default_length * num_repetitions
+    slack_difference = slack_length - num_decoupling_gates
+    if spacing < 0:
+        spacing = slack_difference // num_decoupling_gates
+    uniform_spacing = (
+        1
+        if spacing == 1
+        else spacing * num_decoupling_gates // (num_decoupling_gates + 1)
+    )
+    slack_remainder = slack_difference - num_decoupling_gates * spacing
+    q = LineQubit(0)
+    slack_gates = [I(q) for _ in range(uniform_spacing)]
+    slack_fill = [I(q) for _ in range(slack_remainder)]
+    ddd_circuit = Circuit(
+        slack_fill[: slack_remainder // 2],
+        slack_gates,
+        [
+            (
+                X(q),
+                slack_gates,
+                Y(q),
+                slack_gates,
+                X(q),
+                slack_gates,
+                Y(q),
+                slack_gates,
+            )
+            for _ in range(num_repetitions)
+        ],
+        slack_fill[slack_remainder // 2 :],
+    )
+    return ddd_circuit
+
+
+def yy(
+    slack_length: int, num_repetitions: int = 1, spacing: int = -1
+) -> Circuit:
+    """Returns a YY digital dynamical decoupling sequence, based on inputs.
+
+    Args:
+        slack_length: Length of idle window to fill.
+        num_repetitions: How many repetitions of the dd rule to apply. Default
+            to 1.
+        spacing: How many identity spacing gates to apply between dynamical
+            decoupling gates, as a positive-value int. Defaults to evenly
+            spaced.
+    Returns:
+        An YY digital dynamical decoupling sequence, as a cirq circuit
+    """
+    default_length = 2
+    num_decoupling_gates = default_length * num_repetitions
+    slack_difference = slack_length - num_decoupling_gates
+    if spacing < 0:
+        spacing = slack_difference // num_decoupling_gates
+    uniform_spacing = (
+        1
+        if spacing == 1
+        else spacing * num_decoupling_gates // (num_decoupling_gates + 1)
+    )
+    slack_remainder = slack_difference - num_decoupling_gates * spacing
+    q = LineQubit(0)
+    slack_gates = [I(q) for _ in range(uniform_spacing)]
+    slack_fill = [I(q) for _ in range(slack_remainder)]
+    ddd_circuit = Circuit(
+        slack_fill[: slack_remainder // 2],
+        slack_gates,
+        [
+            (
+                Y(q),
+                slack_gates,
+            )
+            for _ in range(num_decoupling_gates)
+        ],
+        slack_fill[slack_remainder // 2 :],
+    )
+    return ddd_circuit
--- a/mitiq/ddd/__init__.py
+++ b/mitiq/ddd/__init__.py
@@ -14,3 +14,4 @@
 # along with this program.  If not, see <https://www.gnu.org/licenses/>.
 
 """Digital dynamical decoupling (DDD) module."""
+from mitiq.ddd import rules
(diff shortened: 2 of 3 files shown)


mitiq/ddd/rules/tests/test_rules.py
```diff
@@ -14,3 +14,89 @@
...
+    @pytest.mark.parametrize(
+        "rule",
+        [
+            xx(slack_length),
+            xyxy(slack_length),
+            yy(slack_length),
+        ],
+    )
+    def test_rule(rule):
+        assert len(rule) == slack_length
```
andreamari, line 38:
```suggestion
    assert len(rule) == slack_length
```

In other Mitiq tests we usually don't define test sub functions when using multiple `pytest.mark.parametrize`, we just stack multiple `pytest.mark.parametrize` decorators and define a unique test function with multiple arguments. But actually I am not sure what is the best practice.

mitiq/ddd/rules/tests/test_rules.py
```diff
@@ -14,3 +14,89 @@
...
+
+    @pytest.mark.parametrize("spacing", [i for i in range(slack_length)])
+    def user_spacing(spacing):
```
andreamari, line 41:
```suggestion
    def test_spacing_option_in_rules(spacing):
```
Same comment about nested functions vs unique test function with multiple parameters.

andreamari (changes requested):
Great @Aaron-Robertson, very good implementation of ddd rules! 
In addition to inline comments I have two important suggestions:

1) In tests we just check the length of sequences. Can we also add a few tests checking that sequences produced by rule are exactly what we expect? You could use [the _equal function](https://github.com/unitaryfund/mitiq/blob/fd33cc7abf07ed631d4a69396dc3212670c43908/mitiq/utils.py#L129)  _equal() check equality between output sequences and the associated expected circuits.

2) All functions return Cirq circuits only. Unless you want to make them multi-frontend in this PR, we can open a new issue for this. We already did something similar for other DDD functions.

mitiq/ddd/rules/tests/test_rules.py
Aaron-Robertson: I included the sub functions here because the test parametrization(s) require `slack_length`. I don't remember best practice either at the moment.

mitiq/ddd/rules/rules.py
```diff
@@ -16,3 +16,110 @@
...
+        spacing = slack_difference // (num_decoupling_gates + 1)
+    slack_remainder = slack_length - (
+        spacing * (num_decoupling_gates + 1) + num_decoupling_gates
+    )
+    if slack_remainder < 0:
+        raise ValueError(
+            "Rule too long for given slack window by {} moments.".format(
+                slack_remainder * -1
+            )
+        )
```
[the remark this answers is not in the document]
Aaron-Robertson: So, the `slack_remainder` accounts for spacing per window and number of gates in a rule. Given `n` gates there are `n+1` spacing windows, and the `spacing` length is a user input. `slack_difference` doesn't cover a user's attempts to insert too much spacing for a given slack window.

mitiq/ddd/rules/rules.py
```diff
@@ -16,3 +16,110 @@
...
+
+
+def construct_rule(
+    slack_length: int, spacing: int, gates: List[Gate]
+) -> Circuit:
+    """Returns a digital dynamical decoupling sequence, based on inputs.
+
+    Args:
+        slack_length: Length of idle window to fill.
+        spacing: How many identity spacing gates to apply between dynamical
```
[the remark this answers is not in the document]
Aaron-Robertson: Hmm, it is confusing. I wouldn't say that even spacing is always used, e.g.:

If `spacing=2` is provided with `slack_length=20` then the behavior is as follows
`print(xx(20, 2))`
yields
`0: ───I───I───I───I───I───I───I───I───X───I───I───X───I───I───I───I───I───I───I───I───`

mitiq/ddd/rules/tests/test_rules.py
Aaron-Robertson: Apparently sub function tests obfuscated some errors, so I de-nested.

mitiq/ddd/rules/rules.py
```diff
@@ -22,22 +22,31 @@
...
+            given slack_length = 20, gates [X, Y, X, Y] the spacing defaults
+            to 3 and the rule returns the sequence:
+            ──I──I──I──I──X──I──I──I──Y──I──I──I──X──I──I──I──Y──I──I──I──
         gates: A list of Cirq gates to build the rule. E.g. [X, X] is the xx
             sequence, [X, Y, X, Y] is the xyxy sequence
             - Note: To repeat the sequence, specify a repeated gateset
     Returns:
         A digital dynamical decoupling sequence, as a cirq circuit
     """
+    if slack_length < 2 or slack_length < len(gates):
```
andreamari, line 48: I realized some potential errors that we should probably check:
- Error if gates are not single-qubit gates. I think this is important.
- (Maybe) Error if `cirq.unitary(Circuit(gates))` is not np.close to the identity matrix. 
- (Maybe) Error if `len(gates) < 2` , unless we want to support the limit case `gates=[ I ]`.

andreamari (commented):
Hi @Aaron-Robertson, docs are perfect now! 
Two doubts came to my mind only now:

1. We should check if the custom gates of the sequence make sense. At least the should be single-qubit gates based on the current code structure. 
2. An important decision that we must take before merging is what happens if a gate sequence doesn't fit a slack window. Now we give an error, however it may be more practical if a rule function just returns an empty circuit (or `None`) instead of giving an error. Otherwise, when using the rule one needs to use many times the `try-except` statement and I am not sure if it is a good practice.

Sorry for these late comments, I should have thought about these questions in my first review. 
If you think these are not important concerns I am ready to change my mind.
I am also at the Wednesday/Friday calls if you want to have a live discussion.

mitiq/ddd/rules/rules.py
```diff
@@ -16,3 +16,131 @@
...
+            to 6 and the rule returns the sequence:
+            ──I──I──I──I──I──I──X──I──I──I──I──I──I──X──I──I──I──I──I──I──
+            given slack_length = 20, gates [X, Y, X, Y] the spacing defaults
+            to 3 and the rule returns the sequence:
+            ──I──I──I──I──X──I──I──I──Y──I──I──I──X──I──I──I──Y──I──I──I──
+        gates: A list of Cirq gates to build the rule. E.g. [X, X] is the xx
+            sequence, [X, Y, X, Y] is the xyxy sequence
+            - Note: To repeat the sequence, specify a repeated gateset
+    Returns:
+        A digital dynamical decoupling sequence, as a cirq circuit
```
andreamari, line 46:
```suggestion
        A digital dynamical decoupling sequence, as a Cirq circuit.
```

mitiq/ddd/rules/rules.py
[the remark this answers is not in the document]
andreamari: I see. Makes sense.

Aaron-Robertson:
> Two doubts came to my mind only now:
> 
> 1. We should check if the custom gates of the sequence make sense. At least the should be single-qubit gates based on the current code structure.
> 2. An important decision that we must take before merging is what happens if a gate sequence doesn't fit a slack window. Now we give an error, however it may be more practical if a rule function just returns an empty circuit (or `None`) instead of giving an error. Otherwise, when using the rule one needs to use many times the `try-except` statement and I am not sure if it is a good practice.

Agreed, let's discuss on the call today.  I think an empty circuit works for most cases. 

Another case we may want to (re)consider is how to handle multiple rules (and repetitive rules) in the insert strategy. E.g. I don't see a good way to reproduce the sequences from https://arxiv.org/abs/2101.08870:

> The difference between the two cases is that the ancilla qubit remains completely idle during Uˆ and Uˆ † in Fig. S10A, whereas a train of spin echo pulses X − X − Y − Y − X − X... is applied to the ancilla qubit during Uˆ and Uˆ † in Fig. S10B.

We could accept sequences of rules in the `insert_ddd_sequences()` method, where shorter rules act as a fallback in the strategy when long rules are provided. Say an 8-gate rule is provided, then a 2-gate fallback could be provided to fill shorter slack windows. I haven't considered that idea enough yet though.

andreamari:
> > Two doubts came to my mind only now:
> > 
> > 1. We should check if the custom gates of the sequence make sense. At least the should be single-qubit gates based on the current code structure.
> > 2. An important decision that we must take before merging is what happens if a gate sequence doesn't fit a slack window. Now we give an error, however it may be more practical if a rule function just returns an empty circuit (or `None`) instead of giving an error. Otherwise, when using the rule one needs to use many times the `try-except` statement and I am not sure if it is a good practice.
> 
> Agreed, let's discuss on the call today. I think an empty circuit works for most cases.
> 
> Another case we may want to (re)consider is how to handle multiple rules (and repetitive rules) in the insert strategy. E.g. I don't see a good way to reproduce the sequences from https://arxiv.org/abs/2101.08870:
> 
> > The difference between the two cases is that the ancilla qubit remains completely idle during Uˆ and Uˆ † in Fig. S10A, whereas a train of spin echo pulses X − X − Y − Y − X − X... is applied to the ancilla qubit during Uˆ and Uˆ † in Fig. S10B.
> 
> We could accept sequences of rules in the `insert_ddd_sequences()` method, where shorter rules act as a fallback in the strategy when long rules are provided. Say an 8-gate rule is provided, then a 2-gate fallback could be provided to fill shorter slack windows. I haven't considered that idea enough yet though.

Summary of the discussion we had during the community call on April 13:

- Fine to have empty circuits instead of giving errors
- Fallback sequences can be obtained with current code using two passes of DDD: first filling long sequences then filling short sequeces.
- It may be still useful to define rules that fill slack windows with as many repetitions as possible. If so, it is probably better to introduce them as additional rules (functions) instead of adding a lot of options in the current functions.

mitiq/ddd/rules/rules.py
Aaron-Robertson: Currently the error thrown when a multi-qubit gate is provided as input bubbles up from `cirq` (because we assume a single qubit):
`[ValueError: Wrong number of qubits for <cirq.CNOT>. Expected 2 qubits but got <[cirq.LineQubit(0)]>.]()`

Do you think that's sufficient? I think that's fine, though we might want to check that the gate(s) don't have qubit assignment. Currently the error thrown is `[AttributeError: 'GateOperation' object has no attribute 'on']()` when a `GateOperation` is passed to a rule (in the 1 or 2 qubit gate case).

mitiq/ddd/rules/rules.py
```diff
@@ -16,3 +16,169 @@
...
+    until the slack is filled without spacing, up to a complete repetition.
+
+    Args:
+        slack_length: Length of idle window to fill.
+        gates: A list of single qubit Cirq gates to build the rule. E.g. [X, X]
+            is the xx sequence, [X, Y, X, Y] is the xyxy sequence.
+    Returns:
+        A repeated digital dynamical decoupling sequence, as a Cirq circuit.
+    Note:
+        Where the general rule requires pre-knowledge of the amount of slack,
```
andreamari, line 165:
```suggestion
        Where :fun:`.general_rule()` fills a slack window with single sequence,
```

I know what you mean but in practice both functions need to know `slack_length` as input.

mitiq/ddd/rules/rules.py
```diff
@@ -16,3 +16,169 @@
...
+        this rule attempts to fill every moment with sequence repetitions
+        (up to a complete repetition of the gate set).
+        E.g. given slack_length = 17 and gates = [X, Y]
+            this rule returns the sequence:
+            ──I──X──Y──X──Y──X──Y──X──Y──X──Y──X──Y──X──Y──X──Y──
+    """
+    num_decoupling_gates = len(gates)
+    sequence = general_rule(
```
andreamari, line 173: Good idea passing through `general_rule()` here!

andreamari (commented):
Thanks @Aaron-Robertson! 
Minor comments below.

Aaron-Robertson:
@andreamari this one should be ready to go! One final concern is if I added too many tests. Could you ensure locally that the suite isn't too onerous?

andreamari:
> @andreamari this one should be ready to go! One final concern is if I added too many tests. Could you ensure locally that the suite isn't too onerous?

DDD tests are almost instantaneous!

andreamari (approved):

PR #1225, opened 2022-04-13
Fix compilation to native gateset in Braket notebook

andreamari:
This shouldn't have any consequence on Mitiq code and Mitiq docs. But it's worth fixing it.
@vprusso the same should be fixed in the "metric notebook". Simulations should not change anyway. Important for future hardware runs.


vprusso (approved):
Nice catch. I've made sure to propagate this change to our other notebooks that make use of this compilation step.

Issue #1194, opened 2022-03-30
MEP description in docs needs to be updated


purva-thakre:
Considering most of the accepted enhancement proposals so far have been in docs, I think it's better to link to [another template in docs](https://docs.google.com/document/d/1adomheXpbqp4YIBFQ49IsAJzuJKWyr75GRO1NeWg0Fo/edit?usp=sharing). 

It would also be better to avoid the confusion between RFC and MEP by updating[ the documentation](https://mitiq.readthedocs.io/en/stable/contributing.html#proposing-a-new-feature-to-mitiq). I propose keeping both the `md` and `doc` versions just in case. 

The `md` template also has a missing [link](https://github.com/unitaryfund/mitiq/blob/master/docs/mep/0000-feature-proposal-TEMPLATE.md#references-and-footnotes) to another MEP. 

![image](https://user-images.githubusercontent.com/66048318/160828665-1ba183ce-4250-40c3-8a1c-eb690a03aa5d.png)

nathanshammah:
Interesting related news: Markdown supported in google docs https://workspaceupdates.googleblog.com/2022/03/compose-with-markdown-in-google-docs-on.html

Issue #815, opened 2021-07-10
New technique: Measurement error mitigation


crazy4pi314:
### Problem

Mitiq needs more error mitigation techniques.

Measurement error mitigation is based on [Measurement Error Mitigation for Variational Quantum Algorithms](https://arxiv.org/abs/2010.08520), should be relatively easy to implement.

### Rough requirements

- Should be implemented in a way such that it is very easy to compose with existing techniques (e.g., ZNE).
- Requires support for measurement results.

rmlarose:
#844 added on technique for measurement error mitigation, more to come!

rmlarose:
RFC for this issue here, please review:

https://docs.google.com/document/d/1Mb-OoojXBm0k8VTapNUkQQi4YQR5vsUvNJqmVgFpIns/edit?usp=sharing

amirebrahimi:
I'd like to offer to take this on as part of a [final project for QCourse511](https://gitlab.com/qworld/qeducation/qcourse511-1/-/issues/21). Final deliverables are due January 11th though, so my question to the Mitiq team is whether it would be feasible to reach an agreement on the design to allow implementation time before that deadline? At the very least I could perform a preliminary implementation based on the current state of the design doc.

rmlarose:
Thanks @amirebrahimi. You are more than welcome to take this up but I have concerns about the Jan 11th date, especially with the holiday break - there are a few points that need to be worked out in the RFC, then once that is accepted, we need to review the PR(s).

As a practical suggestion for your final project, you may consider doing your own complete implementation on a fork so you are not slowed down by the review process. You can of course make a pull request with this at any time. I hesitate to promise it can be merged before Jan 11th though.

amirebrahimi:
> Thanks @amirebrahimi. You are more than welcome to take this up but I have concerns about the Jan 11th date, especially with the holiday break - there are a few points that need to be worked out in the RFC, then once that is accepted, we need to review the PR(s).
> 
> As a practical suggestion for your final project, you may consider doing your own complete implementation on a fork so you are not slowed down by the review process. You can of course make a pull request with this at any time. I hesitate to promise it can be merged before Jan 11th though.

I think that is a good plan and anything after the 11th can be worked out as the RFC firms up.

andreamari:
[...]

PR #1294, opened 2022-05-17
RCI MEP

amirebrahimi:
Description
-----------
I've followed the guidelines for contributing and have written up a new MEP. I wasn't sure about the MEP type (e.g. "Standards Track" vs. "Process") or the resolution field. 

This work has mostly been completed already, so we can either interact via this MEP or, optionally, I can create a draft PR and we can have a discussion there.

### DIFF
--- a/docs/mep/0001-readout-confusion-inversion.md
+++ b/docs/mep/0001-readout-confusion-inversion.md
@@ -1,0 +1,168 @@
+---
+author: Amir Ebrahimi <github@aebrahimi.com>
+status: Draft
+type: <Standards Track | Process>
+created: 2022-05-16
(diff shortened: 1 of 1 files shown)


andreamari:
Hi @amirebrahimi, this is great thanks!

Given the new contribution policy discussed in https://github.com/unitaryfund/mitiq/issues/1194,
I copied and pasted your md file on a shared google doc (see last comment of #815). This will make discussions and comments easier.

So I am closing this PR in favor of the new shared document in the last comment of #815.

amirebrahimi:
I didn't realize that Google Docs was the new approach! I would have probably just updated the existing RFC. It would definitely be good to mark that as a "good first issue" for the upcoming Unitary Hack. At least removing the old template would be a good idea.

Issue #1094, opened 2022-01-18
[unitaryhack-bounty] Improve H2 example


andreamari:
The current H2 example could be improved in two different aspects:

- [ ] The numerical values of the Hamiltonian coefficients _g_j(R)_ are hard-coded from data extracted from a paper. It would be nice to derive them using some chemistry package e.g.: [PySCF](https://pyscf.org/) or [OpenFermion](https://github.com/quantumlib/OpenFermion). Ideally (but only if not too complicated), it would be better to use the same Hamiltonian in order to reproduce the same results of the Mitiq paper. 
- [ ] Use a `mitiq.Observable` for the Hamiltonian, and just use `mitiq_cirq.compute_density_matrix` for the executor. See the docs for more details on how to define and use observables.

If you only fixed one of the two aspects, please don't close this issue but just mark the corresponding item in the above list.

nathanshammah:
+1 for adding a bounty on UH.

andreamari:
@obliviateandsurrender, thanks! Please note that there is already a PR by @lockwo for the same issue.
On the other hand, it is interesting that both of you solved the issue but in very different ways: @lockwo with OpenFermion and you with PennyLane. So we may decide to keep both examples!

Next week we'll discuss internally and let you know if we'll keep both examples and, if so, how it works with the unitaryhack bounty.

andreamari:
Hi @obliviateandsurrender, I have some good news!
We discussed about this and decided to also keep your H2 example in addition to the previous one by @lockwo. 
If both PRs will be accepted and merged, both of you will the bounty!

PR #1296, opened 2022-05-17
[unitaryhack] Updated molecular hydrogen tutorial

lockwo:
Description
-----------

This is made to address https://github.com/unitaryfund/mitiq/issues/1094. The goal is to address both components of that issue. I made it so the hamiltonian is calculated using openfermion. The hamiltonian is not the same as in the paper (still 2 qubits, just not the same ops). It would be very easy to programmatically change that, I just don't know how to do that exact symmetry reduction in openfermion. Additionally, the hamiltonian is a mitiq.Observable and the executor is no longer a wrapper for the energy function but just the density matrix simulator. I am new to mitiq so this second part is what I think that issue meant, but if it should change just let me know. 

Checklist
---------

- [X] I used [Google-style](https://sphinxcontrib-napoleon.readthedocs.io/en/latest/example_google.html) docstrings for functions.

License
-------

Tips
----


lockwo:
I think the check is unsuccessful (`Couldn't find cache key for notebook file source/examples/molecular_hydrogen.myst`) because OpenFermion isn't installed (based on https://github.com/executablebooks/jupyter-cache/issues/51).

andreamari:
Thanks @lockwo, very interesting! 
You can try to add `openfermion` in  https://github.com/unitaryfund/mitiq/blob/master/dev_requirements.txt  under the _documentation and examples_ section.
Not sure if it works (especially with windows tests) but worth trying. 

As a second option, we can try to add `!pip install openfermion` directly on the notebook.

nathanshammah:
Thanks @lockwo. I like the fact that the hard coded table is replaced with values taken from `generate_molecular_hamiltonian`. 
- It would be nice to add a bit of details about this function and its parameters. 
- It would be nice to keep a reference of {cite}`OMalley_2016_PRX`, even if now the data is not directly taken from there.

lockwo:
> Thanks @lockwo. I like the fact that the hard coded table is replaced with values taken from `generate_molecular_hamiltonian`.
> 
> * It would be nice to add a bit of details about this function and its parameters.
> * It would be nice to keep a reference of {cite}`OMalley_2016_PRX`, even if now the data is not directly taken from there.

I expanded a bit on the function description. Is there anything specific you would like to see? As for the citation, I still see it there (in the image). Did you want that referenced again when talking about the coefficients? 
image

lockwo:
> Thanks @lockwo, very interesting! You can try to add `openfermion` in https://github.com/unitaryfund/mitiq/blob/master/dev_requirements.txt under the _documentation and examples_ section. Not sure if it works (especially with windows tests) but worth trying.
> 
> As a second option, we can try to add `!pip install openfermion` directly on the notebook.

Great, this first suggestion worked!

andreamari:
Hi @lockwo,
as you mentioned, the Hamiltonian produced by openfermion is probably different from the formula visible in the comment above.
Any idea to fix this inconsistency?
Some ideas from my side:

- Change the formula to something more generic e.g. a sum over generic Pauli operators.
- Change the formula to something exactly matching the operators produced by openfermion
- Keep the previous table of parameters and add openfermion as a "second option" to generate the Hamiltonian.
- Other options are welcome.

lockwo:
> Hi @lockwo, as you mentioned, the Hamiltonian produced by openfermion is probably different from the formula visible in the comment above. Any idea to fix this inconsistency? Some ideas from my side:
> 
> * Change the formula to something more generic e.g. a sum over generic Pauli operators.
> * Change the formula to something exactly matching the operators produced by openfermion
> * Keep the previous table of parameters and add openfermion as a "second option" to generate the Hamiltonian.
> * Other options are welcome.

So the formula is the similar it just doesn't have the YY term (hence why the unmitigated case seems to do a tiny bit better it looks), the symmetry reduction comes from: https://arxiv.org/abs/1701.08213. However, this also makes it so the circuit doesn't need the initial X gate on the second qubit. I think a combination of your suggestions might work. Perhaps adding the exact description of the OF Hamiltonian in addition to providing the parameters from the original work might be a good way to do it. I will make those changes in a little bit to see how they look.

andreamari:
Sounds good! Considering that [unitaryhack](https://unitaryhack.dev/) will start in two weeks there is no rush.

lockwo:
Alright, finally got around to making those changes.

andreamari:
Thanks, @lockwo. Next week I'll make a proper review, but looks already very good!

lockwo:
Sounds good!

docs/source/examples/molecular_hydrogen.myst
```diff
@@ -85,31 +87,44 @@ radii = table[:, 0]
 coeffs = table[:, 1:]
 ```
 
-We define a function that evaluates the expectation value of the Hamiltonian on a noisy backend and for a given input circuit (ansatz).
-Here we use a Cirq density matrix simulator with a simple depolarizing noise model.
+We can also calculate their values automatically using OpenFermion (an open source quantum chemistry package).
+This results in slightly modified hamiltonians of the form:
+
+\begin{equation}
+    H = g_0 I + g_1 Z_0 + g_2 Z_1 + g_3 Z_0 Z_1 + g_4 X_0 X_1
```
andreamari, line 94: Do you have any intuition why the last term (g_5 Y_0 Y_1) is missing here?
If not, no worries. 
If yes, I would add a comment about it to make the reader more comfortable with the new different Hamiltonian.

docs/source/examples/molecular_hydrogen.myst
```diff
@@ -85,31 +87,44 @@ radii = table[:, 0]
...
+    """Returns the hamiltonian for the H2 molecule at the given bond length. Uses the STO-6G basis set and the Bravyi-Kitaev
+    transform to map fermionic operators to qubits. Results in a 2 qubit hamiltonian returned via a metriq.Observable."""
+    geometry = [('H', (0., 0., 0.)), ('H', (0., 0., bond_length))]
+    molecular_hamiltonian = generate_molecular_hamiltonian(geometry, basis='sto-6g', multiplicity=1, charge=0)
+    fermionic_op = of.transforms.get_fermion_operator(molecular_hamiltonian)
+    bk_op = of.transforms.symmetry_conserving_bravyi_kitaev(fermionic_op, active_orbitals=4, active_fermions=2)
+    qubit_operator = qubit_operator_to_pauli_sum(bk_op)
+    return Observable(*qubit_operator)
+
+hamiltonians = [get_hamiltonian(i) for i in radii]
```
andreamari, line 124: Is there any simple way to show that these Hemiltonians have indeed the structure mentioned in the latex formula?
E.g. something like 

`print(get_hamiltonian( some_both_lenght))`.

Or any other way to check which Hamiltonian OpenFermion is actually producing.

andreamari (commented):
Thanks @lockwo, I added some comments.
Please have a look when you have some time. 
This is close to be accepted.
Very nice work.

docs/source/examples/molecular_hydrogen.myst
lockwo: I added a print function for the hamiltonian, so you can see it is of the same form. Let me know if that is what you were looking for.

nathanshammah:
Can you add the use of OpenFermion in the title? To differentiate from #1337 that is done with PennyLane.

lockwo:
Hmmm it has somehow broken the build, it works when I build locally. I see an error with the QAOA tutorial, but I'm not sure if that is the source (since this link is now dead: https://quantumai.google/cirq/tutorials/qaoa) I will fix it anyway and see

lockwo:
That was the problem actually (not sure if you want to make that fix a separate PR, but it is in this one).

lockwo:
> Can you add the use of OpenFermion in the notebook title? To differentiate from #1337 that is done with PennyLane.

Done

andreamari (approved):
Thank you @lockwo, LGTM! 
And thanks for fixing the broken link too!

Moreover congratulations for merging a unitaryhack PR and so for getting the associated bounty!

andreamari:
Hi @lockwo, there was a problem with OpenFermion not working with Windows tests on the master branch. So we had to temporarily revert the merging commit. 

Of course we still want to have your example but we need to merge it more carefully.  Could you open a new PR with the same code as this one? Re-opening this one is fine too (if possible).

lockwo:
Just made it

PR #1309, opened 2022-05-24
Update "google-style docstring" URLs

natestemen:
While going through some documentation I found that we use two different links when referring to (presumably) the same thing. I've removed the non-google controlled URL as it seems less helpful, and perhaps slightly different from the Google recommendations, as well as update the URL to a slightly different part of the same page.

If there is a reason we were using the docs from the other URL please let me know!


andreamari (approved):
LGTM! Thanks @natestemen

PR #1316, opened 2022-05-27
Fixes unexpected error from repeated rule

### LINKED ISSUE #1260: Unexpected error from ddd.rules.repeated_rule
When calling:

```python
ddd.rules.repeated_rule(slack_length=3, gates=[X, Y, X, Y])
```
Since the `slack_lentgh` is shorter than the DDD sequence, the expected output is an empty circuit. The function instead gives an error:

```python
ValueError: Gateset too short to make a ddd sequence.
```

Aaron-Robertson:
Description
-----------

Fixes #1260. Returns an empty circuit when the slack length is shorter than the number of gates in a repeated sequence.

License
-------


andreamari (approved):
Thanks, @Aaron-Robertson! LGTM

PR #1320, opened 2022-05-31
Clean up contributing-to-docs doc

natestemen:
## Description

Went through some of this and cleaned up some wording/tried to make it more clear where I thought it wasn't.

**Question**: We seem to inconsistently use 80 character line breaks in some places and not in others. Not urgent, but would be nice to pick something and make it consistent.

### DIFF
--- a/docs/CONTRIBUTING_DOCS.md
+++ b/docs/CONTRIBUTING_DOCS.md
@@ -44,18 +44,17 @@ extensions = ['sphinx.ext.autodoc','sphinx.ext.doctest']
 
 ## Updating the Documentation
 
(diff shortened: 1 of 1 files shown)


natestemen:
@Misty-W any idea why this is failing to build? I think it was first caused by the broken link, but the second failure seems to be something else entirely.

andreamari:
I restarted the docs build on readthedocs. To see if it works. 
I also added the username `Misty-W` as admin on RTD. I hope it's you, @Misty-W  :-) 

I tried adding `natestamen` too, but the name doesn't exist. So, @natestemen, feel free create an account on RTD if you are interested in accessing RTD options.

Misty-W (approved):
LGTM!

Misty-W:
💃🏼

PR #1340, opened 2022-06-05
Loss function for learning biased noise

Misty-W:
Description
-----------

Please explain the changes you made here.

Checklist
---------

License
-------

Tips
----


mitiq/pec/representations/learning.py
```diff
@@ -0,0 +1,86 @@
...
+from mitiq import QPROGRAM, Executor, Observable
+from mitiq.pec import execute_with_pec
+from mitiq.pec.representations.biased_noise import (
+    represent_operation_with_local_biased_noise,
+)
+
+
+def biased_noise_loss_function(
+    params: np.ndarray,
+    operation: Gate,
```
andreamari, line 30:
```suggestion
    operations_to_mitigate: List[QPROGRAM],
```
It would be optimal to have a list of input operations and as a list of `QPROGRAM` objects.
The reason is that PEC in Mitiq is usually applied to operations expressed as QPROGRAM objects because they are
multi-platform while `Gate` is Cirq only.

mitiq/pec/representations/learning.py
```diff
@@ -0,0 +1,86 @@
...
+    """
+    epsilon = params[0]
+    eta = params[1]
+    qreg = LineQubit.range(operation.num_qubits())
+    representations = [
+        represent_operation_with_local_biased_noise(
+            Circuit(operation.on(*qreg)),
+            epsilon,
+            eta,
+        )
```
andreamari, line 69:
```suggestion
    qreg = LineQubit.range(operation.num_qubits())
    representations = [
        represent_operation_with_local_biased_noise(
            Circuit(operation.on(*qreg)),
            epsilon,
            eta,
        )
```
Assume that in the circuit there are many CNOT gates applied to different qubit pairs. With the current code, we only represent the CNOT gate applied to qubits 0 and 1.  If in the circuit there is a CNOT gate applied to qubit 2 and 3, this will remain unmitigated. 

Note that this problem can be easily avoided if you agree to have a list of QPROGRAM objects as input argument as suggested in the previous comment. In this case, you don't need to manually define a `qreg`.

mitiq/pec/representations/learning.py
```diff
@@ -0,0 +1,86 @@
...
+    represent_operation_with_local_biased_noise,
+)
+
+
+def biased_noise_loss_function(
+    params: np.ndarray,
+    operation: Gate,
+    training_circuits: List[QPROGRAM],
+    ideal_values: np.ndarray,
+    noisy_executor: Executor,
```
andreamari, line 33: With the current code, it is impossible to set options for `execute_with_pec`, as, for example, the number of pec samples, the accuracy, etc....
This is a quite strong restriction. For example, setting a smaller number of PEC samples can speedup the computation. Not sure what is the best way to solve this problem. Maybe we can add another argument like
`pec_kwargs : Dict["str", Any]`,
such that, in the body of the function, we can call `execute_with_pec( ....  ,  **pec_kwargs)`.

mitiq/pec/representations/tests/test_learning.py
```diff
@@ -0,0 +1,124 @@
...
+    return ops.MixedUnitaryChannel(mix)
+
+
+@pytest.mark.parametrize("epsilon", [0, 0.7, 1])
+@pytest.mark.parametrize("eta", [0, 1])
+@pytest.mark.parametrize("gate", [CNOT, Rx_ops[0][2]])
+def test_biased_noise_loss_function(epsilon, eta, gate):
+    """Test that the biased noise loss function value (calculated with error
+    mitigation) is smaller than the loss calculated with the noisy
+    (unmitigated) executor"""
```
andreamari, line 85:
```suggestion
    """Test that the biased noise loss function value (calculated with error
    mitigation) is smaller than the loss calculated with the noisy
    (unmitigated) executor"""
```
Nice test!

mitiq/pec/representations/learning.py
Misty-W: Good point! I applied your suggestion to add the argument `pec_kwargs` in the `biased_noise_loss_function`. 

Also I'm now using`pec_kwargs` to set a looser precision in `test_biased_noise_loss_function` and `test_biased_noise_loss_compare_ideal`

mitiq/pec/representations/learning.py
Misty-W: I changed the input argument to `operations_to_mitigate: List[QPROGRAM]` and removed the manually defined `qreg` but unfortunately now the tests are failing. I thought I synced them with the function but I guess I’m missing something.

Misty-W: Actually I have a guess as to where the error is coming from. I’ll check tomorrow.

Misty-W: I was able to get past the "bad operation" error, but the execution time went up to an hour, at which point I cancelled the workflow and reverted the changes associated with `operations_to_mitigate: List[QPROGRAM]`. The tests then ran through successfully in about the same amount of time as before. It's strange because the "list" of operations in the test is still just a list with one element, so the program shouldn't run longer than it did before.

Misty-W: Update: I reapplied the suggestion to change `operation: Gate` to  `operations_to_mitigate: List[QPROGRAM]`  and now test execution time is reasonable after setting `pec_kwargs = {"num_samples", 100}` 🎉

mitiq/pec/representations/learning.py
```diff
@@ -0,0 +1,89 @@
...
+            expectation of the observable.
+
+    Returns: Square of the difference between the error-mitigated values and
+        the ideal values, over the training set
+    """
+    epsilon = params[0]
+    eta = params[1]
+    representations = [
+        represent_operation_with_local_biased_noise(
+            Circuit(operation),
```
andreamari, line 67:
```suggestion
            Circuit(operation),
```
Based on the annotation, we can assume that operation is already a circuit and not a `cirq.Operation`. I know the name is confusing but this choice is important for conversions.

I practice Circuit(.) should be removed from here and called outside the function in tests to convert operations in circuits.

andreamari (commented):
Hi @Misty-W, LGTM up to a minor change and up to the following decision.

Currently this functions works only for Cirq circuits and so we have two alternative options:
1. Making this function multi-frontend with conversions decorators as we often did for other functions. 
2. Keeping this function as Cirq-only but private. In this case we should rename is `_biased_noise_loss_funciton` and maybe open a new issue for transforming it to a public  multi-frontend function in the future. 

Both options are valid for me.

Misty-W:
> Hi @Misty-W, LGTM up to a minor change and up to the following decision.
> 
> Currently this functions works only for Cirq circuits and so we have two alternative options:
> 
> 1. Making this function multi-frontend with conversions decorators as we often did for other functions.
> 2. Keeping this function as Cirq-only but private. In this case we should rename is `_biased_noise_loss_funciton` and maybe open a new issue for transforming it to a public  multi-frontend function in the future.
> 
> Both options are valid for me.

@andreamari, I implemented Option 2 and created a separate issue for adding conversions, #1356.

mitiq/pec/representations/learning.py
```diff
@@ -0,0 +1,88 @@
...
+            `QuantumResult`.
+        pec_kwargs: Options to pass to `execute_w_pec` for the error-mitigated
+            expection value obtained from executing the training circuits
+        observable (optional): Observable to compute the expectation value of.
+            If None, the `executor` must return an expectation value. Otherwise
+            the `QuantumResult` returned by `executor` is used to compute the
+            expectation of the observable.
+
+    Returns: Square of the difference between the error-mitigated values and
+        the ideal values, over the training set
```
andreamari, line 60:
```suggestion
    r"""Loss function for optimizing quasi-probability representations
    assuming a biased noise model depending on two real parameters.

    Args:
        params: Array of optimization parameters epsilon
            (local noise strength) and eta (noise bias between reduced
            dephasing and depolarizing channels).
        operations_to_mitigate: List of ideal operations to be represented by
            a combination of noisy operations.
        training_circuits: List of training circuits for generating the
            error-mitigated expectation values.
        ideal_values: Expectation values obtained by noiseless simulations.
        noisy_executor: Executes the circuit with noise and returns a
            `QuantumResult`.
        pec_kwargs: Options to pass to `execute_w_pec` for the error-mitigated
            expectation value obtained from executing the training circuits.
        observable (optional): Observable to compute the expectation value of.
            If None, the `executor` must return an expectation value. Otherwise
            the `QuantumResult` returned by `executor` is used to compute the
            expectation of the observable.

    Returns: Mean squared error between the error-mitigated values and
        the ideal values, over the training set.
```

A few small changes. Not sure why the diff is not resolving them.

andreamari (approved):
LGTM!  :100: 
I left small optional suggestions.

PR #1348, opened 2022-06-09
Test on windows and mac on each PR and add openfermion to dev requirements

andreamari:
Description
-----------
Often PRs pass all tests but they fail when merged into master since Windows and Mac are tested only on master.
The original reason for non testing Mac and Windows in each PR was to save time.
It may be worth to reconsider this decision depending on the testing time and considering that the main bottleneck at the moment is the docs build and not pytest.

License
-------

Tips
----


.github/workflows/test.yml
```diff
@@ -138,3 +138,24 @@ jobs:
         uses: codecov/codecov-action@v3.1.0
         with:
           fail_ci_if_error: true
+
+  test-nonlinux:
+    runs-on: ${{ matrix.os }}
+    strategy:
+      matrix:
+        os: [macos-latest, windows-latest]
+        python-version: [3.7, 3.8]
```
natestemen, line 147: Minor nitpick: I think this is slightly more clear if the `runs-on` attribute comes _after_ the strategy matrix as is done in the [github documentation](https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#example-using-a-multi-dimension-matrix) on this.

natestemen (approved):
Looks good, just one minor nitpick that's not essential.

The main question I had is whether the new `test-nonlinux` step can be merged with `test-linux` so that the testing is the same regardless of OS. We would still only want to submit the code coverage report once, but that should work just fine I think. Is there a reason to keep these separate?

andreamari:
> Looks good, just one minor nitpick that's not essential.
> 
> The main question I had is whether the new `test-nonlinux` step can be merged with `test-linux` so that the testing is the same regardless of OS. We would still only want to submit the code coverage report once, but that should work just fine I think. Is there a reason to keep these separate?

The reason I kept them separated is that there are several differences in the types of instructions (`make test` vs `make test-all`, using Docker vs not using it, codecove). The original idea was to test _everything_ on linux and to run a lighter test suite on non-linux platforms. 

It is worth exploring the possibility of running exactly the same types of tests on linux ad non-linux. But it is probably a topic for a new issue/PR.

natestemen:
Makes sense, thanks for the explanation. :shipit:

Issue #1136, opened 2022-02-17
ASV report as a part of docs


crazy4pi314:
We have setup ASV as a part of our testing pipelines, and it can generate benchmarking webpages like: https://pv.github.io/numpy-bench/

Is this something we want for Mitiq? It can also give diffs of benchmarks from commit to commit.

@rmlarose I think you were using this the most, thoughts?

rmlarose:
+1 from me. I like the idea of expanding the current benchmarks and hosting the results online (whether in the documentation or on a separate page). asv makes it quite easy to do.

crazy4pi314:
I think the first step for this is to separate out the running of ASV from the regular test suite, then have a separate action to trigger this part of the gh pages site to build.

PR #1354, opened 2022-06-13
Deploy asv benchmark data to web

natestemen:
## Description

The added github action runs our existing [`asv`](https://asv.readthedocs.io/en/stable/) benchmarks and pushes the benchmark data to a [new repository](https://github.com/unitaryfund/mitiq-bench) where it then deploy to github pages. The basic architecture is laid out as follows.

```mermaid
flowchart LR
  subgraph mitiq-bench
    direction TB
    db[(previous data)]
    db -- asv publish\nasv gh-pages --> gh[Github Pages]
  end
  subgraph mitiq
    direction TB
    commit[commit to master] --> run[Run asv]
  end
  run -- push --> db
```

I had originally hoped to keep this confined entirely in the mitiq repo, with the results displayed on our current URL, but after having some technical difficulty with the solution, I found that many others use a separate repo for benchmark data:
- https://github.com/pv/numpy-bench
- https://github.com/pv/scipy-bench
- https://github.com/pybamm-team/pybamm-bench
- https://github.com/scikit-learn/scikit-learn-benchmarks

Hence, I think we are in good company with this method.

While working on this ticket, I also came across [this blog post](https://labs.quansight.org/blog/2021/08/github-actions-benchmarks/) which analyzes the effectiveness of using github actions as a benchmarking tool. The main flaw of using the shared compute of github actions is that you never quite know if you're getting the same level of computational resources. The authors conclude by saying it's _likely_ okay, however.

This solution also has the drawback of not having the entire historical benchmark data since `asv` was added to the project in https://github.com/unitaryfund/mitiq/pull/1047. We could run `asv` from commit c9933e6f70a471cbd6c7d2be3e29dcd97fea070a to current using `asv run c9933e6..`, but the compute power of my machine will likely be different than that of github actions, resulting in potentially different results.

fixes https://github.com/unitaryfund/mitiq/issues/1136

### Example

I've forked the `mitiq` repo and set this up with my own repos:
- my fork: https://github.com/natestemen/mitiq
- `mitiq-bench`: https://github.com/natestemen/mitiq-bench
- displayed benchmarks: https://natestemen.github.io/mitiq-bench/

### Pre-deployment
In order for this action to work appropriately, before deployment the following items must be completed.

- [x] Create `unitaryfund/mitiq-bench` repo (https://github.com/unitaryfund/mitiq-bench/)
- [x] initialize it with empty `results` directory using `.gitkeep` file: e.g. `mkdir results && touch results/.gitkeep`
- [x] add an access token to this repo under name `BENCH_PAT` so that it can have push/pull access to `mitiq-bench` repo
- [x] create empty `gh-pages` branch
- [ ] add appropriate DNS records for `benchmarks.mitiq.dev`
- [ ] add `CNAME` file to `mitiq-bench` repo


asv.conf.json
```diff
@@ -1,16 +1,9 @@
...
-    "branches": ["master"],
-    "dvcs": "git",
-    "environment_type": "virtualenv",
-    "show_commit_url": "https://github.com/unitaryfund/mitiq/commit/",
-    "pythons": ["3.7"],
-     "benchmark_dir": "benchmarks",
-    "env_dir": ".asv/env",
-    "results_dir": ".asv/results",
-    "html_dir": ".asv/html",
-    "hash_length": 7
```
rmlarose, line 15: Any reason to axe these?

rmlarose (commented):
Looking good. Agree with separate repo based on other projects. As per using GH actions, there's actually no timing benchmarks right now - just "EM factor" - so different compute nodes shouldn't matter in this respect.

asv.conf.json
natestemen: Not **great** reasons, but here are some.
- `benchmarks` is the default for `benchmark_dir`
- `hash_length` defaults to 8
- Since `asv` will primarily run via github actions, the custom `env`/`results`/`html` directories don't seem to provide any benefit
- On second thought, `pythons` might be worth keeping just to keep explicit track of what python version we are using to run the benchmarks

If this is a [chesterton's fence](https://fs.blog/chestertons-fence/) scenario, sorry!

natestemen:
> no timing benchmarks right now - just "EM factor" - so different compute nodes shouldn't matter in this respect.

Ah good point, thanks for clarifying that. In that case I should be able to upload historical data to the `mitiq-bench` repo, which will be nice. That said, if you check out the example site I put up while testing this (https://natestemen.github.io/mitiq-bench/), the EM factors change, despite the commits not touching any related code (only github actions code changes). What would you attribute these changes in EM factor to?

asv.conf.json
rmlarose: There wasn't really strong asv rationale behind including these - just for sake of being explicit. I would definitely encourage being explicit about the python version, but the rest sounds good.

rmlarose:
> That said, if you check out the example site I put up while testing this (https://natestemen.github.io/mitiq-bench/), the EM factors change, despite the commits not touching any related code (only github actions code changes). What would you attribute these changes in EM factor to?

Primarily variance due to small sample (problem) size and em technique, but also the figure of merit ("em factor") that I picked somewhat arbitrarily is rather sensitive. The plots are interesting to see and mostly expected, although ZNE jumping several orders of magnitude is a bit disconcerting. 

Ideally these tests would run a lot of problems (circuits) to get better statistics. I made the runtime parameters very small (one qubit, depths 1, 2, and 3) because I wanted these to run (quickly) in CI. I think the best solution is using small runtime parameters for CI but larger ones for better statistics on the benchmarks site. I think asv supports some notion of parameter substitutions but I couldn't find an easy way when I looked into it, so this route may require some engineering, say a short script to replace benchmark runtime parameters.

natestemen:
I've come up with a way to be able to run with different parameters on CI/when running on master. Mainly implemented in [`6a65759`](https://github.com/unitaryfund/mitiq/pull/1354/commits/6a65759a71a8196952e6f5ed60ccb1a758266c7c). Let me know what you think, and also what parameters you might want to see when running for the website.

I've also doubled the number of rounds it runs with. [The docs](https://asv.readthedocs.io/en/stable/tuning.html?highlight=rounds#tuning-timing-measurements) indicate that this helps get more precise numbers, but it's unclear what exactly a round is, and also if this only applies to timing measurements. Again, implemented in [`2d18276`](https://github.com/unitaryfund/mitiq/pull/1354/commits/2d18276577056cc4b0a701a5611e7a736df19250), and I'm happy to remove it if we don't think it actually helps (i.e. adding variable params for the benchmarks is enough).

.github/workflows/test.yml
```diff
@@ -88,22 +87,18 @@ jobs:
...
+          pip install asv virtualenv
+          export BENCHMARK_CI=1
           asv machine --yes
-          result=$(asv dev)
-          echo "${result}"
-          if [[ $result == *"failed"* ]]; then
-            exit 1
-          fi
-          exit 0
+          asv run --quick --strict
```
rmlarose, line 101: Does this exit 0 if there is an error?

benchmarks/benchmarks.py
```diff
@@ -25,14 +26,25 @@
 from mitiq.interface import mitiq_cirq
 
 
+if os.environ.get("BENCHMARK_CI"):
+    params = {"nqubits": [1], "depth": [1, 2, 3], "num_samples": [10]}
+else:
+    params = {"nqubits": [2], "depth": [2, 4, 6], "num_samples": [20]}
```
rmlarose, line 32: I think `num_samples = 100` is better. Good to call this `num_pec_samples` or similar to be clear.

rmlarose: Might need to have the `timeout` (currently 300 seconds) be larger for these tests.

rmlarose (approved):
> I've come up with a way to be able to run with different parameters on CI/when running on master. Mainly implemented in [6a65759](https://github.com/unitaryfund/mitiq/pull/1354/commits/6a65759a71a8196952e6f5ed60ccb1a758266c7c). Let me know what you think, and also what parameters you might want to see when running for the website.

Nice! Would ask @andreamari et al what parameters to use for benchmarks. LGTM modulo a few small questions.

nathanshammah (approved):
Looks very neat. 
- Can we add documentation about the existence of these benchmarks? Even just in the Readme file, if the references in the Sphinx-built documentation are not appropriate at this time. Or open an issue about it for the future. For example, the plot on the top comment in this issue is informative. 
- Would suggest adding some issues about (some examples, checking with you if it makes sense): 
- 1. improving the documentation of the webpages themselves; 
- 2. adding the proper url; find a way to specify [machine properties](https://asv.readthedocs.io/en/stable/using.html#running-benchmarks), e.g., CPU;
- 3. discussion on the metrics benchmarked, i.e. mitigation factor vs time, etc. 
Great addition

.github/workflows/test.yml
natestemen: It will return a non-zero error code when there is an error. From the [`asv run`](https://asv.readthedocs.io/en/stable/commands.html?highlight=--quick#asv-run) docs:

```
 --strict              When set true the run command will exit with a non-
                       zero return code if any benchmark is in a failed state
```

asv.conf.json
```diff
@@ -4,6 +4,7 @@
   "project_url": "https://github.com/unitaryfund/mitiq",
   "repo": "https://github.com/unitaryfund/mitiq.git",
   "branches": ["master"],
+  "pythons": ["3.8"],
```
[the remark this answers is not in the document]
natestemen: I'm also not totally sure how we would test with multiple versions of python on github actions since we specify the python version we want to use in the action. Can probably come up with a workaround, but I don't think it's 100% necessary.

natestemen:
Thanks Nathan for the comments. I've addressed some of your comments in the recent batch of commits.

> find a way to specify [machine properties](https://asv.readthedocs.io/en/stable/using.html#running-benchmarks), e.g., CPU;

From what I understand, the [`asv machine`](https://asv.readthedocs.io/en/stable/commands.html#asv-machine) command figures out what the computers attributes are if you pass `--yes` to it. If we tried to set them manually, they might be incorrect in github changes how actions are run.

Issues I need to raise:
- [ ] add option to support global folding vs local folding (current default) in ZNE (will have a PR open shortly)
- [ ] more documentation surrounding the benchmarks
- [ ] add timing benchmarks

benchmarks/benchmarks.py
natestemen: The benchmarks with increased parameters only took 115 seconds on my computer (2020 mbp), so I think 300 should be fine for github actions.

asv.conf.json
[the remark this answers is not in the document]
natestemen: On second attempt, it seems github actions does not play well python 3.7. [e076cb2](https://github.com/unitaryfund/mitiq/pull/1354/commits/e076cb2044130faba725f4ae043811f0938f0202) failed due to python 3.7 being missing, even though I tried to setup python 3.7.

I think using 3.8 might be okay as many of our test run only on 3.8, and we are planning to support newer versions of python soon (and hence likely drop support for older versions).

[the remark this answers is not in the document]
rmlarose: SGTM!

PR #1360, opened 2022-06-17
[unitaryhack] Added qiskit backends and tested

### LINKED ISSUE #1163: Include backend option for real hardware in mitiq.interface.mitiq_qiskit.qiskit_utils.execute_with_shots
Pre-Request Checklist
---------------------

- [X ] I checked to make sure that this feature has not already been requested.

Issue Description
-----------------
Currently the function execute_with_shots or execute_with_shots_and_noise do not allow for specific backend, e.g real hardware rather than simulator or different noise model from the simulator, so the calculation of the expectation value requires specifies a circuit with different basis measurement X, Y, Z, etc when the Hamiltonian is complex. 

Proposed Solution
-----------------

It would be nice to allow real backend input in the qiskit_utils.execute_with_shots in the qiskit interface so it's a lot easier to calculate the expectation value of an operator.

RamAIbot:
I have added the code to use qiskit backend simulators in the "execute_with_shots" function. The corresponding test file is in "testing.py"

I have tested using 3 simulators of the qiskit. The circuit I tested with is the bell state circuit.

image

Fixes #1163

### DIFF
--- a/mitiq/interface/mitiq_qiskit/qiskit_utils.py
+++ b/mitiq/interface/mitiq_qiskit/qiskit_utils.py
@@ -23,6 +23,7 @@ from qiskit import Aer
 from qiskit import IBMQ
 from qiskit.providers.jobstatus import JOB_FINAL_STATES
 from qiskit import QuantumCircuit, assemble, transpile
+from qiskit.tools.monitor import job_monitor
 import time
 
 # Noise simulation packages
@@ -86,6 +87,7 @@ def execute_with_shots(
     """
 
     """ Added code to use qiskit backends"""
+    
     if(simulator):
         #runs on a qiskit simulator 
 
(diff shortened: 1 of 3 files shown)


Misty-W:
thanks for the PR, @RamAIbot, you're off to a good start! I see the validation is failing- looks like there are a couple issues with types. 

Also, I noticed you commented out some lines, can you remove them? Commented out lines can make the code "cluttered", and we can always get the original back, there are multiple ways to do that in Git.

Misty-W:
Fixes #1163

Misty-W:
Sure, I can meet you tomorrow. I’m in the US Pacific time zone

natestemen:
superseded by https://github.com/unitaryfund/mitiq/pull/1371

PR #1369, opened 2022-06-27
Prevent idle qubits with qiskit barriers from being lost after conversions

### LINKED ISSUE #1368: Unexpected behavior of folding functions with Qiskit circuits with barriers
Pre-Report Checklist
--------------------

- [ ] I am running the latest version of mitiq
- [ ] I checked to make sure that this bug has not already been reported

Issue Description
-----------------

Insert a short description of the bug here, along with what you expected the behavior to be.

Thanks for helping us improve mitiq! 🙂

How to Reproduce
----------------

If useful, provide a numbered list of the steps that result in the error.

Otherwise, just fill out the "Code Snippet" and "Error Output" sections below.

### Code Snippet

```python
Include a snippet of the mitiq code that produces the error here.
```

### Error Output

```
Additionally, please copy and paste the output of the above code here.
```

Environment Context
-------------------

Use the `about()` function to summarize information on operating system, python version and dependencies.
```python
from mitiq import about
about()
```

Additional Python Environment Details (`pip freeze` or `conda list`):

```
Copy and paste the output of `pip freeze` or `conda list` here.
```

andreamari:
Description
-----------

Fixes #1368 

Qiskit circuits with idle qubits and qiskit barriers have unexpected results when modified by folding functions.
This is related to a bug in the Mitiq `noise_scaling_converter` decorator.

The following code snippet:
```python
test_circuit_qiskit = QuantumCircuit(4)
test_circuit_qiskit.x(0)
test_circuit_qiskit.x(2)
test_circuit_qiskit.barrier(0,1,2)
test_folded = zne.scaling.fold_gates_at_random(test_circuit_qiskit, 1.0)
print(f"Input circuit:\n{test_circuit_qiskit}")
print(f"Folded circuit:\n{test_folded}")
```
before fixing the bug gives the unexpected result:
```
Input circuit:
     ┌───┐ ░ 
q_0: ┤ X ├─░─
     └───┘ ░ 
q_1: ──────░─
     ┌───┐ ░ 
q_2: ┤ X ├─░─
     ├───┤ ░ 
q_3: ┤ I ├───
     └───┘   
Folded circuit:
     ┌───┐
q_0: ┤ X ├
     ├───┤
q_1: ┤ X ├
     ├───┤
q_2: ┤ I ├
     └───┘
q_3: ─────
```
after fixing this bug, the correct expected result is recovered:

```
Input circuit:
     ┌───┐ ░ 
q_0: ┤ X ├─░─
     └───┘ ░ 
q_1: ──────░─
     ┌───┐ ░ 
q_2: ┤ X ├─░─
     └───┘ ░ 
q_3: ────────
             
Folded circuit:
     ┌───┐
q_0: ┤ X ├
     └───┘
q_1: ─────
     ┌───┐
q_2: ┤ X ├
     └───┘
q_3: ─────
```

License
-------

### DIFF
--- a/mitiq/interface/conversions.py
+++ b/mitiq/interface/conversions.py
@@ -240,9 +240,14 @@ def noise_scaling_converter(
             from mitiq.interface.mitiq_qiskit.conversions import (
                 _add_identity_to_idle,
             )
+            from qiskit.transpiler.passes import RemoveBarriers
 
             # Avoid mutating the input circuit
             circuit = deepcopy(circuit)
+            # Removing barriers is necessary to correctly identify idle qubits
+            circuit = RemoveBarriers()(circuit)
+            # Apply identity gates to idle qubits otherwise they get lost
+            # when converting to Cirq. Eventually, identities will be removed.
             idle_indices = _add_identity_to_idle(circuit)
 
         scaled_circuit = atomic_converter(noise_scaling_function)(
--- a/mitiq/interface/conversions.py
+++ b/mitiq/interface/conversions.py
@@ -16,6 +16,7 @@
 """Functions for converting to/from Mitiq's internal circuit representation."""
 from functools import wraps
 from typing import Any, Callable, cast, Iterable, Tuple
+from copy import deepcopy
 
 from cirq import Circuit
 
@@ -232,6 +233,7 @@ def noise_scaling_converter(
     def new_scaling_function(
         circuit: QPROGRAM, *args: Any, **kwargs: Any
     ) -> QPROGRAM:
+
         # Pre atomic conversion
         idle_indices = set()
         if "qiskit" in circuit.__module__:
@@ -239,6 +241,8 @@ def noise_scaling_converter(
                 _add_identity_to_idle,
             )
 
+            # Avoid mutating the input circuit
+            circuit = deepcopy(circuit)
             idle_indices = _add_identity_to_idle(circuit)
 
         scaled_circuit = atomic_converter(noise_scaling_function)(
--- a/mitiq/interface/conversions.py
+++ b/mitiq/interface/conversions.py
@@ -16,7 +16,6 @@
 """Functions for converting to/from Mitiq's internal circuit representation."""
 from functools import wraps
 from typing import Any, Callable, cast, Iterable, Tuple
-from copy import deepcopy
 
 from cirq import Circuit
 
@@ -243,7 +242,7 @@ def noise_scaling_converter(
             from qiskit.transpiler.passes import RemoveBarriers
 
             # Avoid mutating the input circuit
-            circuit = deepcopy(circuit)
+            circuit = circuit.copy()
             # Removing barriers is necessary to correctly identify idle qubits
             circuit = RemoveBarriers()(circuit)
             # Apply identity gates to idle qubits otherwise they get lost
--- a/mitiq/interface/mitiq_qiskit/conversions.py
+++ b/mitiq/interface/mitiq_qiskit/conversions.py
@@ -129,6 +129,9 @@ def _add_identity_to_idle(
 
     Returns:
         An unordered set of the indices that were altered
+
+    Note: An idle qubit is a qubit without any gates (including Qiskit
+        barriers) acting on it. 
     """
 
     data = copy.deepcopy(circuit._data)
@@ -136,8 +139,6 @@ def _add_identity_to_idle(
     idle_bit_indices = set()
     for op in data:
         gate, qubits, cbits = op
-        if gate.name == "barrier":  # Skip barriers
-            continue
         bit_indices.update(set(bit.index for bit in qubits))
     for index in range(circuit.num_qubits):
         if index not in bit_indices:
--- a/mitiq/interface/mitiq_qiskit/conversions.py
+++ b/mitiq/interface/mitiq_qiskit/conversions.py
@@ -131,7 +131,7 @@ def _add_identity_to_idle(
         An unordered set of the indices that were altered
 
     Note: An idle qubit is a qubit without any gates (including Qiskit
-        barriers) acting on it. 
+        barriers) acting on it.
     """
 
     data = copy.deepcopy(circuit._data)
--- a/mitiq/interface/mitiq_qiskit/conversions.py
+++ b/mitiq/interface/mitiq_qiskit/conversions.py
@@ -130,11 +130,14 @@ def _add_identity_to_idle(
     Returns:
         An unordered set of the indices that were altered
     """
+
     data = copy.deepcopy(circuit._data)
     bit_indices = set()
     idle_bit_indices = set()
     for op in data:
         gate, qubits, cbits = op
+        if gate.name == "barrier":  # Skip barriers
+            continue
         bit_indices.update(set(bit.index for bit in qubits))
     for index in range(circuit.num_qubits):
         if index not in bit_indices:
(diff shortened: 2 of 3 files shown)


mitiq/interface/mitiq_qiskit/conversions.py
```diff
@@ -130,11 +130,14 @@ def _add_identity_to_idle(
...
         An unordered set of the indices that were altered
     """
+
     data = copy.deepcopy(circuit._data)
     bit_indices = set()
     idle_bit_indices = set()
     for op in data:
         gate, qubits, cbits = op
+        if gate.name == "barrier":  # Skip barriers
+            continue
```
rmlarose, line 140: Is my understanding correct?

- Qiskit barriers have no Cirq analogue and so are removed in the Qiskit -> Cirq conversion.
- If a qubit in a Qiskit circuit only has a barrier acting on it, this will get removed in the Qiskit -> Cirq conversion, and so the qubit will be idle / not present in the Cirq circuit.

If so this LGTM. I wonder if there are more Qiskit operations which have no Cirq analogue and so would need to be treated in a similar manner here.

mitiq/tests/test_conversions.py
```diff
@@ -176,3 +176,36 @@ def test_atomic_one_to_many_converter(to_type):
...
+    q_2: ┤ X ├
+         └───┘
+    q_3: ─────
+    """
+    test_circuit_qiskit = qiskit.QuantumCircuit(4)
+    test_circuit_qiskit.x(0)
+    test_circuit_qiskit.x(2)
+    test_circuit_qiskit.barrier(0, 1, 2)
+    scaled = scaling_function(test_circuit_qiskit)
+    # Mitiq expected to remove qiskit barriers
```
rmlarose, line 207: Actually, if the expected behavior is to remove barriers, isn't it simpler to first remove all barriers before adding identities?

mitiq/interface/mitiq_qiskit/conversions.py
andreamari: Yes, your interpretation is correct. Basically, the scope of the `if gate.name == "barrier"` condition was to mark as idle qubit any qubit in which there are only "barrier" gates.

However I agree with you that it is more intuitive to simply remove barriers before calling this function. (See last commit).

Note that if there are other operations that have no Cirq analogue, the user will get a conversion error. Barriers represent a special case since they are intentionally removed by Mitiq without giving an error.

mitiq/tests/test_conversions.py
andreamari: Applied this approach in the last commit.

andreamari:
I am adding also @nathanshammah  as a last-minute reviewer since it would be good to have this bug fixed within the upcoming release. If we don't make it, it's not the end of the world.

mitiq/interface/mitiq_qiskit/conversions.py
```diff
@@ -129,15 +129,16 @@ def _add_identity_to_idle(
 
     Returns:
         An unordered set of the indices that were altered
+
+    Note: An idle qubit is a qubit without any gates (including Qiskit
+        barriers) acting on it.
```
rmlarose, line 134: Good point. This is technically not a bugfix as qubits with barriers aren't idle.

rmlarose: I would actually update the PR title and issue to reflect this. (Not to save face, it's actually an important subtlety.)

rmlarose (commented):
LGTM, thanks @andreamari, but not approving yet because of the question below:

I was trying to remember: is there a reason unrelated to this PR that Qiskit barriers are expected to be removed? I think it's because there are no barriers in Cirq? 

Ideally the circuit would be kept as close as possible to the original. We already apply a bunch of special cases to Qiskit circuits to try and keep them as close as possible to the original. Would it be worthwhile to add barriers back as well?

I ask this here not to say it should be done in this PR, but because it may influence how the change in this PR is implemented. But again this LGTM, I just wanted to raise this before explicitly approving.

mitiq/interface/mitiq_qiskit/conversions.py
andreamari: The main point is not whether or not we should consider quits with just barriers as idle qubits or active qubits. This more a matter of definition/convention.

The main problem is that there is a class of qiskit circuits (e.g. the example in the description of this PR) for which Mitiq erroneously moves operations from one qubit to another one. So I think this is a quite dangerous bug!

andreamari:
Thanks @rmlarose, this is a good point.

> I was trying to remember: is there a reason unrelated to this PR that Qiskit barriers are expected to be removed? I think it's because there are no barriers in Cirq?
> 

Yes.

> Ideally the circuit would be kept as close as possible to the original. We already apply a bunch of special cases to Qiskit circuits to try and keep them as close as possible to the original. Would it be worthwhile to add barriers back as well?
> 

That would be the ideal solution. However it's quite difficult to re-insert barriers in the right positions after the circuit has been modified by ZNE/PEC/DDD. So for the moment it's better to fix the problem and I'll open a new issue for exploring how to support  qiskit barriers.

> I ask this here not to say it should be done in this PR, but because it may influence how the change in this PR is implemented. 

If at some point Mitiq will support barriers, we just need to remove this line:
https://github.com/unitaryfund/mitiq/blob/06c6f19f7541ced60626a3df1a51b8ade5d894f9/mitiq/interface/conversions.py#L248

mitiq/interface/conversions.py
```diff
@@ -232,13 +233,21 @@ def noise_scaling_converter(
...
         idle_indices = set()
         if "qiskit" in circuit.__module__:
             from mitiq.interface.mitiq_qiskit.conversions import (
                 _add_identity_to_idle,
             )
-
+            from qiskit.transpiler.passes import RemoveBarriers
+
+            # Avoid mutating the input circuit
+            circuit = deepcopy(circuit)
```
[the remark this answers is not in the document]
andreamari: Makes sense. Done here [fe864c0](https://github.com/unitaryfund/mitiq/pull/1369/commits/fe864c0720380491bdbee85c85b1828b9ee18608)

mitiq/tests/test_conversions.py
```diff
@@ -176,3 +176,40 @@ def test_atomic_one_to_many_converter(to_type):
...
+    test_copy = test_circuit_qiskit.copy()
+
+    scaled = scaling_function(test_circuit_qiskit)
+    # Mitiq is expected to remove qiskit barriers
+    expected = qiskit.QuantumCircuit(4)
+    expected.x(0)
+    expected.x(2)
+    assert scaled == expected
+    # Mitiq should not mutate the input circuit
+    test_circuit_qiskit == test_copy
```
natestemen, line 215: Should this have an `assert`?

andreamari: oops! Yes

rmlarose (approved):
Thanks Andrea, LGTM

mitiq/interface/mitiq_qiskit/conversions.py
rmlarose: Correct me if I'm wrong, but my understanding of the behavior in the example in https://github.com/unitaryfund/mitiq/pull/1369#issue-1285873175 was:

1. A barrier touches qubit `q_1` so it is not considered idle.
2. Because qubit `q_1` is not considered idle, no identity is added to it.
3. When converted to Cirq, the barrier is lost, and there is no identity on qubit `q_1`, so this is lost in the Cirq circuit.

This is why the operations get shuffled on the qubits. So what I meant in https://github.com/unitaryfund/mitiq/pull/1369#discussion_r915261811 is that if (1) above wasn't true, the folded circuit would be as expected. So this can (and should I think) be thought of as refining the definition of "idle qubit".

andreamari: Remed PR title with a more explicit one: _Prevent idle qubits with qiskit barriers from being lost after conversions_

PR #1374, opened 2022-06-29
Add fold methods as parameters to benchmarks

natestemen:
# Description

In showing the benchmark work to Andrea, he suggested we benchmark both `fold_global` and `fold_gates_at_random`. We did this work together while on a call.

**Note:** This work is being merged into `nts-asv-benchmarks` as that branch is not yet merged into master.

**Note 2:** It seems that the full suite of tests do not run on PR's that are not onto `master`. Hence please do not merge this until https://github.com/unitaryfund/mitiq/pull/1354/ is merged, so we can ensure the tests are passing.

### DIFF
--- a/benchmarks/benchmarks.py
+++ b/benchmarks/benchmarks.py
@@ -20,20 +20,33 @@ import os
 
 import networkx as nx
 import numpy as np
+from typing import Callable
 
 import cirq
 from mitiq import benchmarks, raw, pec, zne, Observable, PauliString
 from mitiq.interface import mitiq_cirq
+from mitiq.zne.scaling import fold_gates_at_random, fold_global
 
 
 if os.environ.get("BENCHMARK_CI"):
-    params = {"nqubits": [1], "depth": [1, 2, 3], "num_pec_samples": [10]}
+    params = {
+        "nqubits": [1],
+        "depth": [1, 2, 3],
+        "num_pec_samples": [10],
+        "fold_method": [fold_gates_at_random],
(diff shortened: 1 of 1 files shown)


benchmarks/benchmarks.py
```diff
@@ -20,20 +20,33 @@
...
+        "depth": [1, 2, 3],
+        "num_pec_samples": [10],
+        "fold_method": [fold_gates_at_random],
+    }
 else:
-    params = {"nqubits": [2], "depth": [1, 5, 10], "num_pec_samples": [100]}
+    params = {
+        "nqubits": [2],
+        "depth": [1, 5, 10],
+        "num_pec_samples": [100],
```
andreamari, line 42:
```suggestion
        "num_pec_samples": [100, 500],
```
If you think this is too crazy in terms of execution time, I would propose  "num_pec_samples": [50, 100].
Just to have a rough idea of how the result depends on the number of PEC samples

andreamari (approved):
LGTM! 
I left a small comment which is actually quite unrelated to the PR :-). 
So feel free to apply it or to ignore it. Up to you.

benchmarks/benchmarks.py
natestemen: 500 slowed things down quite a bit. Did 50, and 100 instead in https://github.com/unitaryfund/mitiq/pull/1374/commits/3e595cd213315f3c72f972522a5019a75512cab7

PR #1379, opened 2022-06-30
Support the latest Cirq version (0.15.0)

andreamari:
Description
-----------

Apply changes such that Mitiq is compatible with the latest Cirq version.
Main changes are:

- Notation of Pauli operators. E.g. using `Z(q(1))` instead of `Z(0)`.
- Update of `mitiq.utils._equal()` function. A function to test if two circuits are equal. Now the native `circuit_A == circuit_B` is used instead of comparing DAG (directed acyclic graphs) objects. This means that two circuits with the same gates but different moment structure are now considered considered as different. So, in some tests, the function `_squash_moments()` is used to before comparing two circuits. In practice using `_squash_moments()` + the new `_equal` function is equivalent to using the previous (more tolerant)  `_equal` function.

License
-------


mitiq/zne/scaling/tests/test_folding.py
```diff
@@ -1831,12 +1835,12 @@ def test_apply_fold_mask():
     assert _equal(folded, correct)
 
     folded = _apply_fold_mask(circ, [0, 3, 0, 0, 0, 0])
-    correct = Circuit([ops.H.on(qreg[1])] * 6) + circ
-    assert _equal(folded, correct)
+    correct = Circuit(*[ops.H.on(qreg[1])] * 6) + circ
```
natestemen, line 1838: Is the unpacking `*` needed here? If so, what for?

mitiq/observable/pauli.py
```diff
@@ -84,7 +84,7 @@ def __init__(
 
     @property
     def coeff(self) -> complex:
-        return self._pauli.coefficient
+        return self._pauli.coefficient  # type: ignore
```
natestemen, line 87: Why is this needed now?

natestemen (approved):
Looks great Andrea! Thanks for getting this done so quickly. Just two questions, but nothing that prevents merging.

mitiq/zne/scaling/tests/test_folding.py
andreamari: It's not actually. Removed in [2fdc9ce](https://github.com/unitaryfund/mitiq/pull/1379/commits/2fdc9cec0fe904d046c3ce476fd8f2072aa5eda2)

mitiq/observable/pauli.py
andreamari: I guess `cirq.PauliString` now supports symbolic coefficients beyond complex coefficients. In Mitiq instead we only support /assume complex coefficients. `mypy` don't like this fact.

Issue #1330, opened 2022-06-03
Add Error Mitigation example with PennyLane


Rahul-Mistri:
Issue Description
-----------------
Currently we only have one example on [_Error Mitigation_](https://mitiq.readthedocs.io/en/stable/examples/ibmq-backends.html) which is written with Qiskit as the frontend. PennyLane users would find it useful if we had the same example implemented with PennyLane as the frontend.

Proposed Solution
-----------------
- Implement a PennyLane variant of Qiskit example [(the source file can be found here)](https://github.com/unitaryfund/mitiq/blob/master/docs/source/examples/ibmq-backends.myst)

Additional References
---------------------
- Please consult the [contributing guidelines](https://mitiq.readthedocs.io/en/latest/contributing_docs.html) for specific details on contributing to the documentation, and the [general contributing guidelines](https://mitiq.readthedocs.io/en/latest/toc_contributing.html) as well.

nathanshammah:
The guidelines for the example are:
- Keep the example "experiment" as close as possible to the original one (ZNE, scaling factors, noise level, general syntax and text).
- Change the front end to be PennyLane instead of qiskit. 
- Keep IBMQ device simulator in the back end, using as much as possible pennylane, e.g., [`qml.device`](https://pennylaneqiskit.readthedocs.io/en/latest/devices/ibmq.html)

JLenssen:
Hi, I noticed that Pennylane already provides an example of ZNE using mitiq in their [documentation](https://pennylane.readthedocs.io/en/stable/code/api/pennylane.transforms.mitigate_with_zne.html). 
Should an example for another technique be provided instead (with a link to the ZNE example in the docs)?

andreamari:
Hi @JLenssen, our aim is to translate this basic ZNE example to all supported frontends, so I would keep the current ZNE + PennyLane issue in the current form.

On the other hand, additional examples for other error mitigation techniques (e.g. PEC/DDD/CDR) are welcome too!

PR #1384, opened 2022-07-04
add ZNE example using pennylane

natestemen:
## Description

Add ZNE example using pennylane.

fixes https://github.com/unitaryfund/mitiq/issues/1330


docs/source/examples/pennylane-ibmq-backends.myst
```diff
@@ -0,0 +1,172 @@
...
+    LinearFactory.extrapolate, 
+)()
+
+print(f"Mitigated result {mitigated:.3f}")
+```
+
+To specify a different noise scaling method, we can pass a different function for the argument ``scale_noise``. This
+function should input a circuit and scale factor and return a circuit. The following code block shows an example of
+scaling noise by global folding (instead of local folding, the default behavior for
+{func}`zne.execute_with_zne`).
```
andreamari, line 134:
```suggestion
function should input a circuit and scale factor and return a circuit. The following code block shows an example of
scaling noise by local folding instead of global folding.
```

andreamari (commented):
@natestemen, looks great!
Two important points and an optional one.
1. A comment about local vs global folding is inconsistent with the code. See code suggestion.
2. We should probably add a link to the PennyLane [demo on error mitigation](https://pennylane.ai/qml/demos/tutorial_error_mitigation.html) somewhere in the text.
3.[Optional] The thumbnail of the example could be customized as we did for similar examples. If making a release is more urgent, we can update the thumbnail in the next milestone.

andreamari (approved):
LGTM thanks @natestemen !
Minor aspects:

-  If I am not wrong, a [link to the pennylane demo](https://pennylane.ai/qml/demos/tutorial_error_mitigation.html) is missing.

-  RTD seems frozen, it may be necessary to restart a new build.

PR #1398, opened 2022-07-15
update version.txt to dev status

natestemen:
## Description

Does what is says on the 🥫.

Forgot to do this just after the release.


andreamari (approved):
Thanks!

PR #1409, opened 2022-07-25
Add executor with a backend argument in mitiq_qiskit utils

### LINKED ISSUE #1163: Include backend option for real hardware in mitiq.interface.mitiq_qiskit.qiskit_utils.execute_with_shots

andreamari:
Description
-----------

Adds a `sample_bitstrings` executor function in mitiq_qiskit. This function has a backend argument which makes it easier to run circuits on real devices or mock simulated devices in Qiskit.

Fixes #1163 

Fixes #1360 (stale unitary hack PR).  

License
-------

### DIFF
--- a/mitiq/interface/mitiq_qiskit/tests/test_qiskit_utils.py
+++ b/mitiq/interface/mitiq_qiskit/tests/test_qiskit_utils.py
@@ -253,7 +253,7 @@ def test_compute_expectation_value_on_noisy_backend_with_noise_model():
         qiskit_circuit,
         obs,
         noise_model=initialized_depolarizing_noise(0),
-    )[0]
+    )
 
     assert isinstance(noiseless_expval, complex)
     assert np.isclose(np.imag(noiseless_expval), 0.0)
@@ -264,7 +264,7 @@ def test_compute_expectation_value_on_noisy_backend_with_noise_model():
         qiskit_circuit,
         obs,
         noise_model=initialized_depolarizing_noise(0.01),
-    )[0]
+    )
 
     assert isinstance(expval, complex)
     assert np.isclose(np.imag(expval), 0.0)
@@ -282,7 +282,7 @@ def test_compute_expectation_value_on_noisy_backend_with_qiskit_backend():
         qiskit_circuit,
         obs,
         backend=FakeLima(),
-    )[0]
+    )
 
     assert isinstance(expval, complex)
     assert np.isclose(np.imag(expval), 0.0)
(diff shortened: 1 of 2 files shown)


natestemen (commented):
This looks good, but would you mind adding a bit of explanation as to how this fixes the linked issue? As of now it seems like you've added a function which performs some of the desired functionality (choosing specific backend), but does not return an expectation value like `execute_with_shots` does. Perhaps there's a reason to not include this functionality in that function?

andreamari:
> This looks good, but would you mind adding a bit of explanation as to how this fixes the linked issue? As of now it seems like you've added a function which performs some of the desired functionality (choosing specific backend), but does not return an expectation value like `execute_with_shots` does. Perhaps there's a reason to not include this functionality in that function?

Good point. The reason I didn't modify the old function is that it has been created before the `mitiq.Observable` class was introduced in Mitiq. Moreover it is based on a brute-force diagonalization of a global observable matrix. It's not wrong, but it is also not the optimal approach.

A better approach is to define a function that just returns the noisy bitstrings, such that an expectation value can manually deduced or computed from a `mitiq.Observable` object.

To make the evaluation of an expectation value more clear, I also added a new function that does it.

mitiq/interface/mitiq_qiskit/tests/test_qiskit_utils.py
```diff
@@ -174,3 +180,136 @@ def test_circuit_is_not_mutated_by_executors():
...
+    measurement_result = sample_bitstrings(
+        circuit=two_qubit_circ,
+        backend=None,
+        noise_model=initialized_depolarizing_noise(0),
+        shots=2,
+        measure_all=True,
+    )
+    assert measurement_result.result == [[0, 0], [0, 0]]
+    assert measurement_result.qubit_indices == (0, 1)
+    assert isinstance(measurement_result, MeasurementResult)
```
natestemen, line 214: Is this standard to have what kind of seems like 2 tests in a single function? (First test checks the return value is of the form we expect, and the second ensures `measure_all` works as expected)

andreamari: I guess it's always better to have small tests. I was too lazy. Test splitted.

andreamari:
> Do you think we should deprecate the functions the user was writing about in #1163, or are they still useful alongside the new ones you've added?

I think so. Just opened this issue: https://github.com/unitaryfund/mitiq/issues/1434

PR #1433, opened 2022-08-09
0.18.0 in dev

natestemen:
## Description

for next release.

Is it really necessary to do this after every release? I know it's rare, but for the times we do a patch-release or similar, it is actually wrong for some time.


andreamari (approved):
> Is it really necessary to do this after every release? I know it's rare, but for the times we do a patch-release or similar, it is actually wrong for some time.

Maybe the current method is not perfect but it's useful for me. For example, when calling `mitiq.about()` I one can easily distinguish if the installed version of Mitiq is the released one (e.g. coming from `pip install mitiq`) or the dev version (e.g. coming from `pip install -e .`).

natestemen:
Ah okay, that's a good enough reason for me! Just wanted to make sure that it is indeed being used.

PR #1435, opened 2022-08-10
remove patch release docs

natestemen:
## Description

We recently did a patch release and I followed the instructions for a standard release, only to later find that we have patch release specific docs. I understand it would be nice to only include the commits for the patch, but it seems like more work than is necessary. WDYT about just doing patch releases the same way we do 'regular' releases?

### DIFF
--- a/docs/source/release.myst
+++ b/docs/source/release.myst
@@ -143,24 +143,3 @@ following release, meaning that if `vX.Y.Z` was just released, then
 there should be a section for `vX.(Y+1).Z` that is marked \"In
 Development\". Also, change the version in the `VERSION.txt` file from
 `vX.Y.Z` to `vX.(Y+1).0dev`.
-
-# Releasing a version patch
-
-The steps for the patch should be basically identical to a release,
-however, the commits for the patch should be pushed/cherry-picked onto a
-branch that starts from the tag of the version it is patching. So if you
-had just made the 3.14.0 release (which would have a tag on `master`)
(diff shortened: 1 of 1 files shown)


andreamari (approved):
I always like to make things as simple as possible, so sounds good to me! 

However,  with this PR we basically remove the possibility of doing a patch-release with just a few cherry-picked commits. I don't think there are many use cases so it's fine for me. 
But I would ask a second opinion by anyone in the team before merging.

natestemen:
Sounds good. I've assigned @Misty-W to take a look when she's back.

nathanshammah:
In qutip there is a cherry-pick option for bug fixes, maybe changing the docs to reflect this (rather than tying this to a semantic versioning standard) could help have more freedom (e.g., in disregarding cherry-picking if not needed)?
https://qutip.org/docs/latest/development/release_distribution.html#create-a-bug-fix-release

natestemen:
The qutip docs are great, and very detailed. I think we could use a modified version of their release protocol with [our workflow](https://github.com/unitaryfund/mitiq/blob/6bcf21282942206b2ae12ae2166d95a274e6d333/.github/workflows/gh-release.yml), but it would require some changes as we currently release from master and qutip release from feature branches. I don't think it would be a major change, but it would require some work. With the testing required to ensure the docs are correct, I'm not sure it's worth our time to do this since we don't seem to patch release frequently. In fact, have we ever done a patch release following these docs?

Issue #1448, opened 2022-08-21
REM: Write documentation


amirebrahimi:
Per #815, write the documentation sections for the RFC approved in #1387.

**Requirements**:
Write documentation for the following sections:
1. How do I use REM?
1. When should I use REM?
1. What additional options are available when using REM?
1. What happens when I use REM?
1. What is the theory behind REM?

These are the corresponding sub-issues:
- [x] #1451 
- [x] #1452
- [x] #1453
- [x] #1454
- [x] #1455

**Considerations**:
- A rough outline for the documentation is [here](https://docs.google.com/document/d/1buO5PrO5sS02VXjcaYf37RuR0rF6xpyr4J9H1tI4vN4/edit?pli=1#heading=h.m66w3ecdwtem)

nathanshammah:
Thanks @amirebrahimi! This looks great. Possibly this could be a "super-issue" with sub-issues for each of the five sections, see for example https://github.com/unitaryfund/mitiq/issues/1210.

amirebrahimi:
> Thanks @amirebrahimi! This looks great. Possibly this could be a "super-issue" with sub-issues for each of the five sections, see for example #1210.

Thanks. I've added the sub-issues.

amirebrahimi:
@Misty-W - could you please assign this super ticket and sub-issues to me?

Misty-W:
> @Misty-W - could you please assign this super ticket and sub-issues to me?

Done!

amirebrahimi:
All sub-tasks complete, so closing this ticket.

andreamari:
Thanks @amirebrahimi for all the code and docs support on REM!

Issue #1387, opened 2022-07-06
RFC for Readout Error Mitigation


Misty-W:
Pre-Request Checklist
---------------------

- [X] I checked to make sure that this feature has not already been requested.

Issue Description
-----------------

Review and approve [RFC for Readout Error Mitigation (REM)]
(https://docs.google.com/document/d/1buO5PrO5sS02VXjcaYf37RuR0rF6xpyr4J9H1tI4vN4/edit#heading=h.c072otyej9hw)

Feature request issue #815

Misty-W:
Tried to assign @amirebrahimi but he doesn't appear in the list of possible assignees when I type in his username.

amirebrahimi:
Thanks for the update. I have comments in the RFC that I'll be addressing soon.

amirebrahimi:
I've updated the RFC doc and my branch to address some of the changes requested:
https://github.com/amirebrahimi/mitiq/commit/464578b247c6c5a517b556e94f4541149528b26f
https://github.com/amirebrahimi/mitiq/commit/5d7e51e878c3dd8a3e0c2f1e9a80b585505f4e0c

andreamari:
Thanks @amirebrahimi, I left some comments on the RFC document. 
**For my side, the current version of the RFC can be accepted**! 
However a formal acceptance decision can only be taken after a proper discussion with other members of the Mitiq team. I'll keep you posted!

natestemen:
Nice job Amir, this looks really great. I've made a comment about naming similar to Andrea's, but this is ready to move on to the next steps from me. I've marked it as accepted. Congrats! 🥳 🎉 

I'll close this out now as this ticket was just for tracking the RFC, and we can use https://github.com/unitaryfund/mitiq/issues/815 for the implementation whenever we decide to do that.

PR #1449, opened 2022-08-21
REM implementation

### LINKED ISSUE #1447: REM: RFC implementation
Per #815, implement the RFC approved in #1387

**Requirements**:
- Implement `mitigate_measurements`, `execute_with_rem`, `mitigate_executor`, `rem_decorator`.

**Considerations**:
- Circular imports have been challenging with this module because `_typing` uses `MeasurementResult`.

amirebrahimi:
Resolves #1447 

Description
-----------

This is an implementation for the REM RFC approved in #1387. I'm anticipating that some of the validation checks will fail and will fix accordingly.

Checklist
---------

- [ ] ~I [updated the documentation](../blob/master/docs/CONTRIBUTING_DOCS.md) where relevant.~
  This is a separate ticket #1448 
- [ ] ~Added myself / the copyright holder to the AUTHORS file~
  Already added.

License
-------

Tips
----


natestemen:
@amirebrahimi it looks like you've bumped numpy which is why mypy is still complaining (we have some work to do to upgrade). Don't feel like you have to do that in this PR, unless you need 1.21.5.

amirebrahimi:
> @amirebrahimi it looks like you've bumped numpy which is why mypy is still complaining (we have some work to do to upgrade). Don't feel like you have to do that in this PR, unless you need 1.21.5.

Thanks @natestemen - I think I might have done it when I was testing with a newer version of Cirq, but I'll revert those changes.

amirebrahimi:
> Hi @amirebrahimi, before reviewing this PR I have a preliminary question. Is there a reason for moving `MeasurementResult` in a new submodule? A new module seems a bit overkilling and it's creation is not directly related to this PR.

Agreed that it was not directly related to this PR. It was the simplest way to resolve circular dependencies.

> A better solution (that we can address in a new PR) is maybe to introduce a new `mitiq.types` module with all Mitiq types including those in `mitiq._typing`. Alternatively, `MeasurementResult` could go into mitiq.executors since it's the output of an executor. Whatever we decide it is probably better to do it in a different PR. Do you agree?

If you would like to go the former, then  think a new PR is good. If you want to go with the latter, then I could try moving it into mitiq.executors in this PR and see if that fixes the issue.

andreamari:
> Agreed that it was not directly related to this PR. It was the simplest way to resolve circular dependencies.
>
Ok, I see. 

> > A better solution (that we can address in a new PR) is maybe to introduce a new `mitiq.types` module with all Mitiq types including those in `mitiq._typing`. Alternatively, `MeasurementResult` could go into mitiq.executors since it's the output of an executor. Whatever we decide it is probably better to do it in a different PR. Do you agree?
> 
> If you would like to go the former, then I think a new PR would be good. If you want to go with the latter, then I could try moving it into mitiq.executors in this PR and see if that fixes the issue.

Any of the following are good solutions (if the fix your circular dependency problem). You can choose what works better for you:

- Just moving `MeasurementResult` into `mitiq._types` (if this is enough, it's fine).
- Moving all Mitiq types into a new `mitiq.types` module.
- Moving `MeasurementResult` into `mitiq.executors`.

In any case, it's better to do this preliminary change in a **different PR** since, if there are will be any problems, they will be easier to isolate. Moreover, after mergin the preliminary PR, this error mitigation PR will become much easier to read (now there are 19 changed files).

amirebrahimi:
> Any of the following are good solutions (if they fix your circular dependency problem). You can choose what works better for you:
> 
> * Just moving `MeasurementResult` into `mitiq._types` (if this is enough, it's fine).
> * Moving all Mitiq types into a new `mitiq.types` module.
> * Moving `MeasurementResult` into `mitiq.executors`.
>
> In any case, it's better to do this preliminary change in a **different PR** since, if there are will be any problems, they will be easier to isolate. Moreover, after mergin the preliminary PR, this error mitigation PR will become much easier to read (now there are 19 changed files).

I tried moving it into `mitiq.executors` first, but this did not work, so I opted to move it into `mitiq._typing` in #1484.

.all-contributorsrc
```diff
@@ -355,6 +355,17 @@
...
+    {
+      "login": "amirebrahimi",
+      "name": "Amir Ebrahimi",
+      "avatar_url": "https://avatars.githubusercontent.com/u/904110?v=4",
+      "profile": "https://github.com/amirebrahimi/",
+      "contributions": [
+        "doc",
+        "test",
+        "code"
+      ]
```
andreamari, line 368: See this comment: https://github.com/unitaryfund/mitiq/pull/1484#discussion_r968442938

amirebrahimi:
I accidentally merged the other PR, but once that lands to master it should remove that part of the diff.

natestemen:
You can always remove unwanted commits using `git rebase`, or the simpler method of running `git reset --hard HEAD~N` if you want to remove the last _N_ commits.

More info [here](https://stackoverflow.com/questions/1338728/how-do-i-delete-a-commit-from-a-branch).

amirebrahimi:
> You can always remove unwanted commits using `git rebase`, or the simpler method of running `git reset --hard HEAD~N` if you want to remove the last _N_ commits.
> 
> More info [here](https://stackoverflow.com/questions/1338728/how-do-i-delete-a-commit-from-a-branch).

Thanks, I had already pushed and didn't think it was a good thing to re-write history. Perhaps it's not as big a deal as I think though.

mitiq/rem/rem.py
```diff
@@ -0,0 +1,173 @@
...
+) -> MeasurementResult:
+    """Applies the inverse confusion matrix against the noisy measurement
+    result and returns the adjusted measurements.
+
+    Args:
+        noisy_results: The unmitigated ``MeasurementResult``.
+        inverse_confusion_matrix: The inverse confusion matrix to apply to the
+            probability vector estimated with noisy measurement results.
+    """
+    assert noisy_result.qubit_indices is not None
```
andreamari, line 51: In addition to this assertion, we could raise errors if the size of the inverse confusion matrix is not compatible with the measurement results. E.g. different number of qubits...

mitiq/rem/rem.py
```diff
@@ -0,0 +1,173 @@
...
+from mitiq.executor.executor import Executor
+from mitiq.observable.observable import Observable
+
+MatrixLike = Union[
+    np.ndarray,
+    Iterable[np.ndarray],
+    List[np.ndarray],
+    Sequence[np.ndarray],
+    Tuple[np.ndarray],
+]
```
andreamari, line 36: Is allowing for this freedom in the type useful/necessary?
If so, we should probably test all types. E.g. does the code work if the matrix is given as a `Tuple[np.ndarray]` ?

Otherwise, the simplest option is to just require that all matrices are `np.ndarray` objects, i.e., without introducing `MatrixLike` at all.

mitiq/rem/rem.py
```diff
@@ -0,0 +1,173 @@
...
+
+"""Readout Confusion Inversion."""
+
+from typing import Callable, Iterable, List, Optional, Sequence, Tuple, Union
+
+from functools import partial, wraps
+import numpy as np
+
+from cirq.sim import sample_state_vector as sample_prob_dist
+from cirq.qis.states import to_valid_state_vector as to_prob_dist
```
andreamari, line 24: A state vector and a probability distribution are different mathematical objects. Roughly speaking there is a square root difference between the elements of a state and the element of a probability distribution. 

Isn't this a problem here? It seems that we are using functions that are defined for sampling and generating quantum states for a different task (sampling and generating probability distributions).

andreamari (commented):
Hi @amirebrahimi, this is a first step of the review. I looked at the code but I still need to look at the tests.

Meanwhile we can focus on the code which, btw, looks very good and clean.

mitiq/rem/rem.py
amirebrahimi: I've added an additional assert and a new test.

mitiq/rem/rem.py
amirebrahimi: I've replaced these calls with custom helper methods. You are correct about the semantics of this though. The calls from were Cirq were symmetric, so the result was still the same, but it's unnecessary to turn the measurement result back into a state vector.

amirebrahimi:
Looks like the ubuntu 3.8 test that failed was related to zne; Not sure why that failed though.

mitiq/interface/mitiq_cirq/cirq_utils.py
```diff
@@ -100,3 +104,29 @@ def execute_with_depolarizing_noise(
     rho = simulator.simulate(circuit).final_density_matrix
     expectation = np.real(np.trace(rho @ obs))
     return expectation
+
+
+def generate_inverse_confusion_matrix(
```
andreamari, line 109: This function simulates the experimental estimation of the confusion matrix which fine but a quite overkilling.
Given p0 and p1 there should be a fixed analytical solution for the confusion matrix and its inverse without the need to call `cirq.measure_confusion_matrix`.

Fine to keep the function as it is now in this PR, as long as we open a new issue to simplify this function in a future step.

mitiq/interface/mitiq_cirq/cirq_utils.py
```diff
@@ -100,3 +104,29 @@ def execute_with_depolarizing_noise(
...
+    qubits: Sequence["cirq.Qid"],
```
andreamari, line 110: For this function the specific choice of qubits looks irrelevant for the returned matrix.
Maybe we can simplify this argument with a simpler `num_qubits: int` ?

mitiq/rem/rem.py
```diff
@@ -0,0 +1,182 @@
...
+from mitiq.executor.executor import Executor
+from mitiq.observable.observable import Observable
+
+
+def to_probability_vector(measurement: np.ndarray) -> np.ndarray:
+    """Convert a raw measurement to a probability vector.
+
+    Args:
+        measurement: A single measurement.
+    """
```
andreamari, line 33: I think we need a function that returns the probability vector associated to the full list of measurement results.
Something like:

```python
def bitstrings_to_probabiliy_vector(bitstrings: List[Bitstring]) -> np.ndarray:
    """Converts a list of measured bitstrings to a probability vector estimated as
the empirical frequency of each bitstring (ordered with increasing binary value). 
    Args:
        bitstrings: All measured bitstrings.
    """
```

The code of this new function is probably very similar to the code of your function but **averaged** over all bitstrings.

mitiq/rem/rem.py
```diff
@@ -0,0 +1,182 @@
...
+    index = int("".join([str(m) for m in measurement]), base=2)
+    pv = np.zeros((2 ** measurement.shape[0]), dtype=np.uint8)
+    pv[index] = 1
+    return pv
+
+
+def sample_probability_vector(probability_vector: np.ndarray) -> np.ndarray:
+    """Sample a probability vector and return a raw measurement.
```
andreamari, line 41:
```suggestion
    """Sample a probability vector and returns a bitstring.
```

mitiq/rem/rem.py
```diff
@@ -0,0 +1,182 @@
...
+    result and returns the adjusted measurements.
+
+    Args:
+        noisy_results: The unmitigated ``MeasurementResult``.
+        inverse_confusion_matrix: The inverse confusion matrix to apply to the
+            probability vector estimated with noisy measurement results.
+    """
+    assert noisy_result.qubit_indices is not None
+    num_qubits = len(noisy_result.qubit_indices)
+    assert inverse_confusion_matrix.shape == (2**num_qubits, 2**num_qubits)
```
andreamari, line 67: In general we prefer to raise a `ValueError("Some error message")` if the input of a function is not good.

mitiq/rem/rem.py
```diff
@@ -0,0 +1,182 @@
...
+
+    empirical_prob_dist = np.apply_along_axis(
+        to_probability_vector, 1, noisy_result.asarray
+    )
+
+    adjusted_prob_dist = (inverse_confusion_matrix @ empirical_prob_dist.T).T
+
+    adjusted_result = np.apply_along_axis(
+        sample_probability_vector, 1, adjusted_prob_dist
+    )
```
andreamari, line 77: Here, you are basically applying REM to each individual measurement.
This is however problematic since, for a single measurement, the adjusted_prob_dist will typically contain negative elements. 
If I am not wrong, the correct way of applying REM is:

-  get a **single** empirical probability distribution associated to all the noisy measurements. 
- Multiply the inverse confusion matrix.
- Hopefully the result doesn't contain negative elements. If it does, we need to think about a way to handle this case.
- Resample new bitstrings from the adjusted distribution. The number of resampled bitstrings should equal to the number of the original noisy bistrings. 
- [Possible alternative to the previous point] Deduce a list of new bitstrings whose frequency best approximate the adjusted distribution.

mitiq/rem/rem.py
```diff
@@ -0,0 +1,182 @@
...
+    result = MeasurementResult(adjusted_result, noisy_result.qubit_indices)
+    return result
+
+
+def execute_with_rem(
+    circuit: QPROGRAM,
+    executor: Union[Executor, Callable[[QPROGRAM], MeasurementResult]],
+    inverse_confusion_matrix: np.ndarray,
+    *,
+    observable: Optional[Observable] = None,
```
andreamari, line 88: For all techniques we always use` circuit, executor, observable` as the first 3 positional arguments. The other keyword-only arguments (after the `*`) are specific to the technique. 
So in this case we need something like:
```python
def execute_with_rem(
    circuit: QPROGRAM,
    executor: Union[Executor, Callable[[QPROGRAM], MeasurementResult]],
    observable: Optional[Observable] = None,    
    *,
    inverse_confusion_matrix: np.ndarray,
```

mitiq/rem/rem.py
```diff
@@ -0,0 +1,182 @@
...
+        probability_vector: A probability vector.
+    """
+    result = np.random.choice(
+        len(probability_vector), size=1, p=probability_vector
+    )[0]
+    value = [int(i) for i in bin(result)[2:]]
+    return np.array(value, dtype=np.uint8)
+
+
+def mitigate_measurements(
```
andreamari, line 53: It is better if this function and its utilities go into a different file called e.g. `confusion_inversion.py`.
`rem.py` should contain just the high-level functions `execute_with_rem`, `mitigate_executor`, `rem_decorator`.

Note that in Mitiq we use different test files `test_filename.py` associated to different code files. This helps find test of specific functions.

So we probably need a new test file if we create a new file.

andreamari (commented):
This is a second round of review about the code. Still not about tests yet.

mitiq/interface/mitiq_cirq/cirq_utils.py
amirebrahimi: Unfortunately, as this function currently is it does require the actual Cirq `qubits` to be passed due to line 125:
```
qubit_measures = [[q] for q in qubits]
```

mitiq/rem/rem.py
amirebrahimi: Can you share more about where we get negative elements?

mitiq/rem/rem.py
amirebrahimi: I can make the change, but in this case we should require an inverse confusion matrix or would you rather leave it as a named parameter and throw a value error if one is not passed?

mitiq/rem/rem.py
andreamari: **The issue**
Even if the confusion matrix contains positive elements, its inverse (inverse confusion matrix) $M^{-1}$ typically contains negative elements. This means that a positive empirical distribution $p$, can be mapped to an adjusted distribution $p' = M^{-1} p$, where some elements of $p'$ are negative. 

Typically negative terms are very small since, by hypothesis, $p$ is not an arbitrary distribution but a noisy distribution having approximately the structure $p=M p_{\rm ideal}$. Unfortunately, negative elements are still possible due to statistical fluctuations or noise miscalibrations.

This effect is much stronger if REM is applied to individual shots (this is the current approach if I am not wrong), instead of applying REM to the averaged empirical distribution (i.e. the vector of frequencies of measured bitstrings).

You can reproduce this _negative probabilities_ issue, if you replace `p0=1` in your tests with something more intermediate e.g.  `p0=0.1`.

**Possible solution**
A possible solution is to clip small negative values of $p'$ to zero and renormalize. The pseudo code I have in mind is:
```python
# Clip negative numbers
adjusted_prob = np.array(adjusted_prob).clip(min=0)
# Renormalize
adjusted_prob = adjusted_prob / np.sum(adjusted_prob)
```

mitiq/rem/rem.py
andreamari: My proposal is that the argument `inverse_confusion_matrix` should be explicitly passed as a named parameter. Since it doesn't have a default value, the user will get an error if  `inverse_confusion_matrix` is not passed.

We have the same situation with PEC, where `representations` is a necessary parameter but it's not a positional argument.

mitiq/interface/mitiq_cirq/cirq_utils.py
amirebrahimi: Added an issue as #1562

mitiq/rem/rem.py
amirebrahimi: Thanks for this explanation. I tried with p0=0.1 and I can see now what you were concerned about. I'll need a little more time with this to correct it.

One question for now:
> get a single empirical probability distribution associated to all the noisy measurements.

I get that this will be a single vector corresponding to the possible outcomes (e.g. measurements for 2 qubits would have four values corresponding to 00, 01, 10, 11). However, should these be raw counts or should they sum to 1?

`[128, 256, 512, 128]` vs. `[0.125, 0.25, 0.5, 0.125]`

mitiq/rem/rem.py
amirebrahimi: Replaced the asserts with `ValueError`

mitiq/rem/rem.py
amirebrahimi: Fixed.

mitiq/rem/rem.py
andreamari: > Thanks for this explanation. I tried with p0=0.1 and I can see now what you were concerned about. I'll need a little more time with this to correct it.
> 
> One question for now:
> 
> > get a single empirical probability distribution associated to all the noisy measurements.
> 
> I get that this will be a single vector corresponding to the possible outcomes (e.g. measurements for 2 qubits would have four values corresponding to 00, 01, 10, 11). However, should these be raw counts or should they sum to 1?
> 
> `[128, 256, 512, 128]` vs. `[0.125, 0.25, 0.5, 0.125]`

I would maybe normalize such that sum=1. But I don't have a strong preference.  Up to you. 

Note that even if using integer counts (without normalization), after multiplying for $M^{-1}$ the result is a vector with non-integer elements. So I don't see a strong advantage in using integer counts compared to normalized distributions.

mitiq/rem/inverse_confusion_matrix.py
```diff
@@ -0,0 +1,82 @@
...
+def to_probability_vector(measurement: np.ndarray) -> np.ndarray:
+    """Convert a raw measurement to a probability vector.
+
+    Args:
+        measurement: A single measurement.
+    """
+    index = int("".join([str(m) for m in measurement]), base=2)
+    pv = np.zeros((2 ** measurement.shape[0]), dtype=np.uint8)
+    pv[index] = 1
+    return pv
```
natestemen, line 30: I wonder if this function should go in the `MeasurementResult` class? WDYT?

mitiq/rem/rem.py
```diff
@@ -0,0 +1,128 @@
...
+
+    result = executor._run([circuit])
+    noisy_result = result[0]
+    assert isinstance(noisy_result, MeasurementResult)
+
+    mitigated_result = mitigate_measurements(
+        noisy_result, inverse_confusion_matrix
+    )
+    assert observable is not None
+    return observable._expectation_from_measurements([mitigated_result])
```
natestemen, line 57:
```suggestion
    noisy_result = executor._run([circuit])[0]
    assert isinstance(noisy_result, MeasurementResult)

    mitigated_result = mitigate_measurements(
        noisy_result, inverse_confusion_matrix
    )
    assert observable is not None
    return observable._expectation_from_measurements([mitigated_result])
```

We don't have many `assert` statements in our code outside of tests. Is there a reason to use them here over more explicitly raising an error?

natestemen (commented):
Nice work Amir. I've left a few small comments.

Also just FYI we may try and integrate with [mthree](https://github.com/Qiskit-Partners/mthree) in the future (https://github.com/unitaryfund/mitiq/issues/1567). Just thought I'd let you know in case you like the measurement error mitigation stuff and want to continue it.

mitiq/rem/rem.py
amirebrahimi: This has been added.

mitiq/rem/rem.py
amirebrahimi: Please have a look. I clip the negative values as you suggested.

mitiq/rem/inverse_confusion_matrix.py
```diff
@@ -0,0 +1,82 @@
...
+    """Sample a probability vector and returns a bitstring.
+
+    Args:
+        probability_vector: A probability vector.
+    """
+    result = np.random.choice(
+        len(probability_vector), size=1, p=probability_vector
+    )[0]
+    value = [int(i) for i in bin(result)[2:]]
+    return np.array(value, dtype=np.uint8)
```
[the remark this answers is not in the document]
amirebrahimi: This method has been removed.

mitiq/rem/inverse_confusion_matrix.py
amirebrahimi: This function was removed.

mitiq/rem/rem.py
amirebrahimi: I've replaced these with exceptions

amirebrahimi:
> Nice work Amir. I've left a few small comments.
> 
> Also just FYI we may try and integrate with [mthree](https://github.com/Qiskit-Partners/mthree) in the future (#1567). Just thought I'd let you know in case you like the measurement error mitigation stuff and want to continue it.

Thanks, @natestemen. Looks like a useful package. For the time being, I'm mostly busy with my Master's program, so just trying to finish up this feature until I take on any more work.

amirebrahimi:
Any idea why the `asv` check is failing? Seems related to an environment config.
I merged latest and see that pennylane tests are failing.

Misty-W:
> Any idea why the `asv` check is failing? Seems related to an environment config. I merged latest and see that pennylane tests are failing.

Correct, the problem `asv` was in our asv config file. It's fixed now, so once you update or rebase your branch `asv` should pass.

natestemen:
It looks like the tests are failing on python 3.8 due to an error related to using `list` vs `List` in the type hints. Using `list` in type hints is only available in newer versions of python (I thought 3.10, but maybe 3.9 as well). You'll have to use the `List` type from the `typing` module instead.

amirebrahimi:
> It looks like the tests are failing on python 3.8 due to an error related to using `list` vs `List` in the type hints. Using `list` in type hints is only available in newer versions of python (I thought 3.10, but maybe 3.9 as well). You'll have to use the `List` type from the `typing` module instead.

Thanks, Nate. A lesson to keep scrolling further -- I stopped once I saw the errors related to pyquil and pennylane tests and figured it was test instability or something outside of my changes.

So, I'll wait to hear if the current changes are sufficient for the implementation and if so will move on to adding additional tests.

mitiq/rem/rem.py
```diff
@@ -0,0 +1,134 @@
...
+    noisy_result = result[0]
+    if not isinstance(noisy_result, MeasurementResult):
+        raise TypeError("Results are not of type MeasurementResult.")
+
+    mitigated_result = mitigate_measurements(
+        noisy_result, inverse_confusion_matrix
+    )
+
+    if observable is None:
+        raise ValueError("An observable is required.")
```
natestemen, line 60: In the function declaration `observable` defaults to `None` so I would expect this error to be raised very often. I think ideally this function wouldn't require an observable, right?

mitiq/rem/rem.py
```diff
@@ -0,0 +1,134 @@
...
+from mitiq.rem.inverse_confusion_matrix import mitigate_measurements
+
+
+def execute_with_rem(
+    circuit: QPROGRAM,
+    executor: Union[Executor, Callable[[QPROGRAM], MeasurementResult]],
+    observable: Optional[Observable] = None,
+    *,
+    inverse_confusion_matrix: npt.NDArray[np.float64],
+) -> float:
```
natestemen, line 36: I only recently learning about this `*` syntax to make arguments keyword only! Very cool. Just out of curiosity, is there a reason you want `inverse_confusion_matrix` to be keyword only?

mitiq/rem/inverse_confusion_matrix.py
```diff
@@ -0,0 +1,108 @@
...
+    """Converts a list of measured bitstrings to a probability vector estimated
+    as the empirical frequency of each bitstring (ordered with increasing
+    binary value).
+
+    Args:
+        bitstrings: All measured bitstrings.
+    """
+    pv = np.zeros(2 ** len(bitstrings[0]))
+    for bs in bitstrings:
+        index = int("".join([str(bit) for bit in bs]), base=2)
```
natestemen, line 61: I _think_ this should be marginally faster, and IMO a bit cleaner, but it's also unlikely to make much of an effect since the bitstrings we are likely to deal with are relatively short. Take it or leave it!
```suggestion
        index = int("".join(map(str, bs)), base=2)
```

mitiq/rem/inverse_confusion_matrix.py
```diff
@@ -0,0 +1,108 @@
...
+from typing import List
+import numpy as np
+import numpy.typing as npt
+
+from mitiq._typing import MeasurementResult, Bitstring
+
+
+def sample_probability_vector(
+    probability_vector: npt.NDArray[np.float64], samples: int
+) -> npt.NDArray[np.uint8]:
```
natestemen, line 25: Should this function returns `Bitstring`s?

natestemen (commented):
Nice work Amir. I think this is looking great. I left a few comments, but nothing blocking from adding some more tests if you want to go ahead and do that.

- There are a few docstrings with a missing `Returns:` section to describe what the function returns. Can you make sure all added functions have those, and that they are up to date.
- To check my understanding, the reason we ask the user to pass the `inverse_confusion_matrix` into `execute_with_rem` is because it can be measured once, and then reused many times. If we didn't ask the user to supply it, we would have to calculate it every time `execute_with_rem` is called, which would be very inefficient. Is that right?

andreamari (commented):
Hi @amirebrahimi, thanks again for this great contribution! 
The code looks good to me! However I would add the tests I suggested in the comment below. 

If the suggested additional tests don't fail, this is approved from my side!

mitiq/rem/rem.py
andreamari: @natestemen, sorry to step in, but I asked @amirebrahimi to make `inverse_confusion_matrix`  a keyword-only argument.
The reason is for consistency with all other `execute_with_xxx` functions, e.g. PEC and DDD. The motivation is that we want `observable` to be the third positional argument but with a default value of `None`. 
This fact forces all the arguments on the right side to be keyword-only.

Another reason is that keyword-only arguments are great to avoid position mistakes, especially when there are many options. But this is just my opinion and I am not sure if it is a best practice.

natestemen: I'm in favor of it! I was just curious if there was a reason we used it here. Thanks for the info.

mitiq/rem/rem.py
amirebrahimi: Yes that would make sense, but that would be breaking convention from other mitigation techniques (e.g. zne, pec, etc.). I was just following convention. If you are okay with making this required, then I will change it.

mitiq/rem/inverse_confusion_matrix.py
amirebrahimi: I like that much better. Thanks.

mitiq/rem/inverse_confusion_matrix.py
amirebrahimi: Yes, that makes more sense since I was calling `.tolist()` where it was used.

amirebrahimi:
> * There are a few docstrings with a missing `Returns:` section to describe what the function returns. Can you make sure all added functions have those, and that they are up to date.

I've fixed these.

> * To check my understanding, the reason we ask the user to pass the `inverse_confusion_matrix` into `execute_with_rem` is because it can be measured once, and then reused many times. If we didn't ask the user to supply it, we would have to calculate it every time `execute_with_rem` is called, which would be very inefficient. Is that right?

Yes, the confusion matrix would be measured every time you wanted to calibrate to the device, then inverted, and re-used many times to mitigate measurements.

mitiq/rem/tests/test_rem.py
```diff
@@ -0,0 +1,175 @@
...
+
+    assert second_executor.__doc__ == first_executor.__doc__
+
+
+def test_mitigate_executor():
+    true_rem_value = -2.0
+
+    # test with an executor that completely flips results
+    p0 = 1
+    p1 = 1
```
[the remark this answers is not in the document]
amirebrahimi: I've added more tests.

mitiq/rem/rem.py
amirebrahimi: I've added a new file: `test_inverse_confusion_matrix.py` with additional unit tests (although we were already covered 100% without them)

amirebrahimi:
Okay, I think I'm ready for a final review @natestemen and @andreamari

mitiq/rem/__init__.py
```diff
@@ -16,3 +16,9 @@
 """Readout error mitigation (REM) techniques."""
 
 from mitiq.rem.post_select import post_select
+from mitiq.rem.inverse_confusion_matrix import mitigate_measurements
+from mitiq.rem.rem import (
+    execute_with_rem,
+    mitigate_executor,
+    rem_decorator,
```
andreamari, line 23:
```suggestion
    rem_decorator,
    mitigate_measurements,
```

andreamari (approved):
LGTM! :champagne:

nathanshammah:
@natestemen does this look good to review / merge? Excited to see this in the codebase.

natestemen (approved):
Nice work Amir! This is going to be great to have.

Misty-W:
@amirebrahimi when you're ready please click "squash and merge" so it can go out with the upcoming release (milestone 0.21).

Thanks and excited to have this in Mitiq!

Misty-W:
> Thanks, @Misty-W. I don't have write access, so I will need someone on the Mitiq team to merge.

Got it, merged now!

PR #1315, opened 2022-05-27
Simplify issue/PR templates

natestemen:
## Description

This PR moves the majority of the instructions in issue and PR templates to comments, and removes some extraneous material.

### Motivation

As it stands, I think we have too much boilerplate in our issue and PR templates. This both increases the barrier to open an issue, and makes it more difficult to skim issues to obtain an understanding of the problem. For example (but not to pick on), https://github.com/unitaryfund/mitiq/issues/1214 includes 2 lines describing the issue, and a bulk of text that's been left unedited from the boilerplate.

In making these changes I references what other major open source projects use for these templates such as [cirq](https://github.com/quantumlib/Cirq/blob/master/.github/ISSUE_TEMPLATE/bug_report.md), [pennylane](https://github.com/PennyLaneAI/pennylane/blob/master/.github/ISSUE_TEMPLATE/bug_report.yml), [qiskit](https://github.com/Qiskit/qiskit/blob/master/.github/ISSUE_TEMPLATE/bug_report.md), [vscode](https://github.com/microsoft/vscode/blob/main/.github/ISSUE_TEMPLATE/bug_report.md), and [nix](https://github.com/NixOS/nixpkgs/blob/master/.github/ISSUE_TEMPLATE/bug_report.md).

**Question**: I'd like to remove the "License" checkbox in the PR template as well (and perhaps add a line to the hidden comments saying something to the affect of "by opening a PR you are agreeing to ..."), but I'm not sure if there's a reason to keep the checkbox around.

---

It's likely easiest to review this PR commit by commit.

cc @nathanshammah, @andreamari, and @Misty-W since this affects all of us.

---

### License

- [x] I license this contribution under the terms of the GNU GPL, version 3 and grant Unitary Fund the right to provide additional permissions as described in section 7 of the GNU GPL, version 3.

PR #1456, opened 2022-08-22
Add whitepaper publication info/links

natestemen:
## Description

Now that the whitepaper is published, we can update the references to the arXiv. Two other notable changes that are made in this PR

- I've removed the `AUTHORS` file as it does not accurately reflect the authors to this project. It is not nearly as complete as https://github.com/unitaryfund/mitiq/graphs/contributors, and is an additional file that needs to be maintained if kept.
- Instead of continuing to use the `CITATION.cff` format, I've replaced the file with `CITATION.bib` which is the bibtex entry we'd like people to cite. This still works with GitHub's "Cite this repository" feature ([docs](https://docs.github.com/en/repositories/managing-your-repositorys-settings-and-features/customizing-your-repository/about-citation-files#other-citation-files)), but does not require us to deal with the intricacies of `cff`. I believe this is much better as including our own `bib` file allows us to control _exactly_ how we'd like the citation to look, rather than the `cff -> bib` converter making arbitrary choices. Here's an example of what GitHub's "cite this repo" feature will look like when using a `bib` file: https://github.com/natestemen/citation-test.

There is [one line](https://github.com/unitaryfund/mitiq/blob/6bcf21282942206b2ae12ae2166d95a274e6d333/README.md?plain=1#L10) in the readme that still uses the arXiv link for the badge. I'm not aware of an equivalent we could replace this with, however, so I recommend we keep it.

---

This is a follow on from https://github.com/unitaryfund/unitary-fund/pull/255.


natestemen:
The test is failing due to me adding a link to where the new `CITATIONS.bib` file will live once merged, but all else looks good.

nathanshammah (changes requested):
Thanks @natestemen looks good and thank you for the clean-up. Some comments:
- Please keep the AUTHORS file as that is what we're asking to update in the template PR (I commented about this in https://github.com/unitaryfund/mitiq/pull/1315#issuecomment-1148360610). It's related to licensing and a different matter from citation.
- Using the citation.bib looks fine to me. We switched to .cff to be able to have the "Cite this repository" feature in home page. I guess the one caveat is that I hope it does not discourage contributions; the other minor one is this "GitHub will link to these files in the Cite this repository prompt, but will not attempt to parse them into other citation formats." from the page you linked. We can try.
- We also have a [refs.bib](https://github.com/unitaryfund/mitiq/blob/master/docs/source/refs.bib) file with all citations. I think we should cite Mitiq in the Docs's References section too. Can you add the bibitem to the refs.bib file too and include it in the references? If we can list the citation.bib in the latex build, even better, but I don't know how to list two citation files for latex. It would be nice to have it show up as first reference at https://mitiq.readthedocs.io/en/latest/zz_bibliography.html. That's actually what I had first in mind, but good to uniform this.

natestemen:
> Please keep the AUTHORS file

Done in https://github.com/unitaryfund/mitiq/pull/1456/commits/d2b190ca16be861cda1db20ef58f078a5eb72e3e.

> It would be nice to have it show up as first reference at https://mitiq.readthedocs.io/en/latest/zz_bibliography.html.

Agreed. However, due to how we sort the bibliography in https://github.com/unitaryfund/mitiq/blob/ea2b8cb074d3b3efb93f113e1dd5c5198626d1c5/docs/source/zz_bibliography.rst, it will not show up on the top. Instead it is citation # 40.

natestemen:
ping @nathanshammah

nathanshammah:
Thanks @natestemen, LGTM, if tests pass.

andreamari:
> The test is failing due to the linkcheck failing as mentioned in [#1456 (comment)](https://github.com/unitaryfund/mitiq/pull/1456#issuecomment-1225045102). If there is a proper way to add new github file links, let me know!

Have you tried this: `[`CITATION.bib`](CITATION.bib)` .  
Otherwise I think it's ok to merge with the current absolute link, even if linkcheck will fail at the first attempt.

natestemen:
@andreamari I did have something very similar in https://github.com/unitaryfund/mitiq/pull/1456/commits/2f5a3f9520ad33f798fd88582505c759c2dc28d3, but myst/sphinx doesn't seem to understand local file references. 😞

nathanshammah (approved):

Issue #1488, opened 2022-09-14
Add BQSkit <> Mitiq tutorial to docs


nathanshammah:
## Issue Description

It would be great to add a BQskit <> Mitiq tutorial to the docs to show the use of Mitiq with a compiler.

## Proposed Solution

Rework the existing notebook (https://github.com/unitaryfund/research/blob/main/ieee-quantum-week/compilation-with-error-mitigation-tutorial/bqskit.ipynb) into myst format, addressing the remaining comments in [that PR](https://github.com/unitaryfund/research/pull/8#pullrequestreview-1102627669) and updating docs requirements if they need to be updated with bqskit.

Issue #1214, opened 2022-04-08
Remove Mypy errors due to numpy ndarray arg-type check


nathanshammah:
Issue Description
-----------------

Currently the dependabot PR updating numpy is failing due to some mypy issues. 
One of the issue is that numpy type check is now required to define the ndarray datatype.

PR #1489, opened 2022-09-15
BQSKit tutorial

natestemen:
## Description

Follow on from https://github.com/unitaryfund/research/pull/8 to port this integration to our docs. Many improvements included as recommended in https://github.com/unitaryfund/research/pull/8#pullrequestreview-1102627669.

fixes https://github.com/unitaryfund/mitiq/issues/1488

### TODO:
- [x] add references
- [x] thumbnail


nathanshammah:
The error seems to be:
```
WARNING: Execution Failed: /home/runner/work/mitiq/mitiq/docs/source/examples/bqskit.ipynb
WARNING: Couldn't find cache key for notebook file source/examples/bqskit.ipynb. Outputs will not be inserted.
  Last execution failed with traceback saved in /home/runner/work/mitiq/mitiq/docs/build/html/reports/bqskit.log
```
The output is present in the html.

**Thumbnail**
Let's add an image. I suggest using the histogram as thumbnail. If you want, we could even include a logo of BQSkit in that png, as it can be separately uploaded, just like for [pennylane](https://github.com/unitaryfund/mitiq/blob/5e53c83f4043f6fdab44b5b229e5c1ce1327832f/docs/source/_thumbnails/mol-h2-vqe-pl.png).
You need to add the image here: https://github.com/unitaryfund/mitiq/tree/master/docs/source/_thumbnails
and add in the conf.py a line with a link to it https://github.com/unitaryfund/mitiq/blob/5e53c83f4043f6fdab44b5b229e5c1ce1327832f/docs/source/conf.py#L340

natestemen:
#### Update as some things have changed

The BQSKit notebook, converted to myst, failed to build on readthedocs due to an obscure unicode error. It looked like it was coming from a transitive dependency of BQSKit. Perhaps an unpinned dependency that upgraded over the course of the few weeks since we wrote this notebook and now. That means that if someone tries to exactly recreate this notebook, it will likely fail.

That said, I think this is okay. As of two days ago, BQSKit [released](https://github.com/BQSKit/bqskit/releases/tag/1.0.0) their 1.0 version which includes many improvements that will drastically simplify this notebook. Once we upgrade numpy (tracked in https://github.com/unitaryfund/mitiq/issues/1214) we will be able to use the latest version of bqskit and all will work. It's an unfortunate situation to be in, but I do not see a way around this currently. If you have ideas, I'm all ears!

natestemen:
Thanks for the recommendations, it reads more clearly now.

PR #1198, opened 2022-03-31
Raise error if executor and observable not compatible

Misty-W:
Description
-----------

Replaces PR #1186, attempted workaround for GH Action issue

Fixes #1140 

Checklist
---------

Check off the following once complete (or if not applicable) after opening the PR. The PR will be reviewed once this checklist is complete and all tests are passing.

- [X] I added unit tests for new code.
- [X] I used [type hints](https://www.python.org/dev/peps/pep-0484/) in function signatures.
- [X] I used [Google-style](https://sphinxcontrib-napoleon.readthedocs.io/en/latest/example_google.html) docstrings for functions.
- [X] I [updated the documentation](../blob/master/docs/CONTRIBUTING_DOCS.md) where relevant.
- [X] Added myself / the copyright holder to the AUTHORS file

If some items remain, you can mark this a [draft pull request](https://github.blog/2019-02-14-introducing-draft-pull-requests/).

License
-------

- [X] I license this contribution under the terms of the GNU GPL, version 3 and grant Unitary Fund the right to provide additional permissions as described in section 7 of the GNU GPL, version 3.

Tips
----

- If the validation check fails:

    1. Run `make check-types` (from the root directory of the repository) and fix any [mypy](https://mypy.readthedocs.io/en/stable/) errors.

    2. Run `make check-style` and fix any [flake8](http://flake8.pycqa.org) errors.

    3. Run `make format` to format your code with the [black](https://black.readthedocs.io/en/stable/index.html) autoformatter.

  For more information, check the [Mitiq style guidelines](https://mitiq.readthedocs.io/en/stable/contributing.html#style-guidelines).

- Write "Fixes #XYZ" in the description if this PR fixes Issue #XYZ.

PR #1491, opened 2022-09-17
Unmitigated version of learning-based PEC

Misty-W:
Alternate implementation of learning-based PEC: uses noisy (unmitigated) data to learn noise model

## Description

---

### License

### DIFF
--- a/mitiq/executor/executor.py
+++ b/mitiq/executor/executor.py
@@ -150,16 +150,7 @@ class Executor:
                 for circuit_with_measurements in observable.measure_in(circuit)
             ]
             result_step = observable.ngroups
-        elif (
-            observable is not None
-            and self._executor_return_type not in MeasurementResultLike
-            and self._executor_return_type not in DensityMatrixLike
-        ):
-            raise ValueError(
-                """Executor and observable are not compatible. Executors
-                returning expectation values as float must be used with
-                observable=None"""
-            )
+
         else:
             all_circuits = circuits
             result_step = 1
--- a/mitiq/pec/representations/learning.py
+++ b/mitiq/pec/representations/learning.py
@@ -101,14 +101,18 @@ def biased_noise_loss_function(
 
 
 def unmitigated_loss_function(
-    epsilon_guess, noisy_executor, training_circuits, observable=None
-):
+    epsilon_guess: float,
+    noisy_executor: Executor,
+    training_circuits: List[QPROGRAM],
+    observable: Optional[Observable] = None,
+) -> float:
     def noise_model_execute(circ: Circuit) -> np.ndarray:
         circuit = convert_to_mitiq(circ)[0]
         noisy_circ = circuit.with_noise(
             biased_noise_channel(epsilon=epsilon_guess[0], eta=0)
         )
-        return compute_density_matrix(noisy_circ, noise_level=(0.0,))
+        density_matrix = compute_density_matrix(noisy_circ, noise_level=(0.0,))
+        return density_matrix
 
     noise_model_executor = Executor(noise_model_execute)
     noise_model_values = np.array(
@@ -146,13 +150,13 @@ def biased_noise_channel(epsilon: float, eta: float) -> MixedUnitaryChannel:
 
 
 def learn_noise_parameters_from_unmitigated_data(
-    circuit,
-    epsilon0,
-    noisy_executor,
-    num_training_circuits,
-    fraction_non_clifford,
-    training_random_state=None,
-    observable=None,
+    circuit: QPROGRAM,
+    epsilon0: float,
+    noisy_executor: Executor,
+    num_training_circuits: int = 5,
+    fraction_non_clifford: float = 0.2,
+    training_random_state: np.random.RandomState = None,
+    observable: Optional[Observable] = None,
 ):
     training_circuits = generate_training_circuits(
         circuit,
--- a/mitiq/pec/representations/learning.py
+++ b/mitiq/pec/representations/learning.py
@@ -17,14 +17,28 @@ learning-based technique."""
 
 from typing import Optional, Dict, Any, List
 import numpy as np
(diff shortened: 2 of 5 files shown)


Misty-W:
@andreamari, I implemented the "unmitigated" learning routine in this PR. Tests of the learning function are passing and execution time is looking reasonable. 

I realize I forgot to add a test for the new loss function. I did test it in a notebook and just need to add it to the test file.

Note: I had to remove the check in `executor.evaluate` for compatibility of `Observable` and `Executor` to get the Qiskit test to pass. Looking at the original issue (which I closed with PR #1198 ) I think that check doesn't belong in `executor.evaluate` but rather in `execute_w_[xxx]` or `mitigate_executor`. Should I open a separate PR to remove that error message? I'm not 100% sure where it should go instead, so for now I'd just like to remove it.

mitiq/pec/representations/learning.py
```diff
@@ -84,3 +98,78 @@ def _biased_noise_loss_function(
...
+
+
+def learn_noise_parameters_from_unmitigated_data(
+    circuit: QPROGRAM,
+    epsilon0: float,
+    noisy_executor: Executor,
+    num_training_circuits: int = 5,
+    fraction_non_clifford: float = 0.2,
+    training_random_state: np.random.RandomState = None,
+    observable: Optional[Observable] = None,
```
andreamari, line 161: This function could take `operations_to_learn` as an input, as we did for the PEC learning function.
In that case, `_unmitigated_loss_function` should simulate noise only after the operations in `operations_to_learn` .

mitiq/pec/representations/learning.py
```diff
@@ -84,3 +98,78 @@ def _biased_noise_loss_function(
     return np.sum((mitigated_values - ideal_values) ** 2) / len(
         training_circuits
     )
+
+
+def _unmitigated_loss_function(
+    epsilon_guess: np.ndarray,
```
andreamari, line 104: Docstring is missing

mitiq/pec/representations/learning.py
```diff
@@ -84,3 +98,78 @@ def _biased_noise_loss_function(
...
+        (a, unitary(I)),
+        (b, unitary(Z)),
+        (c, unitary(X)),
+        (c, unitary(Y)),
+    ]
+    return ops.MixedUnitaryChannel(mix)
+
+
+def learn_noise_parameters_from_unmitigated_data(
+    circuit: QPROGRAM,
```
andreamari, line 155: docstring is missing

mitiq/executor/executor.py
```diff
@@ -150,16 +150,7 @@ def evaluate(
...
-        elif (
-            observable is not None
-            and self._executor_return_type not in MeasurementResultLike
-            and self._executor_return_type not in DensityMatrixLike
-        ):
-            raise ValueError(
-                """Executor and observable are not compatible. Executors
-                returning expectation values as float must be used with
-                observable=None"""
-            )
```
andreamari, line 162: Could you give more details why you had to remove this check?
What is the part of the code that is incompatible with this?

andreamari (commented):
This looks promising! Do you think we should keep both the original PEC-based learning method and this new unmitigated learning method? Or are we eventually going to choose one of the two?

nathanshammah:
+1 for first porting just this simplified model at first. This could also simplify some naming of functions.

mitiq/executor/executor.py
Misty-W: > This looks promising! Do you think we should keep both the original PEC-based learning method and this new unmitigated learning method? Or are we eventually going to choose one of the two?

I think we should choose one or the other- testing and maintaining both is probably more overhead than it's worth. I implemented this "unmitigated" version because I thought the original, mitigated loss function had reached somewhat of a dead end. In working out the unmitigated version I found an approach that also worked for the mitigated loss function. We can keep this open as a backup option until we merge #1358 .

Misty-W: > Could you give more details why you had to remove this check? What is the part of the code that is incompatible with this?

If we don't test on Qiskit, removing this check isn't needed as far this PR is concerned, but I think it might be a problem in general. When using `qiskit_utils.execute`, `execute_with_noise`, etc, `executor.evaluate` will raise an error that the executor and observable are not compatible. Shouldn't the compatibility check be done when applying error mitigation, e.g. in `execute_with_pec`? Btw, I was the one who added the compatibility check to `evaluate` in #1198.

andreamari: Looking better your qiskit test, the fix is probably to set `observable=None` in `learn_noise_parameters_from_unmitigated_data`. Indeed, after you fix the observable in the executor, it becomes and observable-less one (i.e. it returns a float) and so it must be used with `observable=None` in all Mitiq techniques.

mitiq/pec/representations/learning.py
```diff
@@ -84,3 +98,78 @@ def _biased_noise_loss_function(
...
+        [
+            noisy_executor.evaluate(training_circuit, observable)
+            for training_circuit in training_circuits
+        ]
+    )
+
+    return np.sum(
+        abs(noise_model_values.reshape(-1, 1) - noisy_values.reshape(-1, 1))
+        ** 2
+    ) / len(training_circuits)
```
natestemen, line 137: I believe the `reshape` calls are unneeded here. The return value can be simplified as well if you prefer this:
```py
return np.linalg.norm(noise_model_values - noisy_values) ** 2 / len(training_circuits)
```

Misty-W:
@andreamari, @nathanshammah, @natestemen, closing this PR as I believe it's no longer needed. 
This was opened as an alternative to training on PEC-mitigated expectation values, but `learn_depolarizing_noise_parameters` is already released in Mitiq and soon we'll have `learn_biased_noise_parameters` from #1510 merged.

Let me know if any objections and I can reopen :)

PR #1497, opened 2022-09-20
Restore the titles of the examples in the Mitiq docs

### LINKED ISSUE #1450: Titles with thubmnails in examples in documentation
I remember that an issue previously in the documentation were titles going out of bounds in the thumbnails.  I didn't notice the change that now makes the thumbnails in the documentation take a full row of space on https://mitiq.readthedocs.io/en/latest/examples/examples.html

Having large thumbnails looks ok to me, but we need to have titles (even shortened ones are fine) next to the thumbnails, otherwise it's too cryptic for some examples, i.e. those that just show plots.  Also because hovering with the mouse on the titles in the index on the left does not highlight the corresponding thumbnail, so there is no easy way to know the title of an example shown in a thumbnail. examples-mitiq

andreamari:
## Description

Restores the layout problems of the Mitiq gallery of examples.
Fixes #1450 

---

### License

### DIFF
--- a/dev_requirements.txt
+++ b/dev_requirements.txt
@@ -19,7 +19,7 @@ sphinx-autodoc-typehints~=1.12.0
 myst-nb~=0.12.3
 pydata-sphinx-theme~=0.8.1
 jupytext==1.11.1
-sphinx-gallery
+sphinx-gallery==0.10.1
 nbsphinx
 openfermion==1.5.1; sys_platform != 'win32'
 openfermionpyscf==0.5; sys_platform != 'win32'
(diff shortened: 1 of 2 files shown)


natestemen (approved):
Nice find!

Out of curiosity, how did you pin it down?

andreamari:
> Nice find!
> 
> Out of curiosity, how did you pin it down?

It worked in my local environment (where I had an old version of sphinx-gallery). It didn't work after upgrading `sphinx-gallery`. So I pinned it in `dev_requirements.txt`

Issue #1254, opened 2022-04-28
Consider dropping support for complex expectation values.


andreamari:
We currently assume that observables are generic and so expectation values can be complex. 
This has some pros and cons.

**Pros:**
- More general setting

**Cons:**
- The code and the annotations of functions are less readable. Often there is confusion between floats and complex return values.
- Most importantly, some techniques are not compatible with the mitigation of complex expectation values and
often suspicious warnings like the following appear when using mitiq:
```
 /opt/hostedtoolcache/Python/3.7.12/x64/lib/python3.7/site-packages/numpy/core/_asarray.py:102: ComplexWarning: Casting complex values to real discards the imaginary part
    return array(a, dtype, copy=False, order=order)
```
Two alternative solutions are proposed below.

## Proposed solution A (real expectation values)

- Always assume that user-defined executors return real expectation values or list of real expectation values.
- Always assume that `Executor.evaluate()` returns real expectation values.
- Always assume that a `mitiq.Observable` is Hermitian, such that its expectation value is real.
- Raise an error if the user defines an non-Hermitian observable.

## Note
The user can still decompose any non hermitian observable A as a linear combination of two Hermitian observables (A1, A2), such that: A = A1+ i A2 .
So the restriction to Hermitian observables is not strong as it seems.

## Proposed solution B (complex expectation values)
- Keep using complex expectation values
- Make sure executors that return complex expectation values are always compatible with Mitiq
- Make sure all techniques are compatible with complex expectation values. When this is not possible, split complex expectation values into real and imaginary parts, and mitigate both terms independently. 
- Make sure that all function annotations are consistent with the possibility of complex expectation values.

WrathfulSpatula:
For Hermitian observables, I believe expectation values are _strictly theoretically real-valued_, by the way. Is there even a need for non-Hermitian operator expectation values? I'm not sure that would make sense, in the context of quantum theory.

natestemen:
I found this review comment thread from last year on this issue: https://github.com/unitaryfund/mitiq/pull/1001#discussion_r736396474.

PR #1501, opened 2022-09-22
(partially) upgrade numpy

natestemen:
## Description

Upgrade to `numpy==1.21.6`. This requires adding new type hints for numpy arrays. 64-bit precision was chosen for all type specifications as 64-bit precision is python's default floating point precision. There are some places where I saw `np.complex128`. Is there a need for this high of precision?

I've done my best to try and understand where arrays should be int vs float vs complex, but it's hard to tell in some cases. I've also had to add `# type: ignore` comments in a few places as I have not been able to work out how to satisfy `mypy`. Most of these cases seem to be surrounding code where we are passing around random seeds that are optional.

---

fixes https://github.com/unitaryfund/mitiq/issues/1214

This upgrade will also close out the dependabot security alert related to [CVE-2021-33430](https://nvd.nist.gov/vuln/detail/CVE-2021-33430).


andreamari (approved):
Great job @natestemen ! 
LGTM up to a minor simplification. 

Note: There is an intrinsic inconsistency in Mitiq since a large part of the old code (e.g. factories) assume that expectation values are real. After introducing observables, we extended expectation values to complex numbers. I am surprised `mypy` is not complaining about this :smile: .  This is quite confusing and at some point it should be fixed in some way see e.g. #1254  .   Anyway this PR has a different scope, so it's good to merge it now.

andreamari:
I tried a couple of changes in executor.evaluate() by I could not find any good solution to simplify `DensityMatrixLike`,  `MeasurementResultLike` etc..
So for me it's fine to revert the last commit such that we can merge this PR.

mitiq/executor/executor.py
```diff
@@ -85,6 +87,9 @@ def __init__(
 
         executor_annotation = inspect.getfullargspec(executor).annotations
         self._executor_return_type = executor_annotation.get("return")
+        print(executor)
+        print(executor_annotation)
+        print(self._executor_return_type)
```
andreamari, line 92:
```suggestion
```
This is probably the reason for the failing test

andreamari: I'll self-commit this just to start CI tests.

nathanshammah (approved):
@natestemen looks great. I am not aware of the need of higher precision, so fine with complex64. With regards to the choice of float vs complex, it's mainly when a density matrix could be passed. We may spot later on that some typings need a switch, but since checks are passing and all looks good, we can merge. Thanks @andreamari for the bug fix.

Issue #1493, opened 2022-09-19
Update sphinx config


natestemen:
When running `make docs` I'm getting some deprecation warnings:
```
WARNING: 'jupyter_execute_notebooks' is deprecated for 'nb_execution_mode' [mystnb.config]
WARNING: 'execution_timeout' is deprecated for 'nb_execution_timeout' [mystnb.config]
WARNING: 'nb_render_priority' is deprecated for 'nb_mime_priority_overrides'mystnb.config
```

These will need to be updated in [`docs/source/conf.py`](https://github.com/unitaryfund/mitiq/blob/a48f35f9ba7fab2f491c2baac2a905b6f3706b40/docs/source/conf.py). Documentation is [here](https://www.sphinx-doc.org/en/master/usage/configuration.html).

nathanshammah:
Nice find.

Issue #1292, opened 2022-05-16
Test Mitiq with more recent versions of Python


andreamari:
Mitiq is currently tested with Python 3.7 and Python 3.8.
It would be good, if possible, to upgrade versions by one or two steps. 
I.e., testing Mitiq with:

- Python 3.8 and 3.9
or
- Python 3.8, 3.9 and 3.10

This issue is not just about _testing_ but also about _fixing_  possible issues that will emerge when testing recent Python versions.
For example, dependency conflicts may happen as reported in #1285

saki-zapata:
Any plan to upgrade the `numpy` version to make it compatible with `python 3.10` (#1285)?

andreamari:
This is quite important indeed. I'll add it to the next milestone but I am not sure if it's going to be an easy fix.

Issue #1427, opened 2022-08-05
Speed up and/or remove redundant unit tests


Aaron-Robertson:
Pre-Request Checklist
---------------------

- [x] I checked to make sure that this feature has not already been requested.

Issue Description
-----------------

Local test times are onerous for Mitiq contributors and for the repository actions. Contribution would be easier if tests run faster.

Based on my own local testing, the 15 slowest tests are 
|Time        |     Test                                                                                                                                                                                         |
-----------|--------------------------------------------------------------------------------------------------------------------------------------
|99.69s |     mitiq/pec/tests/test_pec.py::test_pec_decorator_qiskit
|82.92s | mitiq/zne/tests/test_zne.py::test_with_observable_two_qubits[sample_bitstrings]
|49.32s | mitiq/pec/tests/test_pec.py::test_qiskit_noiseless_decomposition_multiqubit[1]
|45.16s | mitiq/pec/tests/test_pec.py::test_qiskit_noiseless_decomposition_multiqubit[2]
|30.31s | mitiq/zne/tests/test_zne.py::test_with_observable_batched_factory[sample_bitstrings]
|28.99s | mitiq/zne/tests/test_zne.py::test_with_observable_adaptive_factory[sample_bitstrings]
|26.77s | mitiq/cdr/tests/test_cdr.py::test_mitigated_execute_with_cdr[5-kwargs1-linear_fit_function-pennylane]
|26.55s | mitiq/cdr/tests/test_cdr.py::test_mitigated_execute_with_cdr[5-kwargs1-linear_fit_function_no_intercept-pennylane]
|26.50s | mitiq/cdr/tests/test_cdr.py::test_mitigated_execute_with_cdr[5-kwargs0-linear_fit_function_no_intercept-pennylane]
|25.77s | mitiq/cdr/tests/test_cdr.py::test_mitigated_execute_with_cdr[5-kwargs0-linear_fit_function_no_intercept-qiskit]
|25.31s | mitiq/cdr/tests/test_cdr.py::test_mitigated_execute_with_cdr[3-kwargs1-linear_fit_function-pennylane]
|24.93s | mitiq/cdr/tests/test_cdr.py::test_mitigated_execute_with_cdr[3-kwargs0-linear_fit_function_no_intercept-pennylane]
|24.74s | mitiq/cdr/tests/test_cdr.py::test_mitigated_execute_with_cdr[5-kwargs0-linear_fit_function-qiskit]
[...]

PR #1504, opened 2022-09-22
Support python 3.10

natestemen:
## Description

This PR adds support for the latest released version of python (3.10). What's new in this version of python can be found [here](https://docs.python.org/3/whatsnew/3.10.html), but I've avoided using the new features as we want to continue support for python 3.8 and 3.9.

The main changes are some dependency upgrades that were previously not compatible with python 3.10, and as a result some changes to documentation as it seems myst has gotten stricter with the updates when it comes to references.

---

fixes https://github.com/unitaryfund/mitiq/issues/1292
fixes https://github.com/unitaryfund/mitiq/issues/1493

FYI: Python EOL schedule: https://endoflife.date/python

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -17,7 +17,7 @@ jobs:
       - name: Set up Python
         uses: actions/setup-python@v4
         with:
-          python-version: 3.9
+          python-version: '3.10'
       - name: Install Python dependencies
         run: |
           python -m pip install --upgrade pip
@@ -40,7 +40,7 @@ jobs:
       - name: Set up Python
         uses: actions/setup-python@v4
         with:
-          python-version: 3.9
+          python-version: '3.10'
 
       - name: Install Python dependencies
         run: |
@@ -68,7 +68,7 @@ jobs:
       - name: Set up Python
         uses: actions/setup-python@v4
         with:
-          python-version: 3.9
+          python-version: '3.10'
       - name: Install Mitiq
         run: |
           python -m pip install --upgrade pip
@@ -87,7 +87,7 @@ jobs:
       - name: Set up Python
         uses: actions/setup-python@v4
         with:
-          python-version: 3.9
+          python-version: '3.10'
       - name: Install Mitiq
         run: |
           python -m pip install --upgrade pip
@@ -104,7 +104,7 @@ jobs:
     strategy:
       matrix:
         os: [ubuntu-latest]
-        python-version: [3.8, 3.9]
+        python-version: ['3.8', '3.9', '3.10']
     runs-on: ${{ matrix.os }}
     steps:
       - name: Check out mitiq
@@ -128,7 +128,7 @@ jobs:
         run: make test-all
       - name: Submit coverage report to Codecov
         # Only submit to Codecov once.
-        if: ${{ matrix.python-version == 3.9 }}
+        if: ${{ matrix.python-version == '3.10' }}
         uses: codecov/codecov-action@v3.1.1
         with:
           fail_ci_if_error: true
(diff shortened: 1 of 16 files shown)


docs/source/conf.py
```diff
@@ -113,6 +113,8 @@
     "smartquotes",
 ]
 
+myst_heading_anchors = 3
+
```
andreamari, line 117: Nice catch!

pyproject.toml
```diff
@@ -1,20 +1,3 @@
...
-    | \.mypy_cache
-    | \.tox
-    | \.venv
-    | _build
-    | buck-out
-    | build
-    | dist
-  )/
-)
-'''
```
andreamari, line 20: Is the removal of `include` and `exclude` intentional? If so, fine for me.

docs/source/conf.py
```diff
@@ -113,45 +113,21 @@
...
-        "application/javascript",
-        "text/html",
-        "image/svg+xml",
-        "image/png",
-        "image/jpeg",
-        "text/markdown",
-        "text/latex",
-        "text/plain",
-    ),
-}
```
andreamari, line 147: Nice cleaning

andreamari (approved):
Great to be up-to-date with Python too! LGTM

pyproject.toml
natestemen: Yes, I removed this additional config in [`aeabb02`](https://github.com/unitaryfund/mitiq/pull/1504/commits/aeabb02a9caf71dac27be0f28ad82f548691f589). The motivation for doing so is to ensure we are using whatever `include`/`exclude` patterns black recommends. The above lists have changed slightly in black since we added this code, so I think it's best if we just default to what they recommend.

docs/source/conf.py
natestemen: Again, this seems to be a case of adding default parameters to our config. If it's default, lets get rid of it! Less code means less maintenance burden 😄

Misty-W:
@natestemen I can look at this tomorrow. I see Andrea already approved so if you want to go ahead and merge that's okay too.

.github/workflows/test.yml
```diff
@@ -137,7 +137,7 @@ jobs:
     strategy:
       matrix:
         os: [macos-latest, windows-latest]
-        python-version: [3.8, 3.9]
+        python-version: ['3.8', '3.9', '3.10']
```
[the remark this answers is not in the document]
natestemen: I think we still want to support users on python 3.8, at least for now. Python 3.8's EOL isn't for another 2 years. We dropped python 3.7 in https://github.com/unitaryfund/mitiq/pull/1505 as
1. It's EOL is in 9 months (a little preemptive of us, but also pt 2/3)
2. 3.8 is very similar to 3.7
3. [Numpy 1.22](https://numpy.org/doc/stable/release/1.22.0-notes.html) drops support for python 3.7

Since we are hoping to be on numpy1.22 soon, it just made sense to drop 3.7.

I don't think we have as much reason to drop 3.8 just yet. Are you saying that despite github actions supposedly running in parallel, in effective it never quite does? I've definitely jobs get started at different times as well. Do you think removing 3.8 would make CI builds faster?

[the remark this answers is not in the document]
natestemen: We do have https://github.com/unitaryfund/mitiq/issues/1427 open to help reduce CI build times, and I agree that dropping 3.8 could help. But I worry that we could break something for python 3.8 users over the next two years. Do you think it's worth the tradeoff?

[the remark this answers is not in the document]
Misty-W: I agree we should continue supporting 3.8 as long as it's practical, so let's keep testing it for now. We can revisit after the changes proposed in #1427, if further reductions in build time become a priority.

PR #1505, opened 2022-09-22
Support python 3.9

natestemen:
## Description

Adds support for python 3.9. This includes building the project on 3.9.

**Question**: Should we drop support for python 3.7? It is EOL in 9 months (https://endoflife.date/python). Seems appropriate to me, especially since if mitiq works with 3.8, it will very likely work with 3.7 due to the similarity of the two versions. The advantage to dropping it is slightly faster CI builds.

---

part of https://github.com/unitaryfund/mitiq/issues/1292. More PRs to come to support python 3.10, but that is proving to be more nontrivial https://github.com/unitaryfund/mitiq/pull/1504.

### DIFF
--- a/.github/workflows/asv.yaml
+++ b/.github/workflows/asv.yaml
@@ -17,7 +17,7 @@ jobs:
       - name: Set up Python
         uses: actions/setup-python@v4
         with:
-          python-version: 3.8
+          python-version: 3.9
 
       - name: Run benchmarks
         run: |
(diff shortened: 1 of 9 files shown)


andreamari (approved):
Thanks @natestemen! 
+1 on dropping tests for Python `3.7`.

nathanshammah (approved):
LGTM, great!

Issue #1280, opened 2022-05-13
Add example for learning-based PEC


Misty-W:
Pre-Request Checklist
---------------------

- [X] I checked to make sure that this feature has not already been requested.

Issue Description
-----------------

Add example showing how to use learning-based PEC

Thanks for helping us improve Mitiq! 🙂

Proposed Solution
-----------------

Let us know what you'd like to see happen, and be sure to mention
any alternatives that you've considered.

Additional References
---------------------

If applicable, provide some references that will help us better understand the request.

PR #1514, opened 2022-09-27
First demo of learning function

Misty-W:
## Description

Adding tutorial example for learning-based PEC workflow. 

fixes https://github.com/unitaryfund/mitiq/issues/1280

---

### License


Misty-W:
Added a label in `pec.myst` for linking to the PEC section of the user guide (instead of re-explaining PEC in the text of the example).

nathanshammah:
Thanks @Misty-W. Just to clarify, a side comment because there are a couple of PRs open: 
- https://github.com/unitaryfund/mitiq/pull/1395: this PR needs to be updated for the user guide section of the docs, correct? Or shall we close it? I would at least add some information in the PEC users guide under "advanced options" file 4. 
- https://github.com/unitaryfund/mitiq/pull/1510: this is WIP
- https://github.com/unitaryfund/mitiq/pull/1491: this has been put on hold, if I understand correctly

I can provide comments on this PR.

Misty-W:
> Thanks @Misty-W. Just to clarify, a side comment because there are a couple of PRs open:
> 
> * [Write user guide section for learning-based PEC #1395](https://github.com/unitaryfund/mitiq/pull/1395): this PR needs to be updated for the user guide section of the docs, correct? Or shall we close it? I would at least add some information in the PEC users guide under "advanced options" file 4.

Let’s keep the PR for the user guide open and I’ll update it after this PR is closed.

> * [Learn biased noise parameters #1510](https://github.com/unitaryfund/mitiq/pull/1510): this is WIP

Yes, I plan to update this in the next week.

> * [Unmitigated version of learning-based PEC #1491](https://github.com/unitaryfund/mitiq/pull/1491): this has been put on hold, if I understand correctly

Yes, let’s revisit the plan after the other open learning-based PEC PRs are closed.

> I can provide comments on this PR.

Great, thanks! I added you as a reviewer.

docs/source/examples/learning-depolarizing-noise.myst
```diff
@@ -0,0 +1,208 @@
...
+print(circuit)
+```
+
+In this example, we will learn the strength of depolarizing noise of the `CNOT` gate in the circuit.
+
+```{code-cell} ipython3
+CNOT_ops = list(circuit.findall_operations_with_gate_type(CXPowGate))
+index = CNOT_ops[0][0]
+op = CNOT_ops[0][1]
+operations_to_learn = [Circuit(op)]
```
natestemen, line 67:
```suggestion
CNOT_op = next(circuit.findall_operations_with_gate_type(CXPowGate))
moment_index, op, _ = CNOT_op
operations_to_learn = [Circuit(op)]
```
WDYT of this?

I was a little lost when I got down to `noisy_execute` because I didn't remember what `op` and `index` were. I wonder if this code should be moved into the same cell as that one, so it's more fresh in the readers mind.

docs/source/examples/learning-depolarizing-noise.myst
```diff
@@ -0,0 +1,208 @@
...
+```
+
+Specify the observable of interest.
+
+```{code-cell} ipython3
+observable = Observable(PauliString("XZ"), PauliString("YY"))
+```
```
natestemen, line 74: Is there anything special about this observable, or was it just chosen at random? If random, can we simplify and go without an explicit observable?

docs/source/examples/learning-depolarizing-noise.myst
```diff
@@ -0,0 +1,208 @@
...
+
+
+def noisy_execute(circ: Circuit) -> np.ndarray:
+    noisy_circ = circ.copy()
+    qubits = op.qubits
+    for q in qubits:
+        noisy_circ.insert(
+            index + 1,
+            DepolarizingChannel(epsilon)(q),
+            strategy=InsertStrategy.EARLIEST,
```
natestemen, line 99: This looks to be the default strategy, so I think we should leave it out.

docs/source/examples/learning-depolarizing-noise.myst
```diff
@@ -0,0 +1,208 @@
...
+        )
+    return ideal_execute(noisy_circ)
+
+
+noisy_executor = Executor(noisy_execute)
+```
+
+Before calling the optimizer, let's plot the loss function that will be minimized in the learning routine, over a small range of noise strength values. The loss function calls {func}`execute_with_pec`, and we can optionally pass keyword arguments to set the number of PEC samples, among other options. Here we set a relatively small value of `num_samples` to obtain a reasonable execution time. However, we avoid using a number of PEC samples that is too small, as it can result in a large statistical error and ultimately cause the optimization process to fail.
```
natestemen, line 107: {func} link is broken. I think the function name needs to start with `.`? I'm not entirely sure.

docs/source/conf.py
```diff
@@ -333,6 +333,7 @@ def get_incollection_template(self, e):
     "examples/mitiq-paper/*": "_static/mitiq-codeblocks.png",
     "examples/zne-braket-ionq": "_static/zne-braket-ionq.png",
     "examples/bqskit": "_static/bqskit.png",
+    "examples/learning-depolarizing-noise": "_static/learn-depolarizing.png",
```
natestemen, line 336: It looks like the thumbnail image needs to be moved into `docs/source/_thumbnails` in order to show on the gallery page.

docs/source/examples/learning-depolarizing-noise.myst
```diff
@@ -0,0 +1,208 @@
...
+```
+
+Comparing with unmitigated and ideal values, we see that the mitigated value, calculated with the depolarizing noise strength learned from Clifford circuit data is closer to the ideal value, as compared with the noisy value.
+
+```{code-cell} ipython3
+unmitigated = noisy_executor.evaluate(circuit, observable)[0].real
+ideal = ideal_executor.evaluate(circuit, observable)[0].real
+print(unmitigated)
+print(ideal)
+```
```
natestemen, line 208: We should be more explicit here about relative errors. It's great for the final line to show that using this technique decreases your error!

natestemen (changes requested):
Nice work Misty. It's cool to get a better understanding of what you've been working on, in action!

Some general feedback:
- Some more explanation between cells might be helpful instead of describing what the code is doing.
- Inserting some references to learning based PEC would be great! Got to help people sink their teeth in if they want to.

Some technical feedback:
- Some of the lines are really long, which makes it slightly harder to read and give feedback on, especially using GitHub's "suggestion" tool. A line limit on this file might be good, but my favorite way to do it in markdown/markup languages is to stick to a one sentence per line rule.
- The git history on this PR seems to have gotten jumbled, and it's hard to follow how we got to the current state. There are a few commits touching files that are not involved with this PR which is very strange. Happy for you (or I) to rewrite history to clean it up.

docs/source/examples/learning-depolarizing-noise.myst
```diff
@@ -0,0 +1,208 @@
+---
+jupytext:
+  text_representation:
+    extension: .myst
+    format_name: myst
+    format_version: 0.13
+    jupytext_version: 1.14.1
+kernelspec:
+  display_name: Python 3.8.13 64-bit ('venv_mitiq_learnrepspec')
```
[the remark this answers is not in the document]
Misty-W: Guess this is an artifact of conversion?

docs/source/examples/learning-depolarizing-noise.myst
```diff
@@ -0,0 +1,208 @@
...
+  language: python
+  name:  python3
+---
+
+# Learning quasiprobability representations with a depolarizing noise model
+
++++
```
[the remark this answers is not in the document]
Misty-W: Which line?

docs/source/examples/learning-depolarizing-noise.myst
Misty-W: At one point I was using the `next` function here, and I forget why I changed it. It should still work and it looks nicer.

docs/source/examples/learning-depolarizing-noise.myst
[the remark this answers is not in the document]
natestemen: line with
```
+++
```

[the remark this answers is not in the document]
Misty-W: It's a block break. It's used for mapping to cell-based formats, e.g. jupyter.

docs/source/examples/learning-depolarizing-noise.myst
Misty-W: Observable has to be defined for `compute_density_matrix`

docs/source/examples/learning-depolarizing-noise.myst
Misty-W: Good point!

docs/source/examples/learning-depolarizing-noise.myst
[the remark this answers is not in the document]
natestemen: To validate my understanding, if we had a myst file that contained
```
some text
+++
more text
```
then if you converted this to a jupyter notebook, you would get two distinct markdown cells?

If so, is there a reason we want that here?

docs/source/examples/learning-depolarizing-noise.myst
Misty-W: that fixed it, thanks!

docs/source/examples/learning-depolarizing-noise.myst
Misty-W: Agreed. Updated to be relative, but I need to fix the formatting.

docs/source/conf.py
Misty-W: I thought it was just a rendering issue on my browser, but moving into `docs/source/_thumbnails` was the right call.

docs/source/examples/learning-depolarizing-noise.myst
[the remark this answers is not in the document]
Misty-W: I guess we don't really need it here- I had separate cells for title and first paragraph, and didn't combine them before conversion to `.myst` format. I can take it out in my next round of edits.

docs/source/examples/learning-depolarizing-noise.myst
Misty-W: Moved to codeblock computing and plotting loss function.

natestemen: I don't think that's true. The definition of [`compute_density_matrix`](https://github.com/unitaryfund/mitiq/blob/1eca124f764d7b29ef5b751cce729479cdcfb92f/mitiq/interface/mitiq_cirq/cirq_utils.py#L61-L65) doesn't take an observable variable, and in the definition of [`evaluate`](https://github.com/unitaryfund/mitiq/blob/1eca124f764d7b29ef5b751cce729479cdcfb92f/mitiq/executor/executor.py#L119-L125) is the observable variable is optional.

https://github.com/unitaryfund/mitiq/blob/1eca124f764d7b29ef5b751cce729479cdcfb92f/mitiq/interface/mitiq_cirq/cirq_utils.py#L61-L65
https://github.com/unitaryfund/mitiq/blob/1eca124f764d7b29ef5b751cce729479cdcfb92f/mitiq/executor/executor.py#L119-L125

Misty-W: Strange, I remember getting an error when setting the `Observable` to `None` when using `evaluate` with `compute_density_matrix`. Anyway, the specified observable happens to work well for the optimization. I've moved it into the code cell for computing the loss function, since I don't think it needs to be emphasized.

Misty-W:
Thanks for the review @natestemen! 
> 
> Some general feedback:
> 
> * Some more explanation between cells might be helpful instead of describing what the code is doing.

Agreed, I've replaced some of the descriptive text with explanations. Let me know what in particular needs additional clarification- it helps to get an outside perspective!

> * Inserting some references to learning based PEC would be great! Got to help people sink their teeth in if they want to.

Thanks for the reminder! I added a citation to the learning-based quantum error mitigation paper, as well as links to the PEC and CDR sections of the guide and the API-doc.

> Some technical feedback:
> 
> * Some of the lines are really long, which makes it slightly harder to read and give feedback on, especially using GitHub's "suggestion" tool. A line limit on this file might be good, but my favorite way to do it in markdown/markup languages is to stick to a one sentence per line rule.

Good point, I've imposed both.

> * The git history on this PR seems to have gotten jumbled, and it's hard to follow how we got to the current state. There are a few commits touching files that are not involved with this PR which is very strange. Happy for you (or I) to rewrite history to clean it up.

Something weird happened to the branch, and it took multiple tries to update it. I didn't think it was possible to rewrite the commit history after pushing to GitHub, but it'd be nice to clean it up.

docs/source/examples/learning-depolarizing-noise.myst
```diff
@@ -0,0 +1,221 @@
...
+from mitiq.cdr import generate_training_circuits
+from mitiq.cdr._testing import random_x_z_cnot_circuit
+from mitiq.pec import execute_with_pec
+from mitiq.pec.representations.depolarizing import (
+    represent_operation_with_local_depolarizing_noise,
+)
+from mitiq.pec.representations.learning import (
+    biased_noise_loss_function,
+    learn_depolarizing_noise_parameter,
+)
```
andreamari, line 54: Please check if we need all the imports. E.g. gates

docs/source/examples/learning-depolarizing-noise.myst
```diff
@@ -0,0 +1,221 @@
...
+
+Since the learning-based workflow uses the function {func}`.cdr.generate_training_circuits` from Mitiq's {ref}`Clifford data regression module
+<guide/cdr/cdr>` to generate the near-Clifford training circuits, the input circuit must be one that is compiled into the Rx-Rz-CNOT gateset. 
+Here we use a simple Rx-Rz-CNOT circuit, with an (optional) seed for reproducibility.
+
+```{code-cell} ipython3
+circuit = random_x_z_cnot_circuit(
+    LineQubit.range(2), n_moments=5, random_state=np.random.RandomState(1)
+)
+print(circuit)
```
andreamari, line 65: Can we show an example with more than 1 CNOT gate?
You can try to change seed or `n_moments`, or just manually define the circuit without calling the CDR function. 

More than 1 CNOT allows us to test the technique one step further in the direction of a real use case.

If for some reason it doesn't work, we come back to the single CNOT case.

docs/source/examples/learning-depolarizing-noise.myst
```diff
@@ -0,0 +1,221 @@
...
+```
+
+
+Define the ideal executor function for simulating the Clifford training circuits without noise. 
+This will be used for comparison to the error-mitigated expectation values in the learning function, and for a final comparison with the
+mitigated and unmitigated values at the end of the workflow.
+
+```{code-cell} ipython3
+def ideal_execute(circ: Circuit) -> np.ndarray:
+    return compute_density_matrix(circ, noise_level=(0.0,))
```
andreamari, line 75:
```suggestion
def ideal_execute(circ: Circuit) -> np.ndarray:
    # For large circuits, this should be replaced by a near-Clifford simulator.
    return compute_density_matrix(circ, noise_level=(0.0,))
```

docs/source/examples/learning-depolarizing-noise.myst
```diff
@@ -0,0 +1,221 @@
...
+```
+
+Define the noisy executor, in this case for simulating depolarizing noise on the `CNOT` gate in each of the training circuits. 
+The optimized value of the noise strength `epsilon` should be close to the value defined for this executor.
+
+```{code-cell} ipython3
+CNOT_ops = list(circuit.findall_operations_with_gate_type(CXPowGate))
+index = CNOT_ops[0][0]
+op = CNOT_ops[0][1]
+operations_to_learn = [Circuit(op)]
```
andreamari, line 88: If we go for multiple CNOTs this part may need to change a bit

docs/source/examples/learning-depolarizing-noise.myst
```diff
@@ -0,0 +1,221 @@
...
+
+epsilon = 0.05
+
+
+def noisy_execute(circ: Circuit) -> np.ndarray:
+    noisy_circ = circ.copy()
+    qubits = op.qubits
+    for q in qubits:
+        noisy_circ.insert(
```
andreamari, line 97: This need to change if there are more than 1 CNOTs. 
[Circuit.batch_insert()](https://quantumai.google/reference/python/cirq/Circuit#batch_insert) or https://quantumai.google/reference/python/cirq/map_operations_and_unroll may be useful.

andreamari (commented):
Thanks @Misty-W, I left some optional suggestions.
Do you have any idea of how long is the execution time?

docs/source/examples/learning-depolarizing-noise.myst
Misty-W: thank you, this is what I was looking for in the Cirq docs and concluded didn't exist!

docs/source/examples/learning-depolarizing-noise.myst
Misty-W: Thanks for the reminder! Just pushed a new commit with cleaned-up imports.

docs/source/examples/learning-depolarizing-noise.myst
Misty-W: Locally this works and takes about 5 min to run. We'll see what it does in CI :)

docs/source/examples/learning-depolarizing-noise.myst
Misty-W: Indeed, this is updated to match the new circuit and generalized executor.

Misty-W:
> Thanks @Misty-W, I left some optional suggestions. Do you have any idea of how long is the execution time? Comparing with docs build time on master this seems 15-20 min longer but maybe I am wrong.

Thanks for the suggestions @andreamari! 

Although I've increased the number of moments in the circuit (5 --> 10), I was able to bring the build time down to 25 min (same as latest commit on master [4ea99b4](https://github.com/unitaryfund/mitiq/commit/4ea99b473888a1d828e4f976a383d1ff1ea6eec6)) by reducing the number of points calculated for the loss function plot. 

Anything else that should be addressed before merging?

andreamari (approved):
LGTM, thanks @Misty-W !
Note: This is our first non-ZNE example!

docs/source/examples/learning-depolarizing-noise.myst
```diff
@@ -0,0 +1,210 @@
...
+from mitiq.pec.representations.depolarizing import (
+    represent_operation_with_local_depolarizing_noise,
+)
+from mitiq.pec.representations.learning import (
+    biased_noise_loss_function,
+    learn_depolarizing_noise_parameter,
+)
+```
+
+Since the learning-based workflow uses the function {func}`.cdr.generate_training_circuits` from Mitiq's {ref}`Clifford data regression module
```
natestemen, line 48: The link to `generate_training_circuits` doesn't link with the API-doc. Should it be `{func}.generate_training_circuits`?

docs/source/examples/learning-depolarizing-noise.myst
```diff
@@ -0,0 +1,210 @@
...
+    noisy_circ.batch_insert(insertions)
+
+    return ideal_execute(noisy_circ)
+
+
+noisy_executor = Executor(noisy_execute)
+```
+
+Before calling the optimizer, let's plot the loss function that will be minimized in the learning routine, over a small range of noise strength values.
+The loss function calls {func}`pec.execute_with_pec`, and we can optionally pass keyword arguments to set the number of PEC samples, among other options.
```
natestemen, line 99: Same problem with the link to API-doc with `execute_with_pec`.

docs/source/examples/learning-depolarizing-noise.myst
```diff
@@ -0,0 +1,210 @@
...
+
+
+ideal_executor = Executor(ideal_execute)
+```
+
+Define the noisy executor, in this case for simulating depolarizing noise on the `CNOT` gate in each of the training circuits. 
+The optimized value of the noise strength `epsilon` should be close to the value defined for this executor.
+
+```{code-cell} ipython3
+CNOT_ops = list(circuit.findall_operations_with_gate_type(CXPowGate))
```
natestemen, line 76: Casting the generator to list is now unneeded.

natestemen (approved):
Nice work Misty! I really like this example. It's clear, and shows a nice clear advantage while using this technique.

I've left 3 comments, all of which are optional changes.

docs/source/examples/learning-depolarizing-noise.myst
Misty-W: Thanks for asking about this! I will leave it as-is because if I don't cast to list, only one noisy operation is added in the `noisy_executor`.

docs/source/examples/learning-depolarizing-noise.myst
Misty-W: Good catch, forgot the `.` before `pec.execute_w_pec`

docs/source/examples/learning-depolarizing-noise.myst
Misty-W: Fixed it to be `.cdr.clifford_training_data.generate_training_circuits`

Misty-W:
> Nice work Misty! I really like this example. It's clear, and shows a nice clear advantage while using this technique.
> 
> I've left 3 comments, all of which are optional changes.

Thanks @natestemen !

docs/source/examples/learning-depolarizing-noise.myst
natestemen: Fine to leave it out, but are you sure? You later iterate over `CNOT_ops`, rather than just grabbing the first element. [`findall_operations_with_gate_type`](https://github.com/quantumlib/Cirq/blob/946f247dc66f1257b463766536a854bb36b4c3c6/cirq-core/cirq/circuits/circuit.py#L778-L780) returns an iterable so I think it should be fine to leave it as that.

Misty-W: I thought so too, but when I tested it, the loss function plot shifted so that the minimum was around 0.03 instead of the true value of 0.05.  I think that's because only one noisy op was added in the executor if I left CNOT_ops as an iterable.

natestemen: Ohhh okay I see. Yeah we can't iterate over a generator multiple times. Forgot about that! Thanks for helping me get there.

PR #1519, opened 2022-09-28
remove `print` calls

natestemen:
removes print statements accidentally added in https://github.com/unitaryfund/mitiq/pull/1501.

Sorry for leaving these in!


Misty-W (approved):
LGTM!

PR #1525, opened 2022-09-29
Implement conversions of IonQ native gates for Braket

yitchen-tim:
## Description

Implement conversions of IonQ native gates for Braket so that Mitiq can understand circuits with IonQ native gates. 

---

### License

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -43,3 +43,4 @@ Luciano Bello
 dependabot[bot]
 allcontributors[bot]
 Isaac Griswold-Steiner
+Tim (Yi-Ting) Chen
--- a/mitiq/interface/mitiq_braket/conversions.py
+++ b/mitiq/interface/mitiq_braket/conversions.py
@@ -22,6 +22,7 @@ from cirq.linalg.decompositions import (
     deconstruct_single_qubit_matrix_into_angles,
     kak_decomposition,
 )
+import cirq_ionq.ionq_native_gates as cirq_ionq_ops
 from braket.circuits import (
     gates as braket_gates,
     Circuit as BKCircuit,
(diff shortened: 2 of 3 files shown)


yitchen-tim:
I'm not sure why the "greeting" test fails. The error message is hard to decode for me.

andreamari:
Many thanks @yitchen-tim! 
Do you think we should wait for the Cirq update (inversions of IonQ gate) to happen before merging this PR?
I don't think merging this PR before a Cirq update is harmful, in the worst case GPI/MS gates won't work in situations where the inverse is necessary. Currently they don't work in all cases.

Second question: In the current code we manually convert other native IonQ gates e.g. V/Vi. Do you think it would be better to use `cirq_ionq.ionq_native_gates` ? 

Taking this reasoning to the extreme, probably Cirq has some conversions functions that can be directly applied to circuits and not to gates.  

We don't have to address those questions in this PR! I am just asking if you think there is space for further improvements.

yitchen-tim:
I don't think we should to convert all gates into native gates at the gate level since the conversion may not be optimal. Doing it at the circuit level is good (something like `cirq.optimize_for_target_gateset`). If there is a transpiler that do anything more than simple gate-to-gate conversions between python packages, I think the pipeline should always be first conversion/transpilation and then circuit folding. If the order is reserve, the transpiler would interfere the with circuit folding.

Also, if there is going to be a transpiler, I think it's better to make it optional rather than automated in the pipeline. There are two types of users: (1) those who directly program in native gates so they don't want their circuit compiled further, and (2) those who program in standard gate set but want to use error mitigation. For (2), users would want transpiler, but not for (1).

yitchen-tim:
For this PR, I think it's good to merge. For those who does `pip install cirq`, they still could not use it because there is no gate inverse, but the gate inverse is already merged in Cirq master branch, so users can use it with github version of Cirq.

Also, do you know why the "greeting" test fails?

purva-thakre:
>  do you have any tip to fix the "greeting" test?

@yitchen-tim I don't think `Greetings` failing is because of something you did. 

This seems to be an issue with Github Actions : https://github.com/actions/first-interaction/issues/101

andreamari (approved):
LGTM! Thanks also for your ideas about transpilation/conversions. I agree with all your comments.

nathanshammah:
Thanks @yitchen-tim!

PR #1530, opened 2022-09-30
replace deprecated `execution_excludepatterns`

natestemen:
## Description

replace deprecated `execution_excludepatterns` with `nb_execution_excludepatterns`. This was not caught when merging https://github.com/unitaryfund/mitiq/pull/1504 as the BQSKit PR (https://github.com/unitaryfund/mitiq/pull/1489) was merged while python 3.10 was being reviewed.

### DIFF
--- a/docs/source/conf.py
+++ b/docs/source/conf.py
@@ -129,11 +129,11 @@ nb_execution_timeout = 600
 # re-run for a subsequent docs build.
 nb_execution_mode = "cache"
 
-execution_excludepatterns = ["bqskit.ipynb"]
+nb_execution_excludepatterns = ["bqskit.ipynb"]
 # If SKIP_PYQUIL is True, do not re-run PyQuil notebooks.
 if os.environ.get("SKIP_PYQUIL"):
     print("Skipping PyQuil notebooks execution since SKIP_PYQUIL is True")
-    execution_excludepatterns.append("*pyquil*.ipynb")
+    nb_execution_excludepatterns.append("*pyquil*.ipynb")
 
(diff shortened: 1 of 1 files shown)


Misty-W:
Not sure how the BQSKit PR is related? Shouldn't this have failed when testing the 3.10 PR?

natestemen:
So I think this is the timeline:
- 3.10 PR was created, commits added, and hence built by CI using master at that point in time
- BQSKit PR is created, commits added (and makes use of `execution_excludepatterns` in `docs/source/conf.py` regardless of `SKIP_PYQUIL` environment variable)
- BQSKit PR merged
- 3.10 PR merged

So the 3.10 PR was not built on CI with the latest version of master by the time it was merged. We could have avoided this if it had been, but usually we don't need to be so careful! Does that make sense, or am I missing something here?

Misty-W:
I'm guessing GitHub didn't detect a merge conflict because there were no new commits on the 3.10 PR?

natestemen:
There were no merge conflicts (even though both PRs modified `docs/source/conf.py`) as the changes did not overlap at all. Different lines were modified in each PR.

Misty-W (approved):
Nice fix! 

Any other keyword changes we should be aware of?

natestemen:
Thanks for the quick review!

There were three others that I ran into (https://github.com/unitaryfund/mitiq/issues/1493), and were fixed as part of https://github.com/unitaryfund/mitiq/pull/1504, but since `SKIP_PYQUIL` is unset locally, it never became aware of the one fixed here.

PR #1532, opened 2022-10-03
[temporary] ignore errors from first-interaction

natestemen:
## Description

The greeting bot is causing all PR builds to "fail", even though most of the time it's just the greeting that's failing. I thought a fix might be out sooner (especially since a [PR](https://github.com/actions/first-interaction/pull/102) is open), but the project doesn't seem to see a whole lot of activity. A workaround was suggested in https://github.com/actions/first-interaction/issues/101#issuecomment-1262487501, so that's what's copied here.

### DIFF
--- a/.github/workflows/greeting.yml
+++ b/.github/workflows/greeting.yml
@@ -7,6 +7,7 @@ jobs:
     runs-on: ubuntu-latest
     steps:
     - uses: actions/first-interaction@v1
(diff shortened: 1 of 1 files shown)


andreamari (approved):
LGTM!

natestemen:
Strangely enough, it still failed...?

PR #1536, opened 2022-10-03
Set master in dev mode

andreamari:
I don't remember if we decided to stop doing this step, or to keep doing it.
I am following the release guide in the docs.


natestemen:
It seems a little unnecessary to me since we don't update the changelog as we go. I guess the main advantage is that if someone downloaded `mitiq` via
```
pip install git+https://github.com/unitaryfund/mitiq.git@master
```
they would know with more certainty what version they are on.

I don't see it as being a big enough gain to continue doing it, but I also recognize it's not a lot of work so fine with me either way.

natestemen:
What you propose (change docs and just update `VERSION.txt` starting with the next release) sounds good to me!

Issue #1179, opened 2022-03-23
Avoid ignoring silently bad inputs in `fold_gates_at_random`


nelimee:
Pre-Request Checklist
---------------------

- [x] I checked to make sure that this feature has not already been requested.

Issue Description
-----------------

The following code
```py
from qiskit import QuantumCircuit
from mitiq.zne.scaling.folding import fold_gates_at_random

circuit = QuantumCircuit(1)
circuit.rx(0, 0)

fold_gates_at_random(circuit, scale_factor=2, fidelities={"invalid_gate_name": 0.06, "rx": 0.1})
```
does not:
1. Warn the user that `invalid_gate_name` and `rx` are not in the valid gates that are allowed in `fidelities` (see the [documentation](https://mitiq.readthedocs.io/en/latest/apidoc.html#mitiq.zne.scaling.folding.fold_gates_at_random) to get this list).
2. Warn the user that, because `rx` is not in the valid gates, the weight that will be used internally by `mitiq` is **not** the expected `1 - 0.1` (i.e. the infidelity, as written in the documentation) but rather the default weight that is here `1 - 0.99**n = 0.01`.

Proposed Solution
-----------------

In order of personal preference:
1. Ideally, the restriction on the allowed gate names that are allowed in `fidelities` should be removed. I am not very familiar with `mitiq` internals but I suspect that this restriction is related to an internal implementation detail of `mitiq` and it might not be easy to remove.
2. Raise an exception if `fidelities` contains gates that are not allowed.
3. Raise an exception if `fidelities` contains gates that are not allowed and at least one of these gate is used in the circuit to fold, leading to the situation "the weight used internally by `mitiq` does not correspond to the weight the caller wrote" which is super unintuitive in my opinion.
4. Write a note in the documentation of the function explaining this behaviour.

Additional References
---------------------

*None*

purva-thakre:
I think a warning in the case of a gate that's not in the input circuit but has fidelities provided could be useful here. 

If `invalid_gate_name` was used in the initial circuit, then either `qiskit` would have raised an error that the gate is not recognized or `mitiq` would have raised an error that the input circuit cannot be folded because[ `has_unitary(circuit)`](https://github.com/unitaryfund/mitiq/blob/d58c8fe015ab9d50db8c2e940b2582c9c8612ec2/mitiq/zne/scaling/folding.py#L52) will fail.

andreamari:
[...]

PR #1542, opened 2022-10-06
Warn when unsupported gate fidelity is passed

natestemen:
## Description

I've gone with the `#2` proposed solution in https://github.com/unitaryfund/mitiq/issues/1179. I.e. warn the user when they pass a fidelity along with a gate that we don't support yet. I would be happy to add more gates to the supported group, but I figured I would get feedback before going down that route. If we do decide to do that, I'd offer to pair on it as I have a few questions about how to add support for gates with parameters.

fixes https://github.com/unitaryfund/mitiq/issues/1179

cc @purva-thakre as you had previously been assigned this issue. Sorry if you were working on this!

---

**EDIT:** I've _also_ added more supported gates in this PR, so not only does it warn now, but it should warn less!

### DIFF
--- a/mitiq/zne/scaling/folding.py
+++ b/mitiq/zne/scaling/folding.py
@@ -40,13 +40,21 @@ _cirq_gates_to_string_keys = {
     ops.Z: "Z",
     ops.T: "T",
     ops.I: "I",
+    ops.S: "S",
+    ops.T: "T",
+    ops.rx: "RX",
+    ops.ry: "RY",
+    ops.rz: "RZ",
     ops.CNOT: "CNOT",
     ops.CZ: "CZ",
+    ops.SWAP: "SWAP",
+    ops.ISWAP: "ISWAP",
+    ops.CSWAP: "CSWAP",
     ops.TOFFOLI: "TOFFOLI",
 }
-_string_keys_to_cirq_gates = dict(
-    zip(_cirq_gates_to_string_keys.values(), _cirq_gates_to_string_keys.keys())
-)
+_string_keys_to_cirq_gates = {
+    opstring: op for op, opstring in _cirq_gates_to_string_keys.items()
+}
 
 _valid_gate_names = list(
     map(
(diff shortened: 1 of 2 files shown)


purva-thakre:
> Sorry if you were working on this!

No worries. I have been distracted by the `id_scaling` issue as well as a RFC for Pauli Twirling/Randomized compiling.

mitiq/zne/scaling/folding.py
```diff
@@ -334,12 +342,20 @@ def _create_weight_mask(
...
-        weights = {k: round(1.0 - f, 12) for k, f in fidelities.items()}
+
+    invalid_fidelities = filter(
+        lambda opname: opname.lower() not in _valid_gate_names, fidelities
+    )
+    for gate_name in invalid_fidelities:
+        warnings.warn(
+            f"You passed a fidelity for the gate '{gate_name}', but we don't "
+            "currently support this gate."
+        )
```
andreamari, line 355: We need a new test for this case.

natestemen: Good thinking, added in [0ae3a77](https://github.com/unitaryfund/mitiq/pull/1542/commits/0ae3a77e54d3a267356d7658b398963dbeae2fd0).

natestemen:
@andreamari I've added support for some more gates. I've taken what I think are the most important ones from https://quantumai.google/cirq/build/gates. Can you take another look?

mitiq/zne/scaling/folding.py
```diff
@@ -39,13 +40,28 @@ class UnfoldableCircuitError(Exception):
     ops.Z: "Z",
     ops.T: "T",
     ops.I: "I",
+    ops.S: "S",
+    ops.T: "T",
+    ops.rx: "RX",
+    ops.ry: "RY",
+    ops.rz: "RZ",
```
andreamari, line 47:
```suggestion
    ops.rx: "rx",
    ops.ry: "ry",
    ops.rz: "rz",
```
For consistency with docstring? I know this is not so important since you convert to lowercase anyway.

andreamari (approved):
Small comments. LGTM!

Issue #1526, opened 2022-09-29
Search engines indexing old versions of docs


natestemen:
For example, when I search for PEC, the first (!) result is our docs, but from version 0.11.1.
<img width="600" alt="Screen Shot 2022-09-29 at 1 59 11 PM" src="https://user-images.githubusercontent.com/12703123/193141091-630d634d-b8d0-4f30-80cf-1533cacd5455.png">

It would be great if search engines only indexed stable releases. Maybe a case can be made for latest as well, but definitely not older versions. Thankfully this issue seems to be solved in https://github.com/readthedocs/readthedocs.org/issues/2430, and the solution is to use a [`robots.txt`](https://developer.mozilla.org/en-US/docs/Glossary/Robots.txt) file along with sphinx's [`html_extra_path`](https://www.sphinx-doc.org/en/master/usage/configuration.html#confval-html_extra_path) config var which gets set in [`docs/source/conf.py`](https://github.com/unitaryfund/mitiq/blob/master/docs/source/conf.py).

An example `robots.txt` we may want to copy can be found here: https://github.com/astropy/astropy/blob/main/docs/robots.txt

andreamari:
Seems like a good idea! This a PR, in astropy, that did what @natestemen suggested: https://github.com/astropy/astropy/commit/53991504e622ceb516d79d37dd562da95f6c845f

purva-thakre:
The version in RTD that shows up is still an older one. I searched for mitiq and if I go to the docs link, I don't directly go to the latest or stable version.
![image](https://user-images.githubusercontent.com/66048318/204333159-a64bd579-1236-46bc-8aeb-5fbe801623f0.png)

natestemen:
Yeah, it can take quite a long time for search engines to refresh their cache. There are ways to ask them to do a refresh, but you first have to prove you own the webpage which can be kind of a pain. IMO it's not worth our time as it stands, unless someone is interested.

PR #1543, opened 2022-10-06
add robots.txt 🤖

natestemen:
## Description

This change should ensure that search engines only index our stable release docs, and nothing else. This is done by using a `robots.txt` file which all major search engines support. Further information about this file can be found [here](https://www.cloudflare.com/learning/bots/what-is-robots.txt/). The `robots.txt` can be seen on the RTD build here: https://mitiq--1543.org.readthedocs.build/en/1543/robots.txt

This change is based on https://github.com/astropy/astropy/commit/53991504e622ceb516d79d37dd562da95f6c845f

This change should solve https://github.com/unitaryfund/mitiq/issues/1526, but we will have to check search engines a few weeks after this is merged to ensure as their search results do not update automatically.

### DIFF
--- a/docs/source/conf.py
+++ b/docs/source/conf.py
@@ -307,6 +307,10 @@ html_logo = "img/mitiq-logo.png"
 
 html_favicon = "img/mitiq.ico"
 
+# Add extra paths that contain custom files here, relative to this directory.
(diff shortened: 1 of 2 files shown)


andreamari (approved):
LGTM thanks!

Issue #1535, opened 2022-10-03
When a test in fails in CI, what about if we still run all the other tests?


andreamari:
## Issue Description

With current settings CI stops as soon as a test fails. In this way we don't know if the other test would have passed or not. 
I understand the motivation: it saves computational power.
On the other hand, it would be helpful to know what are all the failing tests for each commit instead of catching and fixing failing tests one-by-one.

natestemen:
I think the desired effect can be achieved by setting `fail-fast: false` in our test matrices. The keyword is documented [here](https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#jobsjob_idstrategyfail-fast).

PR #1554, opened 2022-10-14
run all tests if one fails

natestemen:
## Description

fixes https://github.com/unitaryfund/mitiq/issues/1535

Documentation for `fail-fast` found here: https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#jobsjob_idstrategyfail-fast.


Misty-W (approved):
Nice fix!

Issue #1520, opened 2022-09-28
Port BQSKit tutorial to bqskit1.0


natestemen:
BQSKit 1.0.0 is a big API change that drastically simplifies our tutorial.

https://github.com/unitaryfund/mitiq/issues/1156 must be completed prior to this as 1.0.0 requires `numpy>=1.22.0`.

Initial PR: https://github.com/unitaryfund/mitiq/pull/1489
Existing file: https://github.com/unitaryfund/mitiq/blob/9f227b13a7c4746c70cb71e91dbad40358cc49be/docs/source/examples/bqskit.ipynb

PR #1557, opened 2022-10-17
Port BQSKit example to 1.0 release

natestemen:
## Description

fixes https://github.com/unitaryfund/mitiq/issues/1520

As part of this PR I have added the `BQSKIT_DOC_CHECK_OVERRIDE` environment variable to RTD. This is a temporary fix issued by the BQSKit devs which might be fixed in a future release. We should watch out for that, and remove accordingly.


nathanshammah:
@natestemen do you want to move this from draft status to open?

nathanshammah:
Happy to review!

natestemen:
I need help fixing the problems on RTD before it's ready for review. Let me know if you can help me because I'm stuck.

nathanshammah:
Just to add also here that build works correctly locally on my machine.

andreamari:
FYI: Another way of setting environment variables for readthedocs is here: https://readthedocs.org/dashboard/mitiq/environmentvariables/

But what you did seems to work too!

PR #1295, opened 2022-05-17
Wait more time before issues are marked as stale and exempt important labels

andreamari:
Description
-----------

This PR doubles the time before issues are marked as stale and exempt important labels: `roadmap`, `priority/p0` and `bug`.
Fixes #1215.

License
-------

- [x] I license this contribution under the terms of the GNU GPL, version 3 and grant Unitary Fund the right to provide additional permissions as described in section 7 of the GNU GPL, version 3.

PR #1558, opened 2022-10-18
remove stale bot

natestemen:
## Description

It's been 5 months since we increased the time-til-stale in https://github.com/unitaryfund/mitiq/pull/1295. I still don't really like this bot. I frequently come across issues that have something like the following history.
Screen Shot 2022-10-17 at 8 36 32 PM

If we had a many new issues every day, then I think this automated pruning makes sense to keep the backlog to a mangable size, but that's not our case. If anything, we can meet once per quarter to go through open issues and close ones we don't want worked on.

We also have [51 closed issues](https://github.com/unitaryfund/mitiq/issues?q=is%3Aissue+is%3Aclosed+label%3Astale) with label `stale`. It would be much nicer if instead of a bot closing these, they had a short explanation as to why they are not planned.

Thoughts? @Misty-W @andreamari @nathanshammah @purva-thakre @Aaron-Robertson


Misty-W:
I support this!

purva-thakre:
> If anything, we can meet once per quarter to go through open issues and close ones we don't want worked on.

Setup the bot to notify us of stale issues for each quarter but let it close issues after a week. 

We can decide on which ones to close within a week, remove `stale` labels for those, let the bot auto-close the ones we do not want to work on.

andreamari (approved):
I think you are right, @natestemen . 
I approve this PR but please wait a bit before merging to see if there are strong objections.

PR #1559, opened 2022-10-18
Docs infrastructure improvements

natestemen:
## Description

Some improvements mostly related to documentation. 91ecfc16a6181573e8355e36200225cd4a1e7189 should remedy the bug that we keep seeing with regards to dependabot opening PRs even though out config says not too. This is an open issue: https://github.com/dependabot/dependabot-core/issues/4989. That said, having these dependencies pinned exactly seems beneficial for reproducibility.

If there are questions about any other actions taken here, happy to explain further!


Makefile
```diff
@@ -13,7 +13,7 @@ check-format:
 
 .PHONY: check-style
 check-style:
-	flake8
+	flake8 mitiq
```
andreamari, line 16: Is this restriction necessary? E.g. what about `docs` (not sure if flake works with myst files)  and `benchmark` folders?

andreamari (commented):
Nice cleaning up

.flake8
```diff
@@ -1,4 +1,4 @@
 [flake8]
 max-line-length = 79
-ignore = E203, E231, W503, F723, F821
```
andreamari, line 3: Could you explain why you removed E231, F723, F821 from `ignore`?

Makefile
natestemen: Thanks for asking about this. It got me to dig a little deeper to understand the problem I was having previously. I was never able to run `flake8` on my machine. Rather the command would hang indefinitely. After some debugging I've found that I had both `.venv/` and `env/` directories it was going through checking style as well (but I no longer use virtualenv). So it was checking something like 20,000 files instead of 86. I've now removed those directories and it works fine. I'll revert this change!

.flake8
natestemen: In short, because we don't have any of those errors! Here's a description of those errors
- [`E231`](https://www.flake8rules.com/rules/E231.html) - Missing whitespace after ',', ';', or ':'
- [`F723`](https://flake8.pycqa.org/en/4.0.1/user/error-codes.html) - syntax error in type comment
- [`F821`](https://www.flake8rules.com/rules/F821.html) - Undefined name name

PR #1243, opened 2022-04-22
Move the gh pages deploy to a manually triggered action

crazy4pi314:
Description
-----------

Moves the gh pages deploy to a manually triggered action, to avoid issues with externally forked PRs.

Checklist
---------

Check off the following once complete (or if not applicable) after opening the PR. The PR will be reviewed once this checklist is complete and all tests are passing.

- [x] I added unit tests for new code.
- [x] I used [type hints](https://www.python.org/dev/peps/pep-0484/) in function signatures.
- [x] I used [Google-style](https://sphinxcontrib-napoleon.readthedocs.io/en/latest/example_google.html) docstrings for functions.
- [x] I [updated the documentation](../blob/master/docs/CONTRIBUTING_DOCS.md) where relevant.
- [x] Added myself / the copyright holder to the AUTHORS file

If some items remain, you can mark this a [draft pull request](https://github.blog/2019-02-14-introducing-draft-pull-requests/).

License
-------

- [x] I license this contribution under the terms of the GNU GPL, version 3 and grant Unitary Fund the right to provide additional permissions as described in section 7 of the GNU GPL, version 3.

PR #1566, opened 2022-10-24
remove gh-pages deploy action

natestemen:
## Description

We don't use the github pages website (https://unitaryfund.github.io/mitiq) and it has potential to be doubly indexed by search engines.

If this is merged, we should also delete the [`gh-pages`](https://github.com/unitaryfund/mitiq/tree/gh-pages) branch and deactivate the `gh-pages` "environment".

---

I was reminded about this by https://github.com/unitaryfund/mitiq/pull/1564.


Misty-W (approved):
Fine to delete this, the gh-pages workflow was created as a test.

natestemen:
As a test for what?

I only found the history of this file through https://github.com/unitaryfund/mitiq/pull/1135 and https://github.com/unitaryfund/mitiq/pull/1243.

Misty-W:
> As a test for what?
> 
> I only found the history of this file through #1135 and #1243.

I should've said it was an "experiment." Sarah was looking into using gh-pages as an alternative to read the docs but we didn't end up migrating.

natestemen:
I guess it does allow for some added flexibility, but I don't think we have a use case for it yet. Easy enough to set back up if we decide to use it!

Issue #1509, opened 2022-09-26
Clean up flaky tests in `test_inference.py`


natestemen:
We seem to have a few flaky tests which have popped up recently giving the dreaded red X. Since our test often take 20-30min to run, restarting them due to a flaky test is undesirable.

- In https://github.com/unitaryfund/mitiq/pull/1492 we saw [`mitiq/zne/tests/test_inference.py::test_plot_data_exp_factory[ExpFactory]`](https://github.com/unitaryfund/mitiq/blob/a48f35f9ba7fab2f491c2baac2a905b6f3706b40/mitiq/zne/tests/test_inference.py#L887-L902) fail.
- In https://github.com/unitaryfund/mitiq/actions/runs/3130210903/jobs/5080201273 we see [`mitiq/zne/tests/test_inference.py::test_plot_fit[LinearFactory]`](https://github.com/unitaryfund/mitiq/blob/55653c73bfa20a796eff8abb6d0eec2ad881174e/mitiq/zne/tests/test_inference.py#L913) fail.

natestemen:
[master](https://github.com/unitaryfund/mitiq/actions/runs/3174423778/jobs/5171198696) just failed on `mitiq/zne/tests/test_inference.py::test_plot_data[LinearFactory]`

natestemen:
[another](https://github.com/unitaryfund/mitiq/actions/runs/3276503155/jobs/5392631981) fail on `mitiq/zne/tests/test_inference.py::test_plot_data[LinearFactory]`

natestemen:
[another](https://github.com/unitaryfund/mitiq/actions/runs/3321940639/jobs/5490277619) fail on `mitiq/zne/tests/test_inference.py::test_plot_data[LinearFactory]`

natestemen:
[another](https://github.com/unitaryfund/mitiq/actions/runs/3358815258/jobs/5566225208) fail with both
- `mitiq/zne/tests/test_inference.py::test_plot_data[LinearFactory]`
- `mitiq/zne/tests/test_inference.py::test_plot_data[RichardsonFactory]`

natestemen:
[another](https://github.com/unitaryfund/mitiq/actions/runs/3363585326/jobs/5576920514) fail on `mitiq/zne/tests/test_inference.py::test_plot_fit[LinearFactory]`

natestemen:
[...]

PR #1580, opened 2022-11-03
use python 3.10 for asv

natestemen:
## Description

[ASV](https://github.com/airspeed-velocity/asv) has been causing our CI to fail due to a mismatch in configuration. The [github action that runs `asv`](https://github.com/unitaryfund/mitiq/blob/3fe70da498213272d64ea753f3809e61f522b090/.github/workflows/asv.yaml) sets up python 3.10, but our [`asv.conf.json`](https://github.com/unitaryfund/mitiq/blob/3fe70da498213272d64ea753f3809e61f522b090/asv.conf.json) requests python 3.8.

This PR ensures both run on the same version. If these were timing benchmarks, swapping the python versions would not be ideal since there could be speed improvements in the language that would make it seem as though our code improved without doing anything. That said, our benchmarks are EM factor-related and hence should not be affected by this.

I think ASV is also smart enough to show new lines on https://unitaryfund.github.io/mitiq-bench/#benchmarks.track_zne when the python version changes anyway.

### Diagnosis

Support for python 3.10 was added in mitiq in https://github.com/unitaryfund/mitiq/pull/1504 (Sep 30th) which would have originally caused this mismatch. However, up until Nov 1, all of our actions running on `ubuntu-latest` defaulted to running on Ubuntu 20.04 which comes pre-installed with a list of software detailed [here](https://github.com/actions/runner-images/blob/main/images/linux/Ubuntu2004-Readme.md), but most importantly has python 3.8 installed as a default. Hence even though we specified use 3.10 in setting up, it was still able to find python 3.8 to run the `asv` benchmarks. On Nov 1 that changed when `ubuntu-latest` was upgraded to run on Ubuntu 22.04 ([runner images release](https://github.com/actions/runner-images/releases/tag/ubuntu22%2F20221027.1)). This finally reared it's head on our CI builds starting Nov 1st since the Ubuntu 22.04 image defaults to python 3.10 as can be seen in the announcement issue: https://github.com/actions/runner-images/issues/6399.

### DIFF
--- a/asv.conf.json
+++ b/asv.conf.json
@@ -4,7 +4,7 @@
   "project_url": "https://github.com/unitaryfund/mitiq",
   "repo": "https://github.com/unitaryfund/mitiq.git",
   "branches": ["master"],
-  "pythons": ["3.8"],
+  "pythons": ["3.10"],
   "environment_type": "virtualenv",
   "show_commit_url": "https://github.com/unitaryfund/mitiq/commit/"
 }


natestemen:
@Misty-W I updated the description since I figured out where the underlying discrepancy came from.

The single failing test is flaky, and is being tracked in https://github.com/unitaryfund/mitiq/issues/1509

Misty-W (approved):
Makes sense, many thanks for figuring this out quickly Nate!

Issue #1466, opened 2022-08-26
Measurements followed by additional gates in a DDD circuit should return a sensible error


Aaron-Robertson:
## Issue Description
Old issue with converted measurements is back, potentially isolated to DDD. Measurement names change resulting in an error:
`qiskit.circuit.exceptions.CircuitError: "Bit 'Clbit(ClassicalRegister(4, 'meas'), 0)' is not in the circuit."`

## How to Reproduce

1. Use circuit:
```
# Qiskit test with measurement
circuit_qiskit_two = qiskit.QuantumCircuit(qreg, creg)
for q in qreg:
    circuit_qiskit_two.x(q)
    circuit_qiskit_two.x(3)
circuit_qiskit_two.measure(0, 3)
circuit_qiskit_two.cx(0, 3)
```
2. Run circuit through DDD:
```
circuit_with_ddd = ddd.insert_ddd_sequences(circuit_qiskit_two, rule=xx_rule)
```

## Expected behavior
Circuit should yield an error, saying that measurements cannot be made terminal (and thus the circuit is incompatible with DDD), per discussion on #1446:

> Circuit has intermediate measurements that are followed by gates. In this case they cannot be pushed to the end. I think in this case we should raise an error since we are not currently able to support this case.

andreamari:
Hi @Aaron-Robertson, do you have any update on this? I am asking since next Friday we are going to make a new release.

natestemen:
Hey Aaron, wondering if you'd like this to go in the next milestone? I know you mentioned you were close, but if you need any help, let us know!

Aaron-Robertson:
@natestemen I've confirmed that I'm available to wrap this up, so let's add it to the milestone!

Misty-W:
hi @Aaron-Robertson, wanted to check if you're planning to work on this issue in the current milestone (0.21, closes Nov 30). If not, no worries, someone on the UF team can pick this up.

Happy to help clear any roadblocks!

PR #1586, opened 2022-11-10
Raise error when inserting DDD sequence with midcircuit measurement

natestemen:
## Description

When inserting DDD sequences into a circuit, all measurements must be terminal. Previously we have used [`synchronize_terminal_measurements`](https://quantumai.google/reference/python/cirq/synchronize_terminal_measurements) to push measurements to the end, but this does not raise an error when it is not possible. This PR raises an error when the circuit contains unavoidable midcircuit measurements so the user more quickly knows this is not supported.

fixes https://github.com/unitaryfund/mitiq/issues/1466

### DIFF
--- a/mitiq/ddd/insertion.py
+++ b/mitiq/ddd/insertion.py
@@ -141,7 +141,7 @@ def _insert_ddd_sequences(
     # Copy to avoid mutating the input circuit
     circuit_with_ddd = circuit.copy()
     qubits = sorted(circuit.all_qubits())
-    for moment_idx, moment in enumerate(circuit):
+    for moment_idx in range(len(circuit)):
         slack_column = slack_matrix[:, moment_idx]
         for row_index, slack_length in enumerate(slack_column):
             if slack_length > 1:
--- a/mitiq/ddd/insertion.py
+++ b/mitiq/ddd/insertion.py
@@ -129,6 +129,12 @@ def _insert_ddd_sequences(
         The circuit with DDD sequences added.
     """
     circuit = synchronize_terminal_measurements(circuit)
+    if not circuit.are_all_measurements_terminal():
+        raise ValueError(
+            "This circuit contains midcircuit measurements which "
+            "are not currently supported by DDD."
+        )
+
     slack_matrix = get_slack_matrix_from_circuit_mask(
         _get_circuit_mask(circuit)
     )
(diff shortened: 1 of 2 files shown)


mitiq/ddd/insertion.py
```diff
@@ -129,13 +129,19 @@ def _insert_ddd_sequences(
...
+        )
+
     slack_matrix = get_slack_matrix_from_circuit_mask(
         _get_circuit_mask(circuit)
     )
     # Copy to avoid mutating the input circuit
     circuit_with_ddd = circuit.copy()
     qubits = sorted(circuit.all_qubits())
-    for moment_idx, moment in enumerate(circuit):
+    for moment_idx in range(len(circuit)):
```
Misty-W, line 144: Nice!

mitiq/ddd/tests/test_insertion.py
```diff
@@ -253,3 +253,17 @@ def test_get_slack_matrix_from_circuit__bad_input_errors():
 def test_insert_sequences(circuit, result, rule):
     circuit_with_sequences = insert_ddd_sequences(circuit, rule)
     assert circuit_with_sequences == result
+
+
+def test_midcircuit_measurement_raises_error():
+    qreg = qiskit.QuantumRegister(2)
```
Misty-W, line 259: Do we need to test this on another frontend besides Qiskit?

natestemen: Hmmm good question! I'm not actually too sure. I think since the code we are ultimately trying to test is written for cirq, and we have a test that we know covers it (even though it's in qiskit), it _should_ be fine, but what do you think?

Misty-W: Agree Cirq doesn't need an explicit test. For the other supported frontends, I think it should be sufficient to check the behavior with other frontends "offline" and record it in a comment on this PR.

natestemen: Added a pyquil test in [`55fae24`](https://github.com/unitaryfund/mitiq/pull/1586/commits/55fae247c747916b01bf3758de4cb928e32dffc1). Looks like neither pennylane, nor braket support mid circuit measurement currently.

Misty-W: That works!

Misty-W (approved):

PR #1587, opened 2022-11-10
color output from pytest on CI

natestemen:
## Description

I recently discovered that the terminal output in github actions can be colorized ([announced](https://github.blog/2020-09-23-a-better-logs-experience-with-github-actions/#opening-the-door-to-a-more-colorful-experience) in 2020, but better late than never)! This should make reading through pytest logs much quicker as the red text is spotted much more easily.

Color output is the default on my terminal, so this should mainly affect CI, but it might force color output for others as well.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -3,6 +3,7 @@ line-length = 79
 target-version = ['py310']
 
 [tool.pytest.ini_options]
+addopts = "--color=yes"
 filterwarnings = [
     # TODO: these are probably too restrictive
     'ignore::UserWarning',


Misty-W:
I like this! Not sure how to test whether it forces color output as it's also my default 😆 . Maybe it's default for most people?

Misty-W (approved):

natestemen:
> Maybe it's default for most people?

I have a feeling it is. I mainly added this so CI also has color output, which can be checked by looking at the pytest logs, and indeed there is color!

PR #1592, opened 2022-11-15
use `\prod` instead of `\Pi`

natestemen:
## Description

Was going through the docs and noticed this. LaTeX has the built-in `\prod` command for products and hence we should use that.


Misty-W (approved):
Nice find, this makes the doc look cleaner for sure. LGTM!

PR #1594, opened 2022-11-16
Fix ZNE guide to reflect that  "observable" is a keyword argument

nickdgardner:
## Description

Issue: The text and accompanying code snippet [here](https://mitiq.readthedocs.io/en/stable/guide/zne-3-options.html) say that `observable` is a **positional argument** "common to all error mitigation techniques", but currently all `execute_with_<QEM method>` functions in the API-doc have observable as a **keyword argument** with default value `None`.

I modified that section of the User Guide to reflect this.

---

### License


nickdgardner:
Woops sorry a bunch of earlier commits are showing up in this PR! The only relevant one is "Fix Users Guide documentation for ZNE options". Is there something I should do on my fork to fix this in the PR?

andreamari:
I think you are looking for a forced [_rebasing_](https://github.blog/changelog/2022-02-03-more-ways-to-keep-your-pull-request-branch-up-to-date/) of your branch on the Mitiq master.
But it's not a big issue since we are going to `squash-marge` all approved PRs anyway. 

On the merit of the change proposed in this PR, I am not sure I agree.
What the guide wants to say is that one can **call** `execute_with_xxx()` with three positional arguments `(circuit, executor, observable)`, while all other options must be necessarily passed as keywords. They are [_keyword-only_](https://peps.python.org/pep-3102/)  arguments.

It is true, `observable` has a default value, but it's not a _keyword-only_ argument.

Probably this is not clear now in the docs, but I am afraid this PR fails to give the correct message.

nickdgardner:
Ah ok! That makes sense sorry I missed this. Agreed, my proposed changes convey the wrong message since the relevant distinction is between arguments callable by position vs. only by keyword. If a change seems necessary, maybe the following would work?

> Beyond the arguments callable by position (`circuit`, `executor` and `observable`) that are common to all
> error mitigation techniques, one can use additional keyword-only arguments for optional settings as shown in next code snippet:
> 
> ```
> from mitiq import zne
> 
> zne_value = zne.execute_with_zne(
>     circuit,
>     executor,
>     observable, # callable by position or keyword
>     scale_noise = <"noise scaling method imported from zne.scaling">,
>     factory = <"extrapolation Factory imported from zne.inference">,
>     num_to_average = <"number of repeated evaluations for each noise-scaled circuit">,
> )
> ```

Or the guide could be left as is.

For context, the reason I found this confusing was that I was reading through the guide pages in order, and earlier on the [How do I use ZNE](https://mitiq.readthedocs.io/en/stable/guide/zne-1-intro.html) page there is a call `execute_with_zne(circuit, execute)`, which then seemed to me not to be allowed when I read here that `observable` is a positional argument (according to my interpretation as "must be called by position "). But others won't be confused if they understand it as "can be called by position".

(And thanks, I'll do a forced rebasing of my branch on Mitiq master.)

natestemen:
I think having optional positional arguments is pretty common in python. All of `range(5)`, `range(1, 5)`, and `range(1, 5, 2)` are completely valid positional arguments. It's can be confusing as you can call functions with positional argument _as keyword arguments_. There's a good explainer video on this subject [here](https://www.youtube.com/watch?v=R8-oAqCgHag).

Happy to clarify the docs in any way you might think would make this more clear, but we don't want to write stuff about python that's already documented elsewhere.

nickdgardner:
Thanks for the clarification (and great video). I'd say leave the docs as is then, since my misinterpretation doesn't seem common. 

Sorry @andreamari @natestemen for the hasty pull request!

natestemen:
No problem Nick! Hope it makes more sense now!

Issue #1562, opened 2022-10-20
REM: Refactor generate_inverse_confusion_matrix


amirebrahimi:
Per [PR feedback](https://github.com/unitaryfund/mitiq/pull/1449#discussion_r972863821) it was suggested to replace the use of the `NoisyReadoutSampler` and `cirq.measure_confusion_matrix` with a fixed analytical solution. Additionally, Revert changes from #1605.

As mentioned in the PR it is okay if this is a follow-up PR to #1449.

Here's a [reference](https://github.com/quantumlib/Cirq/pull/4854/files#diff-6037b8997ce87fcfcfe512167b87b8fb3b420061e5fd835f2fb2f3ce64554dd2R22)

Misty-W:
Hi @amirebrahimi, should we include this issue in milestone 0.21? For context, the target completion date of milestone 0.21 is Nov 30.

amirebrahimi:
I think so as it should be a relatively small PR. #1449 needs to land first.

amirebrahimi:
Would you please assign this to me as well, @Misty-W? Thanks!

PR #1605, opened 2022-11-22
add pytest dependency to asv build

natestemen:
## Description

CI is [failing](https://github.com/unitaryfund/mitiq/actions/runs/3526487137/jobs/5916670876) after the merging of https://github.com/unitaryfund/mitiq/pull/1449 due to the addition of `NoisySingleQubitReadoutSampler` in `mitiq/interface/mitiq_cirq/cirq_utils.py`. https://github.com/unitaryfund/mitiq/blob/dedc6e2f447339261c691fee33f06e4b1795ba1c/mitiq/interface/mitiq_cirq/cirq_utils.py#L24-L26 The file in `cirq` that the `NoisySingleQubitReadoutSampler` comes from [imports `pytest`](https://github.com/unitaryfund/mitiq/blob/dedc6e2f447339261c691fee33f06e4b1795ba1c/mitiq/interface/mitiq_cirq/cirq_utils.py#L24-L26) (as it's a test file) which is causing the asv build to fail as pytest is not installed on the github runner. 

I think we should merge this ASAP, but @amirebrahimi are there any ways around using this class? It seems a bit strange to import an object from a test file, but if it's needed, no problem. Just wanted to ask.

### DIFF
--- a/.github/workflows/asv.yaml
+++ b/.github/workflows/asv.yaml
@@ -21,7 +21,7 @@ jobs:
 
       - name: Run benchmarks
         run: |
-          pip install asv virtualenv
+          pip install asv virtualenv pytest
           asv machine --machine ghrunner
           asv run --attribute rounds=4
           cp asv.conf.json results
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -96,7 +96,7 @@ jobs:
 
       - name: Run ASV benchmarks
         run: |
-          pip install asv virtualenv
+          pip install asv virtualenv pytest
           export BENCHMARK_CI=1
           asv machine --yes
           asv run --quick --strict


amirebrahimi:
Thanks for correcting this and I agree that we should merge this change, @natestemen. There's a follow-on issue #1562 to remove this dependency upon which we can also revert the pytest dependency in asv.

Misty-W:
Seems adding to quick build still doesn't fix the issue. We might be able to work around this by moving the import out of `mitiq_cirq`

Misty-W (commented):
On second look, I think ASV is failing because of what is on master, not on this branch. After all, ASV passed on #1449. We might have to try merging once we're reasonably confident of the solution without actually seeing ASV pass on this PR.

amirebrahimi:
Since the change was small I went ahead and addressed #1562 in PR #1607.

Misty-W:
ASV error fixed by #1607.

PR #1607, opened 2022-11-23
Refactor generate_inverse_confusion_matrix

### LINKED ISSUE #1562: REM: Refactor generate_inverse_confusion_matrix

amirebrahimi:
Resolves #1562

## Description

Replaced the use of `NoisyReadoutSampler` and `cirq.measure_confusion_matrix` with a straightforward approach of inverting a single-qubit confusion matrix and tensoring.

---

### License

### DIFF
--- a/mitiq/interface/mitiq_cirq/cirq_utils.py
+++ b/mitiq/interface/mitiq_cirq/cirq_utils.py
@@ -14,16 +14,13 @@
 # along with this program.  If not, see <https://www.gnu.org/licenses/>.
 """Cirq utility functions."""
 
-from typing import Tuple, Sequence
+from functools import reduce
+from typing import Tuple
 
 import numpy as np
 import numpy.typing as npt
 import cirq
 from mitiq._typing import MeasurementResult
-
-from cirq.experiments.single_qubit_readout_calibration_test import (
-    NoisySingleQubitReadoutSampler,
-)
 
 
 # Executors.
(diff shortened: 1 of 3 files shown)


amirebrahimi:
Oddly enough, asv is failing here even though I removed the dependency, @Misty-W. I think it is because it compares against the main branch, no?

Misty-W:
> Oddly enough, asv is failing here even though I removed the dependency, @Misty-W. I think it is because it compares against the main branch, no?

yes @amirebrahimi I saw the same thing when I tried to remove the dependency, and I think it's because the workflow compares against master.

andreamari (approved):
LGTM! Thanks @amirebrahimi for the quick PR! 

In addition to simplifying the the code, I am quite sure it will also fix the ASV failures that we are seeing everywhere. So I'll merge right now and check if it works. 

@natestemen, it seems that the ASV test in CI always pulls mitiq from the master branch even if the test is triggered by a PR commit (different branch).  Do you think this is normal behavior?

PR #1614, opened 2022-11-30
Calibration prototype

natestemen:
## Description

🦴 skeleton of calibration work as spec'd out in the [calibration RFC](https://docs.google.com/document/d/1EZUJyEEUQUH33UOgSIzCCvXyxP0WLOQn11W0x4Ox4nY/).

This works by specifying a `Settings` object which contains data/parameters to generate a series of "experiments". The `Settings` object is used to create a `Calibration` object which will keep track of all of the data surrounding the subsequent execution of the specified circuits and the associated circuit/mitigation parameters. These results can then be used to understand what techniques/parameters work best for their specified backend. The current implementation only supports running experiments with ZNE.

### Example

```py
import cirq
from mitiq.calibration import Calibration, ZNESettings

def execute(circuit, noise_level=0.001):
    noisy_circuit = circuit.with_noise(cirq.depolarize(p=noise_level))
    return (
        cirq.DensityMatrixSimulator()
        .simulate(noisy_circuit)
        .final_density_matrix[0, 0]
        .real
    )

cal = Calibration(execute, ZNESettings)
cal.get_cost()
>>> {'noisy_executions': 24, 'ideal_executions': 0}

cal.run_circuits()
cal.results
>>> [{'unmitigated': 0.49900118,
  'mitigated': {'zne': {'results': [{'circuit_type': 'ghz',
      'extrapolation_method': 'RichardsonFactory',
      'scale_factors': [1.0, 2.0, 3.0],
      'scale_noise_method': 'fold_global',
      'mitigated_value': 0.5003295540809632}],
    'method_improvement_factor': None}},
  'ideal': 0.49999997}, ... ]

cal.compute_improvements()
cal.get_optimal_strategy()
>>> 'zne'
```

### Questions for reviewers
- What do you think of the user flow?
- Should data be returned by functions like `run_circuits`/`compute_improvements`, or should it be stored internal to the class?
- I think a future improvement could be to have a `Schedule` class which specifies a series of circuits to run and stores the results. Do you think this is a good idea? Is it worth doing in this PR?

I'm aware of the failing mypy issues, and will resolve those.

---

Problems: 
- insert_id_layers seems to break something when circuits have measurements
- LinearFactory and RichardsonFactory are supported, but PolyFactory is not easy, and ExpFactory can error easily.

### DIFF
--- a/mitiq/calibration/calibration.py
+++ b/mitiq/calibration/calibration.py
@@ -14,6 +14,7 @@
 # along with this program.  If not, see <https://www.gnu.org/licenses/>.
 
 from collections import Counter
+from dataclasses import asdict
 from math import prod, sqrt
 from typing import Any, Callable, Union, cast
 
@@ -107,13 +108,15 @@ class Calibrator:
                 )
                 mitigated[strategy.technique]["results"].append(
                     {
-                        "circuit_type": circuit_data.type,
                         "mitigated_value": mitigated_expval,
                         **strategy.technique_params,
                     }
                 )
+            circuit_info = asdict(circuit_data)
+            del circuit_info["circuit"]
             expvals.append(
                 {
+                    "circuit_info": circuit_info,
                     "noisy_value": noisy_distribution.get(
                         bitstring_to_observe, 0
                     ),
--- a/mitiq/calibration/calibration.py
+++ b/mitiq/calibration/calibration.py
@@ -66,7 +66,7 @@ class Calibrator:
             "ideal_executions": ideal,
         }
 
-    def run_circuits(self) -> None:
+    def run_circuits(self) -> list[dict[str, Any]]:
         """Run the calibration circuits and store the ideal and noisy
         expectation values for each circuit in `self.results`."""
         expvals = []
@@ -119,12 +119,14 @@ class Calibrator:
                     "mitigated": mitigated,
                 }
             )
-        self.results = expvals
+        return expvals
 
-    def compute_improvements(self) -> None:
+    def compute_improvements(
+        self, experiment_results: list[dict[str, Any]]
+    ) -> None:
         """Compute the improvement factors for each calibration circuit that
         was run."""
-        for result in self.results:
+        for result in experiment_results:
             ideal_dist = result["ideal_dist"]
             ideal_expval = max(ideal_dist.values())
             noisy_dist = result["noisy_dist"]
@@ -145,12 +147,12 @@ class Calibrator:
                 )
                 di["method_improvement_factor"] = method_improvement_factor
 
-    def get_optimal_strategy(self) -> str:
+    def get_optimal_strategy(self, results: list[dict[str, Any]]) -> str:
         """Finds the optimal error mitigation strategy using the improvement
         factors calculated, and stored in `self.results`."""
         best_val = 0.0
         best_key = ""
-        for result in self.results:
+        for result in results:
             ideal_dist = result["ideal_dist"]
             ideal = max(ideal_dist.values())
             noisy_dist = result["noisy_dist"]
@@ -165,6 +167,13 @@ class Calibrator:
                         best_val = error_diff
                         best_key = method
         return best_key
+
+    def run(self):
+        results = self.run_circuits()
+        self.compute_improvements(results)
+        self.results = results
+        strategy = self.get_optimal_strategy(results)
+        return strategy
 
 
 def bitstrings_to_distribution(
--- a/mitiq/calibration/calibration.py
+++ b/mitiq/calibration/calibration.py
@@ -44,8 +44,8 @@ class Calibration:
         Calibration object, the number of classical simulations is also
         returned."""
         num_circuits = len(self.circuits)
-        num_methods = len(self.settings.mitigation_methods)
-        num_options = prod(map(len, self.settings.method_params.values()))
+        num_methods = len(self.settings.techniques)
+        num_options = prod(map(len, self.settings.technique_params.values()))
 
         noisy = num_circuits * num_methods * num_options
         ideal = noisy if self.ideal_executor else 0
@@ -66,7 +66,7 @@ class Calibration:
             )
             mitigated = {
                 method: {"results": [], "method_improvement_factor": None}
-                for method in self.settings.mitigation_methods
+                for method in self.settings.techniques
             }
             for (
                 method,
--- a/mitiq/calibration/calibration.py
+++ b/mitiq/calibration/calibration.py
@@ -23,8 +23,8 @@ class Calibration:
     def __init__(
         self,
         executor: Executor,
-        ideal_executor: Executor,
         settings: Settings,
+        ideal_executor: Executor | None = None,
     ):
         self.executor = executor
         self.ideal_executor = ideal_executor
@@ -62,7 +62,9 @@ class Calibration:
                 mitigated[
                     f"{method_key}-{factory}-{scale_factors}-{scale_method}"
                 ] = mitiq_func(circuit, self.executor)
-            ideal = self.ideal_executor(circuit)
+            ideal = (
+                self.ideal_executor(circuit) if self.ideal_executor else 1.0
+            )
             expvals.append(
                 {"unmitigated": expval, "mitigated": mitigated, "ideal": ideal}
             )
--- a/mitiq/calibration/calibration.py
+++ b/mitiq/calibration/calibration.py
@@ -15,6 +15,7 @@
 
 from mitiq import Executor
 from mitiq.calibration import Settings
+from math import prod
 
 
 class Calibration:
@@ -37,8 +38,9 @@ class Calibration:
         If method requires simulation, returns number of ideal expectation values needed."""
         num_circuits = len(self.circuits)
         num_methods = len(self.settings.mitigation_methods)
+        num_options = prod(map(len, self.settings.method_params.values()))
 
-        noisy = num_circuits * self.settings.max_executions * num_methods
+        noisy = num_circuits * num_methods * num_options
         ideal = noisy if self.ideal_executor else 0
         return {
             "noisy_executions": noisy,
--- a/mitiq/calibration/calibration.py
+++ b/mitiq/calibration/calibration.py
@@ -23,7 +23,7 @@ from mitiq import QPROGRAM, Executor, MeasurementResult, QuantumResult
 from mitiq.calibration.settings import Settings
 
 
-class Calibration:
+class Calibrator:
     """A calibration object which keeps track of, and aids in Error
     Mitigation parameter tuning."""
 
@@ -53,7 +53,7 @@ class Calibration:
     def get_cost(self) -> dict[str, int]:
         """Returns the expected number of noisy expectation values required
         for calibration. If an ideal_executor was used in specifying the
-        Calibration object, the number of classical simulations is also
+        Calibrator object, the number of classical simulations is also
         returned."""
         num_circuits = len(self.circuits)
         num_methods = len(self.settings.techniques)
--- a/mitiq/calibration/calibration.py
+++ b/mitiq/calibration/calibration.py
@@ -16,12 +16,18 @@
 from collections import Counter
 from dataclasses import asdict
 from math import prod, sqrt
-from typing import Any, Callable, Union, cast, List, Dict, Tuple
+from typing import Any, Callable, Dict, List, Optional, Tuple, Union, cast
 
 import cirq
 
-from mitiq import QPROGRAM, Executor, MeasurementResult, QuantumResult
-from mitiq.calibration.settings import Settings
+from mitiq import (
+    QPROGRAM,
+    Executor,
+    MeasurementResult,
+    Observable,
+    QuantumResult,
+)
+from mitiq.calibration.settings import Settings, Strategy
 
 
 class Calibrator:
@@ -77,6 +83,8 @@ class Calibrator:
                 self.executor, distribution
             )
             noisy_value = expval_executor.evaluate(circuit)[0]
+            ideal_value = distribution[bitstring_to_measure]
+            noisy_error = ideal_value - noisy_value
 
             mitigated_values: Dict[str, Dict[str, Any]] = {
                 technique: {"results": [], "improvement_factor": None}
@@ -86,7 +94,15 @@ class Calibrator:
                 technique, params, mitiq_func = strategy
                 mitigated_value = mitiq_func(circuit, expval_executor)
 
-                result = {"mitigated_value": mitigated_value, **params}
+                improvement_factor = abs(
+                    noisy_error / (ideal_value - mitigated_value)
+                )
+                result = {
+                    "mitigated_value": mitigated_value,
+                    "improvement_factor": improvement_factor,
+                    "strategy": strategy,
+                    **params,
+                }
                 mitigated_values[technique]["results"].append(result)
 
             circuit_info = asdict(problem)
@@ -124,20 +140,19 @@ class Calibrator:
                 )
                 di["improvement_factor"] = improvement_factor
 
-    def get_optimal_strategy(self, results: List[Dict[str, Any]]) -> str:
-        """Finds the optimal error mitigation strategy using the improvement
-        factors calculated, and stored in `self.results`.
-
-        Currently, this function"""
-        best_val = 0.0
-        best_key = ""
-        for result in results:
-            for method, di in result["mitigated_values"].items():
-                improvement_factor = di["improvement_factor"]
-                if improvement_factor > best_val:
-                    best_val = improvement_factor
-                    best_key = method
-        return best_key
+    def best_strategy(self, results: List[Dict[str, Any]]) -> Strategy:
+        """Finds the best strategy by using the parameters that had the
+        largest improvement factor."""
+        best_improvement_factor = 0.0
+        strategy = None
+        for circuit in results:
+            for result in circuit["mitigated_values"]["zne"]["results"]:
+                if result["improvement_factor"] > best_improvement_factor:
+                    best_improvement_factor = result["improvement_factor"]
+                    strategy = result["strategy"]
+        if strategy is None:
+            raise ValueError("None of the improvement factors were > 0")
+        return strategy
 
     def run(self) -> None:
         results = self.run_circuits()
@@ -160,7 +175,9 @@ def bitstrings_to_distribution(
 
 
 def convert_to_expval_executor(
-    ex: Executor, distribution: Dict[str, float]
+    ex: Executor,
+    distribution: Optional[Dict[str, float]] = None,
+    bitstring: Optional[str] = None,
 ) -> Tuple[Executor, str]:
     """Constructs a new executor returning an expectation value given by the
     probability that the circuit outputs the most likely state according to the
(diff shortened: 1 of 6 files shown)


mitiq/calibration/calibration.py
```diff
@@ -0,0 +1,127 @@
...
+            expvals.append(
+                {
+                    "unmitigated": noisy_expval,
+                    "mitigated": mitigated,
+                    "ideal": ideal_expval,
+                }
+            )
+        self.results = expvals
+
+    def compute_improvements(self) -> None:
```
andreamari, line 92: This method is never called.

mitiq/calibration/settings.py
```diff
@@ -0,0 +1,137 @@
...
+from mitiq.zne.inference import ExpFactory, LinearFactory, RichardsonFactory
+from mitiq.zne.scaling import (
+    fold_gates_at_random,
+    fold_global,
+    insert_id_layers,
+)
+
+
+@dataclass
+class CircuitData:
```
andreamari, line 41: Consider renaming `CircuitData` to `BenchmarkProblem` or `BenchmarkCircuit`.

Just to give the idea of something against which a `Strategy` is tested. 

Not essential.

mitiq/calibration/settings.py
```diff
@@ -0,0 +1,137 @@
...
+    circuit: QPROGRAM
+    dimensions: tuple[int, int]
```
andreamari, line 43: What about splitting `dimensions`  into two explicit attributes `num_qubits` and `depth` ?

mitiq/calibration/settings.py
```diff
@@ -0,0 +1,137 @@
...
+    type: str
+    two_qubit_gate_count: int
```
andreamari, line 45: Here we probably need to store also the `ideal_result` (e.g. for mirror circuits this is the correct bitstring).

Optional: We could also consider adding the associated empty attributes `unmitigated_result` and `mitigated_results` that could be populated during the calibration phase.

mitiq/calibration/settings.py
```diff
@@ -0,0 +1,137 @@
...
+
```
andreamari, line 46: What about adding here also a Strategy dataclass?
E.g.
```
@dataclass
class Strategy:
    technique: str
    technique_params: dict[str, Any]
    mitigate_function: Callable
```

This could help to bound together technique, parameters and mitigation function into a single object.

mitiq/calibration/settings.py
```diff
@@ -0,0 +1,137 @@
...
+
+class Settings:
+    """A class to store settings relating to error mitigation calibration."""
+
+    def __init__(
+        self,
+        mitigation_methods: list[str],
```
andreamari, line 53: Optional: we could rename `mitigation_methods` to `techniques`  and `method_params`  to `technique_params`

mitiq/calibration/settings.py
```diff
@@ -0,0 +1,137 @@
...
+        circuit_types: list[str],
+        circuit_dimensions: tuple[int, int],
+        method_params: dict[str, Any],
+    ):
+        self.mitigation_methods = mitigation_methods
+        self.method_params = method_params
+        self.circuit_types = circuit_types
+        self.circuit_dimensions = circuit_dimensions
+
+    def make_circuits(self) -> list[CircuitData]:
```
andreamari, line 63: This function should also save the correct bitstring for each benchmark circuits. E.g. in the `ideal_result` attribute of `CircuitData`.

mitiq/calibration/settings.py
```diff
@@ -0,0 +1,137 @@
...
+                CircuitData(
+                    circuit,
+                    circuit_dimensions,
+                    circuit_type,
+                    two_qubit_gate_count,
+                )
+            )
+        return circuits
+
+    def mitigate_functions(self) -> list[tuple[str, Callable]]:
```
andreamari, line 99: This method could be changed into a `make_strategies()` method, returning `Strategy` objects instead of functions.

andreamari (commented):
Second round. This is second part is about `settings.py.`

**General issue that we may need to discuss. (Not directly related to this review)**
We may force the usage of executors **returning raw bitstings**  rather  than expectation values (floats). This would simplify the comparison of mitigated, noisy and ideal results.

willzeng:
Super excited to see this all coming together. Haven't looked into the PR super deeply but had a mention from looking at the flow the sequence:
```
cal.run_circuits()
cal.compute_improvements()
cal.get_optimal_strategy()
```
raises a bit of concern because there is a lot of hidden state here. Presumably these commands need to be run in exactly this order for things to make sense? If so, perhaps a preferable pattern is:
```
results = cal.run_circuits() # also stored in cal.results
improvements = cal.compute_improvements(results) # also stored in cal.improvements
mitigation_tactic = cal.get_optimal_tactic() 
```

Broadly though this still looks like a lot of steps for the user to go through. Is the idea that these would all be under the hood and the user can do something like (from the RFC):
```
# Initialize a calibration object
cal = Calibration(executor)
# Execute test circuits and compute performances.
cal.run()
 
# Let Mitiq automatically deduce the optimal error mitigation strategy
mitigated_result = execute_with_mitigation(
circuit,
executor,
observable=None, # Optional
calibration=cal, # Full calibration object
)
```
This then only has one step `.run()` which has been executed or not. One could also imagine doing `cal.load(file)` to load a pre-configured setting.

Does that make sense @natestemen? We can discuss live too just let me know

mitiq/calibration/calibration.py
```diff
@@ -0,0 +1,204 @@
...
+                )
+                mitigated_expval = strategy.mitigation_function(
+                    circuit, expval_executor
+                )
+                mitigated[strategy.technique]["results"].append(
+                    {
+                        "circuit_type": circuit_data.type,
+                        "mitigated_value": mitigated_expval,
+                        **strategy.technique_params,
+                    }
```
andreamari, line 113: For each result we are saving the circuit type but additional circuit information is lost (e.g. number of qubits, depth, or the circuit itself). This may be good or bad depending on a trade off between completeness and efficiency. 

In the future we could save here the full `circuit_data` object or, perhaps better, we could save the results `mitigated` inside the existing `circuit_data` object.

This is just a comment to think about for the future. It's probably ok to keep the current structure for now.

mitiq/calibration/calibration.py
```diff
@@ -0,0 +1,204 @@
...
+    def compute_improvements(
+        self, experiment_results: list[dict[str, Any]]
+    ) -> None:
+        """Compute the improvement factors for each calibration circuit that
+        was run."""
+        for result in experiment_results:
+            ideal_dist = result["ideal_dist"]
+            ideal_expval = max(ideal_dist.values())
+            noisy_dist = result["noisy_dist"]
+            noisy_expval = max(noisy_dist.values())
```
andreamari, line 133:
```suggestion
            ideal_value = result["ideal_value"]
            noisy_value = result["noisy_value"]
```

Instead of taking the maximum of the observed frequencies we should take the frequency of `bitstring_to_measure`.  This is not necessarily the same thing, especially if noise is very strong.

mitiq/calibration/calibration.py
```diff
@@ -0,0 +1,204 @@
...
+                mitigated_vals = list(
+                    map(lambda di: di["mitigated_value"], results)
+                )
+                method_improvement_factor = abs(
+                    noisy_expval - ideal_expval
+                ) / sqrt(
+                    len(mitigated_vals)
+                    * sum(
+                        (mitigated_val - ideal_expval) ** 2
+                        for mitigated_val in mitigated_vals
```
andreamari, line 145:
```suggestion
                    noisy_value - ideal_value
                ) / sqrt(
                    len(mitigated_vals)
                    * sum(
                        (mitigated_val - ideal_value) ** 2
                        for mitigated_val in mitigated_vals
```

For consistency between `something_expval` vs `something_value`.
No preference between the two, but better to keep consistency.

mitiq/calibration/calibration.py
```diff
@@ -0,0 +1,204 @@
...
+                    "mitigated": mitigated,
+                }
+            )
+        return expvals
+
+    def compute_improvements(
+        self, experiment_results: list[dict[str, Any]]
+    ) -> None:
+        """Compute the improvement factors for each calibration circuit that
+        was run."""
```
andreamari, line 128:
```suggestion
        """Computes the improvement factors for each calibration circuit that
        was run. Saves the improvement factors in the input dictionary."""
```

mitiq/calibration/calibration.py
```diff
@@ -0,0 +1,204 @@
...
+                method_improvement_factor = abs(
+                    noisy_expval - ideal_expval
+                ) / sqrt(
+                    len(mitigated_vals)
+                    * sum(
+                        (mitigated_val - ideal_expval) ** 2
+                        for mitigated_val in mitigated_vals
+                    )
+                )
+                di["method_improvement_factor"] = method_improvement_factor
```
andreamari, line 148:
```suggestion
                di["improvement_factor"] = method_improvement_factor
```
The improvement factors is always associated to a method. So maybe we can simplify its name.

mitiq/calibration/calibration.py
```diff
@@ -0,0 +1,204 @@
...
+                        best_val = error_diff
+                        best_key = method
+        return best_key
+
+    def run(self) -> str:
+        results = self.run_circuits()
+        self.compute_improvements(results)
+        self.results = results
+        strategy = self.get_optimal_strategy(results)
+        return strategy
```
andreamari, line 176:
```suggestion
     return results
```
Do you think `run()` should compute the optimal strategy? Maybe it could just run the circuits and compute the improvement factors. The optimal strategy could depend on the specific real use-case circuit, so we may postpone its computation at a later step.

mitiq/calibration/calibration.py
```diff
@@ -0,0 +1,204 @@
...
+    def run_circuits(self) -> list[dict[str, Any]]:
+        """Run the calibration circuits and store the ideal and noisy
+        expectation values for each circuit in `self.results`."""
+        expvals = []
+        for circuit_data in self.circuits:
+            circuit = circuit_data.circuit
+            circuit.append(cirq.measure(circuit.all_qubits()))
+
+            noisy_results = cast(
+                MeasurementResult, self.executor._run([circuit])[0]
```
andreamari, line 78:
```suggestion
                MeasurementResult, self.executor._run([circuit])[0]
```
Let's remember to make an issue for upgrading `_run` to become a public method of the `Executor` class. It is strange that there is not a public way of getting bitstrings. Not urgent.

andreamari (commented):
General comment: it's probably the right moment of introducing tests. 
At least a big integration test that mimics the full user experience. Otherwise it is quite hard to imagine how the workflow looks like just by looking at the methods.

mitiq/calibration/calibration.py
```diff
@@ -0,0 +1,204 @@
...
+            noisy = max(noisy_dist.values())
+            unmitigated_error = abs((ideal - noisy) / ideal)
+            for method, di in result["mitigated"].items():
+                for res in di["results"]:
+                    mitigated_expval = res["mitigated_value"]
+                    diff = abs((ideal - mitigated_expval) / ideal)
+                    error_diff = abs(unmitigated_error - diff)
+                    if error_diff > best_val:
+                        best_val = error_diff
+                        best_key = method
```
andreamari, line 168:
```suggestion
            for method, di in result["mitigated"].items():
               improvement_factor = di["improvement_factor]
               if error_diff > best_val:
                   best_val = improvement_factor
                   best_key = method
```
I forgot to add this to the above review.
This is just a sketch but the message is that `get_optimal_strategy()` should not re-compute the improvement factors.

mitiq/calibration/calibration.py
natestemen: Good catch, I like this refactor.

mitiq/calibration/calibration.py
natestemen: Agreed. I had it as more of a place holder, but you're right that it's not needed here.

mitiq/calibration/calibration.py
natestemen: Ah yeah I see what you mean. Originally I was trying to get performance metrics for each result so we could know what parameters went along with the best performance. The suggestion you've made would be to find the best improvement factor for the EM method, but not any parameters.

mitiq/calibration/calibration.py
natestemen: Thanks for the reminder. Done in https://github.com/unitaryfund/mitiq/issues/1639.

mitiq/calibration/calibration.py
andreamari: My main suggestion is to avoid re-computing the improvement factors since, in principle, we have a method `compute_improvements()` for doing it.  Actually it would be much better to know the optimal parameters of a technique. So my code suggestion is probably not good enough and we should improve it. 

The workflow that I have in mind is:

1.  `compute_improvements()`, computes the improvement factors for all combinations of circuit and strategy (strategy = technique + parameters). Schematically something like: **(CircuitData, Strategy) ==> Improvement Factor.** 

2. `get_optimal_strategy()` simply picks the `Strategy` with maximum improvement factor. In a future iteration of the code, the method `get_optimal_strategy() ` could do more advanced things based on the use-case circuit of the user. E.g., picking the best  `Strategy`  after filtering a subset of results associated to benchmark circuits with say $n$ qubits, or with depth $d$, etc...

mitiq/calibration/calibration.py
```diff
@@ -0,0 +1,224 @@
...
+        """Computes the improvement factors for each calibration circuit that
+        was run. Saves the improvement factors in the input dictionary."""
+        for result in experiment_results:
+            ideal_value = result["ideal_value"]
+            noisy_value = result["noisy_value"]
+            for di in result["mitigated_values"].values():
+                results = di["results"]
+                mitigated_values = list(
+                    map(lambda di: di["mitigated_value"], results)
+                )
```
andreamari, line 133: Can we use something like this? 
`mitigated_values = [result["mitigated_value"] for result in results] 
`If not, no problem.

mitiq/calibration/calibration.py
```diff
@@ -0,0 +1,224 @@
...
+                improvement_factor = abs(noisy_value - ideal_value) / sqrt(
```
andreamari, line 134: This is something important for next year. Introducing trials, i.e. repetitions of the same experiment with the same strategy.
```suggestion
                TODO: Use root-mean-square error over multiple trials for equal settings.
                improvement_factor = abs(noisy_value - ideal_value) / sqrt(
```

mitiq/calibration/calibration.py
```diff
@@ -0,0 +1,224 @@
...
+                    len(mitigated_values)
+                    * sum(
+                        (mitigated_value - ideal_value) ** 2
+                        for mitigated_value in mitigated_values
+                    )
+                )
+                di["improvement_factor"] = improvement_factor
```
andreamari, line 141: Is this improvement factor for a specific strategy or for a specific technique?

If the improvement factor is associated to each strategy, as it seems from tests, I would expect a single "mitigated_value"  in the definition of the improvement factor. Why do we have a list of mitigated values instead?

mitiq/calibration/calibration.py
```diff
@@ -0,0 +1,224 @@
...
+
+    def best_strategy(self, results: List[Dict[str, Any]]) -> Strategy:
+        """Finds the best strategy by using the parameters that had the
+        largest improvement factor."""
+        best_improvement_factor = 0.0
```
andreamari, line 146: The base level is 1.0 (corresponding to equal performance with respect to unmitigated strategy).
```suggestion
        best_improvement_factor = 1.0
```

natestemen:
@andreamari despite the coverage % not being where we'd like it due to the last commit, I think we should merge this PR. The last commit [3bd549d](https://github.com/unitaryfund/mitiq/pull/1614/commits/3bd549dd77f23199fb5683e1ccffbb8815dda142) was written quickly to have a start to finish workflow ready for the end of the month. I think it's one of the first things we will have to clean up next year, so writing tests for them now, only to rewrite in less than a month didn't make sense to me. What do you think?

mitiq/calibration/calibration.py
```diff
@@ -0,0 +1,224 @@
...
+        strategy = None
```
andreamari, line 147:
```suggestion
        strategy = Strategy("raw", None, mitiq.raw.execute_with_raw)
```
In the worst case the trivial strategy (no error mitigation) should be returned. It's theoretical improvement factor is 1 by definition.

mitiq/calibration/calibration.py
```diff
@@ -0,0 +1,224 @@
...
+        for circuit in results:
+            for result in circuit["mitigated_values"]["zne"]["results"]:
+                if result["improvement_factor"] > best_improvement_factor:
+                    best_improvement_factor = result["improvement_factor"]
+                    strategy = result["strategy"]
+        if strategy is None:
+            raise ValueError("None of the improvement factors were > 0")
```
andreamari, line 154: I wouldn't raise an error if no technique is useful. We can just return the trivial no-mitigation strategy without halting the code of the user.

mitiq/calibration/calibration.py
```diff
@@ -0,0 +1,224 @@
...
+
+    return (
+        Executor(expval_executor),  # type: ignore [arg-type]
+        bitstring_to_measure,
+    )
+
+
+def execute_with_mitigation(
+    circuit: QPROGRAM,
+    calibrator: Calibrator,
```
andreamari, line 213: We could use the same signature of all the Mitiq execute_with_xxx functions:

```suggestion
    circuit: QPROGRAM,
    executor: Union[Executor, Callable],
    observable: Optional[Observable] = None,
    *,
    calibrator: Calibrator,
```

mitiq/calibration/calibration.py
```diff
@@ -0,0 +1,224 @@
...
+    bitstring: str,
+    observable: Optional[Observable] = None,
```
andreamari, line 215: Different ordering.
```suggestion
```

mitiq/calibration/calibration.py
```diff
@@ -0,0 +1,224 @@
...
+) -> QuantumResult:
+    if not calibrator.results:
+        calibrator.run()
+    strategy = calibrator.best_strategy(calibrator.results)
+    em_func = strategy.mitigation_function
+    expval_executor, _ = convert_to_expval_executor(
+        calibrator.executor, bitstring=bitstring
+    )
+    return em_func(circuit, executor=expval_executor, observable=observable)
```
andreamari, line 224:
```suggestion
    return em_func(circuit, executor=executor, observable=observable)
```

mitiq/calibration/tests/test_calibration.py
```diff
@@ -0,0 +1,167 @@
...
+    cost = cal.get_cost()
+    expected_cost = 2 * 3 * 2 * 2  # circuits * scale * methods * factories
+    assert cost["noisy_executions"] == expected_cost
+    assert cost["ideal_executions"] == 0
+
+
+def test_validate_run_circuits_schema():
+    cal = Calibrator(execute, settings)
+    results = cal.run_circuits()
+    results_schema = Schema(
```
andreamari, line 90: Cool! I didn't know about `Schema`

andreamari (approved):
There are several improvements to keep in mind for next year.
But I don't see any risk in merging what we made up to now to get this in the release.

mitiq/calibration/settings.py
```diff
@@ -0,0 +1,178 @@
...
+)
+
+
+@dataclass
+class BenchmarkProblem:
+    circuit: cirq.Circuit
+    num_qubits: int
+    circuit_depth: int
+    type: str
+    two_qubit_gate_count: int
```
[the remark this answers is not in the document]
natestemen: Great idea!

mitiq/calibration/calibration.py
natestemen: Yup, not sure why i wrote it using a `map`. Much harder to read that way. Thanks for the suggestion.

PR #1630, opened 2022-12-11
Fill in second sub-section of REM docs

### LINKED ISSUE #1452: REM: Write the second sub-section of the user guide
  - When should I use REM?

**Considerations**:

amirebrahimi:
## Description

Resolves #1452. Once #1629 gets merged, then it will make the diff of this PR much simpler. For the time being you can just review `rem-2-use-case.myst`.

---

### License


docs/source/guide/rem-2-use-case.myst
```diff
@@ -0,0 +1,34 @@
...
+  language: python
+  name: python3
+---
+
+# When should I use REM?
+
+## Advantages
+
+Readout-error mitigation is a technique that can correct the final step of a quantum algorithm: readout. 
+These errors are likely device-specific and, as such, can be mitigated specifically that way. If a
```
natestemen, line 19: I think we should assume the reader is familiar with the technique, so the first sentence might be redundant. As for the main advantages, I thought it was mainly that this technique deals with a different type of error than most mitigation techniques do, and it can be composed with other techniques were the main advantages. WDYT of something like this:

> Readout error mitigation is a technique that deals with errors that many other techniques do not. For that reason it can often be combined with existing workflows to produce better results without much modification. The technique can also accept as much information as the user has about the measurement statistics in order to generate more accurate predictions.

natestemen (commented):
Looking good, just want some clarifications as for when it's good/bad to use this technique.

In the future, you _should_ be able to set the target of the PR to another branch so you don't have to include all the changes from the other PR in this one as well. No big deal, just a heads up if you want to try it.

amirebrahimi:
> Looking good, just want some clarifications as for when it's good/bad to use this technique.
> 
> In the future, you _should_ be able to set the target of the PR to another branch so you don't have to include all the changes from the other PR in this one as well. No big deal, just a heads up if you want to try it.

Yes, that would work if the branch was on mitiq and not on a fork, but because these originated in a fork I wasn't able to set the target on mitiq. When working on a fork, if you set the source and target branch both from the fork, then it creates a PR [_on the fork_](https://github.com/amirebrahimi/mitiq/pull/1). Since then I've been added to mitiq, so I suppose I can originate branches on the main repo now.

docs/source/guide/rem-2-use-case.myst
amirebrahimi: I took your verbiage for this. Thanks.

docs/source/guide/rem-2-use-case.myst
```diff
@@ -0,0 +1,34 @@
...
+
+
+
+
+## Disadvantages
+
+The specificity of this technique, which can be an advantage, is also a disadvantage because it means 
+that the mitigation cannot be generally applied to a class of devices. Readout-error mitigation
+requires generating confusion matrices, which require numerous additional measurements in the 
+computational basis (scaling exponentially), are specific to a device. These matrices must then be
```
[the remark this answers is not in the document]
amirebrahimi: Removed.

docs/source/guide/rem-2-use-case.myst
```diff
@@ -0,0 +1,33 @@
...
+## Disadvantages
+
+The specificity of this technique, which can be an advantage, is also a
+disadvantage because it means that the measurement statistics that enable
+readout mitigation for one device cannot be generally applied to other
+devices. Readout-error mitigation requires generating confusion matrices,
+which require numerous additional measurements in the computational basis
+(scaling exponentially). These matrices must then be inverted and provided to
+the technique. For better scaling when generating these matrices, readout-error
+characterization can be kept locally to a qubit or groups of qubits.
```
andreamari, line 33: I would suggest a rewording similar to this: 

```suggestion
Readout-error mitigation requires the preliminary characterization of the measurement errors associated to a specific backend. Measurement errors can be expressed in the form of a confusion matrix and its estimation involves numerous state preparations and measurements in the computational basis.
The confusion matrix must then be inverted and provided as an input to the technique. The complete characterization of a full confusion matrix scales exponentially in the number of qubits. For a better scaling when generating these matrices, readout-error characterization can be simplified under the assumption that measurement errors are local with respect to individual qubits or group of qubits.
```

natestemen (approved):

PR #1631, opened 2022-12-12
Fix extrapolations for ExpFactory in special cases 

### LINKED ISSUE #1596: Unexpected extrapolations for ExpFactory in special cases
## Issue Description

There are cases in which `ExpFactory` fails to "choose" the optimal sign of $A$ in the model $f(x) = A e^{(-B \lambda)} + C$, which determines if  $f(x)$ is a decreasing or an increasing function.  The problems appears **only if the option** `asymptote` is passed (`asymptote` corresponds to fixing $C$ in the model ). 

The reason for this problem is that the sign of $A$ is guessed by Mitiq based on data without taking into account the value of `asymptote`. However the sign of $A$ should be based also on $C$, if $C$ is constrained.

This is ok as long as measured data decay or grow towards the expected direction $C$. If not, the extrapolation fails to get the best fit.

## How to Reproduce

1. 
```
from mitiq.zne.inference import ExpFactory

scale_factors = [1.0, 2, 3, 4]

exp_vals = [0.700000001, 0.7, 0.7, 0.7] # Weakly decreasing data
for infinite_noise_limit in [0.0, 0.2, 0.4, 0.6, 0.8, 1.0, 1.2]:
    print(ExpFactory.extrapolate(scale_factors, exp_vals, asymptote=infinite_noise_limit)) 

>>> 0.7000000009999999
>>> 0.7000000010000003
>>> 0.700000001
>>> 0.700000001
>>> 0.8000010000000001
>>> 1.000001
>>> 1.2000009999999999
```

```
exp_vals = [0.7, 0.7, 0.7, 0.700001] # Weakly increasing data
for infinite_noise_limit in [0.0, 0.2, 0.4, 0.6, 0.8, 1.0, 1.2]:
    print(ExpFactory.extrapolate(scale_factors, exp_vals, asymptote=infinite_noise_limit))

>>> -1.0000000000000112e-06
>>> 0.199999
>>> 0.39999900000000005
>>> 0.599999
>>> 0.6999994999977499
>>> 0.69999949999925
>>> 0.6999994999995498
```

## Expected behavior
In both cases the expected results are, **approximately**:

```
>>> 0.7
>>> 0.7
>>> 0.7
>>> 0.7
>>> 0.7
>>> 0.7
>>> 0.7
```

Misty-W:
## Description

Fixes #1596.

---

### License

### DIFF
--- a/mitiq/zne/inference.py
+++ b/mitiq/zne/inference.py
@@ -1363,7 +1363,15 @@ class PolyExpFactory(BatchedFactory):
 
         # Deduce "sign" parameter of the exponential ansatz
         linear_params, _ = mitiq_polyfit(scale_factors, exp_values, deg=1)
-        sign = np.sign(-linear_params[0])
+
+        if (
+            asymptote is not None
+            and np.sign(-linear_params[0] * (asymptote - linear_params[1]))
+            == 1
+        ):
+            sign = np.sign(linear_params[0])
+        else:
+            sign = np.sign(-linear_params[0])
 
         def _ansatz_unknown(x: float, *coeffs: float) -> float:
             """Ansatz of generic order with unknown asymptote."""


mitiq/zne/inference.py
```diff
@@ -1363,7 +1363,15 @@ def extrapolate(  # type: ignore
...
         # Deduce "sign" parameter of the exponential ansatz
         linear_params, _ = mitiq_polyfit(scale_factors, exp_values, deg=1)
-        sign = np.sign(-linear_params[0])
+
+        if (
+            asymptote is not None
+            and np.sign(-linear_params[0] * (asymptote - linear_params[1]))
+            == 1
+        ):
+            sign = np.sign(linear_params[0])
```
andreamari, line 1372: This part of the code seems equivalent to the following simplified code:

```suggestion
        if asymptote:
            sign = np.sign(asymptote - linear_params[1])
```

But I may be wrong. Please double-check.

mitiq/zne/tests/test_inference.py
```diff
@@ -474,31 +474,41 @@ def test_exp_factory_no_asympt(test_f: Callable[[float], float]):
...
     seeded_f = apply_seed_to_func(test_f, SEED)
-    fac = PolyExpFactory(X_VALS, order=2, asymptote=A, avoid_log=avoid_log)
+    fac = PolyExpFactory(
+        X_VALS, order=2, asymptote=asymptote, avoid_log=avoid_log
+    )
     fac.run_classical(seeded_f)
     assert not fac._opt_params
 
     zne_value = fac.reduce()
-    assert np.isclose(zne_value, seeded_f(0, err=0), atol=POLYEXP_TOL)
```
andreamari, line 493: Why this assertion was removed? 
Is this weakening the existing test conditions?

andreamari (changes requested):
Thanks @Misty-W, looks good!

Can we explicitly add a **new test** in which the code snippets in https://github.com/unitaryfund/mitiq/issues/1596 are executed getting the correct result  (`0.7`) ?

mitiq/zne/tests/test_inference.py
Misty-W: It looks redundant with line 505, where we compare the zne_value and the value from `PolyExpFactory`. If we keep it in, need to modify `seeded_f` to work with the new test cases.

mitiq/zne/inference.py
Misty-W: Almost! Thanks for suggesting I simplify this :)

mitiq/zne/tests/test_inference.py
Misty-W: Thinking about it some more, I will revert this change.

Misty-W:
@all-contributors please add @Misty-W for ideas, tests, maintenance, documentation

andreamari (approved):

Issue #1546, opened 2022-10-07
Identity Insertion Example for Docs


purva-thakre:
How should the identity scaling examples be added ? [On this page](https://mitiq.readthedocs.io/en/stable/examples/ibmq-backends.html), there are some examples for ZNE via Unitary folding. 

1. Create separate examples for identity scaling 
2. Limit the examples to what already exists but separate them into two sections : ZNE via Unitary Folding and ZNE via Identity Scaling

andreamari:
Maybe a separate short example in which unitary folding and identity scaling are compared?
(I would consider this as a nice but not urgent addition, since we already have a lot of ZNE examples.)

natestemen:
Just trying to clean up some issues. I think other issues cover this?

purva-thakre:
@natestemen Not really. The way I see the docs are structured, there's the Users Guide and then there are separate examples for ZNE with different backends etc. I created this issue for the `Examples` page.

PR #1633, opened 2022-12-13
Noise scaling tutorial (identity insertion/folding)

natestemen:
## Description

Quick demo showing off identity insertion alongside unitary folding.

fixes https://github.com/unitaryfund/mitiq/issues/1546

#### Question
1. I would expect these two methods to perform roughly equally for the specified noise model. Any insight as to why folding is getting a better improvement factor? (I ran the experiment with more samples as well, and I'm getting roughly the same results.)
2. Is the improvement factor I calculated the right one?


andreamari:
A quick comment: we have an improvement only if the improvement factor is larger than 1.

From the docs preciew I see this:

```
Avg improvement factor (`fold_global`):       0.55
Avg improvement factor (`insert_id_layers`):  0.1933
```

So, for some reason, both methods are _not improving_.

natestemen:
> both methods are _not improving_.

Ah okay thanks for double checking, I wasn't sure how to interpret that. Does the improvement factor calculation look correct? I'm not 100% sure I got that right.

andreamari:
> > both methods are _not improving_.
> 
> Ah okay thanks for double checking, I wasn't sure how to interpret that. Does the improvement factor calculation look correct? I'm not 100% sure I got that right.

If we want to follow the convention used in [our recent work](https://arxiv.org/abs/2210.07194) , the improvement factor is actually the inverse of what you defined.  I missed this fact in my previous comment. So this is good news since it means that both unitary folding and identity layers insertion are improving the estimation accuracy!

natestemen:
Just a reminder about this @Misty-W @purva-thakre

docs/source/examples/scaling.md
```diff
@@ -0,0 +1,161 @@
...
+print("-----ORIGINAL-----")
+print(demo)
+print("\n-----------------FOLDING------------------")
+print(fold_global(demo, scale_factor))
+print("\n-----------------SCALING------------------")
+print(insert_id_layers(demo, scale_factor))
+```
+
+Theoretically, these circuits should give the same result when measured, but due to noise, this is almost never the case.
+These two methods work both by extending the duration of the circuit, but do so in different ways that might be beneficial for different scenarios.
```
purva-thakre, line 50:
```suggestion
Both methods work by extending the duration of the circuit
```
Structure of the original sentence sounded a bit weird.

docs/source/examples/scaling.md
```diff
@@ -0,0 +1,161 @@
...
+
+|                | Folding              | Identity Insertion |
+| -------------- | -------------------- | ------------------ |
+| Equation       | $G \to GG^\dagger G$ | $G \to I G$        |
+| Mitiq Function | [`fold_global`](https://mitiq.readthedocs.io/en/stable/apidoc.html#mitiq.zne.scaling.folding.fold_global) | [`insert_id_layers`](https://mitiq.readthedocs.io/en/stable/apidoc.html#mitiq.zne.scaling.identity_insertion.insert_id_layers) |
+
+## Comparison
+
+To get started, we can demo what these two functions do to a small GHZ circuit.
+Each function (`fold_global` and `insert_id_layers`) take a circuit, and a specified scale factor which controls how much to increase the depth of the specified circuit.
```
purva-thakre, line 32:
```suggestion
Each function (`fold_global` and `insert_id_layers`) will take a circuit, and a specified scale factor as inputs. The specified scale factor controls how to increase the depth of the input circuit so that the achieved scale factor is exactly equal or very close to the specified scale factor.
```

docs/source/examples/scaling.md
```diff
@@ -0,0 +1,161 @@
...
+            scale_factor, len(demo), folded_depth, id_insert_depth
+        )
+    )
+```
+
+As expected we have $\mathtt{depth} * \mathtt{scale\_factor} = \mathtt{scaled\_depth}$.
+
+## Using noise scaling methods
+
+Here we demo how you can use these noise-scaling technique in ZNE.
```
purva-thakre, line 83:
```suggestion
Here, we demo how you can use these noise-scaling technique in ZNE.
```

docs/source/examples/scaling.md
```diff
@@ -0,0 +1,161 @@
...
+First we define an [executor](https://mitiq.readthedocs.io/en/stable/guide/executors.html) which is needed to tell Mitiq how to run the circuit.
```
purva-thakre, line 84:
```suggestion
First, we define an [executor](https://mitiq.readthedocs.io/en/stable/guide/executors.html) which is needed to tell Mitiq how to run the circuit.
```

docs/source/examples/scaling.md
```diff
@@ -0,0 +1,161 @@
...
+Here we choose a simple density matrix simulation with depolarizing noise which acts between every circuit layer.
```
purva-thakre, line 85:
```suggestion
We choose depolarizing noise via a simple density matrix simulation to act between every circuit layer.
```

docs/source/examples/scaling.md
```diff
@@ -0,0 +1,161 @@
...
+print("Avg improvement factor (`fold_global`):      ", round(np.average(results["fold"]), 4))
+print("Avg improvement factor (`insert_id_layers`): ", round(np.average(results["id"]), 4))
+```
+
+As we can see, both techniques offer an improvement.
+
+
+## Conclusion
+
+In this tutorial we've shown how to use both folding and identity insertion as noise scaling methods for Zero-Noise Extrapolation.
```
purva-thakre, line 160:
```suggestion
In this tutorial, we've shown how to use both folding and identity insertion as noise scaling methods for Zero-Noise Extrapolation.
```

purva-thakre (approved):
Some minor grammar nitpicks. Otherwise, LGTM !

docs/source/examples/scaling.md
```diff
@@ -0,0 +1,161 @@
...
+    scaled_IF = noisy_error / abs(ideal_expval - id_expval)
+
+    results["fold"].append(folded_IF)
+    results["id"].append(scaled_IF)
+
+print("Avg improvement factor (`fold_global`):      ", round(np.average(results["fold"]), 4))
+print("Avg improvement factor (`insert_id_layers`): ", round(np.average(results["id"]), 4))
+```
+
+As we can see, both techniques offer an improvement.
```
Misty-W, line 155: Any idea why noise scaling improvement is 6x greater than folding? Under what conditions would we expect one to perform (much) better than the other?

docs/source/examples/scaling.md
```diff
@@ -0,0 +1,161 @@
...
+    jupytext_version: 1.14.1
+kernelspec:
+  display_name: Python 3
+  language: python
+  name: python3
+---
+
+# Noise Scaling Methods
+
+In this tutorial we will review the differences between noise scaling methods used in [Zero-Noise Extrapolation](https://mitiq.readthedocs.io/en/stable/guide/zne.html) (ZNE).
```
Misty-W, line 16:
```suggestion
In this tutorial we will compare the noise scaling methods available in [Zero-Noise Extrapolation](https://mitiq.readthedocs.io/en/stable/guide/zne.html) (ZNE), identity insertion and unitary folding.
```

docs/source/examples/scaling.md
```diff
@@ -0,0 +1,161 @@
...
+As a reminder, ZNE works by running multiple versions of the desired circuit, each intended to scale the noise up from the base-level achieved by the hardware.
```
Misty-W, line 17:
```suggestion
ZNE works by running multiple versions of the desired circuit, each intended to scale the noise up from the base-level achieved by the hardware.
```

docs/source/examples/scaling.md
```diff
@@ -0,0 +1,161 @@
...
+Experimentally these experiments are often performed by pulse stretching, but as a quantum programmer, we typically do not have access to such low-level control.
+For this reason we use "digital" methods that allow us to scale the noise using gate-based methods.
+To this end, we will study circuit folding and identity insertion as methods to increase the amount of noise present in our computation.
+
+As a reminder, these techniques are summarized by the following equations, and can be performed in Mitiq with the associated functions.
```
Misty-W, line 22:
```suggestion
These techniques are summarized by the following equations, and can be performed in Mitiq with the associated functions.
```

Misty-W (commented):
My main comment is more discussion of the results at the end would be helpful, especially since the improvement is much greater with identity insertion. 

Other than that, some minor wording changes.

docs/source/examples/scaling.md
```diff
@@ -0,0 +1,161 @@
...
+    folded_depth = len(fold_global(demo, scale_factor))
+    id_insert_depth = len(insert_id_layers(demo, scale_factor))
+    print(
+        "{: >12}  {: ^14} {: ^14} {: ^15}".format(
+            scale_factor, len(demo), folded_depth, id_insert_depth
+        )
+    )
+```
+
+As expected we have $\mathtt{depth} * \mathtt{scale\_factor} = \mathtt{scaled\_depth}$.
```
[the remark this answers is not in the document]
natestemen: Good idea!

natestemen:
Thanks for the comments both! I didn't realize I had such a tendency to leave out a comma when introducing a sentence.

docs/source/examples/scaling.md
natestemen: I was also very surprised by this! It's a good sign, but to be truthful, I'm not really sure why it's performing so well! My _guess_ is that is has something to do with applying inverse gates via folding cancels out some of the noise amplification? tagging @andreamari to see if he's got any ideas.

Misty-W: You can also use `mitiq.zne.inference.Factory.plot_data` and `mitiq.zne.inference.Factory.plot_fit` to see the scaling effects.

andreamari: > I was also very surprised by this! It's a good sign, but to be truthful, I'm not really sure why it's performing so well! My _guess_ is that is has something to do with applying inverse gates via folding cancels out some of the noise amplification? tagging @andreamari to see if he's got any ideas.

I think it strongly depends on the simulated noise model.
The Cirq instruction `circuit.with_noise(cirq.depolarize(p=noise_level))` adds noise to **each qubit** after **each moment** independently from the applied gates. For this type of noise model, the insertion of identity layers is  optimal.

On the other hand, for a noise model in which a noise channel is applied after each gate, unitary folding is probably better.

Misty-W: thanks @andreamari! 

@natestemen, WDYT about this as a brief explanation at the end?

```suggestion
Both noise scaling techniques offer an improvement, but for this particular problem identity insertion outperforms unitary folding. 
The noise model adds noise to each qubit after each moment independently from the applied gates. For this type of noise model, the insertion of identity layers is optimal.

On the other hand, for a noise model in which a noise channel is applied after each gate, unitary folding would likely outperform identity insertion.
```

Misty-W (commented):
Suggested a brief explanation at the end based on Andrea's comment.

docs/source/examples/scaling.md
natestemen: I used the suggestion as a basis for [`4de60f9`](https://github.com/unitaryfund/mitiq/pull/1633/commits/4de60f91a51bd0419c221a2a092912a2604bac0b). WDYT?

docs/source/examples/scaling.md
```diff
@@ -0,0 +1,166 @@
...
+    results["fold"].append(folded_IF)
+    results["id"].append(scaled_IF)
+
+print("Avg improvement factor (`fold_global`):      ", round(np.average(results["fold"]), 4))
+print("Avg improvement factor (`insert_id_layers`): ", round(np.average(results["id"]), 4))
+```
+
+As we can see, both techniques offer an improvement, but for this problem identity insertion outperforms folding dramatically.
+This is due to the fact that ZNE assumes one can scale the noise throughout a circuit by doubling, tripling, etc, the noise's strength.
+Identity insertion does exactly this, and hence has great performance on this simple noise model.
```
andreamari, line 159:
```suggestion
On the simple noise model used in our example, identity insertion does exactly this, and hence has great performance.
```

docs/source/examples/scaling.md
```diff
@@ -0,0 +1,166 @@
...
+Noise models with increased complexity would likely see more equal, or perhaps better performance by folding-based techniques.
```
andreamari, line 160:
```suggestion
Noise models with increased complexity and gate-dependent errors would likely see more equal, or perhaps better performance by folding-based techniques.
```

andreamari (approved):

PR #1635, opened 2022-12-14
Fix build badge

andreamari:


natestemen (approved):
Happy to try this out to see if it fixes the problem.

PR #1636, opened 2022-12-14
Update mitiq build badge display name

Misty-W:
Description:
Fixes the badge display name.

### License


natestemen (commented):
I kinda think test is actually more accurate here.

Do you know why it works when using `build` in the URL? The [docs](https://docs.github.com/en/actions/monitoring-and-troubleshooting-workflows/adding-a-workflow-status-badge) that Andrea linked to in the email thread on this topic suggest using the filepath of the workflow. Seeing as we don't have a workflow/file named `build`, how does this work?

Misty-W:
> I kinda think test is actually more accurate here.
> 
> Do you know why it works when using `build` in the URL? The [docs](https://docs.github.com/en/actions/monitoring-and-troubleshooting-workflows/adding-a-workflow-status-badge) that Andrea linked to in the email thread on this topic suggest using the filepath of the workflow. Seeing as we don't have a workflow/file named `build`, how does this work?

We do have a workflow named `build`, see https://github.com/unitaryfund/mitiq/pull/1635/commits/774b58f0d461678fd0d71ff92cbdfcda41536789. 

It’s not a big deal but I think we should be consistent in the naming.

natestemen:
> We do have a workflow named build, see https://github.com/unitaryfund/mitiq/commit/774b58f0d461678fd0d71ff92cbdfcda41536789.

Good point. My guess is that GitHub is being generous and allowing both the use of the filename _and_ the workflow name? Since we don't know (and seeing as the badge just broke while using the workflow name as opposed to filename)  I think we should stick with the documented path. What if we rename the workflow back to `test`?

> we should be consistent in the naming.

Agreed, consistency is always good!

Misty-W:
GitHub is looking at the workflow `build` on the master branch, and it finds the master branch via the path specified. I think we should keep the name as `build` to keep confidence that we haven't changed the _criteria_ for a passing master branch.

andreamari:
The confusion was caused by my PR where I introduced a difference between the workflow file and the workflow name.

@Misty-W is suggesting that, in the current situation, we should point the badge to the workflow name (`build`).
@natestemen is suggesting that it is better to use the same name for the workflow and for the file. Otherwise we are not sure what is the right name that we must use in the badge link. 

I think you are both right!
 
If I am not wrong, we have 3 options:

1. Renaming both the workflow name back to `test`,  keeping the file name to `test.yml`  and using this badge: `https://github.com/unitaryfund/mitiq/workflows/test/badge.svg?branch=master`

2.  Keeping the workflow name `build`,  renaming the file name to `build.yml`  and using this badge: `https://github.com/unitaryfund/mitiq/workflows/build/badge.svg?branch=master`

3.  Merging this PR as it is now despite using different names for the workflow and for its file, but testing that the badge works as expected. E.g. by intentionally introducing a failing test. 

All solutions work for me!

Misty-W:
> The confusion was caused by my PR where I introduced a difference between the workflow file and the workflow name.
> 
> @Misty-W is suggesting that, in the current situation, we should point the badge to the workflow name (`build`). @natestemen is suggesting that it is better to use the same name for the workflow and for the file. Otherwise we are not sure what is the right name that we must use in the badge link.
> 
> I think you are both right!
> 
> If I am not wrong, we have 3 options:
> 
> 1. Renaming both the workflow name back to `test`,  keeping the file name to `test.yml`  and using this badge: `https://github.com/unitaryfund/mitiq/workflows/test/badge.svg?branch=master`
> 2. Keeping the workflow name `build`,  renaming the file name to `build.yml`  and using this badge: `https://github.com/unitaryfund/mitiq/workflows/build/badge.svg?branch=master`
> 3. Merging this PR as it is now despite using different names for the workflow and for its file, but testing that the badge works as expected. E.g. by intentionally introducing a failing test.
> 
> All solutions work for me!

I’m okay with solutions 2 or 3. WDYT @natestemen ?

natestemen:
Let's go with 2, but for the sake of following the docs to the T, let's include the `.yml` in the URL as they recommend. So something like: `https://github.com/unitaryfund/mitiq/workflows/build.yml/badge.svg?branch=master`.

natestemen:
Good catch Andrea! That's potentially another difference from the docs that could be the cause of the failure.

README.md
```diff
@@ -3,7 +3,7 @@
 <!-- ALL-CONTRIBUTORS-BADGE:START - Do not remove or modify this section -->
 [![All Contributors](https://img.shields.io/badge/all_contributors-34-orange.svg?style=flat-square)](#contributors-)
 <!-- ALL-CONTRIBUTORS-BADGE:END -->
-[![build](https://github.com/unitaryfund/mitiq/workflows/test/badge.svg?branch=master)](https://github.com/unitaryfund/mitiq/actions)
+[![build](https://github.com/unitaryfund/mitiq/workflows/build/badge.svg?branch=master)](https://github.com/unitaryfund/mitiq/actions)
```
Misty-W, line 6:
```suggestion
[![build](https://github.com/unitaryfund/mitiq/actions/workflows/build.yml/badge.svg?branch=master)
```

natestemen (approved):

PR #1651, opened 2022-12-23
Use source svg files instead of exported pngs

amirebrahimi:
## Description

All major browsers for the past 10 years have  [_basic_ SVG support](https://caniuse.com/?search=svg). When generating the workflow image for REM it required @andreamari to dig it up. I was thinking that it may be good to have it checked in to the repository, so that new figures can be created easily, but then just realized we could use it directly in place of the png. It's a smaller file size, too.

---

### License


docs/source/img/rem_workflow.svg
```diff
@@ -0,0 +1,383 @@
+<?xml version="1.0" encoding="UTF-8" standalone="no"?>
```
natestemen, line 1: I'm not sure what happened, but it looks like the arrows are missing from this diagram.

amirebrahimi: I'm seeing arrows here:
Screen Shot 2022-12-26 at 12 40 09 PM

You're not?

natestemen: Strangely, no!

Screenshot 2022-12-26 at 11 55 33 PM

After testing on Chromium it seems to work there, but not on firefox. What browser are you on?

amirebrahimi: I'm on Brave, which is Chromium. Do you have the latest Firefox version?

amirebrahimi: I tried with Firefox 72.0.2 on OSX and it rendered correctly:
Screen Shot 2022-12-27 at 7 50 03 PM

WrathfulSpatula: Firefox 108.0.1 (64-bit) on Ubuntu 22.04 LTS doesn't render arrows, either, rem_workflow.svg

natestemen: Thanks for checking Dan. I'm also on Firefox 108.0 (64-bit) on macOS Ventura 13.0.

Amir, if you upgrade firefox does it still work?

amirebrahimi: With Firefox 108.0.2 it does on macOS Catalina 10.15.7 it doesn't render the arrows. I'll see if I can track down why.

amirebrahimi: Should be fixed now. The trick is to use "Stroke to Path" feature in Inkscape, which splits the line and arrow out to separate shapes. I kept the original line/arrow in there, too, as a source and just hid the layer.

amirebrahimi: image

andreamari:
I had the same "_arrow problem_" too. But now it works!

amirebrahimi:
I think the failed test is unrelated. I can push an empty commit if you'd like to see it re-run.

natestemen:
I've re-run the failed action.

natestemen (approved):
Once the build has passed this looks good! Thanks Amir for updating these!

PR #1655, opened 2023-01-03
Fix executor and update wording of DDD tutorial

Misty-W:
## Description

Change noise level to 0 for `is_noisy=False` condition of executor. This wasn't caught earlier because `skip-execution` tags were used to minimize build time.

Also, clarify the improvement is very slight and emphasize it is for circuit depths above 3.

---

### License


andreamari:
Hi @Misty-W, I have some suggestions in  #1658, which is a PR pointing to this PR.

Misty-W:
Thanks for the suggestions in #1658 @andreamari !  I merged them into this PR and updated the thumbnail accordingly. Looking much better!

andreamari (changes requested):
Thanks for updating the figure! I didn't notice the `[skip-execution]` tags and so I expected the plot to be automatically generated.

With the current code the execution takes just a few seconds (10 seconds on my laptop). 
What about if we remove all the `[skip-execution]` tags, such that output text and the plot are automatically generated?

All the rest looks good to me!

Misty-W:
`skip-execution` tags removed. Docs failure is due to broken link to myst parser guide, fixed in #1661

andreamari (approved):

natestemen:
Nice!

PR #1658, opened 2023-01-04
PR towards the DDD PR for code suggestions

andreamari:
This PR is not directed to master but to the DDD PR by Misty. 
The reason is that I wasn't able to give code suggestions on lines of code that are already merged on master. 

I Hope it works.

Main changes:  

- Using a single run with many repetitions (shots) in the executor
- Using a non-Markovian type of noise
- Increasing depth and printing the total number of gates to see if DDD sequences are applied.

### DIFF
--- a/docs/source/examples/ddd_tutorial.myst
+++ b/docs/source/examples/ddd_tutorial.myst
@@ -4,7 +4,7 @@ jupytext:
     extension: .myst
     format_name: myst
     format_version: 0.13
-    jupytext_version: 1.14.1
+    jupytext_version: 1.11.1
 kernelspec:
   display_name: Python 3
   language: python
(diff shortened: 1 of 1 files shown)


docs/source/examples/ddd_tutorial.myst
```diff
@@ -169,28 +167,25 @@ noise_scaled_expectation_values = []
...
 
-        ddd_executor = functools.partial(
+        noisy_executor = functools.partial(
             execute,
             shots=shots,
             correct_bitstring=correct_bitstring,
         )
 
         ddd_value = ddd.execute_with_ddd(
-            circuit, ddd_executor, rule=rule, num_trials = 10,
```
andreamari, line 193: Note that `num_trials > 1` is typically only useful for **random** DDD sequences. Sence the YY rule is deterministic, there is no need to use multiple trials.

Issue #1560, opened 2022-10-19
Support python 3.11


natestemen:
It's supposed to arrive in 5 days ([schedule](https://peps.python.org/pep-0664/)). There are some serious speed improvements.

nathanshammah:
+1 for supporting this, and indeed it's been released.

natestemen:
Thanks Peter for letting us know! I've opened a [PR](https://github.com/aws/amazon-braket-sdk-python/pull/479) on `braket` to loosen the oqpy dependency requirement so we don't have to do this tango so often. Hopefully we can work something out. Once that (or something similar) is merged, we can move forward on this.

karalekas:
I doubt they'll elect to fully unpin OQpy as it's such a critical dependency for the Braket SDK. There shouldn't need to be much of a tango after this though? Supporting a new Python version only happens once a year

natestemen:
> I doubt they'll elect to fully unpin OQpy as it's such a critical dependency for the Braket SDK. 

Yeah, that makes sense. I'll see what they recommend. Maybe there's an acceptable range. If not, it's not a problem, but 2 weeks ago I bumped the oqpy version in braket ([PR](https://github.com/aws/amazon-braket-sdk-python/pull/469)) after the numpy and openpulse versions were loosed (in oqpy) as that was blocking us from upgrading to the latest version of braket.

This is the "tango" of dependency resolution that I referred to. You're right, hopefully this will be smoother going forward.

natestemen:
https://github.com/unitaryfund/mitiq/pull/1626 has been merged, and this should be ready to test.

PR #1663, opened 2023-01-06
Uprade python support to 3.11

natestemen:
## Description

This PR builds Mitiq using python 3.11, as well as ensuring users know our latest, supported python range via the `python_requires` keyword in `setup.py`.

resolves https://github.com/unitaryfund/mitiq/issues/1560

### DIFF
--- a/.github/workflows/asv.yaml
+++ b/.github/workflows/asv.yaml
@@ -17,7 +17,7 @@ jobs:
       - name: Set up Python
         uses: actions/setup-python@v4
         with:
-          python-version: '3.10'
+          python-version: "3.11"
 
       - name: Run benchmarks
         run: |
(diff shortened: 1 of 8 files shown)


natestemen:
Looks like [`h5py`](https://github.com/h5py/h5py/) (which is a dependency of pyscf) does not yet support python 3.11, and hence this is blocked until they provide the requisite wheels. Issue is tracked here: https://github.com/h5py/h5py/issues/2146.

natestemen:
Comment: the reason this PR was opened in January, and is not moving forward until now is that the RTD build was having problem, or taking extremely long. I originally planned to only move forward once everything was passing on python 3.11. However, I think it's best that we add testing for python 3.11 over ensuring the docs build with python 3.11. Hence I have removed that section of the PR, and will open a new issue to address RTD with 3.11.

andreamari (approved):
So, all CI tests pass with python 3.11 but we don't use python 3.11 for RTD, right? 
If so, sounds good and LGTM!

natestemen:
> we don't use python 3.11 for RTD, right?

Yup, exactly! I've opened https://github.com/unitaryfund/mitiq/issues/1790 in case we want to down the line.

PR #1665, opened 2023-01-06
DDD on IBMQ backends

Misty-W:
## Description

Tutorial for running Digital Dynamical Decoupling (DDD) on IBMQ backends

---

### License


Misty-W:
Strangely is the plot looks different from the Cirq example. I think there is some discrepancy in the way the noise model is applied but so far haven't been able to pinpoint it.

andreamari:
The difference between Qiskit and Cirq is probably that Cirq applies noise at each moment (at least in our DDD example), while Qiskit adds noise after gates. So if you have a long idle window, noise is applied many times in Cirq and just once in Qiskit. 

I am not sure if there is a way to simulate idle noise in Qiskit. A workaround could be to convert a circuit to Cirq, add RZ noise (this step will insert Rz unitary gates), convert back to Qiskit and run on a (noiseless) simulator. 

Note that, in any case, the simulated noise model is very _ad-hoc_ and real IBM hardware will behave in a completely different way.

nathanshammah:
Great to see this PR, great work @Misty-W! And great to see the use of a simple non-Markovian noise model at the Hamiltonian level as suggested by @andreamari: Rotation errors are known to have affected also the ~sqrt(iSWAP) gate in the Google Sycamore experiment, so they are quite realistic. 

- One idea in order to try to improve the results of the application of DDD: What about using [ghz circuits](https://mitiq.readthedocs.io/en/latest/apidoc.html#module-mitiq.benchmarks.ghz_circuits) as additional benchmark? They are way more sparse in terms of gates than mirror circuits and thus should provide longer and more windows for DDD sequences. 

- Regarding the issues with Qiskit: Is there a tool/way to run Cirq on IBMQ backends, bypassing the need to convert Qiskit conversion in Mitiq, maybe with some existing plugin? I couldn't find any.

nathanshammah:
Plan for next steps:
1. Run the GHZ circuit workload using the existing function [`generate_ghz_circuit()`](https://mitiq.readthedocs.io/en/latest/_modules/mitiq/benchmarks/ghz_circuits.html#generate_ghz_circuit) on a simulator. This can be Cirq simulator (first) and then a FakeDevice in Qiskit as a IBM Quantum simulator
2. Run on hardware step 1 (IBM Quantum).

The observable can be the ghz circuit + the ghz circuit inverse so that the result should be the original 00..0 state.

Misty-W:
1. Plot looks reasonable, after removing spaces from the noisy bitstring output. The extra spaces were introduced through conversions from Qiskit to Cirq and back.
2. Docs failure is due to a "broken" link in another example, which looks to be a temporary outage.

docs/source/examples/ddd_on_ibmq_ghz.md
```diff
@@ -0,0 +1,243 @@
...
+
+avg_ddd_values = np.average(ddd_values, axis=1)
+std_ddd_values = np.std(ddd_values, axis=1, ddof=1)
+
+plt.figure(figsize=(9, 5))
+
+plt.plot(num_qubits, avg_true_values, "--", label="True", lw=2)
+eb = plt.errorbar(
+    num_qubits, avg_noisy_values, yerr=std_noisy_values, label="Raw", ls="-."
+)
```
andreamari, line 231: Error bars are so small that are practically invisible. 
This is probably due to GHZ circuits being deterministic as opposed to rb/mirror. 

Possible solutions:

- Reducing number of shots to increase error bars.
- Avoiding the loop over multiple trials and the error bars.

andreamari (commented):
Looks very good! Nice to see DDD working with Qiskit.

docs/source/examples/ddd_on_ibmq_ghz.md
```diff
@@ -0,0 +1,243 @@
...
+    return circuit, correct_bitstring
+```
+
+## Define the executor
+
+Now that we have a circuit, we define the `execute` function which inputs a circuit and returns an expectation value - here, the
+frequency of sampling the correct bitstring.
+
+**Importantly**, since DDD is designed to mitigate time-correlated (non-Markovian) noise, we simulate a particular noise model consisting of
+systematic $R_Z$ rotations applied to each qubit after each moment.
```
Misty-W, line 95:
```suggestion
systematic $R_Z$ rotations applied to each qubit after each layer. 
```

docs/source/examples/ddd_on_ibmq_ghz.md
Misty-W: Solved by reducing shots down to 1000.

docs/source/examples/ddd_on_ibmq_ghz.md
```diff
@@ -0,0 +1,243 @@
...
+    noisy_sim: bool = True,
+) -> float:
+    """Returns the expectation value to be mitigated.
+
+    Args:
+        circuit: Circuit to run.
+        shots: Number of times to execute the circuit to compute the expectation value.
+    """
+
+    if USE_REAL_HARDWARE:
```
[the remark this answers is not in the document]
Misty-W: I discovered this problem when preparing to run on IBMQ, but I hadn't had a chance to update it here until now. The workflow is as you listed above.

andreamari (approved):
LGTM! Thanks @Misty-W

PR #1669, opened 2023-01-10
Clean up LaTeX

natestemen:
## Description

Some minor things I saw when we were going over the PEC module.


andreamari (approved):
Many thanks for this!

PR #1670, opened 2023-01-10
Improve and modify MeasurementResult class

### LINKED ISSUE #1570: Support dictionary with counts in the MeasurementResult class
## Issue Description

The [`MeasurementResult`](https://mitiq--1514.org.readthedocs.build/en/1514/_modules/mitiq/_typing.html#MeasurementResult) class in Mitiq represents the measured bitstrings obtained in a quantum computation.
At the moment it can only be initialized with an explicit list of bitstrings (one for each shot).

Another compact way of representing measurement results is via a dictionary of counts.
E.g.  `["01", "11", "11", "11"]  ` can be represented as  --->  `{"01": 1, "11": 3 }`
This method is used by most quantum software libraries.

## Proposed Solution

- It would be good if the `MeasurementResult` class could be initialized **also** by a dictionary (instead of a list of bitstrings). 

- It would be good if the `MeasurementResult` class could return results as a dicitonar (e.g. with a `self.as_dict()` method).

- It would be good if the `MeasurementResult` class used always a dictionary for storing data independently from the initialization input type. Storing a dictionary is typically more efficient than storing raw bitstrings and, for a large number of shots, it can make a big  difference.

andreamari:
## Description

Fixes #1570 .

This PR does the following things:

- Rename `_typing.py` into `typing.py` .  This is the only reason why so many files are changed!
- Generalize `mitiq.Bitstring` to allow for `List[str]` in addition to the current `List[List[int]]`. This is because bitstrings are often  `str` objects in many quantum libraries.
- Makes `MeasurementRresult` compatible with `Bitstring`s of type `str`.
- Allow defining a `MeasurementRresult`  from and to dictionaries of counts.
- Improve printing output of `MeasurementRresult`  to avoid printing millions of bitstrings. 
 
There are other changes to apply but I think it is better to implement them in a **different PR**.
For example, it would be good to take advantage of the new features to simplify the code of Mitiq, e.g. `mitiq.rem`, `mitiq.calibration`, to simplify and clean several examples in the docs.

---

### License


mitiq/typing.py
```diff
@@ -115,11 +124,47 @@ def nqubits(self) -> int:
...
+        """Initializes a MeasurementResult from a dictionary of counts.
+
+        Example:
+        >>> MeasurementResult.from_counts({"00": 175, "11": 177})
+        """
+        counter = Counter(counts)
+        return cls(list(counter.elements()), qubit_indices)
+
+    def get_counts(self) -> Dict[str, int]:
+        """Returns a colleciton.Counter whose keys are the measured
```
andreamari, line 142:
```suggestion
        """Returns a dictionary whose keys are the measured
```

mitiq/typing.py
```diff
@@ -115,11 +124,47 @@ def nqubits(self) -> int:
...
+    @classmethod
+    def from_counts(
+        cls,
+        counts: Dict[str, int],
+        qubit_indices: Optional[Tuple[int, ...]] = None,
+    ) -> "MeasurementResult":
+        """Initializes a MeasurementResult from a dictionary of counts.
+
+        Example:
+        >>> MeasurementResult.from_counts({"00": 175, "11": 177})
```
natestemen, line 136: This looks a little strange on the API-doc. WDYT of these two?
```suggestion
        **Example**::

            MeasurementResult.from_counts({"00": 175, "11": 177})
```
or
```suggestion
        **Example**: ``MeasurementResult.from_counts({"00": 175, "11": 177})``
```

mitiq/typing.py
```diff
@@ -115,11 +124,47 @@ def nqubits(self) -> int:
...
+        """
+        counter = Counter(counts)
+        return cls(list(counter.elements()), qubit_indices)
```
natestemen, line 139: TIL the `elements` method! Cool.

mitiq/typing.py
natestemen: The following line also needs a tweak as well FYI.

mitiq/typing.py
```diff
@@ -115,11 +124,47 @@ def nqubits(self) -> int:
...
+        data = {
+            "qubit_indices": self.qubit_indices,
+            "shots": self.shots,
+            "counts": self.get_counts(),
+        }
+        return data
+
+    def filter_qubits(self, qubit_indices: List[int]) -> npt.NDArray[np.int64]:
+        """Returns the bitstrings associated to a subset of qubits."""
+        return np.array([self._measurements[i] for i in qubit_indices]).T
```
natestemen, line 163: Is there a reason you like making a custom method, over the previously used `getitem` way? I don't have a preference, just curious if there is a reason.

mitiq/tests/test_measurement_result.py
```diff
@@ -69,16 +71,69 @@ def test_empty():
...
+    """Try using strings instead of lists of integers."""
+    bitstrings = ["00", "01", "10"]
+    int_bitstrings = [[0, 0], [0, 1], [1, 0]]
+
+    result = MeasurementResult(bitstrings, qubit_indices=qubit_indices)
+
+    assert result.nqubits == 2
+    assert result.qubit_indices == qubit_indices
+    assert result.shots == 3
+    assert np.allclose(result.result, int_bitstrings)
```
natestemen, line 93: Does exact equality not hold? I would think it should, as `result.result` should be a list of ints.

mitiq/tests/test_measurement_result.py
```diff
@@ -69,16 +71,69 @@ def test_empty():
...
+
+    int_bitstrings = [[0, 0], [0, 1], [0, 1]]
+    result = MeasurementResult(
+        result=int_bitstrings,
+        qubit_indices=(1, 20),
+    )
+    assert result.get_counts() == counts
+    # Info about qubit indices is expected to be lost
+    new_res = MeasurementResult.from_counts(result.get_counts())
+    assert new_res.qubit_indices != result.qubit_indices
```
natestemen, line 124: The qubit indices should not only not be equal, but it should be None, right?
```suggestion
    assert new_res.qubit_indices is None
```

mitiq/tests/test_measurement_result.py
```diff
@@ -69,16 +71,69 @@ def test_empty():
...
+    result = MeasurementResult.from_counts(
+        counts=counts,
+        qubit_indices=(1, 20),
+    )
+    expected = (
+        "MeasurementResult: {'qubit_indices': (1, 20),"
+        " 'shots': 1200, 'counts': {'00': 1000, '01': 200}}"
+    )
+    assert result.__repr__() == expected
+    assert result.__str__() == expected
```
natestemen, line 139: Any reason for not calling `repr(result)`/`str(result)`?

natestemen (commented):
Nice work Andrea, I think this will be very handy.

I've made some small comments/left some questions.

docs/source/apidoc.myst
```diff
@@ -35,6 +35,32 @@
...
+## Circuit types and result types
+
+### Circuit types
+```{eval-rst}
+.. autoclass:: mitiq.typing.QPROGRAM
+```
+
+```{eval-rst}
+.. autoclass:: mitiq.typing.SUPPORTED_PROGRAM_TYPES
+```
```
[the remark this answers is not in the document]
andreamari: Removed it. I also removed the sub headers `### Circuit types `and `### Result types` to keep it simple.

mitiq/typing.py
andreamari: The reason is that `getitem` is very natural for arrays with contiguous and unambiguous indices. 
In our case the function is used and defined for `qubit_indeces` and so its behavior would be quite ambiguous. 

E.g. is `result[3] ` the third measurement result, the results of the third bit, or the results of the qubit with index 3?

So, I preferred to use a different and more explicit name.

mitiq/tests/test_measurement_result.py
andreamari: Equality works!

mitiq/tests/test_measurement_result.py
andreamari: Changed to `assert new_res.qubit_indices = (0, 1)` since qubit_indeces are initialized to default values in `__post_init__`.

andreamari:
Suggestions applied. Note that I also added a simple `from_dict()` method  for symmetry reasons :-)

mitiq/typing.py
natestemen: Gotcha, that makes sense. Thanks for the explanation, and I agree this seems like the right way to go.

natestemen (approved):
Nice work! 🚢

PR #1673, opened 2023-01-17
Further updates related to the measurement result class

andreamari:
This PR has two main updates:

- A new section on the Executor docs, about executors that return measurement results.
- A new method for MeasurementResult that returns a probability distribution instead of counts.
- Simplification of a couple of executors defined in the docs.

### License

### DIFF
--- a/docs/source/guide/calibrators.md
+++ b/docs/source/guide/calibrators.md
@@ -6,7 +6,7 @@ jupytext:
     format_version: 0.13
     jupytext_version: 1.14.1
 kernelspec:
-  display_name: dev
+  display_name: Python 3 (ipykernel)
   language: python
   name: python3
 ---
@@ -32,17 +32,9 @@ from mitiq import MeasurementResult
 ```{code-cell} ipython3
 def execute(circuit, noise_level=0.001):
     circuit = circuit.with_noise(cirq.amplitude_damp(noise_level))
-
     result = cirq.DensityMatrixSimulator().run(circuit, repetitions=100)
-    return MeasurementResult(
-        result=np.column_stack(list(result.measurements.values())).tolist(),
-        qubit_indices=tuple(
-            # q[2:-1] is necessary to convert "q(number)" into "number"
-            int(q[2:-1])
-            for k in result.measurements.keys()
-            for q in k.split(",")
-        ),
-    )
+    bitstrings = np.column_stack(list(result.measurements.values()))
+    return MeasurementResult(bitstrings)
 ```
 
 We can now import the required objects and functions required for calibration.
--- a/docs/source/guide/rem-1-intro.myst
+++ b/docs/source/guide/rem-1-intro.myst
@@ -35,7 +35,7 @@ from mitiq.observable.observable import Observable
 from mitiq.observable.pauli import PauliString
 
 qreg = [LineQubit(i) for i in range(2)]
-circuit = Circuit(X.on_each(*qreg), measure_each(*qreg))
+circuit = Circuit(X.on_each(*qreg))
 observable = Observable(PauliString("ZI"), PauliString("IZ"))
 
 print(circuit)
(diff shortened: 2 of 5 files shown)


natestemen (commented):
Code portion looks nice! This will be very handy!

docs/source/guide/calibrators.md
```diff
@@ -32,17 +32,9 @@ from mitiq import MeasurementResult
...
     result = cirq.DensityMatrixSimulator().run(circuit, repetitions=100)
-    return MeasurementResult(
-        result=np.column_stack(list(result.measurements.values())).tolist(),
-        qubit_indices=tuple(
-            # q[2:-1] is necessary to convert "q(number)" into "number"
-            int(q[2:-1])
-            for k in result.measurements.keys()
-            for q in k.split(",")
-        ),
-    )
```
Misty-W, line 45: Why are the lines specifying the qubit indices removed? How do we keep track of the qubit indices associated with the bitstring if they aren't part of the `MeasurementResult`?

docs/source/guide/rem-1-intro.myst
```diff
@@ -70,13 +69,7 @@ def noisy_readout_executor(
     result = simulator.run(circuit, repetitions=shots)
 
     return MeasurementResult(
-        result=np.column_stack(list(result.measurements.values())),
-        qubit_indices=tuple(
-            # q[2:-1] is necessary to convert "q(number)" into "number"
-            int(q[2:-1])
-            for k in result.measurements.keys()
-            for q in k.split(",")
-        ),
```
Misty-W, line 79: Same question as in `calibrators.md`.

Misty-W (commented):
The addition of `prob_distribution` to  `typing.py` and the corresponding test look good. Just need clarification on the changes to `calibrators.md` and `rem-1-intro.md`.

andreamari:
`qubit_indices` is optional and is set by default to (0, 1, 2, 3 ....)
However, @Misty-W , you are right that assuming the default value for `qubit_indices` can be dangerous!
This is especially true when using Observables with support on a subset of qubits.

I tried to explain things in these commits ([a762f07](https://github.com/unitaryfund/mitiq/pull/1673/commits/a762f07a4944dad03fe468dd48817bf65318270f) [7468aca](https://github.com/unitaryfund/mitiq/pull/1673/commits/7468acaface990a41282438f5f7c1bb928e72e24))  and I restored explicit `qubit_indeces` in the docs when the default value is dangerous [ebee816](https://github.com/unitaryfund/mitiq/pull/1673/commits/ebee816ef1dc94ae0c8d21c3e0aa3fb358ffa9e3) while trying to simplify the code whenever possible.

Let me know what you think.

docs/source/guide/calibrators.md
andreamari: For the calibrator it is probably ok to use the default value for `qubit_indices` since all qubits must be measured during the calibration process.

mitiq/typing.py
```diff
@@ -76,7 +76,28 @@
...
+            If not given, Mitiq assumes the default ordering
+            ``tuple(range(self.nqubits))``, where ``self.nqubits``
+            is the bitstring length deduced from ``result``.
+
+    Note:
+        Using the default option for ``qubit_indeces`` can be
+        dangerous, especially when estimating an :class:`.Observable`
+        acting on a subset of qubits. In this case Mitiq
+        only applies measurement gates to the specific qubits and, therefore,
+        it is essential to specify the corresponding ``qubit_indeces``.
```
Misty-W, line 99:
```suggestion
        Use caution when selecting the default option for ``qubit_indices``,
        especially when estimating an :class:`.Observable`
        acting on a subset of qubits. In this case Mitiq
        only applies measurement gates to the specific qubits and, therefore,
        it is essential to specify the corresponding ``qubit_indices``.
```

Misty-W (approved):
LGTM @andreamari! 

Just a minor suggestion on wording of the docstring.

PR #1677, opened 2023-01-18
Add H2 image to fix the missing link in the docs

andreamari:
Fix docs build.

### License


natestemen (approved):
Thanks for fixing this up quickly! Looks good, but wondering if we should give credit to the image. We could keep in the line "Extracted from ...", but add the link to the linkcheck ignore:
https://github.com/unitaryfund/mitiq/blob/73acf6602fe09173938d2508b808b594fea19e38/docs/source/conf.py#L157-L160

Other ideas welcome!

andreamari:
> We could keep in the line "Extracted from ...", but add the link to the linkcheck ignore:

I did the image :-D (with inkscape)

natestemen:
Oh, awesome!! Nice work.

PR #1678, opened 2023-01-19
REM: mitigate executors as a first step and expectation values at a second step.

andreamari:
This PR applies a significant change on how REM is applied under the hood by Mitiq.
Currently REM is applied in the body of `execute_with_rem()` which is later called by `mitigate_executor`.
Wit this PR, REM is applied in the body of `mitigate_executor`, which is later called by `execute_with_rem()` .

The two main reasons for this change re:

1. REM acts on bitstrings, so the new workflow seems more natural.
2. The new workflow allows us to map an executor returning `MeasurementResult`s to a mitigated executor with the same signature. With the current workflow this is not possible, since the mitigated executor is necessarily returning a `float`.
3. The new workflow is compatible with non-diagonal Observables which are handled correctly by the `Executor.evaluate ` method. The current workflow based on `observable._expectation_from_measurements()` is buggy for non-diagonal observables.


docs/source/guide/rem-1-intro.myst
```diff
@@ -36,7 +34,7 @@ from mitiq.observable.observable import Observable
 from mitiq.observable.pauli import PauliString
 
 qreg = [LineQubit(i) for i in range(2)]
-circuit = Circuit(X.on_each(*qreg), measure_each(*qreg))
```
amirebrahimi, line 39: How come you removed the measurement? Is it it not necessary when defining observables?

mitiq/rem/tests/test_rem.py
```diff
@@ -14,28 +14,30 @@
...
-from mitiq.rem.rem import execute_with_rem, mitigate_executor, rem_decorator
+from mitiq.rem import execute_with_rem, mitigate_executor, rem_decorator
 from mitiq.raw import execute as raw_execute
 from mitiq.interface.mitiq_cirq import sample_bitstrings
 
 # Default qubit register and circuit for unit tests
 qreg = [cirq.LineQubit(i) for i in range(2)]
-circ = cirq.Circuit(cirq.ops.X.on_each(*qreg), cirq.measure_each(*qreg))
+circ_with_measurements = cirq.Circuit(
+    cirq.ops.X.on_each(*qreg), cirq.measure_each(*qreg)
```
amirebrahimi, line 38: Based on my other question, should we remove the measurements here, too?

mitiq/rem/rem.py
```diff
@@ -48,66 +49,86 @@ def execute_with_rem(
...
             unmitigated ``MeasurementResult``.
-        observable: Observable to compute the expectation value of (required).
         inverse_confusion_matrix: The inverse confusion matrix to apply to the
             probability vector estimated with noisy measurement results.
 
     Returns:
         The error-mitigated version of the input executor.
     """
+    if not isinstance(executor, Executor):
+        executor_obj = Executor(executor)
```
amirebrahimi, line 82: I haven't understood this pattern too well. In the case where an executor is not an `Executor`, then it must be a `Callable[[QPROGRAM], MeasurementResult]`, but what is the distinction here between these two? Can you give examples?

amirebrahimi (approved):
Glad to see this get implemented, since it came up in [previous discussion](https://discord.com/channels/764231928676089909/773957956659052605/959408516903362610) during implementation. I didn't and still don't fully understand the execution model of chaining/wrapping executors. A small write-up for contributing developers might help at some point.

docs/source/guide/rem-1-intro.myst
andreamari: Because when using an executor with to evaluate the expectation value of a Mitiq Observable, Mitiq automatically implements change of basis rotations and **adds measurements** on the qubits associated to the Observable. So we don't need to add measurements otherwise qubits will get measured twice.

The situation is different if instead of evaluating an Observable we want to just get the raw measurement results. In that case, the user must had measurements to the circuit.

I know, this is quite confusing and hopefully we'll make it simpler, but that's the current situation when using  `Observable`s in Mitiq.

mitiq/rem/tests/test_rem.py
andreamari: Since here we are not using the executor to evaluate an Observable but to just get the raw bitstrings, we need to explicitly add measurement gates to the circuit otherwise we get an empty result.

mitiq/rem/rem.py
andreamari: `Executor` is a class that Mitiq uses to wrap executor functions. `Executor` is what Mitiq always uses under the hood to execute circuits.
However, in order to make the user's experience simpler and more direct, we also accept as a valid input a simple Python function instead of an `Executor` object. In this case, we internally wrap the function into an `Executor` object. 

I added some docs and comments in the next commit to help developers and maintainers. But please, if things are not clear, let me know.

amirebrahimi:
LGTM. I'd say go ahead and merge.

PR #1682, opened 2023-01-26
Improve calibration method parameter selection

### LINKED ISSUE #1627: Improve calibration method/parameter selection
We need a smarter method to go from the calibration experiments, to recommending a error mitigation method along with parameters for that method.

More details will come later once https://github.com/unitaryfund/mitiq/pull/1614 is merged.

Misty-W:
## Description

Fixes #1627 

---

### License

### DIFF
--- a/mitiq/calibration/calibrator.py
+++ b/mitiq/calibration/calibrator.py
@@ -72,7 +72,7 @@ class Calibrator:
             else None
         )
         self.settings = settings
-        self.circuits = self.settings.make_circuits(self.settings.circuit_seed)
+        self.circuits = settings.make_circuits()
         self.results: List[Dict[str, Any]] = []
 
     def get_cost(self) -> Dict[str, int]:
@@ -187,11 +187,20 @@ class Calibrator:
         def filter_on_strategy(
             result: Dict[str, Dict[str, Dict[str, Any]]], strategy_id: int
         ) -> Dict[str, Dict[str, Dict[str, Any]]]:
+            """Obtain results corresponding to the strategy of interest.
+            Args:
+                result: Calibration experiment results.
+                strategy_id: Index of the strategy of interest.
+
+            Returns:
+                A dictionary of results corresponding to the strategy of
+                interest.
+            """
             res = result
             res["mitigated_values"]["ZNE"]["results"] = [
                 res["mitigated_values"]["ZNE"]["results"][strategy_id]
             ]
-            return result
+            return res
 
         for strategy_id in range(
             len(results[0]["mitigated_values"]["ZNE"]["results"])
@@ -220,9 +229,9 @@ class Calibrator:
             {"best_improvement_factor": best_improvement_factor}
         )
 
-        if strategy is None:
+        if strategy is DefaultStrategy:
             warnings.warn("None of the improvement factors were > 1")
-            return DefaultStrategy
+
         return strategy
 
     def run(self) -> None:
--- a/mitiq/calibration/calibrator.py
+++ b/mitiq/calibration/calibrator.py
@@ -210,9 +210,16 @@ class Calibrator:
                 ]
                 > best_improvement_factor
             ):
+                best_improvement_factor = strategy_group[0][
+                    "mitigated_values"
+                ]["ZNE"]["results"][0]["improvement_factor"]
                 strategy = strategy_group[0]["mitigated_values"]["ZNE"][
                     "results"
                 ][0]["strategy"]
+        self.results.append(
+            {"best_improvement_factor": best_improvement_factor}
+        )
+
         if strategy is None:
             warnings.warn("None of the improvement factors were > 1")
             return DefaultStrategy
--- a/mitiq/calibration/calibrator.py
+++ b/mitiq/calibration/calibrator.py
@@ -13,6 +13,8 @@
 # You should have received a copy of the GNU General Public License
 # along with this program.  If not, see <https://www.gnu.org/licenses/>.
 
+from copy import deepcopy
+import warnings
 from collections import Counter
 from math import sqrt
 from typing import (
@@ -70,7 +72,7 @@ class Calibrator:
             else None
         )
         self.settings = settings
-        self.circuits = self.settings.make_circuits()
+        self.circuits = self.settings.make_circuits(self.settings.circuit_seed)
         self.results: List[Dict[str, Any]] = []
 
     def get_cost(self) -> Dict[str, int]:
@@ -178,12 +180,42 @@ class Calibrator:
             parameters that performed best.
         """
         best_improvement_factor = 1.0
+        num_circuits = len(self.settings.circuit_types)
+
         strategy = DefaultStrategy
-        for circuit in results:
-            for result in circuit["mitigated_values"]["ZNE"]["results"]:
-                if result["improvement_factor"] > best_improvement_factor:
-                    best_improvement_factor = result["improvement_factor"]
-                    strategy = result["strategy"]
+
+        def filter_on_strategy(
+            result: Dict[str, Dict[str, Dict[str, Any]]], strategy_id: int
+        ) -> Dict[str, Dict[str, Dict[str, Any]]]:
+            res = result
+            res["mitigated_values"]["ZNE"]["results"] = [
+                res["mitigated_values"]["ZNE"]["results"][strategy_id]
+            ]
+            return result
+
+        for strategy_id in range(
+            len(results[0]["mitigated_values"]["ZNE"]["results"])
+        ):
+            strategy_group = []
+            for c in range(num_circuits):
+                result_copy = deepcopy(results)
+                strategy_group.append(
+                    filter_on_strategy(result_copy[c], strategy_id=strategy_id)
+                )
+                self.compute_improvements(strategy_group)
+
+            if (
+                strategy_group[0]["mitigated_values"]["ZNE"]["results"][0][
+                    "improvement_factor"
+                ]
+                > best_improvement_factor
+            ):
+                strategy = strategy_group[0]["mitigated_values"]["ZNE"][
+                    "results"
+                ][0]["strategy"]
+        if strategy is None:
+            warnings.warn("None of the improvement factors were > 1")
+            return DefaultStrategy
         return strategy
 
     def run(self) -> None:
--- a/mitiq/calibration/calibrator.py
+++ b/mitiq/calibration/calibrator.py
@@ -305,7 +305,6 @@ def execute_with_mitigation(
     input circuit, via the application of the best mitigation strategy, as
     determined by calibration.
 
-
     Args:
         circuit: The input circuit to execute.
         executor: A Mitiq executor that executes a circuit and returns the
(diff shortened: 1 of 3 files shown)


Misty-W:
I know there are some files on this PR that really "belong" to #1676, but I wanted to make sure this worked with those changes. Maybe I should've made a PR to the #1676 branch. Hopefully #1676 gets merged soon, and then I can clean this up with a rebase.

mitiq/calibration/settings.py
```diff
@@ -177,6 +177,7 @@ def __init__(
         num_qubits: int,
         circuit_depth: int,
         strategies: List[Dict[str, Any]],
+        circuit_seed: Optional[int] = None,
```
andreamari, line 180: `circuit_seed` should be added to the docstring of `Settings`

mitiq/calibration/calibrator.py
```diff
@@ -70,7 +72,7 @@ def __init__(
             else None
         )
         self.settings = settings
-        self.circuits = self.settings.make_circuits()
+        self.circuits = self.settings.make_circuits(self.settings.circuit_seed)
```
andreamari, line 75:
```suggestion
        self.circuits = settings.make_circuits(settings.circuit_seed)
```
This is a bit simpler and should be equivalent, if I am not wrong.

mitiq/calibration/settings.py
```diff
@@ -186,10 +187,14 @@ def __init__(
...
+        self.circuit_seed = circuit_seed
 
-    def make_circuits(self) -> List[BenchmarkProblem]:
+    def make_circuits(
+        self, seed: Optional[int] = None
+    ) -> List[BenchmarkProblem]:
         """Generate the circuits to run for the calibration experiment.
-
+        Args:
+            seed(optional): a random seed for (mirror) circuit generation.
```
andreamari, line 197:
```suggestion
            seed(optional): A random seed for (mirror) circuit generation.
```

mitiq/calibration/tests/test_calibration.py
```diff
@@ -151,6 +151,47 @@ def test_compute_improvements_modifies_IF():
...
+    )
+    cal = Calibrator(execute, test_strategy_settings)
+    cal.run()
+    strategy = cal.best_strategy(cal.results)
+    strategy_params = strategy.as_dict()
+
+    if cal.results[-1]["best_improvement_factor"] > 1:
+        assert strategy_params != {}
+    else:
+        assert strategy_params == {}
```
andreamari, line 192: This is a bit hard to understand. 
E.g. Why we have `[-1]` in `cal.results[-1]`?
Moreover we can be more explicit e.g.:
 instead of  `assert strategy_params != {}` we could use `assert strategy.technique.name = "zne"` and
instaed of `assert strategy_params == {}` we could use `assert strategy.technique.name = "raw"` .

mitiq/calibration/calibrator.py
```diff
@@ -178,12 +180,49 @@ def best_strategy(self, results: List[Dict[str, Any]]) -> Strategy:
...
+        num_circuits = len(self.settings.circuit_types)
+
         strategy = DefaultStrategy
-        for circuit in results:
-            for result in circuit["mitigated_values"]["ZNE"]["results"]:
-                if result["improvement_factor"] > best_improvement_factor:
-                    best_improvement_factor = result["improvement_factor"]
-                    strategy = result["strategy"]
+
+        def filter_on_strategy(
```
andreamari, line 187: Docstring is missing

mitiq/calibration/tests/test_calibration.py
andreamari: Moreover, can we also check in this test or in a new one, if the best strategy is _really_ the best strategy?

E.g. we could manually compute the expectation value for all strategies (without the calibrator) and check that the optimal strategy gives indeed the best result. Or something along this line.
If the number of things to manually compute is too large, we can make a new test with just 2 strategies and a single circuit type.

andreamari (commented):
Thanks @Misty-W!  My main suggestion is to test that the best strategy is the best strategy :-)

mitiq/calibration/calibrator.py
```diff
@@ -178,12 +180,49 @@ def best_strategy(self, results: List[Dict[str, Any]]) -> Strategy:
...
+            ):
+                best_improvement_factor = strategy_group[0][
+                    "mitigated_values"
+                ]["ZNE"]["results"][0]["improvement_factor"]
+                strategy = strategy_group[0]["mitigated_values"]["ZNE"][
+                    "results"
+                ][0]["strategy"]
+        self.results.append(
+            {"best_improvement_factor": best_improvement_factor}
+        )
```
natestemen, line 221: It seems like this is working (!), but there are a few things I'm hesitant about here.

1. We shouldn't rely on the ordering of the strategies to be the same in each `["mitigated_values"]["ZNE"]["results"]` chain.
2. `compute_improvements` was written to be called on the output of `run_circuits`. Does `strategy_group` have the same structure?
3. There are some really long chains going into this dictionary here, which I think is impossible to avoid, but I think setting some as variables will be helpful for readability and maintainability. E.g. on line 197, that variable we are using `range` on could be called `num_strategies`.
4. I think this loop could be simplified quite a bit by using a dictionary to keep track of things.
```py
strategy_improvement_factors: Dict[Strategy, List[float]] = defaultdict(list) # not sure what the type would be exactly
for circuit in results:
    for result in circuit["mitigated_values"]["ZNE"]["results"]:
        strategy = result["strategy"]
        value = result["mitigated_value"] # or could be improvement factor
        strategy_improvement_factors[strategy].append(value)
```
Then once you have `strategy_improvement_factors` you can write a function to compute the average IF, or do it as in `compute_improvements` if you store the ideal and noisy values in the array as well. What do you think of this approach?

natestemen (changes requested):
Nice work Misty, thanks for diving into this code!

I think there are some simplifications we can make, but I think this is on the right track!

mitiq/calibration/calibrator.py
Misty-W: Wasn't sure if one was needed here given the simplicity of the function, but since it came up I'll add one.

mitiq/calibration/settings.py
Misty-W: Agreed, done.

mitiq/calibration/calibrator.py
Misty-W: This could also work, but I changed`make_circuits` so that it no longer has the seed as an explicit argument.

mitiq/calibration/settings.py
Misty-W: Moved docstring for `seed` to `Settings` since it is no longer an explicit argument of `make_circuits`.

mitiq/calibration/calibrator.py
```diff
@@ -178,12 +180,49 @@ def best_strategy(self, results: List[Dict[str, Any]]) -> Strategy:
...
+
+        if strategy is None:
+            warnings.warn("None of the improvement factors were > 1")
+            return DefaultStrategy
         return strategy
```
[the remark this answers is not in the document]
Misty-W: This works, thanks Nate!

mitiq/calibration/settings.py
```diff
@@ -186,10 +187,14 @@ def __init__(
         self.circuit_types = circuit_types
         self.num_qubits = num_qubits
         self.circuit_depth = circuit_depth
+        self.circuit_seed = circuit_seed
 
-    def make_circuits(self) -> List[BenchmarkProblem]:
+    def make_circuits(
+        self, seed: Optional[int] = None
+    ) -> List[BenchmarkProblem]:
```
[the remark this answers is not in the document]
Misty-W: Thanks for bringing this up, it was unintentional. Now using `self.circuit_seed` inside `make_circuits`.

mitiq/calibration/calibrator.py
```diff
@@ -178,12 +180,49 @@ def best_strategy(self, results: List[Dict[str, Any]]) -> Strategy:
...
-                    strategy = result["strategy"]
+
+        def filter_on_strategy(
+            result: Dict[str, Dict[str, Dict[str, Any]]], strategy_id: int
+        ) -> Dict[str, Dict[str, Dict[str, Any]]]:
+            res = result
+            res["mitigated_values"]["ZNE"]["results"] = [
+                res["mitigated_values"]["ZNE"]["results"][strategy_id]
+            ]
+            return result
```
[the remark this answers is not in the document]
Misty-W: I noticed this too when writing the docstring for `filter_on_strategy`, which should return `res` instead of `result`. I've changed it accordingly.

mitiq/calibration/settings.py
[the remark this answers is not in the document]
Misty-W: Moved this arg docstring to `Settings` and removed the `optional` tag.

mitiq/calibration/tests/test_calibration.py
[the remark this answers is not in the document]
Misty-W: I tried testing with a particular strategy, but there is some randomness that causes others to be selected sometimes.

mitiq/calibration/calibrator.py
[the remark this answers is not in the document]
natestemen: I think even if it returns `res`, `results` will be mutated due to how python stores objects. E.g.
```py
>>> d = {"one": [1,2,3]}
>>> def foo(d):
...     a = d
...     a["one"] = [d["one"][1]]
...     return a
...
>>> foo(d)
{'one': [2]}
>>> d
{'one': [2]}
```

If you take this route, maybe make a copy of results.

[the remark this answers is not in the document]
Misty-W: Input to `filter_on_strategy` is already a copy.

mitiq/calibration/tests/test_calibration.py
Misty-W: `cal.results[-1]` refers to the last entry in the list of results, which is where the average improvement factors (averaged over the circuits) are stored.

I've updated the existing test to `assert strategy.technique.name == "zne"` and `assert strategy.technique.name == "raw"`

mitiq/calibration/tests/test_calibration.py
```diff
@@ -151,6 +151,45 @@ def test_compute_improvements_modifies_IF():
...
+        circuit_seed=1,
+    )
+    cal = Calibrator(execute, test_strategy_settings)
+    cal.run()
+    strategy = cal.best_strategy(cal.results)
+
+    if cal.results[-1]["best_improvement_factor"] > 1:
+        assert strategy.technique.name == "ZNE"
+    else:
+        assert strategy.technique.name == "RAW"
```
natestemen, line 190:
```suggestion
    for _ in range(5):
        cal = Calibrator(execute, test_strategy_settings)
        cal.run()
        strategy = cal.best_strategy(cal.results)

        if cal.results[-1]["best_improvement_factor"] > 1:
            assert strategy.technique.name == "ZNE"
        else:
            assert strategy.technique.name == "RAW"
```

natestemen (approved):
Andrea and I did some testing and feel good that this is doing the thing we expect it to. Nice work!

The fact that this is a relatively complicated change, for a relatively conceptually simple task, indicates that our underlying data storage is probably too complicated. We came up with some ideas for a refactor, but think it's best done in another PR in the next milestone. I've added one suggested change just to make sure the strategy selection isn't too subject to randomness. Not totally necessary however.

Issue #1639, opened 2022-12-19
Support (publicly) executors with bitstring output


natestemen:
Despite executors supporting functions which output acceptable forms of bit strings, there is not public method to get bit string output once the user-defined function is converted to an executor. We do have the [`executor._run`](https://github.com/unitaryfund/mitiq/blob/8aa81350f4550421f9a4cb3bbdda46fa80f4f0df/mitiq/executor/executor.py#L224-L229) function, but this is still semi-private. We should upgrade this function, or write a new one as suggested in https://github.com/unitaryfund/mitiq/pull/1614#discussion_r1052466906 to allow for this functionality.

nathanshammah:
Community call decision is to upgrade the current function.

PR #1684, opened 2023-01-26
upgrade `_run` to a public function

natestemen:
## Description

Upgrade the `_run` function which, like the commonly used [`evaluate`](https://github.com/unitaryfund/mitiq/blob/ecfbfa5481b54ffdc82273a0a3da5251308da50c/mitiq/executor/executor.py#L119) method, runs all of the circuits, but does not post-process the results. The capability to run a series of circuits, without post-processing is potentially useful to users. This PR makes it "public", which includes the function in the API-doc.

resolves https://github.com/unitaryfund/mitiq/issues/1639

Also fixed a few minor docstrings issues while I was poking around here.

### DIFF
--- a/mitiq/calibration/calibrator.py
+++ b/mitiq/calibration/calibrator.py
@@ -235,7 +235,7 @@ def convert_to_expval_executor(
         bitstring_to_measure = bitstring
 
     def expval_executor(circuit: cirq.Circuit) -> float:
-        raw = cast(MeasurementResult, executor._run([circuit])[0]).result
+        raw = cast(MeasurementResult, executor.run([circuit])[0]).result
         raw = cast(List[List[int]], raw)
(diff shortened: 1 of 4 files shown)


Misty-W (approved):
LGTM!

natestemen:
@Misty-W only thing I was thinking was should the signature of the `run` function match that of `evaluate`, or is it fine if the API's are slightly different?

Misty-W:
> @Misty-W only thing I was thinking was should the signature of the `run` function match that of `evaluate`, or is it fine if the API's are slightly different?

I think it's okay that the signature isn't the same because `run` doesn't use `Observable`- it's running the circuits in which the Observable has already been taken into account. We can double-check this in the Mitiq meeting tomorrow.

andreamari (approved):
LGTM!

PR #1688, opened 2023-01-27
Use closest probability distribution in REM

### LINKED ISSUE #1660: Improve the handling of negative probabilities in REM
## Issue Description

At the moment we clip negative probabilities when running REM. 
An alternative solution, typically used in the literature, is to return the closest positive probability distribution.

I.e., after getting the adjusted quasi-prob distribution $p' = C^{-1} p$, where $C$ is the confusion matrix, instead of returning
$p_{\rm positive} = clip(p) / |clip(p)|_1$ it is probably better to return something the optimal solution of:

 $$ \min_{p_{\rm positive}} \||p' - p_{\rm positive}\||_1$$

See e.g. https://arxiv.org/abs/2006.14044

## Additional comment

In principle, if we are just interested in computing expectation values, one can also work with the non-positive distribution $p' = C^{-1} p$. Actually this would be the optimal approach since if we avoid any post-processing we get an unbiased estimate of the expectation value, similar to PEC.  See e.g. https://arxiv.org/abs/2006.14044 . Not sure however how easy is the implementation of this approach in Mitiq.

andreamari:
Fixes #1660 .

When the REM-mitigated distribution of measurement results has negative elements, we look for the closet positive distribution instead of just clipping. For details, see #1660.

### License

- [x ] I license this contribution under the terms of the GNU GPL, version 3 and grant Unitary Fund the right to provide additional permissions as described in section 7 of the GNU GPL, version 3.


amirebrahimi (approved):
Nice improvement.

PR #1693, opened 2023-01-30
REM: Write-up third section of docs

### LINKED ISSUE #1453: REM: Write the third sub-section of the user guide
  - What additional options are available when using REM?

**Considerations**:

nickdgardner:
## Description

Resolves #1453 by filling in the third section of the REM docs ("What additional options are available when using REM").

---

### License

### DIFF
--- a/docs/source/guide/rem-3-options.md
+++ b/docs/source/guide/rem-3-options.md
@@ -15,13 +15,13 @@ kernelspec:
 
 ## Overview
 The main options the user has when using REM concern how to specify the inverse
-confusion matrix that is used to mitigate errors in the raw measurement (or "readout") results. Currently Mitiq does not implement methods for estimating readout-error confusion matrices (a form of measurement noise calibration and therefore a device specific task), so the user must provide enough information to allow Mitiq to construct one. As described below, Mitiq's options support the differing levels of information a user may have about the readout-error characteristics of their device. After the confusion matrix has been constructed, the remaining steps of standard REM are straightforward (compute the pseudoinverse of the confusion matrix and then apply this to the raw measurement results). 
+confusion matrix that is used to mitigate errors in the raw measurement (or "readout") results. Currently Mitiq does not implement methods for estimating readout-error confusion matrices (which is a form of measurement noise calibration and therefore a device specific task), so the user must provide enough information to allow Mitiq to construct one. As described below, Mitiq's options support the differing levels of information a user may have about the readout-error characteristics of their device. After the confusion matrix has been constructed, the remaining steps of standard REM are straightforward (compute the pseudoinverse of the confusion matrix and then apply this to the raw measurement results). 
 
 
 
 ## What is a Confusion Matrix?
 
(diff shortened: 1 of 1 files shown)


andreamari (approved):
This is great Nick! Very clear and informative.
Minor inline suggestions, plus an optional suggestion: What about adding a note after the first code cell in which we say something like:

_Hey, we computed the confusion matrix A to better explain the idea, but to apply REM you actually need to compute just its inverse `A_pinv`._ 

You can add a note with

{note}backthick backthick backthick

NOTE TEXT

backthick backthick backthick

nickdgardner:
Great, thanks for this feedback Andrea. Should be good to go now!

docs/source/guide/rem-3-options.md
```diff
@@ -12,3 +12,153 @@ kernelspec:
...
+A_pinv = generate_inverse_confusion_matrix(2, p0=0.1, p1=0.2)
+
+print(f"Confusion matrix:\n{A}\n")
+print(f"Column-wise sums:\n{A.sum(axis=0)}\n\n")
+print(f"Inverse confusion matrix:\n{A_pinv}")
+```
+
+### `generate_tensored_inverse_confusion_matrix(num_qubits, confusion_matrices)`
+This function allows for any stochastic noise model (i.e. any model representable classically by a $2^{n} \times 2^{n}$ confusion matrix). The general form of an overall $n$-qubit confusion matrix $A$ is a factorization $A = A^{(1)} \otimes \dots \otimes A^{(k)}$ into the tensor product of $k$ smaller confusion matrices (supplied by the user in `confusion_matrices`), one for each subgroup in a partition of the $n$ qubits into $k$ smaller subgroups. The factorization encodes the assumption that there are $k$ independent/uncorrelated noise processes affecting the $k$ disjoint subgroups of qubits (possibly of different sizes), but within each subgroup noise may be correlated between qubits in that subgroup. This model includes the simplest noise model above as the special case where $k=n$ and each of the $n$ single-qubit subgroups has the same confusion matrix $A_1$:
```
amirebrahimi: I'd recommend using "subspaces" rather than subgroups, but don't have a strong opinion about it.

amirebrahimi (approved):
Great job on this and better than what I originally had in mind to write, so glad that you picked this up in the end.

docs/source/guide/rem-3-options.md
andreamari: I agree that _subgroups_ can be confusing due to its mathematical meaning. 
My two cents: _subspaces_ is good, especially if there is a conceptual link to an underlying Hilbert space. If we are referring to subsets of qubits, we can simply use _subsets_.

nickdgardner: Sounds good! I went back and forth on using _subgroup_ because of this. It's helpful to know it sticks out to others--definitely confusing now that I think about it. It seems like _subsets_ would be the right choice here because the connection to _subspaces_ is not crucial in this context (maybe more for "What is the Theory behind REM"). I'll change it to _subsets_ for now, but if _subspaces_ seems better we can revise to that.

Issue #1699, opened 2023-02-01
Re-organize calibration data in better data structures


andreamari:
The data that we need to store in a `mitiq.calibration.Calibrator`  is given by the following three things:

- A list of `BenchmarkProblem`s, where  a `BenchmarkProblem` wraps a circuit and an ideal distribution of counts;
- A list of error mitigation `Strategy`s, where a `Strategy` wraps a technique and its related parameters;
- A list of `BenchmarkResult`'s, where a `BenchmarkResult` contains the error mitigated result for each pair (`BenchmarkProblem`, `Strategy`) taken from the Cartesian product of the previous two lists.   (Note the class `BenchmarkResult` doesn't exist yet, I am just using it here to just to give it an explicit name. If we think it's useful, we can define this class too.)

## Issue Description

The main limitation that we have now is that all the data mentioned above is now stored in a highly nested dictionary, with multiple  hard-to-remember text keys. More or less something like this:
```
[
            {
                "circuit_info": {
                    "circuit_depth": int,
                    "ideal_distribution": {str: float},
                    "num_qubits": int,
                    "two_qubit_gate_count": int,
                    "type": str,
                },
                "ideal_value": float,
                "noisy_value": float,
                "mitigated_values": {
                    str: {
                        "improvement_factor": Or(float, None),
                        "results": [
                            {
                                "factory": str,
                                "scale_factors": [float],
                                "scale_method": str,
                                "mitigated_value": float,
                                "improvement_factor": float,
                                "strategy": Strategy,
                            }
                        ],
                    }
                },
            }
        ]
    )
```

This is hard to maintain and to use. In particular, the computation of the optimal `Strategy` based on all the data (or a filtered subset of it) becomes very involved. 

## Proposed Solution

To be discussed. 
E.g. a simple `List[Tuple[BenchmarkProblem, Strategy, BenchmarkProblem]]` would already be a significant improvement. But maybe there is a better solution.

PR #1706, opened 2023-02-06
Calibration results data structure refactor

natestemen:
## Description

This PR changes the data structure of how the calibration experiment results are stored. This is done to simplify calculation that needs to be done with this data. The `best_strategy` computation also has been shifted from using improvement factors, to absolute errors for simplicity. This should be relatively straightforward to change if we wish, however.

fixes https://github.com/unitaryfund/mitiq/issues/1699

### Manual Testing Performed
- Calibration tests

### Questions
1. I've added a numerical `id` to each of the strategy instances. This helps keep track of which results comes from which strategies. As it stands now, this is sometimes needed since we separate the two, to later bring them back together. Alternatively, we could rely on the ordering of the `calibration.results` attribute, and use indices instead. I'm slightly leaning towards the `id`'s, but it does add a small amount of complexity to the code that is non-obvious to understand why it's there.
2. Mypy is complaining about the following line: https://github.com/unitaryfund/mitiq/blob/ec7cf9a0398dd34a28a8afb00e60143e5f30490b/mitiq/calibration/calibrator.py#L229-L234 I _cannot_ figure out how one is "supposed" to resolve this error[^1]. We can ignore it, of course, but I'd really like to understand what mypy sees as the problem first. Any ideas what's going on?

[^1]:`mitiq/calibration/calibrator.py:234: error: Argument 1 to "Executor" has incompatible type "Callable[[Circuit], float]"; expected "Callable[[Union[Union[Circuit, Program, Any, Any, Any], Sequence[Union[Circuit, Program, Any, Any, Any]]]], Any]"  [arg-type]`


andreamari:
> 2\. Mypy is complaining about the following line: https://github.com/unitaryfund/mitiq/blob/ec7cf9a0398dd34a28a8afb00e60143e5f30490b/mitiq/calibration/calibrator.py#L229-L234

I noticed that it works if you annotate the expval executor as follows:
`def expval_executor(circuit: QPROGRAM) -> float: `
but I am not sure if it is much better compared to just using `#  type ignore`

Thinking about this, the reason is probably that the annotation of the `executor` argument in the `Executor` class is not technically rigorous since it expects a function able to execute **all types** in `QPROGRAM`, but what we actually mean is a function that can execute **at least one** of the types in `QPROGRAM`.

mitiq/calibration/calibrator.py
```diff
@@ -92,208 +94,118 @@ def get_cost(self) -> Dict[str, int]:
...
-            circuit_info = problem.problem_summary_dict()
-            expvals.append(
-                {
-                    "circuit_info": circuit_info,
-                    "noisy_value": noisy_value,
-                    "ideal_value": distribution[bitstring_to_measure],
-                    "mitigated_values": mitigated_values,
+                    "mitigated": mitigated_value,
+                    "noisy": noisy_value,
+                    "ideal": problem.largest_probability(),
```
andreamari, line 126: This information is already in `problem` is it necessary to copy it also in `results`?

mitiq/calibration/calibrator.py
```diff
@@ -92,208 +94,118 @@ def get_cost(self) -> Dict[str, int]:
...
-                improvement_factor = abs(noisy_value - ideal_value) / sqrt(
-                    regularizing_epsilon
-                    + len(mitigated_values)
-                    * sum(
-                        (mitigated_value - ideal_value) ** 2
-                        for mitigated_value in mitigated_values
-                    )
-                )
-                di["improvement_factor"] = improvement_factor
+    def filter_problems(
```
andreamari, line 135: this and other methods right below don't have a docstring

mitiq/calibration/calibrator.py
```diff
@@ -92,208 +94,118 @@ def get_cost(self) -> Dict[str, int]:
...
+        errors = self.compute_errors(results)
 
-        return strategy
+        best_strategy_id = min(errors, key=errors.__getitem__)
+        return self.get_strategy(best_strategy_id)
 
     def run(self) -> None:
         results = self.run_circuits()
-        self.compute_improvements(results)
         self.results = results
```
andreamari, line 182: `run()` seems now redundant with `run_circuits`

mitiq/calibration/calibrator.py
```diff
@@ -38,7 +27,20 @@
...
+from mitiq.calibration.settings import (
+    Settings,
+    Strategy,
+    BenchmarkProblem,
+    DefaultStrategy,
+)
+
+
+@dataclass
+class Result:
```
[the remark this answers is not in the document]
natestemen: Agree, this wasn't the best name. Fixed by further restructuring, and eliminating the need for this class to begin with.

mitiq/calibration/calibrator.py
```diff
@@ -92,208 +94,118 @@ def get_cost(self) -> Dict[str, int]:
...
         """Runs all the circuits required for calibration.
 
         Returns:
             A collection of experimental results along with a summary of each
             :class:`BenchmarkProblem` that was run.
         """
-        expvals = []
+        results = []
+
+        strategies = self.settings.make_strategies()
```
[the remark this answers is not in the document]
natestemen: Good point, let's be consistent.

mitiq/calibration/calibrator.py
natestemen: This is true, but it's often convenient to have all of mitigated/noisy/ideal values in the same place. When this is not the case, it's much less convenient to compute errors due to the fact that we are not storing entire `Strategy` and `BenchmarkProblem` instances, but rather unique IDs. Thus we would have to grab the `BenchmarkProblem` ID, get the actual instance, get the ideal value, and execute it to obtain the noisy value.

While I dislike duplicating data, I think this makes the code more readable, and maintainable.

mitiq/calibration/calibrator.py
```diff
@@ -92,208 +94,118 @@ def get_cost(self) -> Dict[str, int]:
...
             circuit.append(cirq.measure(circuit.all_qubits()))
 
-            expval_executor, bitstring_to_measure = convert_to_expval_executor(
-                self.executor, distribution
+            bitstring_to_measure = problem.most_likely_bitstring()
+            expval_executor = convert_to_expval_executor(
+                self.executor, bitstring_to_measure
             )
+
             noisy_value = expval_executor.evaluate(circuit)[0]
```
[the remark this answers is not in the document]
natestemen: See comment at https://github.com/unitaryfund/mitiq/pull/1706#discussion_r1106570878

mitiq/calibration/calibrator.py
natestemen: Further restructuring resulted in these getting removed.

mitiq/calibration/calibrator.py
natestemen: Indeed it is now redundant. I've gone ahead and renamed `run_circuits -> run` where `run` stores the data internal to the class.

mitiq/calibration/settings.py
```diff
@@ -189,6 +197,14 @@ def __init__(
...
+        self.problem_dict: Dict[int, BenchmarkProblem] = {}
+
+    def get_strategy(self, strategy_id: int) -> Strategy:
+        return self.strategy_dict[strategy_id]
+
+    def get_problem(self, problem_id: int) -> BenchmarkProblem:
+        return self.problem_dict[problem_id]
 
     def make_circuits(self) -> List[BenchmarkProblem]:
         """Generate the circuits to run for the calibration experiment.
```
andreamari, line 210:
```suggestion
    def make_problems(self) -> List[BenchmarkProblem]:
        """Generate the benchmark problems for the calibration experiment.
```
I think it's better to apply this change locally instead of committing this code suggestion. A few tests and other parts of the code need to be updated accordingly.

mitiq/calibration/calibrator.py
```diff
@@ -92,208 +120,68 @@ def get_cost(self) -> Dict[str, int]:
...
-            A collection of experimental results along with a summary of each
-            :class:`BenchmarkProblem` that was run.
-        """
-        expvals = []
+    def run(self) -> None:
+        """Runs all the circuits required for calibration."""
         for problem in self.circuits:
-            circuit, distribution = problem["circuit", "ideal_distribution"]
+            circuit = problem.circuit
             circuit.append(cirq.measure(circuit.all_qubits()))
```
andreamari, line 127: What about adding measurements in `make_problems()`, i.e. in the current `make_circuits()`?
Any disadvantage?

andreamari (commented):
Very clean refactoring! Almost approved. See inline comments.

andreamari:
Something I forgot to mention in the review: unit tests for `ExperimentResults` are missing.

mitiq/calibration/calibrator.py
natestemen: Hmmm we totally can do that, but I think there are some minor tradeoffs.

1. [pro] The thing we are storing is the thing we are actually going to run, so no modifications are needed.
2. [pro] The user doesn't have to guess how circuits are being measured if they inspect them.
3. [con] Computing `two_qubit_gate_count` is slightly more cumbersome, and `depth` becomes somewhat ambiguous. E.g. does `-X-H-M-` have depth 2, or 3?

Aesthetically, I have a preference for storing the circuit without measurements, but that's only very minor. WDYT?

mitiq/calibration/calibrator.py
```diff
@@ -38,7 +26,43 @@
...
 )
-from mitiq.calibration.settings import Settings, Strategy, DefaultStrategy
+from mitiq.calibration.settings import (
+    Settings,
+    Strategy,
+    BenchmarkProblem,
+)
+
+
+class ExperimentResults:
```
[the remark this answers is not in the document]
natestemen: Great idea. I ended up starting the arrays with `np.nan`, and then detect if there are any NaN values in the array once the problem processing has completed ([`9d26c3e`](https://github.com/unitaryfund/mitiq/pull/1706/commits/9d26c3e7b53a5333de12cf2ea4259f218cff0ecc)).

mitiq/calibration/calibrator.py
andreamari: Fine to keep measurements in `run()`

mitiq/calibration/calibrator.py
```diff
@@ -38,7 +26,52 @@
...
+        store the results."""
+        self.mitigated[strategy.id, problem.id] = mitigated_val
+        self.noisy[strategy.id, problem.id] = noisy_val
+        self.ideal[strategy.id, problem.id] = ideal_val
+
+    def squared_errors(self) -> npt.NDArray[np.float32]:
+        """Returns an array of squared errors, one for each (strategy, problem)
+        pair."""
+        return (self.ideal - self.mitigated) ** 2
+
```
andreamari, line 66: Here we can add a new method `check_if_full()` with something like:

```
 if np.isnan(self.results.mitigated).any():
          raise ValueError( "One or more results from calibration experiments are missing. ")
```
such that we can call it in multiple places:

- Inside `best_strategy_id()`. Which I think is the most critical place to avoid the risk of false best strategies. 

- And, if you think it's useful, somewhere in the `Calibrator` . E.g., where you have currently placed this check.

Moreover, it's probably good to have a unit **test** for triggering this error.

andreamari (approved):
Great PR!
I left a small suggestion about the error message. Feel free to merge once fixed.

Misty-W (approved):
PR looks great, nice work @natestemen!

Agree with Andrea's comment about the `check_if_full()` method and test. Otherwise good to go from my side!  🎉

Issue #1659, opened 2023-01-04
Improve OperationRepresentation and NoisyOperation objects


andreamari:
## Issue Description

`OperationRepresentation` and `NoisyBasis` objects are necessary to apply PEC with Mitiq.
The aim of this issue is to improve their structure such that PEC can be applied by users in a simpler way. 

## Example of improvements

- Currently the user needs to define `NoisyOperation` objects,   a `NoisyBasis` and an `OperationRepresentation` .
Is it possible to reduce the number of objects the user needs to initialize? E.g. is the current `NoisyBasis` object useful? Can we replace it with a simple list of `NoisyOperation`(s) ?

- Currently, the elements of the noisy basis are stored as a Python `set`.  The rationale is reasonable but this has some important disadvantages: it would require the definition of an hash function for noisy operations (currently not defined) and, most importantly, we don't have a fixed ordering which would be quite useful feature.
 For example, one often deals with a quasi-prob distribution (coefficients) or with a prob. distribution (`OperationRepresentation.distribution()`), but it is not clear what is the corresponding ordered list of noisy operations. Moreover it is quite confusing that, running the code twice, can produce two OperationRepresentation objects that are equivalent but have different distributions due to the different ordering.

PR #1712, opened 2023-02-10
Refactoring of NoisyOperation, NoisyBasis, OperationRepresentation

andreamari:
Fixes: #1659 

### License


andreamari:
@Misty-W  and/or @natestemen , I am fighting with some silly style issues but all other tests are passing. So feel free to review this PR when you have some time.

mitiq/pec/types/__init__.py
```diff
@@ -14,7 +14,7 @@
 # along with this program.  If not, see <https://www.gnu.org/licenses/>.
 
 from mitiq.pec.types.types import (
-    NoisyBasis,
     NoisyOperation,
     OperationRepresentation,
+    NoisyBasis,
 )
```
natestemen, line 20: Does changing the order of imports affect anything, or just reordering to show level of importance?

natestemen (commented):
Looks pretty good to me, but I don't think I fully understand why switching from `{ op: coeff }` to `[ ops ]` and `[ coeffs ]` is better. I know in the ticket you mentioned having an ordering is nice, but seeing as there is a tight connection between the operations and coefficients, it does seem natural to me to have them stored in a dictionary. Can you just provide a little more detail on what this achieves?

In terms of future plans, we maybe remove `NoisyBasis` altogether in the release after the current one?

andreamari:
> Looks pretty good to me, but I don't think I fully understand why switching from `{ op: coeff }` to `[ ops ]` and `[ coeffs ]` is better. I know in the ticket you mentioned having an ordering is nice, but seeing as there is a tight connection between the operations and coefficients, it does seem natural to me to have them stored in a dictionary. Can you just provide a little more detail on what this achieves?
> 
Some motivations:

- [Main motivation] Using `NoisyOperation`s as a keys of a dictionary, requires having a reliable and well-tested `__hash__()` function for the `NoisyOperation` class. We don't have one, so we have been using the default hash based on the memory id which is quite unsafe. 
- From a practical usage point of view, it's probably a matter of subjective preference. I like separated lists for representing objects in a vector space since one typically fixes the basis with a fixed ordering and, once this is done, one can represent objects via simple real vectors which can be easily transformed, optimized, sampled, etc... .  
- An example use case is when solving a numerical optimization problem to find the optimal coefficients. The optimal solution is a real vector of coefficients, but to what `NoisyOperations` do the coefficients correspond? This ambiguity is absent with the new approach. 
- Note that `NoisyOperation.basis_expansion` still exists and can be used to couple coefficients and basis elements.

mitiq/pec/sampling.py
```diff
@@ -97,7 +104,10 @@ def sample_sequence(
         noisy_op, sign, _ = operation_representation.sample(
             random_state  # type: ignore
         )
-        sequences.append(noisy_op.native_circuit)
+        cirq_circ = noisy_op.circuit
+        cirq_circ = cirq_circ.transform_qubits(qubit_map)
+        native_circ = convert_from_mitiq(cirq_circ, native_type)
```
andreamari, line 109:
```suggestion
        if operation_representation.is_qubit_dependent:
            native_circuit = noisy_op.native_circuit
        else:
            cirq_circ = noisy_op.circuit
            cirq_circ = cirq_circ.transform_qubits(qubit_map)
            native_circ = convert_from_mitiq(cirq_circ, native_type)
```

mitiq/pec/types/__init__.py
andreamari: This is because I first removed `NoisyBasis` from Mitiq and then re-added a deprecated version of the `NoisyBasis` later. Reordering was not intentional.

Misty-W: To confirm, is the idea of re-adding just to warn the user that it's deprecated?

andreamari: Yes. We can completely remove it in 1 or 2 milestones.

Technically, this is not a _deprecation_ intended in the standard way, which would imply warning but still supporting `NoisyBasis`. In our case we just drop support for it and we give some info to the user with an error message.

natestemen (approved):
:shipit:

PR #1713, opened 2023-02-14
Simplify noisy operation

andreamari:
First step of the more general scope of #1712.
Since this is a self-contained step, I opened a new PR such that it can be reviewed more easily.
 
If this will be merged to master, the diff of #1712 will be much simpler to follow.

This PR simplifies the structure of the `NoisyOperation` class without compromising the workflow of PEC. 

Motivations:

- Short code is easier to maintain and less subject to bugs. 
- Some removed methods have inconsistent effects when applied to non-Cirq circuits. 
- Lightweight objects may speedup PEC. (Not sure how strong this effect is).
- Simpler code helps the more general refactoring process of #1712. 

### License

### DIFF
--- a/mitiq/pec/types/tests/test_types.py
+++ b/mitiq/pec/types/tests/test_types.py
@@ -692,3 +692,17 @@ def test_equal_method_of_representations():
         basis_expansion={noisy_xop_b: 0.7, noisy_zop_b: 0.5},
     )
     assert rep_a != rep_b
+
+
+def test_noisy_operation_copy():
+    cnot_qiskit = qiskit.QuantumCircuit(2)
+    cnot_qiskit.cnot(0, 1)
+    matrix = np.random.rand(16, 16)
+    op_a = NoisyOperation(cnot_qiskit, matrix)
+    op_b = op_a.copy()
+    assert np.allclose(op_a.channel_matrix, op_b.channel_matrix)
+    assert op_a.num_qubits == op_b.num_qubits
+    assert op_a.qubits == op_b.qubits
+    assert op_a.circuit == op_b.circuit
+    assert op_a.native_circuit == op_b.native_circuit
+    assert str(op_a) == str(op_b)
--- a/mitiq/pec/types/tests/test_types.py
+++ b/mitiq/pec/types/tests/test_types.py
@@ -692,17 +692,3 @@ def test_equal_method_of_representations():
         basis_expansion={noisy_xop_b: 0.7, noisy_zop_b: 0.5},
     )
     assert rep_a != rep_b
-
-
-def test_noisy_operation_copy():
-    cnot_qiskit = qiskit.QuantumCircuit(2)
-    cnot_qiskit.cnot(0, 1)
-    matrix = np.random.rand(16, 16)
-    op_a = NoisyOperation(cnot_qiskit, matrix)
-    op_b = op_a.copy()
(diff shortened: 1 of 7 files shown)


mitiq/pec/types/tests/test_types.py
```diff
@@ -27,81 +28,61 @@
...
-def test_init_with_gate():
-    ideal_gate = cirq.Z
+
+def test_init_with_cirq_circuit():
     real = np.zeros(shape=(4, 4))
-    noisy_op = NoisyOperation.from_cirq(ideal_gate, real)
+    noisy_op = NoisyOperation(zcirq, real)
     assert isinstance(noisy_op._circuit, cirq.Circuit)
 
-    assert _equal(
```
Misty-W, line 37: Not quite sure why this statement was removed, can you clarify?

mitiq/pec/types/types.py
```diff
@@ -57,239 +56,62 @@ def __init__(
...
-                new_qubits = list(new_qubits)
-
-            qubit_map = dict(zip(base_qubits, new_qubits))
-            new_circuit = base_circuit.transform_qubits(lambda q: qubit_map[q])
-
-            noisy_ops.append(NoisyOperation(new_circuit, channel_matrix))
-
-        return noisy_ops
-
-    def extend_to(
```
Misty-W, line 200: May I ask why this was deleted? I do see that it was only "used "in tests- is it unlikely to be useful to Mitiq users?

mitiq/pec/types/types.py
```diff
@@ -356,23 +178,6 @@ def add(self, *basis_elements: Sequence["NoisyOperation"]) -> None:
                 )
             self._basis_elements.add(noisy_op)
 
-    def extend_to(self, qubits: Sequence[List[cirq.Qid]]) -> None:
```
Misty-W, line 359: Same question as in `NoisyOperation`, although here it's less important since I believe the plan is to remove `NoisyBasis` eventually.

Misty-W (commented):
Nice cleanup on this PR @andreamari! I have a few questions around the removal of some of the methods and tests, but once those are cleared up it should be good to go.

mitiq/pec/types/types.py
andreamari: It doesn't seem to be needed in practice. Moreover it has some problems:

- It takes as input cirq qubits, so it's not multi-platform.
- It creates a mismatch between the `self._native_circuit` (not extended) and self.circuit (extended).
- I don't think `channel_matrix` is correctly extended in a consistent way.

We could fix these problems, but is it worth the effort and the risk of creating subtle bugs?

mitiq/pec/types/types.py
andreamari: Exactly, `NoisyBasis` will be removed anyway in the next PR.

mitiq/pec/types/types.py
Misty-W: Makes sense, agree to remove.

mitiq/pec/types/tests/test_types.py
Misty-W: Ah, okay I see it now. I think it the multi-line formatting threw me off, all the better to simply it.

Misty-W (approved):
LGTM!

PR #1714, opened 2023-02-15
[calibration] usability improvements

natestemen:
## Description

This PR has some minor improvements to the usability of calibration. In particular,

1. adds a logging option when running the calibration experiments. I envision this being useful to give a user more immediate understanding of what's happening, and how techniques are performing. The results can be seen [below](#logging).
2. adds a check if the user would like to proceed when using `execute_with_mitigation` if the `calibrator` object has not yet been run. Running automatically without the user's permission could be unintended if they are using a real backend.
3. Allows calibrators to run the specified experiments multiple times. Previously running `calibrator.run()` twice would cause an error (and an unhelpful one at that). If a user would like to refresh their experiments, I don't see a reason why we shouldn't allow them to.

### Testing performed
Tests, and some further user flow ideas in a notebook.

#### Logging example output
```
Running ghz circuit using: ['ZNE', 'LinearFactory', [1.0, 3.0, 5.0], 'fold_global']
✅ ideal: 0.50	noisy: 0.58	mitigated: 0.50
Running ghz circuit using: ['ZNE', 'RichardsonFactory', [1.0, 2.0, 3.0], 'fold_gates_at_random']
❌ ideal: 0.50	noisy: 0.58	mitigated: 0.38
Running ghz circuit using: ['ZNE', 'RichardsonFactory', [1.0, 3.0, 5.0], 'fold_gates_at_random']
❌ ideal: 0.50	noisy: 0.58	mitigated: 0.72
Running ghz circuit using: ['ZNE', 'LinearFactory', [1.0, 2.0, 3.0], 'fold_gates_at_random']
❌ ideal: 0.50	noisy: 0.58	mitigated: 0.61
Running ghz circuit using: ['ZNE', 'LinearFactory', [1.0, 3.0, 5.0], 'fold_gates_at_random']
✅ ideal: 0.50	noisy: 0.58	mitigated: 0.53
```

### Technical note

The base branch for this PR is that of https://github.com/unitaryfund/mitiq/pull/1706, and should only be merged once that has been merged.

### DIFF
--- a/mitiq/calibration/calibrator.py
+++ b/mitiq/calibration/calibrator.py
@@ -14,6 +14,7 @@
 # along with this program.  If not, see <https://www.gnu.org/licenses/>.
 
 from typing import Callable, Dict, Optional, Union, cast
+import warnings
 
 import cirq
 import numpy as np
@@ -56,12 +57,26 @@ class ExperimentResults:
         ideal_val: float,
         noisy_val: float,
         mitigated_val: float,
+        log: bool = False,
     ) -> None:
         """Add a single result from a (Strategy, BenchmarkProblem) pair and
         store the results."""
         self.mitigated[strategy.id, problem.id] = mitigated_val
         self.noisy[strategy.id, problem.id] = noisy_val
         self.ideal[strategy.id, problem.id] = ideal_val
+        if not log:
+            return
+        mitigated_better = abs(ideal_val - mitigated_val) < abs(
+            ideal_val - noisy_val
+        )
+        performance = "✅" if mitigated_better else "❌"
+        print(
+            f"Running {problem.type} circuit using:",
+            list(strategy.to_dict().values()),
+        )
+        print(
+            f"{performance} ideal: {ideal_val:.2f}\tnoisy: {noisy_val:.2f}\tmitigated: {mitigated_val:.2f}"
+        )
 
     def is_missing_data(self) -> bool:
         """Method to check if there is any missing data that was expected from
@@ -148,7 +163,7 @@ class Calibrator:
             "ideal_executions": ideal,
         }
 
-    def run(self) -> None:
+    def run(self, log: bool = False) -> None:
         """Runs all the circuits required for calibration."""
         for problem in self.circuits:
             circuit = problem.circuit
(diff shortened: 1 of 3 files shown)


mitiq/calibration/calibrator.py
```diff
@@ -49,10 +50,26 @@ def add_result(
...
         self.noisy[strategy.id, problem.id] = noisy_val
         self.ideal[strategy.id, problem.id] = ideal_val
+        if not log:
+            return
+        mitigated_better = abs(ideal_val - mitigated_val) < abs(
+            ideal_val - noisy_val
+        )
+        performance = "✅" if mitigated_better else "❌"
+        print(
+            f"Running {problem.type} circuit using:",
```
andreamari, line 65: Technically we are not "running" the circuit. We completed the execution and we are showing the result.

mitiq/calibration/calibrator.py
```diff
@@ -61,9 +78,19 @@ def smallest_error(self) -> int:
...
         strategy_errors = np.sum(errors, axis=1)
         strategy_errors /= len(errors[0])
-        strategy_id = np.argmin(strategy_errors)
+        strategy_id = int(np.argmin(strategy_errors))
         return strategy_id
 
+    def is_empty(self) -> bool:
+        return np.allclose(
+            self.mitigated, np.zeros((self.num_strategies, self.num_problems))
+        )
```
andreamari, line 87: This method may change depending on the code suggestions given in the parent PR.
In particular we probably need to check if all elements of the data matrices have been filled with associated results.

mitiq/calibration/calibrator.py
```diff
@@ -120,10 +147,13 @@ def get_cost(self) -> Dict[str, int]:
...
 
-    def run(self) -> None:
+    def run(self, log: bool = False) -> None:
         """Runs all the circuits required for calibration."""
+        if not self.results.is_empty():
+            self.results.reset_data()
+
         for problem in self.circuits:
-            circuit = problem.circuit
+            circuit = problem.circuit.copy()
```
andreamari, line 156: This may be unnecessary based on suggestions given in the parent PR.

andreamari:
Very cool!
For the moment, I only gave preliminary comments. After #1706 is merged on master and after master is merged here, I'll make a proper review.

mitiq/calibration/calibrator.py
natestemen: Technically true, but this code would only ever run _just_ after executing a circuit. Do you think a word change is in order, or would you like the logging code moved to the `run` function?

andreamari: The current message can be confusing in some cases: e.g. when there is a long queue in the hardware or when a circuit execution fails.

Both solutions you suggested would work for me.

natestemen: Reworded in [2af1b88](https://github.com/unitaryfund/mitiq/pull/1714/commits/2af1b88a05610ce387af63fe6cff16f61d7f9e2c).

andreamari (approved):
Good for me! Thanks

PR #1718, opened 2023-02-19
REM: Update the fifth section of the docs

### LINKED ISSUE #1455: REM: Write the fifth sub-section of the user guide
  - What is the theory behind REM?

**Considerations**:

amirebrahimi:
## Description

Resolves #1455:
- Adds a small portion about postselection to the intro.
- Moves the explanation about confusion matrices from the third to the fifth section of the docs.
- Fills out the theory section by providing info about applying inverse confusion matrices to the probability distribution

---

### License

### DIFF
--- a/docs/source/guide/rem-1-intro.md
+++ b/docs/source/guide/rem-1-intro.md
@@ -88,7 +88,7 @@ print(f"Error without mitigation: {error:.3}")
 
 ## Apply Postselection
 The simplest version of readout error mitigation that we could do is to postselect specific bitstrings using 
-{mod}`mitiq.rem.post_select`. We observe that the circuit in our example is symmetric with respect to interchanging the
+{mod}`mitiq.rem.post_select`. This strategy is only applicable if, from the structure of the problem, there is some kind symmetry that we can assume and therefore enforce by post-selection. For example, we observe that the circuit in our example is symmetric with respect to interchanging the
 two qubits. Assuming we are given the promise that the ideal result must be a unique bitstring, such a bitstring must
 be symmetric with respect to a bit swap. Therefore, a possible error mitigation strategy is to keep only the results
 where the bits match.
--- a/docs/source/guide/rem-1-intro.md
+++ b/docs/source/guide/rem-1-intro.md
@@ -88,8 +88,10 @@ print(f"Error without mitigation: {error:.3}")
 
 ## Apply Postselection
 The simplest version of readout error mitigation that we could do is to postselect specific bitstrings using 
-{mod}`mitiq.rem.post_select`. In our case we know that we should only have bitstrings of `[1,1]`, so we could discard
-any results that don't match.
+{mod}`mitiq.rem.post_select`. We observe that the circuit in our example is symmetric with respect to interchanging the
+two qubits. Assuming we are given the promise that the ideal result must be a unique bitstring, such a bitstring must
+be symmetric with respect to a bit swap. Therefore, a possible error mitigation strategy is to keep only the results
+where the bits match.
 
 ```{code-cell} ipython3
 from mitiq.rem import post_select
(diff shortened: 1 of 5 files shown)


docs/source/guide/rem-1-intro.md
```diff
@@ -86,8 +86,31 @@ error = abs((ideal_value - noisy_value)/ideal_value)
 print(f"Error without mitigation: {error:.3}")
 ```
 
+## Apply Postselection
+The simplest version of readout error mitigation that we could do is to postselect specific bitstrings using 
+{mod}`mitiq.rem.post_select`. In our case we know that we should only have bitstrings of `[1,1]`, so we could discard
+any results that don't match.
```
andreamari, line 92: Good idea adding a subsection on post-selection!

However, I am not sure about the post-selection example. Post-selecting on the expected ideal result is a too strong type of selection :smile: . Basically in this case the q. computer is completely useless. 

What about something a bit more realistic as this (feel free to rewrite in your own words): 

_We observe that the circuit in our example is symmetric with respect to interchanging the two qubits. Assuming we are given the promise that the ideal result must be a unique bitstring, such a bitstring must be symmetric with respect to a bit swap. Therefore, a possible error mitigation strategy is to discard all results with different bits._

If you agree with this, the associated code cell must be changed accordingly.

docs/source/guide/rem-5-theory.md
```diff
@@ -13,12 +13,50 @@ kernelspec:
...
+Pr(00|01) & Pr(01|01) & Pr(10|01) & Pr(11|01) \\
+Pr(00|10) & Pr(01|10) & Pr(10|10) & Pr(11|10) \\
+Pr(00|11) & Pr(01|11) & Pr(10|11) & Pr(11|11)
+\end{bmatrix}
+$$
+
+
+where $Pr(ij|kl)$ is the probability of observing state $|ij\rangle$ when measuring true state $|kl\rangle$. 
+
+The most straightforward way to empirically estimate a device's full confusion matrix is to go through all the measurement basis states, and for each one $|u\rangle$, repeatedly prepare-then-measure $|u\rangle$ and record the histogram of observed outcomes. This histogram, normalized to give a probability distribution, is an estimate for the $u$th column of the confusion matrix A (i.e. the distribution of measurement outcomes when the true state is $|u\rangle$). Since the number of basis states scales exponentially with the number of qubits $n$, estimating the full confusion matrix in this way requires $O(2^n)$ samples and is therefore only practical for small devices.
```
andreamari, line 46: I know this part was copied and pasted but I just noticed a small inconsistency.

We print a 4x4 matrix where elements $P(ij|kl)$ with equal input state $kl$ are placed on the same row.

but after few lines we say that 
`This histogram, normalized to give a probability distribution, is an estimate for the $u$th column`.

This seems inconsistent. From a quick check I would guess that it's all good if we just transpose the printed 4x4 matrix.
Whatever convention we use in the docs for the entries of A, it should correspond to the convention implicitly assumed by `mitiq.rem` and possibly to what other authors used in the previous literature.

docs/source/guide/rem-5-theory.md
```diff
@@ -13,12 +13,50 @@ kernelspec:
...
+
+## Applying the inverse confusion matrix to results
+
+With our raw measurement results we convert our bitstrings into a probability vector, $p$, representing our empirical
+probability distribution. After obtaining the psuedoinverse matrix $A^{+}$, we can apply it to our empirical probability
+distribution to obtain an adjusted *quasi*-probability distribution, $p' = A^{+} p$, which could possibly be 
+non-positive. As such, we want to find the closest *positive* probability distribution {cite}`Bravyi_2021` to our
+empirical probability distribution:
+
+ $$ p'' = \min_{p_{\rm positive}} \||p' - p_{\rm positive}\||_1$$
```
andreamari, line 60:
```suggestion
 $$ p'' = \min_{p_{\rm positive}} \|p' - p_{\rm positive}\|_1$$
```
`\|`  produces two bars.

andreamari (commented):
I think the theory section is perfect! Clear and straight to the point. 
See minor comments below.

docs/source/guide/rem-5-theory.md
amirebrahimi: I made the changes, but please double-check.

docs/source/guide/rem-1-intro.md
amirebrahimi: Thanks, I incorporated these changes, but made a slight tweak to put the language in the affirmative since that is how you specify the selector w/ post_select.

docs/source/guide/rem-1-intro.md
```diff
@@ -86,8 +86,37 @@ error = abs((ideal_value - noisy_value)/ideal_value)
 print(f"Error without mitigation: {error:.3}")
 ```
 
+## Apply Postselection
+The simplest version of readout error mitigation that we could do is to postselect specific bitstrings using 
+{mod}`mitiq.rem.post_select`. We observe that the circuit in our example is symmetric with respect to interchanging the
```
andreamari, line 91:
```suggestion
The simplest version of readout error mitigation that we could do is to postselect specific bitstrings using 
{mod}`mitiq.rem.post_select`. This strategy is only applicable if, from the structure of the problem, there is some kind symmetry that we can assume and therefore enforce by post-selection. For example, we observe that the circuit in our example is symmetric with respect to interchanging the
```

andreamari (approved):
LGTM! I left some further optional suggestions about post-selection.

The main problem with post-selection is not much the increasing sampling cost (this is true for almost all techniques), but it's the requirement of some symmetry that must be present in the physical problem or artificially encoded by some kind of error correction code. For a generic circuit, post-selection is not useful/applicable.

Issue #1597, opened 2022-11-17
Add W-state/QPE benchmarking circuits


natestemen:
We'd like to add (in [`mitiq/benchmarks`](https://github.com/unitaryfund/mitiq/tree/master/mitiq/benchmarks)) support for two new benchmarking circuits.
- **W state circuits**: $W_n = \frac{1}{\sqrt{n}} |10\cdots 00\rangle + |010\cdots0\rangle + \cdots + |0\cdots 01\rangle$. The circuit for creating these is described here: https://arxiv.org/abs/1807.05572
- **Quantum Phase Estimation**: circuits described on [wikipedia](https://en.wikipedia.org/wiki/Quantum_phase_estimation_algorithm).

nathanshammah:
+1 for W state circuits. It can be similar to the ghz state ones, https://mitiq.readthedocs.io/en/stable/apidoc.html#module-mitiq.benchmarks.ghz_circuits.

purva-thakre:
Does this need to be finished by the end of the current milestone ? If not, I can take it for the next one. 

My proposal is close to being accepted by the end of next week. So, I know I will have more time on my hand as well.

purva-thakre:
Sure ! I can unassign one for now because I got lost and stopped working on it. #746

 I haven't had the chance to look at what has been completed by you and Misty in the documentation for identity scaling. 

#1426 is a quick one but I was waiting for a comment on where to put the tutorial in the docs. 

#987 and #513 are two WIP proposals currently.

Misty-W:
hi @purva-thakre, still planning to work on this in this milestone?

purva-thakre:
Yep !

natestemen:
Removing this from the 0.25 milestone. A major portion of this ticket is now completed with #1723.

purva-thakre:
@natestemen @andreamari For the QPE circuit, is better to restrict the input U to be a `cirq.Gate` ? 

I was thinking of also allowing `np.array` and then using [`cirq.deconstruct_single_qubit_matrix_into_angles`](https://quantumai.google/reference/python/cirq/deconstruct_single_qubit_matrix_into_angles) to describe the unitary in terms of valid quantum gates. This does increase the size of the circuit. 

The other option would be to create a custom gate with the provided input array. I had tried this with the W-state circuit and noticed there were quite a few issues to fix when I created a controlled-U version of my custom gate.

andreamari:
[...]

PR #1723, opened 2023-02-24
W state Benchmarking Circuit

purva-thakre:
One of the benchmarking circuits in #1597

### DIFF
--- a/mitiq/benchmarks/tests/test_w_state_circuits.py
+++ b/mitiq/benchmarks/tests/test_w_state_circuits.py
@@ -36,6 +36,8 @@ def test_bad_qubit_number():
 
 def test_w4_circuit():
     """Tests for a W-state of 3 qubits."""
+
+    # compare the circuits
     output_circuit = generate_w_circuit(4)
     qubits = cirq.LineQubit.range(4)
     correct_circuit = cirq.Circuit(
@@ -54,18 +56,20 @@ def test_w4_circuit():
     )
     assert _equal(output_circuit, correct_circuit)
 
+    # compare the state vector
     w4_state_vector = (
         cirq.Simulator()
         .simulate(output_circuit, initial_state=1000)
         .final_state_vector
     )
-    non_zero_final_vector = np.nonzero(w4_state_vector)
-    for i in range(len(non_zero_final_vector[0])):
-        assert np.isclose(w4_state_vector[non_zero_final_vector[0][i]], 0.5)
+    correct_final_state_vector = np.array([0, 0.5, 0.5, 0, 0.5, 0, 0, 0, 0.5, 0, 0, 0, 0, 0, 0, 0])
+    assert np.allclose(w4_state_vector, correct_final_state_vector)
 
 
 def test_w2_circuit():
     """Tests for W-state of 2 qubits."""
+
+    # compare the output circuit with expected circuit
     output_circuit = generate_w_circuit(2)
     qubits = cirq.LineQubit.range(2)
     correct_circuit = cirq.Circuit(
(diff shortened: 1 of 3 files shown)


mitiq/benchmarks/tests/test_w_state_circuits.py
```diff
@@ -0,0 +1,33 @@
...
+
+def test_w4_circuit():
+    output_circuit = W_circuit_linear_complexity(4)
+    qubits = cirq.LineQubit.range(4)
+    correct_circuit = cirq.Circuit(
+        cirq.ControlledGate(GRotationGate(1/4)).on(qubits[0], qubits[1]), cirq.CNOT(qubits[1], qubits[0]), 
+        cirq.ControlledGate(GRotationGate(1/3)).on(qubits[1], qubits[2]), cirq.CNOT(qubits[2], qubits[1]),
+        cirq.ControlledGate(GRotationGate(1/2)).on(qubits[2], qubits[3]), cirq.CNOT(qubits[3], qubits[2]),
+        )
+    assert _equal(output_circuit, correct_circuit, require_qubit_equality=True)
```
purva-thakre, line 33: This test fails because of the manually created `GRotation` gate. 

`_equal` is comparing object at some_location in both circuits.

mitiq/benchmarks/w_state_circuits.py
```diff
@@ -0,0 +1,85 @@
...
+from typing import Optional, Union
+import numpy.typing as npt
+import numpy as np
+import cirq
+
+from mitiq import QPROGRAM
+from mitiq.interface import convert_from_mitiq
+
+
+class GRotationGate(cirq.Gate):
```
andreamari, line 28: I think this gate equivalent to the existing Ry gate (https://quantumai.google/reference/python/cirq/Ry), up to a change of variable:
`cos(t/2) <--> p`
If so, it's better to use Ry instead of defining a custom gate.

purva-thakre: Thanks for catching this ! Defining the custom gate did cause quite a few issues. 

I think, the equivalence you have should be `cos(t/2) <--> SQRT(p)`

andreamari:
Hi @purva-thakre, is this ready for re-review?
When ready, please change the status of this PR from "draft" to "ready for review".

purva-thakre:
@andreamari Not yet. The paper has 2 circuits of different complexity. What I have so far is the circuit with linear complexity. 

I am trying to get the other one to work.

mitiq/benchmarks/w_state_circuits.py
```diff
@@ -0,0 +1,58 @@
...
+        raise ValueError("{} is invalid for the number of qubits. ", n_qubits)
+
+    qubits = cirq.LineQubit.range(n_qubits)
+    circuit = cirq.Circuit()
+
+    for i, j in zip(range(0, n_qubits), range(1, n_qubits)):
+        N = n_qubits - i
+        circuit.append(
+            cirq.Ry(rads=1 / N * np.pi).controlled().on(qubits[i], qubits[j])
+        )
```
purva-thakre, line 50: verify the relationship between p and t/2 in Ry gate

purva-thakre:
@natestemen @andreamari I am marking this ready for review only for the linear complexity circuit. 

I am going to create a separate issue for the log complexity circuits for both GHZ and W state circuits. The method described in the paper is  not that straightforward and I need some more time to think on generalizing them.

natestemen:
That sounds like a good idea! I don't think I even realized that there _were_ two types of W-state circuits!

andreamari:
> @natestemen @andreamari I am marking this ready for review only for the linear complexity circuit.
> 
> I am going to create a separate issue for the log complexity circuits for both GHZ and W state circuits. The method described in the paper is not that straightforward and I need some more time to think on generalizing them.

I agree. For the purposes of Mitiq, having the linear complexity circuits is more than enough.  I would consider the  implementation of the log complexity alternative as optional / lower priority.

andreamari (commented):
Thanks @purva-thakre. This looks good!
Beyond my inline comments, I would suggest to add 2 more tests:
- A test with a different number of qubits (e.g. 2 or 3).
- A test which checks all possible return types. No need to check the results for all different frontends. But we should check the function runs without errors if we call e.g. generate_w_state(3, "qiskit").

purva-thakre:
> we should check the function runs without errors if we call e.g. generate_w_state(3, "qiskit").

I noticed there was an existing test like this for GHZ circuits. I used the same test for W state circuits.

andreamari (approved):
LGTM!

PR #1725, opened 2023-02-24
Remove PDF build option for docs

natestemen:
## Description

As discussed at the Mitiq meeting (and discovered by @Misty-W), RTD "latest" has been failing for the past 2-3 weeks. The error is related to PDF builds of the docs which we think are not being used by anyone. For that reason we have decided to drop this part of the docs build system, and in particular in our RTD build to workaround this error.

I've removed the code, and associated mentions of PDFs in our docs.

### Testing

As for testing, I'm not sure how to test that this will fix our problems on latest. Any ideas?

### DIFF
--- a/.github/workflows/gh-release.yml
+++ b/.github/workflows/gh-release.yml
@@ -1,11 +1,11 @@
 on:
   push:
     tags:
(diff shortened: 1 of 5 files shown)


Misty-W (approved):
LGTM!

PR #1729, opened 2023-02-27
Fix mypy errors on #1728

Misty-W:
## Description
Type check failing on `master` due to missing default on `benchmark.get("circuit_depth")`

See https://github.com/unitaryfund/mitiq/actions/runs/4286350547/jobs/7465806334#step:5:13

---

### License

### DIFF
--- a/mitiq/calibration/settings.py
+++ b/mitiq/calibration/settings.py
@@ -213,6 +213,8 @@ class Settings:
             circuit_type = benchmark["circuit_type"]
             num_qubits = benchmark["num_qubits"]
             depth = benchmark.get("circuit_depth")
+            if not depth:
+                depth = num_qubits
             if circuit_type == "ghz":
                 circuit = generate_ghz_circuit(num_qubits)
                 ideal = {"0" * num_qubits: 0.5, "1" * num_qubits: 0.5}
--- a/mitiq/calibration/settings.py
+++ b/mitiq/calibration/settings.py
@@ -212,9 +212,8 @@ class Settings:
         for i, benchmark in enumerate(self.benchmarks):
             circuit_type = benchmark["circuit_type"]
             num_qubits = benchmark["num_qubits"]
-            depth = benchmark.get("circuit_depth")
-            if not depth:
-                depth = num_qubits
+            # Circuits only generated when depth is specified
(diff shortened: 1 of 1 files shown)


Misty-W:
#1728 failed type checking because it was missing a conditional statement checking whether `depth` was `None` and giving it a value (`num_qubits`).

natestemen (commented):
So my understanding of the merge queue as being a thing that waits to ensure CI passes is totally wrong. I'm not seeing any benefit thus far.

natestemen:
Also, in the future please add links to failing builds (and even lines from the logs), so they can be found and referenced quickly. It helps add context to PRs.

E.g. it's good to link the failing master build on this PR: https://github.com/unitaryfund/mitiq/actions/runs/4286350547/jobs/7465806334#step:5:13

mitiq/calibration/settings.py
```diff
@@ -213,6 +213,8 @@ def make_problems(self) -> List[BenchmarkProblem]:
             circuit_type = benchmark["circuit_type"]
             num_qubits = benchmark["num_qubits"]
             depth = benchmark.get("circuit_depth")
+            if not depth:
+                depth = num_qubits
```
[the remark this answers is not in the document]
Misty-W: I had forgotten that about `dict.get`. I've implemented the solution of setting the default as -1.

mitiq/calibration/settings.py
```diff
@@ -212,7 +212,8 @@ def make_problems(self) -> List[BenchmarkProblem]:
         for i, benchmark in enumerate(self.benchmarks):
             circuit_type = benchmark["circuit_type"]
             num_qubits = benchmark["num_qubits"]
-            depth = benchmark.get("circuit_depth")
+            # Circuits only generated when depth is specified
```
natestemen, line 215: Sorry if this is a nitpick, but I don't think this comment is correct. Circuits are generated regardless of whether we have the default passed in `dict.get`. We need the default so that the type-checker is happy.

Misty-W: Good point, I changed this to `# Set default to return correct type`.

natestemen (approved):
Looks good! Thanks for taking care of this.

Reminder to use descriptive commit messages in the future :)

PR #1744, opened 2023-03-10
 Consider id as slack in ddd circuit mask

### LINKED ISSUE #1708: Consider identity gates as idle moments when computing idle windows for DDD.
## Issue Description

DDD works by first processing the input circuit to detect all idle windows, i.e., all the empty holes in the circuit where DDD sequences can be inserted. 

The proposal of this issue is to identify identity gates applied to a qubit as idle moments. In the current situation an identity gate is considered as an actual gate and therefore Mitiq considers that moment as "occupied".

## Proposed Solution

- Skip identity gates when building the circuit mask matrix in `ddd.generate_circuit_mask()`. 
- Note that `insert_ddd_sequences()` may need to be changed accordingly.
- This is a small but breaking change. So make sure that it doesn't create big unexpected problems. A good way to check this is having a look at what happens to CI tests and DDD docs examples after the change. 

## Motivation

Two main motivations:

- For some frontends (e.g. Qiskit), adding identities (or delay gates) is often the only way to artificially create idle windows (without creating barriers of 2-qubit gates). So it would be good to consider sequences of identities as proper idle windows.

- DDD adds many identities to create a delay space between the sequence gates. If we implement this feature, we can recursively call  `insert_ddd_sequences()` to also fill such delay spaces with additional sequences.

Aaron-Robertson:
## Description

Fixes #1708.

---

### License

### DIFF
--- a/mitiq/ddd/insertion.py
+++ b/mitiq/ddd/insertion.py
@@ -151,11 +151,13 @@ def _insert_ddd_sequences(
                     {LineQubit(0): qubits[row_index]}
                 )
                 for idx, op in enumerate(ddd_sequence.all_operations()):
-                    circuit_with_ddd._moments[
-                        moment_idx + idx
-                    ] = circuit_with_ddd._moments[
-                        moment_idx + idx
-                    ].with_operations(
+                    moment = circuit_with_ddd[moment_idx + idx]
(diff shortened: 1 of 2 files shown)


natestemen (approved):
Nice work Aaron! Great to have this bug fixed.

PR #1745, opened 2023-03-10
Compose REM + ZNE in Mitiq

Misty-W:
## Description

---

### License


docs/source/examples/combine_techniques.md
```diff
@@ -0,0 +1,107 @@
...
+```
+
+## Noise model and executor
+
+The noise in this example is a combination of depolarizing and readout errors, the latter of which are modeled as bit flips immediately prior to measurement. We use an [`Executor`](../guide/executors.md) function to execute the quantum circuit with the noise model applied.
+
+```{code-cell} ipython3
+def execute(circuit: cirq.Circuit, noise_level: float = 0.005, p0: float = 0.05) -> MeasurementResult:
+    """Execute a circuit and return a ``MeasurementResult``, with depolarizing noise of strength
+    ``noise_level`` and readout errors with probability ``p0``."""
```
natestemen, line 47: These few lines should be formatted so they don't overflow the text box:
Screenshot 2023-05-04 at 11 11 19 PM

One method to do this is copy and paste the code into a mitiq codefile and let `black` format it.

Also, the double ticks are not needed since it's being formatted as code. Maybe not even single backticks?

Finally, I think it flows a little better like this. If you want to mention the return type, probably best to add a new sentence.
```suggestion
def execute(circuit: cirq.Circuit, noise_level: float = 0.005, p0: float = 0.05) -> MeasurementResult:
    """Execute a circuit with depolarizing noise of strength ... and readout errors ...
```

docs/source/examples/combine_techniques.md
```diff
@@ -0,0 +1,107 @@
...
+
+```{code-cell} ipython3
+obs = Observable(PauliString("ZI"), PauliString("IZ"))
+noisy = raw.execute(circuit, execute, obs)
+```
+
+```{code-cell} ipython3
+from functools import partial
+ideal = raw.execute(circuit, partial(execute, noise_level=0, p0=0), obs)
+f"Error without mitigation: {abs(ideal - noisy) :.5f}"
```
natestemen, line 73: Woah... I didn't know you could put strings/f-strings directly into the code cell like this and have them be executed and returned. Cool TIL. It does quote them in the return, however. 🤷

docs/source/examples/combine_techniques.md
```diff
@@ -0,0 +1,107 @@
...
+    jupytext_version: 1.14.1
+kernelspec:
+  display_name: Python 3 (ipykernel)
+  language: python
+  name: python3
+---
+
+# Composing techniques: Readout Error Mitigation and Zero Noise Extrapolation
+
+Noise in quantum computers can arise from a variety of sources, and sometimes applying multiple error mitigation techniques can be more beneficial than applying a single technique alone. Here we apply a combination of readout error mitigation (REM) and zero noise extrapolation (ZNE) to a randomized benchmarking (RB) task. In REM, the inverse transition / confusion matrices are generated and applied to the noisy measurement results. In ZNE, the expectation value of the observable of interest is computed at different noise levels, and subsequently the ideal expectation value is inferred by extrapolating the measured results to the zero-noise limit. More information on the [REM](../guide/rem.md) and [ZNE](../guide/zne.md) techniques can be found in the corresponding sections of the user guide.
```
natestemen, line 17: - I think we typically stylize mitigation techniques with capital letters, so Zero-Noise Extrapolation.
- I really like how you added references for further info about REM/ZNE, but I think it makes sense to move those links to the place they first appear in case a reader hasn't heard of them before. It would probably eliminate the need for the last sentence as well.
- I think there should only be one inverse confusion matrix, not many.

---

Since this paragraph is set on a single line, it makes it hard to make review comments accurately. Hopefully I was able to communicate the sections I think could be modified, but I think using a line-character limit, or a single sentence per line rule[^1] could be helpful.

[^1]: my preferred way.

docs/source/examples/combine_techniques.md
```diff
@@ -0,0 +1,107 @@
...
+```{code-cell} ipython3
+import cirq
+import numpy as np
+from mitiq.benchmarks import generate_rb_circuits
+from mitiq import MeasurementResult, Observable, PauliString, raw
+```
+
+## Task
+
+We will demonstrate using REM + ZNE on RB circuits, which are generated using Mitiq's built-in benchmarking function, `generate_rb_circuits`. More information on the RB protocol is available [here](https://learn.qiskit.org/course/quantum-hardware/randomized-benchmarking). In this example we use a two-qubit RB circuit with a Clifford depth (number of Clifford groups) of 5.
```
natestemen, line 34: Link to API-doc on `generate_rb_circuits` could be helpful: https://mitiq.readthedocs.io/en/stable/apidoc.html#mitiq.benchmarks.randomized_benchmarking.generate_rb_circuits

docs/source/examples/combine_techniques.md
```diff
@@ -0,0 +1,107 @@
...
+
+```{code-cell} ipython3
+circuit = generate_rb_circuits(2, 5)[0]
+```
+
+## Noise model and executor
+
+The noise in this example is a combination of depolarizing and readout errors, the latter of which are modeled as bit flips immediately prior to measurement. We use an [`Executor`](../guide/executors.md) function to execute the quantum circuit with the noise model applied.
```
natestemen, line 42: Hmm we aren't actually using the `Executor` class here, so maybe just something like this
```suggestion
The noise in this example is a combination of depolarizing and readout errors, the latter of which are modeled as bit flips immediately prior to measurement. We use an [executor function](../guide/executors.md) to run the quantum circuit with the noise model applied.
```

docs/source/examples/combine_techniques.md
```diff
@@ -0,0 +1,107 @@
...
+    simulator = cirq.DensityMatrixSimulator()
+
+    result = simulator.run(circuit, repetitions=10000)
+    bitstrings = np.column_stack(list(result.measurements.values()))
+    return MeasurementResult(bitstrings)
+```
+
+## Observable
+
+We then obtain a unmitigated expectation value of the `ZI` + `ZZ` observable. In the ideal case the result is 2, but as we will see, the unmitigated result is impacted by depolarizing and readout errors.
```
natestemen, line 63: I think this part needs a little more motivation. Just something short like
> we now have the ability to run circuits, and we wish to measure an expecation value. In this example we take the following observable...

Also you can use math mode to write out the operator $Z\otimes I + I \otimes Z$ or just $ZI + IZ$.

docs/source/examples/combine_techniques.md
```diff
@@ -0,0 +1,107 @@
...
+```
+
+Finally, we apply a combination of REM and ZNE. REM is applied first to minimize the impact of measurement errors on the extrapolated result in ZNE.
+
+```{code-cell} ipython3
+combined_executor = zne.mitigate_executor(rem_executor, observable=obs)
+
+combined_result = combined_executor(circuit)
+f"Error with REM + ZNE: {abs(ideal - combined_result) :.5f}"
+```
```
natestemen, line 105: Wow cool to see this has such a small error compared to unmitigated! I noticed above it says the ideal expectation value is 2, but it looks like we're approaching 0 as the ideal value.

docs/source/examples/combine_techniques.md
```diff
@@ -0,0 +1,107 @@
...
+
+We can see that, in this example, the combination of REM and ZNE is more effective in mitigating errors than either technique alone.
```
natestemen, line 107: I would maybe just expand this a little bit to summarize what we did.

1. Defined an executor which returns measurement statistics
2. Defined an observable we want to measure
3. Created a new executor by applying REM first, and ZNE on the readout error mitigated values.

docs/source/examples/combine_techniques.md
natestemen, line 1: Since we might have more tutorials in the future which combine techniques, maybe we call this `combine_ZNE_REM` or something to reflect the specificity?

natestemen (commented):
Nice work Misty! I think this will be a very popular example. Cool to see it get flushed out after our initial work with Andrea!

docs/source/examples/combine_techniques.md
```diff
@@ -0,0 +1,107 @@
...
+from functools import partial
+ideal = raw.execute(circuit, partial(execute, noise_level=0, p0=0), obs)
+f"Error without mitigation: {abs(ideal - noisy) :.5f}"
+```
+
+Next we generate the inverse confusion matrix and apply readout error mitigation (REM). More information on generating the inverse confusion matrix is available in the [REM theory](../guide/rem-5-theory.md) section of the user guide. 
+
+```{code-cell} ipython3
+from mitiq import rem
+p0 = p1 = 0.05
```
andreamari, line 80:
```suggestion
from mitiq import rem

p0 = p1 = 0.05
```

docs/source/examples/combine_techniques.md
```diff
@@ -0,0 +1,107 @@
...
+We then obtain a unmitigated expectation value of the `ZI` + `ZZ` observable. In the ideal case the result is 2, but as we will see, the unmitigated result is impacted by depolarizing and readout errors.
+
+```{code-cell} ipython3
+obs = Observable(PauliString("ZI"), PauliString("IZ"))
+noisy = raw.execute(circuit, execute, obs)
+```
+
+```{code-cell} ipython3
+from functools import partial
+ideal = raw.execute(circuit, partial(execute, noise_level=0, p0=0), obs)
```
andreamari, line 72:
```suggestion
from functools import partial

ideal = raw.execute(circuit, partial(execute, noise_level=0, p0=0), obs)
```

andreamari (commented):
Seems all good to me apart from a quite relevant problem, the combined error is larger than the ZNE error:

image

This is not consistent with our final comment: `We can see that, in this example, the combination of REM and ZNE is more effective in mitigating errors than either technique alone.`

This could be the effect of a random fluctuation of the results which changes the errors each time we build the docs. A way to make results more deterministic is by using ZNE with odd `scale_factors` or with  `fold_global` .

docs/source/examples/combine_techniques.md
Misty-W: Some of the other Mitiq examples have the f-strings, but I've changed this example to have `print` statements instead to avoid the weird quote marks.

docs/source/examples/combine_techniques.md
Misty-W: All good points! I believe I've addressed them all.

docs/source/examples/combine_techniques.md
Misty-W: Nice catch, probably a copy/paste error.

docs/source/examples/combine_techniques.md
Misty-W: I changed the result statements back to expectation values instead of errors for clarity.

docs/source/examples/combine_techniques.md
Misty-W: Thanks for the reminder, I meant to do that earlier and forgot.

docs/source/examples/combine_techniques.md
Misty-W: Agreed. Hopefully the way I'm referencing it now works on RTD.

docs/source/examples/combine_techniques.md
Misty-W: I did expand the last sentence slightly, but I've left out the workflow summary as the example is relatively short and the user is unlikely to get "lost" in a limited number of steps.

Misty-W:
> Seems all good to me apart from a quite relevant problem, the combined error is larger than the ZNE error:
> 

Thanks for looking into this @andreamari! I ran the notebook several times locally, but the problem only became visible after pushing the changes, since an "unlucky" random circuit happened to be generated that time. I've since changed the folding method to `fold_global` and increased the depth of the circuit. With those changes the combined error was smaller than the ZNE error 12/12 times, so I think it's safe now.

docs/source/examples/combine_rem_zne.md
```diff
@@ -0,0 +1,121 @@
...
+import numpy as np
+from mitiq.benchmarks import generate_rb_circuits
+from mitiq import MeasurementResult, Observable, PauliString, raw
+```
+
+## Task
+
+We will demonstrate using REM + ZNE on RB circuits, which are generated using Mitiq's built-in benchmarking circuit generation function, {func}`.generate_rb_circuits()`. 
+More information on the RB protocol is available [here](https://learn.qiskit.org/course/quantum-hardware/randomized-benchmarking). 
+In this example we use a two-qubit RB circuit with a Clifford depth (number of Clifford groups) of 5.
```
andreamari, line 42:
```suggestion
In this example we use a two-qubit RB circuit with a Clifford depth (number of Clifford group elements) of 10.
```

andreamari (approved):
Looks good to me! 

> Thanks for looking into this @andreamari! I ran the notebook several times locally, but the problem only became visible after pushing the changes, since an "unlucky" random circuit happened to be generated that time. I've since changed the folding method to fold_global and increased the depth of the circuit. With those changes the combined error was smaller than the ZNE error 12/12 times, so I think it's safe now.

Thanks for looking into this! Sounds safe now

natestemen (approved):
Nice work! Good idea adding a reference to this in the REM guide as well.

I think a thumbnail should be added before shipping.

PR #1748, opened 2023-03-14
Make the Calibrator multi-platform

### LINKED ISSUE #1746: Ensure calibration workflow can run with any supported `QPROGRAM`
As it stands, the calibration workflow only works with Cirq. This is likely due to two reasons:

1. When generating test circuits in the [`Settings.make_problems`](https://github.com/unitaryfund/mitiq/blob/50a1b2c7574507fdb36ce20a552b141b42847102/mitiq/calibration/settings.py#L207) function, we use the default circuit type which is Cirq with the underlying assumption that the executor can run Cirq.
2. In the [`Calibrator.run`](https://github.com/unitaryfund/mitiq/blob/50a1b2c7574507fdb36ce20a552b141b42847102/mitiq/calibration/calibrator.py#L174) method, we append measurements using the Cirq interface.

Two solutions come to mind:
1. Track what frontend the user is using. Either by looking at typehints (personally, I do not like this approach), or by asking the user to pass a string specifying (also don't love this, but it might be the most simple). This method seems like it would fix point 1 above quite well, but I'm not sure how it would fix point 2 as after a cursory search we don't seem to have any sort of `QPROGRAM` agnostic `measure_all` function, which could be part of this solution.
2. Convert the users executor to accept any `QPROGRAM` type using our [`accept_any_qprogram_as_input`](https://github.com/unitaryfund/mitiq/blob/50a1b2c7574507fdb36ce20a552b141b42847102/mitiq/interface/conversions.py#L150) function.

This would be great to try and get done this (march) milestone, so we can make a launch of calibration without the asterisk that not all frontends are supported yet.

cc @andreamari @nathanshammah since y'all unearthed this problem. Which, by the way, thank you!

andreamari:
## Description

This is a possible solution to fix #1746. Me and @Misty-W drafted this in a live coding call :rocket: !

Tests are missing.
Design could be changed. E.g. I am not sure what is the best place to ask the user for  `frontend_type`. Now this is an argument of `Calibrator`.

### License

### DIFF
--- a/mitiq/calibration/calibrator.py
+++ b/mitiq/calibration/calibrator.py
@@ -16,7 +16,6 @@
 from typing import Callable, Dict, Optional, Union, cast
 import warnings
 
-import cirq
 import numpy as np
 import numpy.typing as npt
 
@@ -234,7 +233,7 @@ def convert_to_expval_executor(executor: Executor, bitstring: str) -> Executor:
         the most likely bitstring, according to the passed ``distribution``
     """
 
-    def expval_executor(circuit: cirq.Circuit) -> float:
+    def expval_executor(circuit: QPROGRAM) -> float:
         raw = cast(MeasurementResult, executor.run([circuit])[0])
         distribution = raw.prob_distribution()
         return distribution.get(bitstring, 0.0)
--- a/mitiq/calibration/calibrator.py
+++ b/mitiq/calibration/calibrator.py
@@ -133,6 +133,7 @@ class Calibrator:
         self,
         executor: Union[Executor, Callable[[QPROGRAM], QuantumResult]],
         settings: Settings,
+        frontend_type: str = "cirq",
         ideal_executor: Union[
             Executor, Callable[[QPROGRAM], QuantumResult], None
         ] = None,
@@ -141,17 +142,18 @@ class Calibrator:
         self.executor = (
             executor if isinstance(executor, Executor) else Executor(executor)
         )
+        self.frontend_type = frontend_type
         self.ideal_executor = (
             Executor(ideal_executor)
             if ideal_executor and not isinstance(ideal_executor, Executor)
             else None
         )
         self.settings = settings
-        self.circuits = settings.make_problems()
+        self.problems = settings.make_problems()
         self.strategies = settings.make_strategies()
         self.results = ExperimentResults(
             num_strategies=len(self.strategies),
-            num_problems=len(self.circuits),
+            num_problems=len(self.problems),
         )
 
     def get_cost(self) -> Dict[str, int]:
@@ -161,7 +163,7 @@ class Calibrator:
         Returns:
             A summary of the number of circuits to be run.
         """
-        num_circuits = len(self.circuits)
+        num_circuits = len(self.problems)
         num_options = len(self.settings.technique_params)
 
         noisy = num_circuits * num_options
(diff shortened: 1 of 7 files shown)


mitiq/calibration/calibrator.py
```diff
@@ -236,7 +264,7 @@ def convert_to_expval_executor(executor: Executor, bitstring: str) -> Executor:
         the most likely bitstring, according to the passed ``distribution``
     """
 
-    def expval_executor(circuit: cirq.Circuit) -> float:
+    def expval_executor(circuit: QPROGRAM) -> float:
```
natestemen, line 267: After this change, is the type ignore comment still needed in the return statement of this function?

mitiq/calibration/settings.py
```diff
@@ -81,6 +82,20 @@ def most_likely_bitstring(self) -> str:
...
+
+        Args:
+            circuit_type: The circuit type as a string.
+                For supported circuit types see mitiq.SUPPORTED_PROGRAM_TYPES.
+        Returns:
+            The converted circuit with final measurements.
+        """
+        circuit = self.circuit.copy()
+        circuit.append(cirq.measure(circuit.all_qubits()))
+        return convert_from_mitiq(circuit, circuit_type)
```
natestemen, line 97: I like this small, modular function! Nice job breaking stuff up into manageable bits.

mitiq/calibration/tests/test_calibration.py
```diff
@@ -84,8 +99,26 @@ def execute(circuit, noise_level=0.001):
...
+
+
+def non_cirq_execute(circuit):
+    # Ensure test circuits are converted to user's frontend by the Calibrator
+    assert not isinstance(circuit, cirq.Circuit)
+    circuit, circuit_type = convert_to_mitiq(circuit)
+    # Pennylane and Braket conversions discard measurements so we re-append
+    if circuit_type in ["braket", "pennylane"]:
+        circuit.append(cirq.measure(q) for q in circuit.all_qubits())
+    return execute(circuit)
```
natestemen, line 117: Ha! Really interesting executor to test across all frontend types. Very nifty.

Do we know why pennylane and braket discard measurements in the conversion? IIRC braket doesn't have measurement gates, but I'm unsure about pennylane.

mitiq/interface/mitiq_braket/conversions.py
```diff
@@ -122,6 +123,11 @@ def _translate_cirq_operation_to_braket_instruction(
     Raises:
         ValueError: If the operation cannot be converted to Braket.
     """
+    # Measurement gates do not exist in Braket.
+    if isinstance(op.gate, cirq_ops.MeasurementGate):
+        warn("Measurement gate removed when converting from Cirq to Braket.")
+        return []
```
natestemen, line 129: Confirming my recollection that braket doesn't have measurements. I wonder why. Do they just assume everything gets measured at the end?

Anyway, great to have this warning as well.

natestemen (commented):
Looking great! Thanks for taking this on Andrea, and @Misty-W! Just had a question or two, no major changes requested.

mitiq/calibration/calibrator.py
```diff
@@ -13,13 +13,13 @@
...
+from typing import Callable, Dict, Optional, Union, cast, Sequence
 import warnings
 
-import cirq
 import numpy as np
 import numpy.typing as npt
 
+import cirq
 from mitiq import (
     QPROGRAM,
```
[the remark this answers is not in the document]
andreamari: Is [this](https://github.com/unitaryfund/mitiq/pull/1748/commits/9c09685121c7f46b923d3572ab066750607e5a9d) better?

mitiq/calibration/settings.py
andreamari: Actually I wasn't sure if keeping this method or removing it, given that now we have a cirq_executor in the Calibrator.  So this method is now not essential.
I think it doesn't harm and it may help if the user wants to apply a more low-level and manual workflow.

mitiq/interface/mitiq_braket/conversions.py
andreamari: AFAIK, Braket assumes all qubits are measured.

PR #1751, opened 2023-03-15
Update DDD Qiskit tutorial with hardware experiments

Misty-W:
## Description

---

### License


andreamari:
Hi @Misty-W, the tutorial is great!
However we are skipping the execution of many code cells and so we risk to show untested code which can become outdated.
What about this:

- We keep USE_REAL_HARDWARE = True as it is now.
- Right after that code cell we add a new code cell with `USE_REAL_HARDWARE = False` but with the tag `hide-cell` such that it won't be shown in the docs.
- Run all cells. I.e. remove all `skip-execution` tags.
- Tag the code cell with the plot with `hide-output` such that the simulated plot won't be shown in the docs.
- Keep the hard-coded hardware plot.

For a list of myst tags see [here](https://myst-nb.readthedocs.io/en/latest/configuration.html#cell-tags).

docs/source/examples/ddd_on_ibmq_ghz.md
```diff
@@ -11,236 +11,253 @@ kernelspec:
...
+name: ddd-qiskit-ghz-plot-ibmq
+---
+Plot of the unmitigated and DDD-mitigated expectation values obtained from executing the corresponding circuits.
+```
+
+
+We can see that DDD improves the expectation value at each circuit depth, and the repeated XX sequence is the best at mitigating the errors
+occurring during idle windows. 
+This is expected because the idle windows are long compared to the length of the other operations in the circuit, so repeated pulses close
+together in time would more effectively decouple the qubits from the environment.
```
andreamari, line 263: I don't feel we have enough certainty on the reason why XX is doing better. According to your comment, repeated XX rule should **always** be better than non-repeated sequences. But don't think this is always true in practical experiments.

I propose we remove or we put it in a more speculative way.  E.g. something like: "A possible reason why repeated xx is more effective is ...." 


```suggestion

```

andreamari (approved):
Great, LGTM!
I left a minor comment.

PR #1755, opened 2023-03-17
Set default force_run_all as True in PEC

andreamari:
Set default force_run_all as True in PEC.
This is safer because the procedure used when `force_run_all = False` is too unreliable (due to conversions) and statistically questionable since it would require scaling the number of shots proportionally to the counts of equal circuits.

### License

### DIFF
--- a/mitiq/executor/executor.py
+++ b/mitiq/executor/executor.py
@@ -271,14 +271,12 @@ class Executor:
                 batch = to_run[i * step : (i + 1) * step]
                 self._call_executor(batch, **kwargs)
 
-        these_results = self._quantum_results[start_result_index:]
+        results = self._quantum_results[start_result_index:]
 
-        if force_run_all:
-            return these_results
-
-        # Expand computed results to all results using counts.
-        results_dict = dict(zip(collection.keys(), these_results))
-        results = [results_dict[key] for key in hashable_circuits]
+        if not force_run_all:
+            # Expand computed results to all results using counts.
+            results_dict = dict(zip(collection.keys(), results))
+            results = [results_dict[key] for key in hashable_circuits]
 
         return self._post_run(results)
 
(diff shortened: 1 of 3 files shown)


mitiq/rem/rem.py
```diff
@@ -87,7 +87,7 @@ def mitigate_executor(
         executor_obj = deepcopy(executor)
 
     def post_run(
-        self: MeasurementResult,
+        self: Executor,
```
natestemen, line 90: Is this change intended? It seems unrelated to the other changes, but maybe I'm missing something.

If it is unrelated, happy to keep it in this PR. I imagine you happened upon it while working on this.

mitiq/executor/executor.py
```diff
@@ -271,14 +271,12 @@ def run(
...
-        if force_run_all:
-            return these_results
-
-        # Expand computed results to all results using counts.
-        results_dict = dict(zip(collection.keys(), these_results))
-        results = [results_dict[key] for key in hashable_circuits]
+        if not force_run_all:
+            # Expand computed results to all results using counts.
+            results_dict = dict(zip(collection.keys(), results))
+            results = [results_dict[key] for key in hashable_circuits]
```
natestemen, line 279: It looks like this change modifies the return type in what is now the default case of `force_run_all=True` since now everything is going through `self._post_run`. Maybe `_post_run` doesn't do anything in the `force_run_all=True` case, though? Can you clarify slightly?

natestemen (approved):
Agree with the change, and happy to merge whenever. Just had two questions that I'd like answered just for my understanding's sake.

mitiq/rem/rem.py
andreamari: In fact it's unrelated. It's a small type fix, since `post_run()` is defined to be used as a method of `Executor`s and not of `MeasurementResult`s.  For some reason, this small inconsistency was never detected by `mypy`.

mitiq/executor/executor.py
andreamari: The scope of `_post_run` is to always post-process the measurement results, independently on the value of `force_run_all`. 

What you noticed is correct: before this PR,  `_post_run` was not called in the case `force_run_all=True`. But that was wrong :smile: !   It's a proof that using multiple `return` statements is handy but can be dangerous.

PR #1756, opened 2023-03-17
Add calibration tutorial

### LINKED ISSUE #1616: Calibration tutorial
Once we have a more stable Calibration API, we'll need a tutorial!

Calibration PR: https://github.com/unitaryfund/mitiq/pull/1614

nathanshammah:
## Description
Add calibration tutorial, close #1616. 

---

### License


docs/source/examples/calibration-tutorial.md
```diff
@@ -0,0 +1,204 @@
...
+## Using calibration to improve the results
+
++++
+
+ Let's consider as executor a noisy quantum circuit using Qiskit noisy backend simulators, `FakeJakarta`. Right now the calibration module does not natively support Qiskit circuits, so in the executor, we use Mitiq's conversion functions to convert the Qiskit circuit with `mitiq.interface.mitiq_qiskit.conversions.to_qiskit`.
+
+```{code-cell} ipython3
+def execute_calibration(cirq_circuit):
+    """Execute the input circuits and return the measurement results."""
+    qiskit_circuit = to_qiskit(cirq_circuit)
```
andreamari, line 112:
```suggestion
def execute_calibration(qiskit_circuit):
    """Execute the input circuits and return the measurement results."""
```

natestemen (commented):
Nice work Nathan! Thanks for being a bit of a guinea pig and testing this out and finding some kinks. Your feedback for this module has been really helpful.

Aside from the comments I left, we need to convert the latest notebook to markdown, and remove the jupyter notebook (or vice versa) as the double-up seems to be confusing sphinx.

Excited to get this out!

nathanshammah:
Hi @natestemen @andreamari  the tutorial is now failing. I cannot install current dev version of Mitiq, seemingly due to an issue with `pyscf`, recently updated. 

In a version I could run, with `pip instal mitiq` installing `v=0.24.0` there are some issues in the tutorial:
- 1. `cal.run(log=True)` fails seemingly due to "Invalid input object %s, must be either a QuantumCircuit, Schedule, or a list of either". Could it be that something changed in the way calibrator gets as input `circuit`/`circuit[0]`?
- 2. The final cell now does not run, asking a `yes/no` answer. I also suggest to replace `yes/no` with `y/n` in the suggested reply, as these are the expected keys. I would like to bypass it though, to avoid having this interactive step prevent running the cell (snapshot below)
cell-calibration

andreamari:
It seems to run now. Looks good from my side! 

Something to be aware of,  which I think should not block this PR, is the content of final code cell output:
```
ideal = 	 	 1
unmitigated = 	 	 0.948
mitigated = 	 	 1.021
calibrated_mitigated = 	 0.9703333333333335
```
Note that default `mitigated` is slightly better than `calibrated_mitigated`. This is probably due to a combination of random fluctuations and the fact that the default ZNE parameters are quite good in the first place for this particular example. 

There must be other regimes in which the default parameters are not good, such that `calibrated_mitigated` is likely to be better. If we find such type of regime (different noise level, different circuit depths, etc.. ) we can always update this tutorial.

Issue #1475, opened 2022-08-30
Identity Scaling : Theory


purva-thakre:
https://mitiq.readthedocs.io/en/stable/guide/zne-5-theory.html

The theory is specific to Unitary Folding. Add a separate subsection behind identity insertion and decoherence.

natestemen:
closed by https://github.com/unitaryfund/mitiq/pull/1759

Issue #1474, opened 2022-08-30
Identity Scaling : What happens when I use ZNE ?


purva-thakre:
https://mitiq.readthedocs.io/en/stable/guide/zne-4-low-level.html

First step is only for using unitary folding. Make changes for identity scaling. 

Heere, add a table (or diagram) differentiating between Unitary Folding and Identity Insertion methods.

natestemen:
closed by https://github.com/unitaryfund/mitiq/pull/1759

PR #1759, opened 2023-03-17
Remaining Stuff related to Identity Insertion Scaling in Docs

### LINKED ISSUE #1471: Identity Scaling : How do I use ZNE ?
In [`Select Noise Scaling Method`](https://mitiq.readthedocs.io/en/stable/guide/zne-1-intro.html#select-a-noise-scaling-method), add discussion on how to use identity scaling instead of unitary folding.

purva-thakre:
Fixes #1471 
#1474 
#1475

[Link to updated diagrams](https://docs.google.com/presentation/d/1jWKEaJQFpJsaLtoN1NARi19QWlxe7565GZ88-q3rQcE/edit?usp=sharing).


docs/source/guide/zne-5-theory.md
```diff
@@ -50,11 +52,26 @@ name: figzne_local
...
+```{figure} ../img/zne_id_scaling_layers.png
+---
+width: 500
+name: figid
+---
+The diagram demonstrates how identity gates are inserted after a circuit layer when identity insertion scaling function is applied.
+```
+
+Additional details on the theory of identity insertion scaling are similar to those in unitary folding. The only difference 
+is that instead of scaling gate noise, the insertion of an identity gate increases the wait time after each circuit layer is executed. This allows the qubits to interact with the environment through some noisy process. The decoherence time for the qubits in a quantum system is determined by the amount of time our system of interest remains coherent and uncouples from the external environment. More details on decoherence can be found in {cite}`Schlosshauer_2019, Zurek_2003_arxiv`.
```
purva-thakre, line 69:
```suggestion
is that instead of scaling gate noise, the insertion of an identity gate increases the wait time after each circuit layer is executed. This allows the qubits to interact with the environment through some noisy process and decohere if the system-environment interaction is strong. The decoherence time for the qubits in a quantum system is determined by the amount of time our system of interest remains coherent and uncouples from the external environment. More details on decoherence can be found in {cite}`Schlosshauer_2019, Zurek_2003_arxiv`.
```

docs/source/examples/molecule_pyscf_sto-6g.hdf5
natestemen, line 1: Is this file intended to be included?

docs/source/guide/zne-3-options.md
```diff
@@ -225,14 +225,18 @@ $$G \longrightarrow I G$$
 
 Here, $G$ is a single circuit layer containing individual that can be performed simultaneously.
 
-To use this method, call {func}`.zne.scaling.insert_id_layers()` with both a circuit to scale, and a scale factor.
+To use this method, call {func}`.zne.scaling.identity_insertion.insert_id_layers()` with both a circuit to scale, and a scale factor.
```
natestemen, line 228: Since `insert_id_layers` is imported in [`zne/scaling/__init__.py`](https://github.com/unitaryfund/mitiq/blob/master/mitiq/zne/scaling/__init__.py), this change isn't totally necessary. Is there a reason you prefer this? I personally prefer a shorter function/module chain.

natestemen (commented):
Nice work finding all the spots where we assume unitary folding is the only scaling method. Great to have this last push on id scaling docs done after this.

Just two minor question/comments.

docs/source/examples/molecule_pyscf_sto-6g.hdf5
purva-thakre: No. It's not. I'll remove it. 

I don't know where this comes from ! This is the second time it has happened. Maybe something from Visual Studio Code.

docs/source/guide/zne-3-options.md
purva-thakre: Yes, I understand this and I would have used `zne.scaling.insert_id_layers()` as well. I noticed that in the docs, it was not linking to the API docs. It is not in blue. 

image

natestemen: Ahhh okay sphinx must not be able to find it otherwise. Thanks for checking.

natestemen (approved):

PR #1762, opened 2023-03-22
Update qiskit tutorial with ddd functions

### LINKED ISSUE #1783: DDD tutorial broken thumbnail
Navigate to the [examples](https://mitiq.readthedocs.io/en/stable/examples/examples.html) and scroll to the bottom. The [DDD with Qiskit on GHZ circuits](https://mitiq.readthedocs.io/en/stable/examples/ddd_on_ibmq_ghz.html) demo has a broken thumbnail.

Screenshot 2023-04-12 at 1 30 25 PM

@Misty-W

Misty-W:
## Description

1. Replaced user-defined sequence functions with built-in DDD functions.

2. Added axis titles on plot (embarrassingly they were missing on the first PR for results from HW).

Also (to do):

Fixes #1783

---

### License


Misty-W:
Rebase didn't behave as anticipated. Superseded by #1766.

docs/source/examples/ddd_on_ibmq_ghz.md
```diff
@@ -198,14 +192,20 @@ def ibm_executor(
...
-    for rule in rules:
-        print(f"DDD sequence: {rule}.")
-        circuit = get_circuit_with_sequence(depth, rule)
-        noisy_value = ibm_executor(
+    circuit = get_circuit(depth)
+    noisy_value = ibm_executor(
             circuit, shots=shots, correct_bitstring=[0]
+    )
+    print("Result:", noisy_value)
+    data.append((depth, rule, noisy_value))
```
andreamari, line 200: Does this run even if `rule` is not yet defined?

andreamari (commented):
Sorry for the late review, the PR looks good to me! Up to a small doubt about the `rule` variable in the inline comment.

However, when reading again the tutorial, I noticed a small unrelated problem. In the text we say that we measure the frequency of the 00... bitstring, but we actually only measure the first qubit.  So, in practice we should:
- Measure both qubits (optimal solution but would require re-running the experiments :scream: ). 
- Change the text and say that we measure just the first qubit.

Feel free to fix the problem within this PR or to just open a new issue about it.

natestemen (commented):
Always good to use our own functions where possible 😊

EDIT: sorry, there was supposed to be two comments attached here. Not sure what happened. My internet connection is unstable, so maybe related to that.

docs/source/examples/ddd_on_ibmq_ghz.md
Misty-W: I still don't know how it was able to run before without defining the `rule`, but it should be fixed now. 

Also changed the wording of the text in the beginning and now it reads, "the frequency of the $|0 \rangle$ bitstring is our target metric (in this example we only measure the first qubit)."
Not ideal but consistent at least 😅

andreamari (approved):
Looks good to me!

PR #1763, opened 2023-03-24
Calibration API simplifications

### LINKED ISSUE #1757: Make execute_with_mitigation() a method of Calibrator
## Proposed Solution

In the current code, `execute_with_mitigation()` is an independent function which takes as input a calibrator object.
My proposal is to re-define it as a method of `Calibrator`.

cal.run()
cal.execute_with_mitigation(...)

Motivations:
- Less things to import for the user.
- It makes clear that the applied mitigation is "decided" by the calibrator.

Do you agree? Am I missing any drawbacks?

P.S.: If useful, we could also have both: the method and the standalone function. But I am not sure we need both.

natestemen:
## Description

This PR does 3 things:

1. Makes `ZNESettings` the default settings object for calibration (feedback from Will).
2. Allows the `Calibrator` class to be imported from the top-level of mitiq. This is important if we want to brand it as the main way into mitiq.
3. Adds an `execute_with_mitigation` method to the `Calibrator` class, so that the function does not need to be imported separately (fixes #1757).

**Tests are currently missing, I know.**


andreamari (approved):
LGTM. If there is no easy fix to the mypy issue, you can add a "type ignore" flag to unblock this (and all other PRs).

natestemen:
Docs failure is a broken link unrelated to this work.

PR #1767, opened 2023-04-01
Support noise scaling by layer

### LINKED ISSUE #1722: Support noise scaling by layer
## Issue Description

It could be nice to allow the user to fold circuits by selecting the specific layer and providing the number of times to fold it. Right now there is no function that allows this level of control in `global_folding`. What exists now is, besides the strategies `from_left`, right, and randomly, is that, assigning a fidelity lower than 1 to one or more gate, folding takes that into account. 

## Proposed Solution
I propose to define a stand-alone folding function similar to the one shown by @vprusso in the recent Quantum Wednesday talk about circuit debugging. Thanks to @andreamari for discussing this a bit already. 
A possible structure could be (pseudocode):
```py
def zne.scaling.layer_scaling.layer_folding(num_fold_list: list[int]): ...
```
where `num_fold_list = [0, 0, 1, 0, 0]` or `num_fold_list = [1, 2, 0, 0, 0]`, etc., with the index, $n$, of the list referring to the layer number, and the element filled by an integer would represent the number of times the layer is folded. In the two lists above, in the first case layer 3 is folded once, while in the second list, layer 1 is folded once and layer 2 is folded twice. 
 
For each layer the scale factor is $2n +1$. 

The proposed solution could be a module in `mitiq.zne`, similarly to what done for the identity insertion scaling module.

## Additional References
- @vprusso's [talk](https://docs.google.com/presentation/d/1THv95c9IBXVHA-mHen-M5KemCMm7enI6SGlRGLA94dQ/) given at Quantum Wednesday about https://arxiv.org/abs/2204.06056.
- Identity insertion scaling module ([code](https://github.com/unitaryfund/mitiq/blob/master/mitiq/zne/scaling/identity_insertion.py), [docs](https://mitiq.readthedocs.io/en/latest/guide/zne-3-options.html#identity-scaling)).

vprusso:
Closes #1722 

## Description

1. Adding a feature to fold circuit by layer. The function `layer_folding` allows one to provide a `cirq.Circuit` object and supply a list of integers of size equal to the depth of the circuit where each integer in the list corresponds to how many folds to apply on a given layer. 

2. Adding corresponding unit tests for folding functionality.

3. Adding documentation for layerwise folding.
---

### License


mitiq/zne/scaling/layer_scaling.py
```diff
@@ -0,0 +1,74 @@
...
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program.  If not, see <https://www.gnu.org/licenses/>.
+
+"""Functions for layer-wise unitary folding on supported circuits."""
+from typing import List
+from cirq import Circuit, inverse, Moment
```
natestemen, line 18: To reduce typing ambiguity what do you think about something like this?
```suggestion
import cirq
from cirq import inverse, Moment
```
This way in type hints we will use `cirq.Circuit`, as other frontends can call their abstraction `Circuit` as well. Pretty minor, however. WDYT?

mitiq/zne/scaling/layer_scaling.py
```diff
@@ -0,0 +1,74 @@
...
+
+
+def layer_folding(circuit: Circuit, layers_to_fold: List[int]) -> Circuit:
```
natestemen, line 21: Do you think we should apply [`noise_scaling_converter`](https://github.com/unitaryfund/mitiq/blob/c849314bc73c2e7b1b71d312639c5e8417ab2e74/mitiq/interface/conversions.py#L221-L223) here as a decorator? It should allow for this function to be used with other frontends.

If we decide to add this, we should test with an additional frontend as well.

natestemen, line 21: For consistency with the [folding](https://github.com/unitaryfund/mitiq/blob/c849314bc73c2e7b1b71d312639c5e8417ab2e74/mitiq/zne/scaling/folding.py#L219-L223) methods, and the [identity insertion](https://github.com/unitaryfund/mitiq/blob/c849314bc73c2e7b1b71d312639c5e8417ab2e74/mitiq/zne/scaling/identity_insertion.py#L82) scaling method, what if we separated `layers_to_fold` into two separate variables:

1. `layers_to_fold`: To record layer indices
2. `scale_factors`: to record the number of times to fold

I do like how the current API is very dense in it's input data. Not a clear winner for me here, but I lean slightly towards having consistent API if it doesn't seem like a disadvantage to you.

mitiq/zne/scaling/layer_scaling.py
```diff
@@ -0,0 +1,74 @@
...
+        by ``layers_to_invert``.
+    """
+    layers = []
+    for i, layer in enumerate(circuit):
+        layers.append(layer)
+        # Apply the requisite number of folds to each layer.
+        num_fold = layers_to_fold[i]
+        for _ in range(num_fold):
+            layers.append(Moment(inverse(layer)))
+            layers.append(Moment(layer))
```
natestemen, line 41: Is there a reason we need to call `Moment` in the two append calls, but not at the top? My local testing indicates they might be unnecessary.

mitiq/zne/scaling/layer_scaling.py
```diff
@@ -0,0 +1,74 @@
...
+
+    # We combine each layer into a single circuit.
+    combined_circuit = Circuit()
+    for layer in layers:
+        combined_circuit.append(Circuit(layer))
```
natestemen, line 46: Similar to above, is this `Circuit(layer)` call needed, or can we append `layer` directly?

mitiq/zne/scaling/layer_scaling.py
```diff
@@ -0,0 +1,74 @@
...
+    return combined_circuit
+
+
+def layer_folding_all(
+    circuit: Circuit, layers_to_fold: int = 1
+) -> List[Circuit]:
```
natestemen, line 52:
```suggestion
def layer_folding_all(
    circuit: Circuit, num_folds: int = 1
) -> List[Circuit]:
```

Might be a better name?

docs/source/guide/zne-3-options.md
```diff
@@ -216,6 +216,66 @@ folding method. Fidelity values must be between zero and one.
...
+
+Circuit debugging techniques (as proposed in
+[arXiv:2204.06056](https://arxiv.org/abs/2204.06056)) utilize the ability to
+apply layerwise folding of a circuit.
+
+A quantum circuit $C$ may be represented as a series of $d$ layers. Each layer
+$L$ contains one or more quantum gates acting concurrently on an $n$-qubit
+system:
+
+$$C = L_d L_{d-1} \ldots L_2 L_1$$.
```
natestemen, line 229: Potential stylistic preference, but `\cdots` seems more appropriate here since it's multiplication.

docs/source/guide/zne-3-options.md
```diff
@@ -216,6 +216,66 @@ folding method. Fidelity values must be between zero and one.
...
+$i$-inversion is defined as
+
+$$C^{(i)} = L_d \ldots L_{i+1} L_i \left[L_i^{-1} L_i\right]^m L_{i-1} \ldots L_1$$
+
+where we apply a given $i$-inversion $m$ times.
+
+For example, consider the following circuit
+
+```{code-cell} ipython3
+from mitiq.zne.scaling.layer_scaling import layer_folding
```
natestemen, line 249:
```suggestion
from mitiq.zne.scaling import layer_folding
```

I think we should be able to import directly from `scaling` now.

natestemen (commented):
Really nice work Vincent! Thanks for studying this technique, and getting it worked in so quickly.

I've left a few comments, most of which are completely optional, and I of course will not be offended if you don't agree with a suggestion. My bigger question is do we want this PR to add capability to use this layer folding technique within ZNE? I _think_ it's very close, and so I lean yes, but I have not tested it yet, and there may be unforeseen limitations/challenges. Let me know what you think!

mitiq/zne/scaling/layer_scaling.py
vprusso: `mypy` seems to throw a fit if this isn't explicitly stated here. Removing the `Moment()` wrapper and running `make check-types` yields:

```
Argument 1 to "append" of "list" has incompatible type "Union[Operation, OpTree]"; expected "Moment"  [arg-type]
````

We could also explicitly ignore the line as well. Any preference?

File: mitiq/zne/scaling/layer_scaling.py (hunk 4 above)
Comment (vprusso) [reply mitiq/zne/scaling/layer_scaling.py:47]: This seems to be something we can do. Change made.

File: docs/source/guide/zne-3-options.md (hunk 6 above)
Comment (vprusso) [reply docs/source/guide/zne-3-options.md:229]: Done!

File: mitiq/zne/scaling/layer_scaling.py (hunk 3 above)
Comment (natestemen) [reply mitiq/zne/scaling/layer_scaling.py:46]: Weird, is it complaining about appending the layer, or the inverse (or both)? From my testing they should all be moments!

```py
c = cirq.Circuit()
q = cirq.LineQubit.range(2)

c.append(cirq.ops.H.on_each(*q))
c.append(cirq.ops.X.on_each(*q))

for layer in c:
    print(type(layer))
    print(type(cirq.inverse(layer)))

>>> <class 'cirq.circuits.moment.Moment'>
>>> <class 'cirq.circuits.moment.Moment'>
>>> <class 'cirq.circuits.moment.Moment'>
>>> <class 'cirq.circuits.moment.Moment'>
```

What does [`reveal_type(layer)`](https://mypy.readthedocs.io/en/stable/type_inference_and_annotations.html?highlight=reveal_type#type-inference) show?

natestemen:
Thanks for the nice chat today Vincent. As discussed, I think this PR should aim to implement a folding method that can be used within ZNE. What does this look like in practice (modulo function naming)?

```py
layer_to_fold = 999
fold_layer_func = mitiq.zne.scaling.fold_layer(layer_to_fold)

mitiq.zne.execute_with_zne(circuit, executor, scale_noise=fold_layer_func)
```

The `scale_noise` parameter has the following signature: https://github.com/unitaryfund/mitiq/blob/d6dc058384b5b2b8eac76d448ae13d3dbf0d8e6f/mitiq/zne/zne.py#L31-L33

Hence, however we go about this (refactoring one of the existing functions, or creating a new one), we need a function with the following signature:
```
Callable[
    [int], # layer
    Callable[[
            QPROGRAM, # circuit to fold
            float # scale factor
        ],
        QPROGRAM # folded circuit
    ]
]
# in one line for readability:
Callable[[int], Callable[[QPROGRAM, float], QPROGRAM]]
```

Does this make sense? Let me know if you have any questions, or disagree that this should be the approach.

docs/source/guide/zne-3-options.md
```diff
@@ -216,6 +216,66 @@ folding method. Fidelity values must be between zero and one.
...
+    cirq.H(q0),
+    cirq.CNOT(q0, q1),
+)
+print(circuit)
+```
+
+This circuit consists of two layers (the first layer has a Hadamard gate and the
+second layer has a CNOT gate).
+
+We can make use of the `layer_folding` function where the argument
```
nathanshammah, line 262:
```suggestion
We can make use of the {func}`.layer_folding()` function where the argument
```

docs/source/guide/zne-3-options.md
```diff
@@ -216,6 +216,66 @@ folding method. Fidelity values must be between zero and one.
...
+[arXiv:2204.06056](https://arxiv.org/abs/2204.06056)) utilize the ability to
+apply layerwise folding of a circuit.
+
+A quantum circuit $C$ may be represented as a series of $d$ layers. Each layer
+$L$ contains one or more quantum gates acting concurrently on an $n$-qubit
+system:
+
+$$C = L_d L_{d-1} \cdots L_2 L_1$$.
+
+The $i$-inverse operation of a circuit $C$ is defined as
```
nathanshammah, line 231: Something is off here:
latex-bug-mitiq
Note you can preview the docs relative to this PR by clicking on the CI check for readthedocs. Direct link [here](https://mitiq--1767.org.readthedocs.build/en/1767/).

nathanshammah (changes requested):
@vprusso looks very neat to me, really excited to have this in.
I added some small comments on the documentation.
- Please also add the new module to the API doc file, and include a subtitle like for the identity insertion scaling, [link](https://mitiq--1767.org.readthedocs.build/en/1767/apidoc.html#module-mitiq.zne.scaling.identity_insertion).

mitiq/zne/scaling/layer_scaling.py
vprusso: Yeah, it's complaining about both the layer and the inverse. Indeed, when I check the types of the layer and inverse, I see `Moment` as well. I think this might be a quirk of `mypy` or perhaps the version that is pinned is getting confused as without this explicit typecast it's failing in the pipeline.

mitiq/zne/scaling/layer_scaling.py
vprusso: Agreed. Made this change. Thank you!

docs/source/guide/zne-3-options.md
vprusso: I would think so as well, but attempting to import this seems to yield an error when running on the pipeline for mitiq. I'm keeping the longer import style here, but let me know if you think it's possible to use the shorter one without the pipeline failure.

docs/source/guide/zne-3-options.md
vprusso: Hmm. Oddly enough, this is rendering fine for me locally and I'm not seeing any obvious LaTeX issues with the source. Maybe on the next build it will fix itself. Will see.

vprusso:
> @vprusso looks very neat to me, really excited to have this in. I added some small comments on the documentation.
> 
> * Please also add the new module to the API doc file (`apidoc.md`), and include a subtitle like for the identity insertion scaling, [link](https://mitiq--1767.org.readthedocs.build/en/1767/apidoc.html#module-mitiq.zne.scaling.identity_insertion).

Added to `apidoc.md`. Nice catch!

docs/source/guide/zne-3-options.md
```diff
@@ -216,6 +216,72 @@ folding method. Fidelity values must be between zero and one.
...
+*Calderon et al. Quantum (2023)* {cite}`Calderon_2023_Quantum`, utilize the
+ability to apply layerwise folding of a circuit.
+
+A quantum circuit $C$ may be represented as a series of $d$ layers. Each layer $L$
+contains one or more quantum gates acting concurrently on an $n$-qubit
+system:
+
+$$
+C = L_d L_{d-1} \cdots L_2 L_1
+$$.
```
natestemen, line 231: Looks like the error is caused by the period _after_ the `$$`. If you move it directly inside it works.

```suggestion
$$C = L_d L_{d-1} \cdots L_2 L_1.$$
```

vprusso: Excellent.  Saved me some headache there. Thank you! :)

docs/source/guide/zne-3-options.md
vprusso: Resolved. Thanks again @natestemen

andreamari:
@vprusso, for some reason the diff of this PR contains a lot of unrelated previous PRs. Could you try merging master on this PR, or rebasing this PR on the latest Mitiq master?

@natestemen: I am not sure there is a simple way of casting `layer_folding` as a standard noise scaling function compatible with `execute_with_zne`. They have a different number of input parameters. Even if we decide to quantify the amount of noise scaling with a `scale_factor` instead of `num_folds`, the main problem is that `layer_folding` wants a lot of scale factors (one for each layer).

Maybe we should consider `layer_folding` more like a self-contained tool that the user can apply for circuit debugging or for a low-level custom implementation of ZNE.

Note also that what is now called `layer_folding_all` has the same effect as local folding (`fold_gates_at_random`) which, with odd integer scale factors, scales all the gates (and therefore all the layers) in the same way of `layer_folding_all`. So I am not sure if it's useful in the current form.

vprusso:
> @vprusso, for some reason the diff of this PR contains a lot of unrelated previous PRs. Could you try merging master on this PR, or rebasing this PR on the latest Mitiq master?

Hmm. That's odd. I merged the `master` branch into the PR branch and it looks okay with the exception of one doc failure. Not entirely sure why this is failing (maybe it's obvious and either @natestemen or @andreamari has seen this one before?)
https://github.com/unitaryfund/mitiq/actions/runs/4692483610/jobs/8318255429#step:6:881

andreamari:
> Hmm. That's odd. I merged the `master` branch into the PR branch and it looks okay with the exception of one doc failure. Not entirely sure why this is failing (maybe it's obvious and either @natestemen or @andreamari has seen this one before?) https://github.com/unitaryfund/mitiq/actions/runs/4692483610/jobs/8318255429#step:6:881

We can ignore the error. (We check for broken links in CI and a link to a paper was temporarily broken)

natestemen:
I think it is a straightforward change that makes this change more useful, faster. I implemented (quickly) what I had in mind in my [previous comment](https://github.com/unitaryfund/mitiq/pull/1767#issuecomment-1498177636) here: https://github.com/unitaryfund/mitiq/commit/c80124a6369d400949c0609152d6b2efe6742655. Feel free to cherry pick it, or modify as you see fit. It's a relatively small diff, but definitely needs some more testing. It should work as I outlined above, and hopefully with `execute_with_zne` as well, but I did not test that yet.

mitiq/zne/scaling/layer_scaling.py
natestemen: WDYT of this?

```py
if not cirq.is_measurement(layer):
    layers.append(layer**-1)
    layers.append(layer)
```

The advantage of using [`cirq.inverse`](https://github.com/quantumlib/Cirq/blob/v1.1.0/cirq-core/cirq/protocols/inverse_protocol.py#L70-L121) seems to be that it ensures `__pow__` is defined, but I believe this should always be the case for moments that are not measurements, right?

docs/source/guide/zne-3-options.md
natestemen: Is there a build where this was tested and it failed? I see https://github.com/unitaryfund/mitiq/pull/1767/commits/a547bc285e14c7b488a2d57b5c7187736d1b1ba2, but when inspecting the build, it seems all the failures I saw was from an unrelated error that was fixed elsewhere.

andreamari:
> I think it is a straightforward change that makes this change more useful, faster. I implemented (quickly) what I had in mind in my [previous comment](https://github.com/unitaryfund/mitiq/pull/1767#issuecomment-1498177636) here: 
[c80124a](https://github.com/unitaryfund/mitiq/commit/c80124a6369d400949c0609152d6b2efe6742655). Feel free to cherry pick it, or modify as you see fit. It's a relatively small diff, but definitely needs some more testing. It should work as I outlined above, and hopefully with `execute_with_zne` as well, but I did not test that yet.

If we implement something like [c80124a](https://github.com/unitaryfund/mitiq/commit/c80124a6369d400949c0609152d6b2efe6742655) (I like it), can we call `fold_layer` as `get_layer_folding` or `fold_layer_function`, or anything which gives the idea of returning a folding function instead of actually applying the folding operation?

@natestemen and @vprusso, based on this discussion, do you agree that `layer_folding_all()` can be removed since it is  equivalent to `fold_gates_at_random` up to a `num_folds` --> odd `scale_factor` conversion?

vprusso:
> > I think it is a straightforward change that makes this change more useful, faster. I implemented (quickly) what I had in mind in my [previous comment](https://github.com/unitaryfund/mitiq/pull/1767#issuecomment-1498177636) here:
> > [c80124a](https://github.com/unitaryfund/mitiq/commit/c80124a6369d400949c0609152d6b2efe6742655). Feel free to cherry pick it, or modify as you see fit. It's a relatively small diff, but definitely needs some more testing. It should work as I outlined above, and hopefully with `execute_with_zne` as well, but I did not test that yet.
> 
> If we implement something like [c80124a](https://github.com/unitaryfund/mitiq/commit/c80124a6369d400949c0609152d6b2efe6742655) (I like it), can we call `fold_layer` as `get_layer_folding` or `fold_layer_function`, or anything which gives the idea of returning a folding function instead of actually applying the folding operation?
> 
> @natestemen and @vprusso, based on this discussion, do you agree that `layer_folding_all()` can be removed since it is equivalent to `fold_gates_at_random` up to a `num_folds` --> odd `scale_factor` conversion?

- Incorporating/cherry-picking @natestemen  suggestion with the name change as suggested by @andreamari 
- Removing `fold_layer_all` function as suggested by @andreamari 
- Tests for layerwise folding decorator in `test_zne.py`

Let me know what you both think, and specifically, if the tests at present make sense.

vprusso:
An additional question is regarding the typing of things. the `get_layer_folding` function takes in a `scale_factor` argument that is a float, but the way in which this argument is used in `layer_folding` only makes sense if the argument is an integer. I wonder if we want to either rethink the way that we perform the folding or treat the scale factors for this instance. Any thoughts, @natestemen / @andreamari ?

mitiq/zne/scaling/layer_scaling.py
```diff
@@ -0,0 +1,76 @@
...
+    for layer in layers:
+        combined_circuit.append(layer)
+    return combined_circuit
+
+
+def get_layer_folding(layer_index: int) -> Callable[[QPROGRAM, float], QPROGRAM]:
+    @noise_scaling_converter
+    def fold_ith_layer(
+        circuit: cirq.Circuit, scale_factor: float
+    ) -> cirq.Circuit:
```
natestemen, line 70:
```suggestion
    def fold_ith_layer(
        circuit: cirq.Circuit, scale_factor: int
    ) -> cirq.Circuit:
```

Since `layer_folding` doesn't currently support float scale factors, lets make this an int for now.

vprusso: Makes sense. Removed and adapted as suggested!

mitiq/zne/scaling/layer_scaling.py
```diff
@@ -0,0 +1,78 @@
...
+    circuit: cirq.Circuit, layers_to_fold: List[int]
+) -> cirq.Circuit:
+    """Applies a variable amount of folding to select layers of a circuit.
+
+    Args:
+        circuit: The input cirq circuit.
+        layers_to_fold: A list with the index referring to the layer number,
+                        and the element filled by an integer represents the
+                        number of times the layer is folded.
+
```
andreamari, line 36: There is a small ambiguity about the length of `layers_to_fold`: do we count measurements as a layer?
Based on the code the answer seems to be :  yes, but the associated value should be 0 since in practice we cannot fold measurement layers.

Another option is to copy the workflow used in other folding functions where `check_foldable` and `pop_measurements` and `append_measurements` are called. In this case, `layers_to_fold` should be defined without considering the measurement layer (if present).

Whatever approach we use, we should probably say in the docstring what is the expected structure of `layers_to_fold` when measurements are present in the circuit.

mitiq/zne/scaling/layer_scaling.py
```diff
@@ -0,0 +1,78 @@
...
+    combined_circuit = cirq.Circuit()
+    for layer in layers:
+        combined_circuit.append(layer)
+    return combined_circuit
+
+
+def get_layer_folding(
+    layer_index: int,
+) -> Callable[[QPROGRAM, float], QPROGRAM]:
+    @noise_scaling_converter
```
andreamari, line 69: Docstring missing.

mitiq/zne/scaling/layer_scaling.py
```diff
@@ -0,0 +1,78 @@
...
+    def fold_ith_layer(
+        circuit: cirq.Circuit, scale_factor: int
+    ) -> cirq.Circuit:
```
andreamari, line 72: Docstring missing.

mitiq/zne/tests/test_zne.py
```diff
@@ -492,6 +493,34 @@ def test_execute_with_zne_with_supported_circuits(circuit_type):
...
+    unmitigated = generic_executor(circuit)
+    # Use odd scale factors for deterministic results
+    fac = RichardsonFactory([1, 3, 5])
+    # Layerwise-fold
+    layer_to_fold = 0
+    fold_layer_func = get_layer_folding(layer_to_fold)
+
+    zne_value = execute_with_zne(
+        circuit, generic_executor, factory=fac, scale_noise=fold_layer_func
+    )
```
andreamari, line 518: Nice to see this is compatible with `execute_with_zne` !

docs/source/guide/zne-3-options.md
```diff
@@ -216,6 +216,72 @@ folding method. Fidelity values must be between zero and one.
...
+print(layer_folding(circuit, [1, 0]))
+```
+
+We can also fold specific layers a variable number of times. Here is an example
+of folding the first layer twice and the second layer three times.
+
+```{code-cell} ipython3
+print(layer_folding(circuit, [2, 3]))
+```
+
```
andreamari, line 284: We should mention that `layer_folding` is not a standard noise scaling function which can be directly used with  `execute_with_zne`  since its signature is not compatible.
 
We could add that:

 `layer_folding` could be used for noise diagnostics `cite Calderon_2023_Quantum` or for a low-level implementation of ZNE in a similar way as shown in the `[next section](zne-4-low-level.md)`.
To get a noise scaling function which is instead directly compatible with execute_with_zne one can use  `{func}.get_layer_folding`.

mitiq/zne/scaling/layer_scaling.py
```diff
@@ -0,0 +1,78 @@
...
+        layers = [0] * len(circuit)
+        layers[layer_index] = scale_factor
```
andreamari, line 74: This change is important. It maps the scale_factor to the right number of foldings. 

```suggestion
        layers = [0] * len(circuit)
        num_folds = (scale_factor - 1)  /  2
        if np.isclose(num_folds, int(num_folds)):
            num_folds = int(num_folds)
        layers[layer_index] = num_folds
```

The following part allows us using `scale_factor` equal to 1.0, 3.0 , 5.0` instead of just 1, 3, 5):
```
        if np.isclose(num_folds, int(num_folds)):
            num_folds = int(num_folds)
```
It can be removed if you don't like it.

andreamari (commented):
I like the addition of the new function and the removal of `fold_all_layers`. Additional  comments below.

mitiq/zne/scaling/layer_scaling.py
vprusso: Added.

mitiq/zne/scaling/layer_scaling.py
vprusso: Added.

mitiq/zne/tests/test_zne.py
vprusso: Indeed! Thanks to @natestemen !

mitiq/zne/scaling/layer_scaling.py
vprusso: I like it! Applied.

docs/source/guide/zne-3-options.md
vprusso: *Think* this should be okay now.

docs/source/guide/zne-3-options.md
vprusso: I like it. Added!

mitiq/zne/scaling/layer_scaling.py
vprusso: I think I like the second option you proposed here for the sake of consistency. I attempted to mimick the way in which it was done elsewhere within the folding function. Hopefully, that's what you had in mind. Thanks for the suggestion!

vprusso:
Thanks again for the round of comments, @andreamari ! Hopefully, the updates there make sense, and of course, let me know if not. 

@natestemen if you want to give the PR another look over as we briefly chatted about, have at it!

mitiq/zne/scaling/layer_scaling.py
```diff
@@ -0,0 +1,115 @@
...
+
+
+@noise_scaling_converter
+def layer_folding(
+    circuit: cirq.Circuit, layers_to_fold: List[int]
+) -> cirq.Circuit:
+    """Applies a variable amount of folding to select layers of a circuit.
+
+    Args:
+        circuit: The input cirq circuit.
```
andreamari, line 39:
```suggestion
        circuit: The input circuit.
```
The function is now multi-platform

mitiq/zne/scaling/layer_scaling.py
```diff
@@ -0,0 +1,115 @@
...
+        layers_to_fold: A list with the index referring to the layer number,
+                        and the element filled by an integer represents the
```
andreamari, line 41:
```suggestion
                        and the element filled by an integer representing the
```

mitiq/zne/scaling/layer_scaling.py
```diff
@@ -0,0 +1,115 @@
...
+                        number of times the layer is folded.
+
+    Returns:
+        A cirq ``Circuit`` with layers and number of times to fold specified
+        by ``layers_to_invert``.
```
andreamari, line 46:
```suggestion
        The folded circuit.
```

mitiq/zne/scaling/layer_scaling.py
```diff
@@ -0,0 +1,115 @@
...
+        layers[layer_index] = int(num_folds)
+
+        _check_foldable(circuit)
+
+        folded = deepcopy(circuit)
+        measurements = _pop_measurements(folded)
+
+        folded = layer_folding(folded, layers_to_fold=layers)
+
+        _append_measurements(folded, measurements)
```
andreamari, line 111:
```suggestion
        folded = layer_folding(folded, layers_to_fold=layers)
```
Isn't better to move the following lines:
```
        folded = deepcopy(circuit)
        measurements = _pop_measurements(folded)
        ...body of the function
        _append_measurements(folded, measurements)
```
into the low-level function `layer_folding()` instead of the high-level function `fold_ith_layer`? In this way they would be applied in both cases.

docs/source/guide/zne-3-options.md
```diff
@@ -216,6 +216,79 @@ folding method. Fidelity values must be between zero and one.
...
+The `layer_folding` function could be used for noise diagnostics `cite
+Calderon_2023_Quantum` or for a low-level implementation of ZNE in a similar way
+as shown in the `[next section](zne-4-low-level.md)`.
+
+To get a noise scaling function which is instead directly compatible with
+execute_with_zne one can use  `{func}.get_layer_folding`.
```
andreamari, line 290:
```suggestion
`{func}.execute_with_zne` one can use  `{func}.get_layer_folding`.
```

mitiq/zne/scaling/layer_scaling.py
```diff
@@ -0,0 +1,115 @@
...
+    Returns:
+        A cirq ``Circuit`` with layers and number of times to fold specified
+        by ``layers_to_invert``.
+    """
+    circuit = cirq.synchronize_terminal_measurements(circuit)
+    layers = []
+
+    if len(layers_to_fold) != len(circuit):
+        raise ValueError(
+            "Length of `layers_to_fold` must be equal to length of circuit."
```
andreamari, line 53:
```suggestion
            f"Length of layers_to_fold must be equal to the circuit depth excluding measurements, i.e., {len(circuit_without_meas)})"
```
Up to style, can we say something like this? Note that this suggestion assumes `circuit_without_meas` has been created.

A reason to provide the exact depth is that the user may not know what is the length of the internal circuit representation, since a conversion from say Qiskit to Cirq could change the depth.

mitiq/zne/scaling/layer_scaling.py
```diff
@@ -0,0 +1,115 @@
...
+        circuit: The input cirq circuit.
+        layers_to_fold: A list with the index referring to the layer number,
+                        and the element filled by an integer represents the
+                        number of times the layer is folded.
+
+    Returns:
+        A cirq ``Circuit`` with layers and number of times to fold specified
+        by ``layers_to_invert``.
+    """
+    circuit = cirq.synchronize_terminal_measurements(circuit)
```
andreamari, line 48:
```suggestion
    circuit = cirq.synchronize_terminal_measurements(circuit)
```
This seems dangerous since it could mutate the input variable `circuit`. But, if it doesn't, ignore this comment.

andreamari (commented):
A few other comments. I promise that, in the next round, I'll approve!  :-D

mitiq/zne/scaling/layer_scaling.py
vprusso: Right. I think this did make sense before but perhaps now can be taken out. Removed!

vprusso:
> A few other comments. I promise that, in the next round, I'll approve! :-D

Thank you for the comments, @andreamari , and thank you for the suggested fixes. Of course, don't approve if you think it's still lacking in some way, but hopefully this is closer to ready in any case!

vprusso:
After a bit of wrestling with the tests, I think we are good!
https://github.com/unitaryfund/mitiq/actions/runs/4760447950

andreamari:
> After a bit of wrestling with the tests, I think we are good! https://github.com/unitaryfund/mitiq/actions/runs/4760447950

andreamari:
I think you missed the text of this [comment](https://github.com/unitaryfund/mitiq/pull/1767#discussion_r1172408568) in which the suggestion was to move some lines of code from the `fold_ith_layer` function into the `layer_folding` function. They instead completely disappeared. 

Moreover, it would be good to have back the error message that was removed in the recent commit [bfa9389](https://github.com/unitaryfund/mitiq/pull/1767/commits/bfa9389782994267cb899964c79dccd971a68e99). That message is useful to the user in order to understand how long `layers_to_fold` should be.

If I am missing some aspect or if there are any problems, we can sync on Discord.

nathanshammah (approved):

docs/source/refs.bib
```diff
@@ -88,6 +88,16 @@ @article{Cafaro_2014_PRA
...
 
+@article{Calderon_2023_Quantum,
+  title       = {Quantum circuit debugging and sensitivity analysis via local inversions},
+  author      = {Calderon-Vargas, Fernando A and Proctor, Timothy and Rudinger, Kenneth and Sarovar, Mohan},
+  journal     = {Quantum},
+  volume      = {7},
+  pages       = {921},
+  year        = {2023},
+  publisher   = {Verein zur F{\"o}rderung des Open Access Publizierens in den Quantenwissenschaften},
+}
```
natestemen, line 99: Adding a URL (perhaps https://doi.org/10.22331/q-2023-02-09-921) would be helpful here to make the bibliography more useful.

docs/source/guide/zne-3-options.md
```diff
@@ -216,6 +216,79 @@ folding method. Fidelity values must be between zero and one.
...
+We can also fold specific layers a variable number of times. Here is an example
+of folding the first layer twice and the second layer three times.
+
+```{code-cell} ipython3
+print(layer_folding(circuit, [2, 3]))
+```
+
+The `layer_folding` function could be used for noise diagnostics `cite
+Calderon_2023_Quantum` or for a low-level implementation of ZNE in a similar way
+as shown in the `[next section](zne-4-low-level.md)`.
```
natestemen, line 287:
```suggestion
The `layer_folding` function could be used for noise diagnostics {cite}`Calderon_2023_Quantum`
or for a low-level implementation of ZNE in a similar way
as shown in the [next section](zne-4-low-level.md).
```

I'll let you determine how to break lines, but the citation and "next section" links are broken.

docs/source/guide/zne-3-options.md
```diff
@@ -216,6 +216,79 @@ folding method. Fidelity values must be between zero and one.
...
+
+To get a noise scaling function which is instead directly compatible with
+`{func}.execute_with_zne` one can use  `{func}.get_layer_folding`.
```
natestemen, line 290:
```suggestion
To get a noise scaling function which is instead directly compatible with
{func}`.execute_with_zne` one can use  {func}`.get_layer_folding`.
```

**untested**, but I think this is the syntax.

mitiq/zne/scaling/layer_scaling.py
```diff
@@ -0,0 +1,107 @@
...
+    layer_index: int,
+) -> Callable[[QPROGRAM, float], QPROGRAM]:
+    """Applies a variable amount of folding to select layers of a circuit.
+
+    Args:
+        layer_index: The layer of the circuit to apply folding to.
+
+    Returns:
+        fold_ith_layer: The function for folding the ith layer.
+    """
```
natestemen, line 80: This function does not apply the folding, rather it returns the function to do folding. I think the docstring could be clarified, and might be worth mentioning this is intended to be used in something like ZNE. Doesn't need to be long.

mitiq/zne/scaling/layer_scaling.py
```diff
@@ -0,0 +1,107 @@
...
+            scale_factor: Factor to scale the circuit by.
+
+        Returns:
+            layer_folding: the folded quantum circuit.
+        """
+        layers = [0] * len(circuit)
+        num_folds = (scale_factor - 1) / 2
+        if np.isclose(num_folds, int(num_folds)):
+            num_folds = int(num_folds)
+        layers[layer_index] = int(num_folds)
```
natestemen, line 99: It looks like `num_folds` is getting cast as an int regardless of the situation here. Is that intended? If so, lines 97/98 can likely be cut.

mitiq/zne/scaling/layer_scaling.py
```diff
@@ -0,0 +1,107 @@
...
+
+        _check_foldable(circuit)
```
natestemen, line 101: Might as well do this before everything else. I wouldn't expect it to affect timing since the above computation should be fast, but conceptually we should check foldability first.

mitiq/zne/tests/test_zne.py
```diff
@@ -31,6 +31,7 @@
     fold_gates_at_random,
     insert_id_layers,
 )
+from mitiq.zne.scaling.layer_scaling import get_layer_folding
```
natestemen, line 34: If we add `get_layer_folding` to the `scaling/__init__.py` file, we can add this import to the group directly above.

mitiq/zne/tests/test_zne.py
```diff
@@ -492,6 +493,34 @@ def test_execute_with_zne_with_supported_circuits(circuit_type):
...
+    # Layerwise-fold
+    layer_to_fold = 0
+    fold_layer_func = get_layer_folding(layer_to_fold)
+
+    zne_value = execute_with_zne(
+        circuit, generic_executor, factory=fac, scale_noise=fold_layer_func
+    )
+
+    # Test zero noise limit is better than unmitigated expectation value
+    assert abs(unmitigated - expected) > abs(zne_value - expected)
```
natestemen, line 521: I was a little worried this might not be deterministic enough, but I ran the test 2k times and it didn't fail, so that's good enough for me!

mitiq/zne/scaling/tests/test_layer_scaling.py
```diff
@@ -0,0 +1,88 @@
...
+# You should have received a copy of the GNU General Public License
+# along with this program.  If not, see <https://www.gnu.org/licenses/>.
+
+"""Unit tests for scaling by layer."""
+from cirq import (
+    Circuit,
+    LineQubit,
+    ops,
+)
+from mitiq.zne.scaling.layer_scaling import layer_folding
```
natestemen, line 22:
```suggestion
from mitiq.zne.scaling import layer_folding
```

This worked for me!

natestemen (approved):
Awesome work Vincent. Very complete addition to Mitiq. I hope adding the `get_layer_folding` function makes your research code a little cleaner.

I left a few comments, but happy for this to go out once addressed without an additional round of feedback.

mitiq/zne/scaling/layer_scaling.py
andreamari: Good catch! 
I think line 99 should be `layers[layer_index] = num_folds`  without the `int`. 
The idea is that inputs like `num_folds = 3` or  ` num_folds = 3.0 ` will be cast to `int` , and so the function will work.
On the other hand, something like `num_folds = 3.1` won't be cast to an int and so an error will occur as expected.

mitiq/zne/tests/test_zne.py
vprusso: Haha, appreciate the check!

mitiq/zne/tests/test_zne.py
vprusso: Good call. Changed!

docs/source/refs.bib
vprusso: Done!

mitiq/zne/scaling/layer_scaling.py
vprusso: I put in a concise description. Let me know if that's not what you had in mind.

mitiq/zne/scaling/layer_scaling.py
vprusso: Definitely makes sense. Changed!

mitiq/zne/scaling/layer_scaling.py
vprusso: Makes sense. Changed!

vprusso:
Final ping to @natestemen before merging. I believe all of the above items have been addressed but wanted to ensure before pressing the big green button.

mitiq/zne/scaling/layer_scaling.py
```diff
@@ -0,0 +1,109 @@
...
+) -> Callable[[QPROGRAM, float], QPROGRAM]:
+    """Return function to perform folding. The function return can be used as
+    an argument to define the noise scaling within the `execute_with_zne`
+    function.
+
+    Args:
+        layer_index: The layer of the circuit to apply folding to.
+
+    Returns:
+        fold_ith_layer: The function for folding the ith layer.
```
natestemen, line 81:
```suggestion
    Returns:
        The function for folding the ith layer.
```

mitiq/zne/scaling/layer_scaling.py
```diff
@@ -0,0 +1,109 @@
...
+        circuit: cirq.Circuit, scale_factor: int
+    ) -> cirq.Circuit:
+        """Returns a circuit folded according to integer scale factors.
+
+        Args:
+            circuit: Circuit to fold.
+            scale_factor: Factor to scale the circuit by.
+
+        Returns:
+            layer_folding: the folded quantum circuit.
```
natestemen, line 95:
```suggestion
        Returns:
            The folded quantum circuit.
```

natestemen (approved):
Congrats again Vincent! 🎉

Left two minor comments to help fix the API-doc. Also, no need to tell the user the internal variable name we use for the returned object.

| before | after |
| ------ | ----- |
| Screenshot 2023-04-28 at 2 43 34 PM | Screenshot 2023-04-28 at 2 43 40 PM |

vprusso:
Awesome! Merging post-changes. Thanks again for all of the help @natestemen and @andreamari !

vprusso:
Build is failing but it seems to be something unrelated to the contributions of this diff? Any thoughts @andreamari @natestemen. Want to avoid pushing broken code

natestemen:
Looks like an unrelated failure, but I've triggered it to build again to ensure this is the case.

andreamari:
Tests are green now and so we can finally merge! 
Congrats @vprusso!  :partying_face:

vprusso:
Woo-hoo! Thank you for all the help again, @andreamari and @natestemen !

PR #1778, opened 2023-04-07
Use Calibration workflow for Quickstart guide

natestemen:
## Description

This PR has two components:

1. updating the README to use the calibration workflow
2. remove the requirement that a user pass an expectation value executor when using `execute_with_mitigation`. This requirement isn't 100% needed, and removing it simplifies the workflow by quite a bit. The user no longer needs to define two executors in order to use calibration.

### Testing

I've run the code in the quickstart guide and it works!

### DIFF
--- a/README.md
+++ b/README.md
@@ -3,6 +3,7 @@
 <!-- ALL-CONTRIBUTORS-BADGE:START - Do not remove or modify this section -->
 [![All Contributors](https://img.shields.io/badge/all_contributors-34-orange.svg?style=flat-square)](#contributors-)
 <!-- ALL-CONTRIBUTORS-BADGE:END -->
+
 [![build](https://github.com/unitaryfund/mitiq/actions/workflows/build.yml/badge.svg?branch=master)](https://github.com/unitaryfund/mitiq/actions)
 [![Documentation Status](https://readthedocs.org/projects/mitiq/badge/?version=stable)](https://mitiq.readthedocs.io/en/stable/)
 [![codecov](https://codecov.io/gh/unitaryfund/mitiq/branch/master/graph/badge.svg)](https://codecov.io/gh/unitaryfund/mitiq)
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -41,28 +42,38 @@ Define a function which inputs a circuit and returns an expectation value you wa
...
+
+from mitiq import Calibrator, MeasurementResult, benchmarks
+
 
+def execute(circuit, noise_level=0.001):
+    """Returns measurement statistics from circuit with amplitude damping noise."""
+    circuit = circuit.with_noise(cirq.amplitude_damp(noise_level))
+    result = cirq.DensityMatrixSimulator().run(circuit, repetitions=1000)
+    bitstrings = np.column_stack(list(result.measurements.values()))
+    return MeasurementResult(bitstrings)
```
andreamari, line 55: This is a simple way to define the expectation value executor (if we need it).

```suggestion
def execute(circuit, noise_level=0.001):
    """Returns measurement results from circuit with amplitude damping noise."""
    circuit = circuit.with_noise(cirq.amplitude_damp(noise_level))
    result = cirq.DensityMatrixSimulator().run(circuit, repetitions=1000)
    bitstrings = np.column_stack(list(result.measurements.values()))
    return MeasurementResult(bitstrings)
    
def get_expval(circuit):
    """Returns the expectation value of |0⟩⟨0|."""
    execute(circuit).prob_distribution()["0"]
```

andreamari (commented):
Thanks @natestemen, 

As a first comment, I would remark that we are replacing the _getting started_ from something simple and well tested (ZNE), with something more complex and less tested (Calibrator).  I am not against it, especially if this is the long-term vision form Mitiq. I just want to be sure we are aware of the risk.

If we go for it, my main comment is that assuming the "0...00" bitstring as default observable works only for an RB-like circuit. I expect users will copy and paste the code of the README and replace the `circuit` with their problem-specific circuit. In this case the code will run without errors but it will give strange results because of the "0...00" hidden assumption.
In the code suggestions below I wrote a minimal expectation value executor which is probably not-too-bad and which would allow us to remove the dangerous "0...00" assumption.

README.md
```diff
@@ -44,53 +47,70 @@ import cirq
...
 ```
-Error (w/o  Mitiq): 0.0688
-Error (with Mitiq): 0.0002
+Error w/o  Mitiq: 0.264
+Error w Mitiq:    0.073
 ```
 
+### Calibration
+
+Unsure what error mitigation technique or parameters to use?
```
Misty-W, line 81:
```suggestion
Unsure which error mitigation technique or parameters to use?
```

README.md
```diff
@@ -44,53 +47,70 @@ import cirq
...
+Try out the calibration module demonstrated below to help find the best parameters for you particular backend!
+
+<video width="520" height="400" controls>
```
Misty-W, line 84: Great to embed the video! Unfortunately I don't see it when I view this version of the readme.

Misty-W:
> If we go for it, my main comment is that assuming the "0...00" bitstring as default observable works only for an RB-like circuit. I expect users will copy and paste the code of the README and replace the `circuit` with their problem-specific circuit. In this case the code will run without errors but it will give strange results because of the "0...00" hidden assumption. In the code suggestions below I wrote a minimal expectation value executor which is probably not-too-bad and which would allow us to remove the dangerous "0...00" assumption.

@natestemen, is it safe to assume this no longer applies because the Quickstart still uses ZNE?

README.md
natestemen: Good catch, I replaced it with an uploaded movie in [513392c](https://github.com/unitaryfund/mitiq/pull/1778/commits/513392c697fabf68b4bf77eb8eecf5a133dfa561) which works on github, but not RTD. Using a gif ([48d84d2](https://github.com/unitaryfund/mitiq/pull/1778/commits/48d84d2de9c74e49b0489ff24a12e7016011618a)), seems to work for both.

mitiq/calibration/calibrator.py
```diff
@@ -252,34 +254,50 @@ def best_strategy(self) -> Strategy:
     def execute_with_mitigation(
         self,
         circuit: QPROGRAM,
-        expval_executor: Union[Executor, Callable[[QPROGRAM], QuantumResult]],
+        expval_executor: Optional[
+            Union[Executor, Callable[[QPROGRAM], QuantumResult]]
+        ] = None,
```
andreamari, line 259: Having `None` here as default is dangerous since it is easy to miss the fact that there is a hidden default bitstring under the hood. Having `expval_executor` as a required argument seems safer to me.

andreamari (approved):
I still think that the hidden default bitstring 00....0 is dangerous (even if now this fact is not used in the README).

The README is very good!

Misty-W (approved):

PR #1780, opened 2023-04-07
Remove survey link from README

andreamari:
Remove survey


natestemen (approved):
Thanks Andrea, good catch.

PR #1785, opened 2023-04-13
W State and QPE Benchmarking circuits in API Doc

purva-thakre:
https://github.com/unitaryfund/mitiq/issues/1597#issuecomment-1507281611

### DIFF
--- a/docs/source/apidoc.md
+++ b/docs/source/apidoc.md
@@ -35,6 +35,18 @@
    :members:
 ```
 
+### Quantum Phase Estimation Circuits
+```{eval-rst}
+.. automodule:: mitiq.benchmarks.qpe_circuits
+   :members:
+```
+
+### W State Circuits
+```{eval-rst}
+.. automodule:: mitiq.benchmarks.w_state_circuits
+   :members:
+```
+
 ## Circuit types and result types
 
 ```{eval-rst}


purva-thakre:
@natestemen This is ready. For whatever reason, wiley link for one of the articles was causing some issues. So, I changed it to the arxiv link.

There's one test that's failing due to codecov.

natestemen:
Thanks for the quick addition! Looks good, but also looks like we need to change the formatting of the docstring for [`generate_w_circuit`](https://github.com/unitaryfund/mitiq/blob/7302c8b48821f86a796df533abff6c023c36e49e/mitiq/benchmarks/w_state_circuits.py#L28) so that sphinx is happy. Additionally, there should be a period at the end of this sentence:
https://github.com/unitaryfund/mitiq/blob/7302c8b48821f86a796df533abff6c023c36e49e/mitiq/benchmarks/w_state_circuits.py#L16-L17
Can we remove the space starting the string as well?

purva-thakre:
> Thanks for the quick addition! Looks good, but also looks like we need to change the formatting of the docstring for [`generate_w_circuit`](https://github.com/unitaryfund/mitiq/blob/7302c8b48821f86a796df533abff6c023c36e49e/mitiq/benchmarks/w_state_circuits.py#L28) so that sphinx is happy. Additionally, there should be a period at the end of this sentence:
> 
> https://github.com/unitaryfund/mitiq/blob/7302c8b48821f86a796df533abff6c023c36e49e/mitiq/benchmarks/w_state_circuits.py#L16-L17
> 
> 
> Can we remove the space starting the string as well?

That's strange. I did not see this when I built the docs locally.

natestemen (approved):
Thanks for cleaning up quickly! 🚢

Issue #1774, opened 2023-04-05
Pauli Twirling of CNOT Gate


nathanshammah:
## Issue Description
It could be useful to include tools to implement randomized compiling (RC), as noted already in #1245, as the technique is proving useful and widely adopted. 
## Proposed Solution
A simple approach could be that of applying a subset of the fully general technique, limiting to applying RC to CNOT gates with the introduction of Pauli rotations. 

<img width="743" alt="randomized-compiling" src="https://user-images.githubusercontent.com/14573436/230027966-5d6caf54-c313-4c85-9506-1d1da81212e2.png">

Pseudocode below
```
mitiq.rcc.execute_with_randomized_compiling(circuit, executor, rule="cnot")
```
where the protocol is similar to digital dynamical decoupling (DDD): it finds, based on a rule, the gates to substitute. It has a difference with respect to DDD, more similar to ZNE or PEC, in the sense that it generates multiple circuits based on one original and includes an extrapolation part (simple average) after the execution.

## Additional References
- "Hypothesis Testing for Error Mitigation: How to Evaluate Error Mitigation" https://arxiv.org/abs/2301.02690; the slides (relevant snapshot above) from @natestemen talk in the journal club of [Quantum Wednesday](https://github.com/unitaryfund/quantum-wednesday). 

- "Mitigating depolarizing noise on quantum computers with noise-estimation circuits", https://arxiv.org/abs/2103.08591

purva-thakre:
I can take this as part of [my previous attempt at a RFC](https://docs.google.com/document/d/1TORbBhvJfSaPykAmqX2BersOQHwiUD2Gn3Aum4rDa2g/edit?usp=sharing). 

I have got one issue that I am currently working on and after that I was planning on linking a newer version of the randomized compiling RFC anyway. This was my plan for the upcoming milestone.

nathanshammah:
Focusing sounds great to me, @purva-thakre!

purva-thakre:
@nathanshammah What was your motivation behind using `rule="cnot"` ? 

To me, it looks like you want to apply randomized compiling to the CNOT gate only ? Or just 1 layer in the circuit ?

Misty-W:
Consider reusing PEC modules- idea of replacing a gate. 
@natestemen, @purva-thakre, @Aaron-Robertson, @andreamari

PR #1802, opened 2023-04-25
CNOT twirling

natestemen:
## Description

Add ability to execute circuits with Pauli twirling on CNOT gates as suggested in https://github.com/unitaryfund/mitiq/issues/1774.

fixes https://github.com/unitaryfund/mitiq/issues/1774

---

### Notes until ready for review:

So far I have only tested simple cases with `add_paulis` which was made very simple using Cirq's [`map_operations`](https://quantumai.google/reference/python/cirq/map_operations) function.

Currently the `sample_paulis` function is executed _inside_ `twirl_CNOT`. I can imagine it being useful to have this outside, and $P, Q, R$ and $S$ passed in as variables. It might separate the logic a little more in case we wanted to modify how these unitaries are sampled as the circuits are generated. If not, this is probably unnecessary.

#### TODO:
- [x] docstrings
- [x] tests
- [x] `sample_paulis` implementation / table
- [ ] some sort of `check_twirlable` function (_can do this later_, no rush)
- [x] mypy issue (haven't been able to figure out the difference between `floating[Any]` and `float`)

### DIFF
--- a/mitiq/pt/__init__.py
+++ b/mitiq/pt/__init__.py
@@ -5,5 +5,6 @@
 
 from mitiq.pt.pt import (
     execute_with_pauli_twirling,
-    sample_paulis,
+    twirl_CNOT_gates,
+    twirl_CZ_gates,
 )
--- a/mitiq/pt/__init__.py
+++ b/mitiq/pt/__init__.py
@@ -1,0 +1,6 @@
+# Copyright (C) Unitary Fund
+#
+# This source code is licensed under the GPL license (v3) found in the
+# LICENSE file in the root directory of this source tree.
+
+from mitiq.pt.pt import execute_with_pauli_twirling
--- a/mitiq/pt/__init__.py
+++ b/mitiq/pt/__init__.py
@@ -5,6 +5,5 @@
 
 from mitiq.pt.pt import (
     execute_with_pauli_twirling,
-    _generate_lookup_table,
     sample_paulis,
 )
--- a/mitiq/pt/__init__.py
+++ b/mitiq/pt/__init__.py
@@ -3,4 +3,4 @@
 # This source code is licensed under the GPL license (v3) found in the
 # LICENSE file in the root directory of this source tree.
 
-from mitiq.pt.pt import execute_with_pauli_twirling
+from mitiq.pt.pt import execute_with_pauli_twirling, _generate_lookup_table, sample_paulis
--- a/mitiq/pt/__init__.py
+++ b/mitiq/pt/__init__.py
@@ -3,4 +3,8 @@
 # This source code is licensed under the GPL license (v3) found in the
 # LICENSE file in the root directory of this source tree.
 
-from mitiq.pt.pt import execute_with_pauli_twirling, _generate_lookup_table, sample_paulis
+from mitiq.pt.pt import (
+    execute_with_pauli_twirling,
+    _generate_lookup_table,
+    sample_paulis,
+)
--- a/mitiq/pt/pt.py
+++ b/mitiq/pt/pt.py
@@ -77,7 +77,10 @@ def execute_with_pauli_twirling(
     executor = (
         executor if isinstance(executor, Executor) else Executor(executor)
     )
-    twirled_circuits = twirl_CNOT_gates(circuit, num_circuits)
+    CNOT_twirled_circuits = twirl_CNOT_gates(circuit, num_circuits)
+    twirled_circuits = [
+        twirl_CZ_gates(c, num_circuits=1)[0] for c in CNOT_twirled_circuits
+    ]
     expvals = executor.evaluate(twirled_circuits, observable)
     return cast(float, np.average(expvals))
 
--- a/mitiq/pt/pt.py
+++ b/mitiq/pt/pt.py
@@ -58,7 +58,8 @@ def execute_with_pauli_twirling(
     *,
     num_circuits: int = 10,
 ) -> float:
-    """Estimates the expecation value by averaging pauli twirled circuits.
+    """Estimates the expectation value of the input circuit by averaging
+    expectation values obtained from Pauli twirled circuits.
 
     Args:
         circuit: The input circuit to execute with twirling.
@@ -112,8 +113,8 @@ def _twirl_CZ_qprogram(circuit: cirq.Circuit) -> cirq.Circuit:
 
 
 def _twirl_single_CNOT_gate(op: cirq.Operation) -> cirq.OP_TREE:
-    """Function which converts a CNOT to a logical equivalent. That is, it
-    converts CNOT operations into the following, leaving all other cirq
+    """Function which converts a CNOT gate to a logical equivalent. That is,
+    it converts CNOT operations into the following, leaving all other cirq
     operations alone.
 
         --P---⏺---R--
@@ -138,8 +139,8 @@ def _twirl_single_CNOT_gate(op: cirq.Operation) -> cirq.OP_TREE:
 
 
 def _twirl_single_CZ_gate(op: cirq.Operation) -> cirq.OP_TREE:
-    """Function which converts a CNOT to a logical equivalent. That is, it
-    converts CNOT operations into the following, leaving all other cirq
+    """Function which converts a CZ gate to a logical equivalent. That is, it
+    converts CZ operations into the following, leaving all other cirq
     operations alone.
 
         --P---⏺---R--
--- a/mitiq/pt/pt.py
+++ b/mitiq/pt/pt.py
@@ -31,23 +31,22 @@ CNOT_twirling_gates = [
     (cirq.Z, cirq.Y, cirq.I, cirq.Y),
     (cirq.Z, cirq.Z, cirq.I, cirq.Z),
 ]
-# TODO: check the CZ gates. should be in PQRS order
 CZ_twirling_gates = [
     (cirq.I, cirq.I, cirq.I, cirq.I),
-    (cirq.X, cirq.I, cirq.I, cirq.X),
-    (cirq.Y, cirq.I, cirq.I, cirq.Y),
-    (cirq.Z, cirq.I, cirq.I, cirq.Z),
-    (cirq.I, cirq.Z, cirq.Z, cirq.I),
-    (cirq.X, cirq.Z, cirq.Z, cirq.X),
-    (cirq.Y, cirq.Z, cirq.Z, cirq.Y),
-    (cirq.Z, cirq.Z, cirq.Z, cirq.Z),
-    (cirq.I, cirq.Y, cirq.Y, cirq.I),
+    (cirq.I, cirq.X, cirq.Z, cirq.X),
+    (cirq.I, cirq.Y, cirq.Z, cirq.Y),
+    (cirq.I, cirq.Z, cirq.I, cirq.Z),
+    (cirq.X, cirq.I, cirq.X, cirq.Z),
+    (cirq.X, cirq.X, cirq.Y, cirq.Y),
     (cirq.X, cirq.Y, cirq.Y, cirq.X),
-    (cirq.Y, cirq.Y, cirq.Y, cirq.Y),
-    (cirq.Z, cirq.Y, cirq.Y, cirq.Z),
-    (cirq.I, cirq.Z, cirq.Z, cirq.I),
-    (cirq.X, cirq.Z, cirq.Z, cirq.X),
-    (cirq.Y, cirq.Z, cirq.Z, cirq.Y),
+    (cirq.X, cirq.Z, cirq.X, cirq.I),
+    (cirq.Y, cirq.I, cirq.Y, cirq.Z),
+    (cirq.Y, cirq.X, cirq.X, cirq.Y),
+    (cirq.Y, cirq.Y, cirq.X, cirq.X),
+    (cirq.Y, cirq.Z, cirq.Y, cirq.I),
+    (cirq.Z, cirq.I, cirq.Z, cirq.I),
+    (cirq.Z, cirq.X, cirq.I, cirq.X),
+    (cirq.Z, cirq.Y, cirq.I, cirq.Y),
     (cirq.Z, cirq.Z, cirq.Z, cirq.Z),
 ]
 
@@ -104,9 +103,7 @@ def twirl_CZ_gates(circuit: QPROGRAM, num_circuits: int) -> List[QPROGRAM]:
         circuit: The circuit to generate twirled versions of
         num_circuits: The number of sampled circuits to return
     """
-    return [
-        _twirl_CZ_qprogram(circuit, num_circuits) for _ in range(num_circuits)
-    ]
+    return [_twirl_CZ_qprogram(circuit) for _ in range(num_circuits)]
 
 
 @noise_scaling_converter
--- a/mitiq/pt/pt.py
+++ b/mitiq/pt/pt.py
@@ -1,0 +1,61 @@
+# Copyright (C) Unitary Fund
+#
+# This source code is licensed under the GPL license (v3) found in the
+# LICENSE file in the root directory of this source tree.
+
+from typing import Callable, List, Optional, Tuple, Union
+
+import cirq
+import numpy as np
+
+from mitiq import QPROGRAM, Executor, Observable, QuantumResult
+from mitiq.interface import atomic_converter
+
+
+def execute_with_pauli_twirling(
+    circuit: QPROGRAM,
+    executor: Union[Executor, Callable[[QPROGRAM], QuantumResult]],
+    observable: Optional[Observable] = None,
+    *,
+    num_circuits: int = 100,
+) -> float:
+    executor = (
+        executor if isinstance(executor, Executor) else Executor(executor)
+    )
+    twirled_circuits = pauli_twirled_circuits(circuit, num_circuits)
+    expvals = executor.evaluate(twirled_circuits, observable)
+    return np.average(expvals)
+
+
+def pauli_twirled_circuits(
+    circuit: QPROGRAM, num_circuits: int
+) -> List[QPROGRAM]:
+    return [add_paulis(circuit) for _ in range(num_circuits)]
+
+
+Paulis = Union[cirq.IdentityGate, cirq.XPowGate, cirq.YPowGate, cirq.ZPowGate]
+
+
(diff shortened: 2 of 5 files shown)


mitiq/pt/pt.py
```diff
@@ -0,0 +1,61 @@
...
+) -> List[QPROGRAM]:
+    return [add_paulis(circuit) for _ in range(num_circuits)]
+
+
+Paulis = Union[cirq.IdentityGate, cirq.XPowGate, cirq.YPowGate, cirq.ZPowGate]
+
+
+def sample_paulis() -> Tuple[Paulis, Paulis, Paulis, Paulis]:
+    # P, Q, R, S
+    return cirq.ops.X, cirq.ops.I, cirq.ops.X, cirq.ops.X
```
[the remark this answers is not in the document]
natestemen: Yup! This is just a stub so I could test other stuff.

mitiq/pt/pt.py
```diff
@@ -0,0 +1,61 @@
...
+from mitiq import QPROGRAM, Executor, Observable, QuantumResult
+from mitiq.interface import atomic_converter
+
+
+def execute_with_pauli_twirling(
+    circuit: QPROGRAM,
+    executor: Union[Executor, Callable[[QPROGRAM], QuantumResult]],
+    observable: Optional[Observable] = None,
+    *,
+    num_circuits: int = 100,
```
andreamari, line 20: In principle this is a good number but it will make experiments `100` times slower. Maybe `10` is a better compromise?

mitiq/pt/pt.py
```diff
@@ -0,0 +1,61 @@
...
+@atomic_converter
+def add_paulis(circuit: cirq.Circuit) -> cirq.Circuit:
+    return circuit.map_operations(twirl_CNOT)
+
+
+def twirl_CNOT(op: cirq.Operation) -> cirq.OP_TREE:
+    if op.gate != cirq.CNOT:
+        return op
+
+    P, Q, R, S = sample_paulis()
```
andreamari, line 53: I guess `sample_paulis()` will sample a single pair Paulis and deduce the other pair assuming they will be applied to a CNOT gate.

It would be conceptually a bit clearer if such "deduction" happened inside `twirl_CNOT` such that `sample_paulis()` can simply return 2 to random pauli gates.

mitiq/pt/pt.py
```diff
@@ -0,0 +1,61 @@
...
+    return cirq.ops.X, cirq.ops.I, cirq.ops.X, cirq.ops.X
+
+
+@atomic_converter
+def add_paulis(circuit: cirq.Circuit) -> cirq.Circuit:
+    return circuit.map_operations(twirl_CNOT)
+
+
+def twirl_CNOT(op: cirq.Operation) -> cirq.OP_TREE:
+    if op.gate != cirq.CNOT:
```
andreamari, line 50: We could twirl CZ gates too, in a similar way.
If so `twirl_CNOT`--> `twirl_CNOT_and_CZ` or it could be a new function.

mitiq/pt/pt.py
natestemen: Ah I was thinking of a slightly more brute force approach, which is how I interpret what happens in this [reference](https://arxiv.org/pdf/2301.02690.pdf) which is to choose $P, Q, R, S$ randomly from the following table.

Screenshot 2023-04-26 at 6 30 35 AM

This is exactly what you suggest, but pre-computed. Do you have a preference? I'm not actually sure how I would go about computing $R$ and $S$ given $P$ and $Q$ if we take that route. Do you have ideas there?

mitiq/pt/pt.py
natestemen: Yeah, Cirq's `map_operations` method makes this a pretty straightforward change. I think having an additional function for $\mathrm{C}Z$ (that could be combined with this one) would be great.

natestemen: @Aaron-Robertson I noticed you combined the `twirl_CNOT` function and the proposed `twirl_CZ` function. Seeing as this is a new technique that we don't fully understand, I think it might be best to keep them separate for now. Do you have a preference to keeping them as a single function?

mitiq/pt/pt.py
```diff
@@ -3,28 +3,64 @@
...
-    num_circuits: int = 100,
+    num_circuits: int = 10,
 ) -> float:
     executor = (
         executor if isinstance(executor, Executor) else Executor(executor)
     )
     twirled_circuits = pauli_twirled_circuits(circuit, num_circuits)
     expvals = executor.evaluate(twirled_circuits, observable)
-    return np.average(expvals)
+    return cast(float, np.average(expvals))
```
natestemen: I saw this mypy error as well, but I don't quite understand it.
```
mitiq/pt/pt.py:27: error: Incompatible return value type (got "floating[Any]", expected "float")  [return-value]
```
Do you know what can be classified as `floating[Any]` that's not a `float`?

Totally okay with having a `cast` call here, but I like to understand if it's needed.

mitiq/pt/pt.py
```diff
@@ -3,28 +3,64 @@
...
+from typing import Callable, List, Optional, Tuple, Union, cast
 
+import random
 import cirq
 import numpy as np
 
 from mitiq import QPROGRAM, Executor, Observable, QuantumResult
 from mitiq.interface import atomic_converter
 
+Paulis = Union[cirq.IdentityGate, cirq.XPowGate, cirq.YPowGate, cirq.ZPowGate]
```
natestemen, line 15: Looks like this type isn't being used anymore. I tried to be specific when defining this type, but perhaps it's not needed as `cirq.Gate` covers all the Pauli's (and more). What I've seen about type-hinting is "accept general, return specific" meaning be as general as you can type-hinting function arguments (that users may use), and as specific as possible type-hinting return types. WDYT?

mitiq/pt/__init__.py
```diff
@@ -3,4 +3,8 @@
 # This source code is licensed under the GPL license (v3) found in the
 # LICENSE file in the root directory of this source tree.
 
-from mitiq.pt.pt import execute_with_pauli_twirling
+from mitiq.pt.pt import (
+    execute_with_pauli_twirling,
+    _generate_lookup_table,
+    sample_paulis,
+)
```
natestemen, line 10: Probably don't need to expose the "private" `_generate_lookup_table` function at this level, right?

natestemen (commented):
Nice changes Aaron! Already getting pretty close to having a testable technique!

mitiq/pt/pt.py
Aaron-Robertson: I wasn't sure why; I just saw the same cast on `np.average()` in `scale_factor_to_expectation_value()` and `execute_with_pec()` and matched the pattern

mitiq/pt/pt.py
Aaron-Robertson: I found Paulis wasn't functioning as a type, so we didn't get any inheritance or annotation. I think if we want to define a specific return type we'll want to build it in cirq with [`GateFamily`](https://quantumai.google/reference/python/cirq/GateFamily)

mitiq/pt/pt.py
Aaron-Robertson: From a code perspective I like the single function, since we get code reusability into the future. As a new `mitiq` technique I certainly see where you're coming from. How about we keep what we have and hide the general case, while exposing the specific?

natestemen: > How about we keep what we have and hide the general case, while exposing the specific?

What do you mean by this? Can you clarify?

Throwing another idea out there, we could have a function `generate_twirl_func(gate) -> Callable[[cirq.Operation], cirq.OP_TREE]` which you could then use to create `twirl_CNOT = generate_twirl_func(cirq.CNOT)`, and `twirl_CZ = generate_twirl_func(cirq.CZ)`. Don't love this approach, but don't hate it either.

Aaron-Robertson: That's essentially what I meant, but I like the simpler call into the existing `_twirl_two_qubit_gate`:
```
def _twirl_two_qubit_gate(op: cirq.Operation) -> cirq.OP_TREE:...

def twirl_CZ() -> cirq.OP_TREE:
    return _twirl_two_qubit_gate(cirq.CZ)
    
def twirl_CNOT() -> cirq.OP_TREE:
    return _twirl_two_qubit_gate(cirq.CNOT)
```

mitiq/pt/pt.py
```diff
@@ -12,40 +12,44 @@
...
+    (cirq.X, cirq.I, cirq.X, cirq.X),
+    (cirq.X, cirq.X, cirq.X, cirq.I),
+    (cirq.X, cirq.Y, cirq.Y, cirq.Z),
+    (cirq.X, cirq.Z, cirq.Y, cirq.Y),
+    (cirq.Z, cirq.I, cirq.Z, cirq.I),
+    (cirq.Z, cirq.X, cirq.Z, cirq.X),
+    (cirq.Z, cirq.Y, cirq.I, cirq.Y),
+    (cirq.Z, cirq.Z, cirq.I, cirq.Z),
+]
+# TODO: check the CZ gates. should be in PQRS order
```
natestemen, line 34: I noticed when modifying the lookup tables that the gates were not in $P, Q, R, S$ order. I think that makes most sense since that's what the reference we are referring to uses. Could you ensure the `CZ_twirlings_gates` object is defined in that order as well, and then fill in the `test_twirl_CZ` test as well?

mitiq/pt/pt.py
```diff
@@ -0,0 +1,111 @@
...
+        executor if isinstance(executor, Executor) else Executor(executor)
+    )
+    twirled_circuits = twirl_CNOT_gates(circuit, num_circuits)
+    expvals = executor.evaluate(twirled_circuits, observable)
+    return cast(float, np.average(expvals))
+
+
+def twirl_CNOT_gates(circuit: QPROGRAM, num_circuits: int) -> List[QPROGRAM]:
+    twirl_qprogram = atomic_converter(
+        lambda circuit: circuit.map_operations(_twirl_single_CNOT_gate)
```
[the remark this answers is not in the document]
natestemen: I'm pretty sure this problem is unavoidable in the current state of things when using `atomic_converter` since it destroys most information about typehints anyway:
https://github.com/unitaryfund/mitiq/blob/4e94c995816aa372402f16e6c104a33dac3ac6b5/mitiq/interface/conversions.py#L155-L157

Even with your suggestion `twirl_qprogram` is `(...) -> Any`. 😞

mitiq/pt/pt.py
```diff
@@ -0,0 +1,111 @@
...
+    )
+    return [twirl_qprogram(circuit) for _ in range(num_circuits)]
+
+
+def twirl_CZ_gates(circuit: QPROGRAM, num_circuits: int) -> List[QPROGRAM]:
+    twirl_qprogram = atomic_converter(
+        lambda circuit: circuit.map_operations(_twirl_single_CZ_gate)
```
andreamari, line 79:
```suggestion
        lambda c: c.map_operations(_twirl_single_CZ_gate)
```
to avoid confusion with the `circuit` argument of `twirl_CZ_gates()`

mitiq/pt/pt.py
```diff
@@ -0,0 +1,111 @@
...
+    )
+    return [twirl_qprogram(circuit) for _ in range(num_circuits)]
```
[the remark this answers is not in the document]
andreamari: Fixed by @Aaron-Robertson  in [de50806]. Thanks !

Misty-W:
hi @natestemen, any help needed getting this PR ready for review?

mitiq/pt/pt.py
```diff
@@ -0,0 +1,163 @@
...
+            is used to compute the expectation of the observable.
+        num_circuits: Number of circuits to be randomly twirled, and averaged.
+
+    Returns:
+        The expectation value estimated with Pauli twirling.
+    """
+    executor = (
+        executor if isinstance(executor, Executor) else Executor(executor)
+    )
+    twirled_circuits = twirl_CNOT_gates(circuit, num_circuits)
```
Misty-W, line 79: Why is only `twirl_CNOT_gates` used in `execute_with_pauli_twirling` and not `twirl_CZ_gates`?

mitiq/pt/pt.py
```diff
@@ -0,0 +1,163 @@
...
+
+    Args:
+        circuit: The input circuit to execute with twirling.
+        executor: A Mitiq executor that executes a circuit and returns the
+            unmitigated ``QuantumResult`` (e.g. an expectation value).
+        observable: Observable to compute the expectation value of. If
+            ``None``, the ``executor`` must return an expectation value
+            (float). Otherwise, the ``QuantumResult`` returned by ``executor``
+            is used to compute the expectation of the observable.
+        num_circuits: Number of circuits to be randomly twirled, and averaged.
```
Misty-W, line 71: What is meant by "randomly twirled"?

mitiq/pt/pt.py
```diff
@@ -0,0 +1,163 @@
...
+
+
+def execute_with_pauli_twirling(
+    circuit: QPROGRAM,
+    executor: Union[Executor, Callable[[QPROGRAM], QuantumResult]],
+    observable: Optional[Observable] = None,
+    *,
+    num_circuits: int = 10,
+) -> float:
+    """Estimates the expecation value by averaging pauli twirled circuits.
```
Misty-W, line 61:
```suggestion
    """Estimates the expectation value from the input circuit by averaging the expectation values obtained from Pauli twirled circuits.
```

mitiq/pt/pt.py
```diff
@@ -0,0 +1,163 @@
...
+        P.on(control_qubit),
+        Q.on(target_qubit),
+        op,
+        R.on(control_qubit),
+        S.on(target_qubit),
+    ]
+
+
+def _twirl_single_CZ_gate(op: cirq.Operation) -> cirq.OP_TREE:
+    """Function which converts a CNOT to a logical equivalent. That is, it
```
Misty-W, line 141:
```suggestion
    """Function which converts a CZ gate to a logical equivalent. That is, it
```

mitiq/pt/pt.py
```diff
@@ -0,0 +1,163 @@
...
+    converts CNOT operations into the following, leaving all other cirq
```
Misty-W, line 142:
```suggestion
    converts CZ operations into the following, leaving all other cirq
```

mitiq/pt/pt.py
```diff
@@ -0,0 +1,163 @@
...
+    return [_twirl_CZ_qprogram(circuit) for _ in range(num_circuits)]
+
+
+@noise_scaling_converter
+def _twirl_CZ_qprogram(circuit: cirq.Circuit) -> cirq.Circuit:
+    return circuit.map_operations(_twirl_single_CZ_gate)
+
+
+def _twirl_single_CNOT_gate(op: cirq.Operation) -> cirq.OP_TREE:
+    """Function which converts a CNOT to a logical equivalent. That is, it
```
Misty-W, line 115:
```suggestion
    """Function which converts a CNOT gate to a logical equivalent. That is, it
```

Misty-W (commented):
Nice work @natestemen, great to have twirling in this Mitiq release! I have some minor comments and questions to address before merge.

Misty-W (commented):
Nice work @natestemen, great to have twirling in this Mitiq release! I have some minor comments and questions to address before merge.

mitiq/pt/pt.py
```diff
@@ -0,0 +1,163 @@
...
+              |
+        --Q---⏺---S--
+
+    Where P, Q, R, and S are Pauli operators sampled so as to not effect
+    change on the underlying unitary.
+    """
+    if op.gate != cirq.CZ:
+        return op
+
+    P, Q, R, S = random.choice(CZ_twirling_gates)
```
Misty-W, line 155: Any idea why Codecov complains about `_twirl_single_CZ_gate` but not `_twirl_single_CZ_gate`? 

IMO we don't have to resolve it on this PR, it just raised my curiousity.

mitiq/pt/pt.py
natestemen: Good catch. I added CZ twirling in https://github.com/unitaryfund/mitiq/pull/1802/commits/4d1e9b1cfaadfd19d48026f3f16877b37a2fa3a9.

mitiq/pt/pt.py
natestemen: I think I was trying to emphasize that there is some random sampling that occurs, but I don't think it actually does that, or clarifies. I've removed "randomly" in https://github.com/unitaryfund/mitiq/pull/1802/commits/6b873326f04d90b1d64f3c645f92eb0bab5480a7.

mitiq/pt/pt.py
purva-thakre: At first glance, it looks like there's only 1 test for `twirl_CZ_gates` and 3 additional tests for `twirl_CNOT_gates`. 

Maybe Codecov's issue is related to those additional tests? Haven't looked at it in more detail though.

andreamari (approved):
LGTM, nice teamwork!

mitiq/pt/tests/test_pt.py
```diff
@@ -0,0 +1,127 @@
...
+    CZ_twirling_gates,
+)
+from mitiq.interface.mitiq_cirq import compute_density_matrix
+from mitiq.benchmarks import generate_mirror_circuit
+
+num_qubits = 2
+qubits = cirq.LineQubit.range(num_qubits)
+circuit = cirq.Circuit()
+circuit.append(cirq.CNOT.on_each(zip(qubits, qubits[1:])))
+
```
andreamari, line 27: This should fix coverage errors for CZ functions:

```suggestion
circuit.append(cirq.CNOT.on_each(zip(qubits, qubits[1:])))
circuit.append(cirq.CZ.on_each(zip(qubits, qubits[1:])))
```
or,  maybe equivalently:

```python
circuit.append(cirq.CNOT(*qubits))
circuit.append(cirq.CZ(*qubits))
```

natestemen: Niiiice! Looks like that's the reason why coverage was complaining.

I originally wrote line 26 when I had `len(qubits) > 2` so calling `cirq.CNOT(*qubits)` does not work, but since it's just 2 I like your simplified approach. I'm not sure I even knew you could call the gates directly. I thought you had to use `cirq.GATE.on`. Good to know.

Misty-W (approved):
LGTM @natestemen!

PR #1803, opened 2023-04-25
use 1.7.0 link for grove

natestemen:
## Description

Had a spurious docs build failure on this link and noticed it was set for master, despite being a reference to a specific version of Grove. Bring link in line with text.


Misty-W (approved):
LGTM!

PR #1814, opened 2023-04-30
Supports observable multiplication with observable and PauliString

### LINKED ISSUE #1810: Support multiplication operator for Observable objects
## Issue Description

Support multiplication operator for `Observable` (and/or `PauliString`) objects. 

This is a sub-issue of the Quantum Subspace Expansion (QSE) technique #818 .

bubakazouba:
closes #1810 

## Description

Supports observable multiplication with observable, PauliString and numeric types.

I realized that combining duplicates would be necessary to ensure the compactness of the observable and efficiency of computing its expectation value when we multiply an observable by an observable. So I added a private `_combine_duplicates` function to be used after the full set of product PauliString is computed.

`_combine_duplicates` required PauliString to support a `with_coeff` method to allow us to normalize the states to a fixed coeff before computing the hash which will allow us to dedupe the PauliString's using a dictionary.

---

### License

### DIFF
--- a/mitiq/observable/observable.py
+++ b/mitiq/observable/observable.py
@@ -185,8 +185,8 @@ class Observable:
 
 
 def _combine_duplicate_pauli_strings(
-    paulis: list[PauliString],
-) -> list[PauliString]:
+    paulis: Iterable[PauliString],
+) -> List[PauliString]:
     """Combines duplicate PauliStrings by adding their coefficients.
     Discards paulis with zero coefficients.
 
--- a/mitiq/observable/observable.py
+++ b/mitiq/observable/observable.py
@@ -197,8 +197,7 @@ def _combine_duplicate_pauli_strings(
     )
     for pauli_string in paulis:
         cache_key = pauli_string.with_coeff(1)
-        new_coeff = pauli_string.coeff + pauli_string_coefficients[cache_key]
-        pauli_string_coefficients[cache_key] = new_coeff
+        pauli_string_coefficients[cache_key] += pauli_string.coeff
     return list(
         [
             pauli_string.with_coeff(coeff)
--- a/mitiq/observable/observable.py
+++ b/mitiq/observable/observable.py
@@ -82,10 +82,11 @@ class Observable:
         if isinstance(other, (PauliString, complex, float, int)):
             return Observable(*[pauli * other for pauli in self._paulis])
         elif isinstance(other, Observable):
-            all_paulis = []
-            for other_pauli in other._paulis:
-                all_paulis += [pauli * other_pauli for pauli in self._paulis]
-            new_observable = Observable(*all_paulis)
+            new_observable = Observable(
+                pauli * other_pauli
+                for pauli in self._paulis
+                for other_pauli in other._paulis
+            )
             return new_observable.combine_duplicates()
         return NotImplemented
 
--- a/mitiq/observable/observable.py
+++ b/mitiq/observable/observable.py
@@ -155,7 +155,13 @@ class Observable:
         return Executor(execute).evaluate(circuit, observable=self)[0]
 
     def _combine_duplicates(self) -> None:
-        pauli_string_coefficients = defaultdict(int)
+        """Modifies self._paulis in place.
+        Combines duplicate PauliStrings by adding their coefficients.
+        Discards paulis with zero coefficients.
+        """
+        pauli_string_coefficients: defaultdict[
+            PauliString, complex
+        ] = defaultdict(complex)
         for pauli_string in self._paulis:
             cache_key = pauli_string.with_coeff(1)
             new_coeff = (
--- a/mitiq/observable/observable.py
+++ b/mitiq/observable/observable.py
@@ -14,7 +14,7 @@
 # along with this program.  If not, see <https://www.gnu.org/licenses/>.
 
 import copy
-from typing import Callable, cast, List, Optional, Set, Union, Any
+from typing import Callable, cast, List, Optional, Set, Union, Any, Iterable
 
 import numpy as np
 import numpy.typing as npt
@@ -39,7 +39,7 @@ class Observable:
             paulis: PauliStrings used to define the observable.
 
         """
-        self._paulis = _combine_duplicate_pauli_strings(list(paulis))
+        self._paulis = _combine_duplicate_pauli_strings(paulis)
         self._groups: List[PauliStringCollection]
         self._ngroups: int
         self.partition()
@@ -198,10 +198,8 @@ def _combine_duplicate_pauli_strings(
     for pauli_string in paulis:
         cache_key = pauli_string.with_coeff(1)
         pauli_string_coefficients[cache_key] += pauli_string.coeff
-    return list(
-        [
-            pauli_string.with_coeff(coeff)
-            for (pauli_string, coeff) in pauli_string_coefficients.items()
-            if not np.isclose(coeff, 0.0)
-        ]
-    )
+    return [
+        pauli_string.with_coeff(coeff)
+        for (pauli_string, coeff) in pauli_string_coefficients.items()
+        if not np.isclose(coeff, 0.0)
+    ]
--- a/mitiq/observable/observable.py
+++ b/mitiq/observable/observable.py
@@ -14,12 +14,14 @@
 # along with this program.  If not, see <https://www.gnu.org/licenses/>.
 
 import copy
-from typing import Callable, cast, List, Optional, Set, Union, Dict
+from typing import Callable, cast, List, Optional, Set, Union, Any
 
 import numpy as np
 import numpy.typing as npt
 import cirq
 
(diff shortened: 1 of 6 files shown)


mitiq/observable/observable.py
```diff
@@ -73,6 +77,27 @@ def qubit_indices(self) -> List[int]:
...
+    def __mul__(
+        self, other: Union["Observable", "PauliString", complex, float, int]
+    ) -> "Observable":
+        if isinstance(other, (PauliString, complex, float, int)):
+            return Observable(*[pauli * other for pauli in self._paulis])
+        elif isinstance(other, Observable):
+            all_paulis = []
+            for other_pauli in other._paulis:
+                all_paulis += [pauli * other_pauli for pauli in self._paulis]
+            new_observable = Observable(*all_paulis)
```
bubakazouba, line 89: this line has been covered by the unit tests, i'm not sure why it's saying it's uncovered. (same for the one before this is just the continuation of the function definition)

mitiq/observable/observable.py
```diff
@@ -128,6 +153,23 @@ def expectation(
...
+                d[cache_key] = d[cache_key].with_coeff(new_coeff)
+            else:
+                d[cache_key] = pauli_string
+        self._paulis = list(
+            [
+                pauli_string
+                for pauli_string in d.values()
+                if pauli_string.coeff != 0
+            ]
+        )
```
andreamari, line 171: It is probably safer to `return` a new `Observable` based on the new combined `PauliString`s instead of mutating the internal private attribute `self._paulis`. In the first case `self.partition()` is called again by the __init__ method, in the second case it's not re-called and this seems dangerous.

mitiq/observable/observable.py
```diff
@@ -128,6 +153,23 @@ def expectation(
...
+            if cache_key in d:
+                new_coeff = pauli_string.coeff + d[cache_key].coeff
+                d[cache_key] = d[cache_key].with_coeff(new_coeff)
+            else:
+                d[cache_key] = pauli_string
+        self._paulis = list(
+            [
+                pauli_string
+                for pauli_string in d.values()
+                if pauli_string.coeff != 0
```
andreamari, line 169:
```suggestion
                if not np.isclose(pauli_string.coeff, 0.0)
```
This works better with floats and complex numbers.

mitiq/observable/pauli.py
```diff
@@ -146,10 +162,21 @@ def _expectation_from_measurements(
...
-        result = PauliString()
-        result._pauli = self._pauli * other._pauli
-        return result
+    def __mul__(
+        self, other: Union["PauliString", complex, float, int]
+    ) -> "PauliString":
+        if isinstance(other, PauliString):
+            result = PauliString()
+            result._pauli = self._pauli * other._pauli
+            return result
```
andreamari, line 171: Same comment as before:
Here the private variable`result._pauli` is changed but the `__init__` method of the `PauliString` object is not called.
Is this safe? If not, returning a new `PauliString` directly  initialized from the string associated to `self._pauli * other._pauli` may be necessary.

mitiq/observable/pauli.py
```diff
@@ -146,10 +162,21 @@ def _expectation_from_measurements(
...
+        elif isinstance(other, (complex, float, int)):
+            result = PauliString()
+            result._pauli = self._pauli * other
+            return result
+        return NotImplemented
+
+    def __rmul__(self, other: Union[complex, float, int]) -> "PauliString":
+        return self.__mul__(other)
```
andreamari, line 179: Good idea defining `__rmul__` for scalars!

andreamari (commented):
Very nice PR @bubakazouba !  All seems to work, but I'm a bit worried about changing private variables of some objects (see code comments).

Additional minor point: Shall we remove `# TODO: Add option to Combine duplicates. E.g. [Z(0, Z(0)] -> [2*Z(0)].`  now that we have a method for it?

mitiq/observable/pauli.py
bubakazouba: all resolved, good catch I didn't pay enough attention to `__init__`

natestemen (approved):
Nice work. Clean and easy to understand. A few comments, but again, awesome job.

mitiq/observable/observable.py
```diff
@@ -37,8 +39,8 @@ def __init__(self, *paulis: PauliString) -> None:
             paulis: PauliStrings used to define the observable.
 
         """
-        # TODO: Add option to Combine duplicates. E.g. [Z(0, Z(0)] -> [2*Z(0)].
         self._paulis = list(paulis)
+        self._combine_duplicates()
```
bubakazouba, line 43: I wasn't sure if its best to let this method modify `_paulis` in place or if I should move the method under some `utils` file and let it return a new list of paulis.

Modifying in place can be dangerous if gets used somewhere else within the class.

andreamari: I agree! 
It is safer if `_combine_duplicates()` is just some function from an input list of paulis to an output list of paulis.

bubakazouba: sounds good, there's no use of having this as a member method anymore, so I pulled it out of the class.

andreamari (approved):
Up to the small problem about failing tests,  looks good to me! 
Thanks  @bubakazouba (Abdulrahman Sahmoud) and thanks to your collaborators Tim (Yi-Ting) Chen, Ammar Jahin, Dariel Mok, Preksha Naik. Congrats for your first PR on Mitiq!

Tip: You should try to import import `List` from ` typing` instead of using `list`  in the function annotation. `list` was not supported as a type annotation in python 3.8 and we are trying to remain compatible with all versions $\ge$` 3.8`.

mitiq/observable/observable.py
```diff
@@ -154,3 +179,30 @@ def _expectation_from_density_matrix(
...
+
+    Returns: deduped list of PauliStrings.
+    """
+    pauli_string_coefficients: defaultdict[PauliString, complex] = defaultdict(
+        complex
+    )
+    for pauli_string in paulis:
+        cache_key = pauli_string.with_coeff(1)
+        new_coeff = pauli_string.coeff + pauli_string_coefficients[cache_key]
+        pauli_string_coefficients[cache_key] = new_coeff
```
natestemen, line 201: I think we can use the `defaultdict` functionality here to simplify the new coefficient calculation:
```suggestion
    for pauli_string in paulis:
        cache_key = pauli_string.with_coeff(1)
        pauli_string_coefficients[cache_key] += pauli_string.coeff
```

mitiq/observable/observable.py
```diff
@@ -154,3 +179,30 @@ def _expectation_from_density_matrix(
...
+    return list(
+        [
+            pauli_string.with_coeff(coeff)
+            for (pauli_string, coeff) in pauli_string_coefficients.items()
+            if not np.isclose(coeff, 0.0)
+        ]
+    )
```
natestemen, line 208: Looks like a double cast to `list`. Removing the outer `list` call should be fine, right?

mitiq/observable/observable.py
```diff
@@ -154,3 +179,30 @@ def _expectation_from_density_matrix(
...
     def __str__(self) -> str:
         return " + ".join(map(str, self._paulis))
+
+    def __eq__(self, other: Any) -> bool:
+        return np.allclose(self.matrix(), other.matrix())
+
+
+def _combine_duplicate_pauli_strings(
+    paulis: list[PauliString],
+) -> list[PauliString]:
```
natestemen, line 189:
```suggestion
def _combine_duplicate_pauli_strings(
    paulis: Iterable[PauliString],
) -> List[PauliString]:
```

Looks like `paulis` is only being iterated over, so we can use the typehint to reflect that. You'll have to import `Iterable` from `typing`. Then like Andrea mentioned, `list` is the newer syntax that we aren't currently using so use `typing.List` instead.

If you take this change, you can also remove the cast to list on line 42.

natestemen (approved):
Nice work again!

I'm very confused why codecov thinks line 84 and 89 are not being tested by your (very thorough) tests. Any idea why? Not a blocker, just curious.

Left a few minor simplifications (I hope). But looking ready to go once we get the tests passing.

bubakazouba:
> Nice work again!
> 
> I'm very confused why codecov thinks line 84 and 89 are not being tested by your (very thorough) tests. Any idea why? Not a blocker, just curious.
> 
> Left a few minor simplifications (I hope). But looking ready to go once we get the tests passing.

I'm not sure why codecov thinks these lines are not being tested, for example line 85 can't run without line 84 running. But somehow it thinks line 85 is being tested but line 84 wasn't.

```
| 84 | if isinstance(other, (PauliString, Number)):
| 85 |             return Observable(*[pauli * other for pauli in self._paulis])
```

The same thing seems to be happening in line 89.

I want to say this is a bug on the code coverage tool. Do you have any suggestions other than filing an issue there?

PR #1825, opened 2023-05-08
Clean up types

natestemen:
## Description

A "quick" pass over all all of our `# type: ignore` comments, as well as `typing.cast` calls. Was able to remove many without any mypy error. It's a good reminder to check when refactoring code.

Things I thought about while making these changes:
1. We should have a consistent style for setting random states/seeds. Often we support `int` or `np.random.RandomState`, in some places and just a `np.random.RandomState` in others. Maybe we allow user facing functions to take both (maybe in addition to [`random.Random`](https://docs.python.org/3/library/random.html#random.Random)?), and then expect a single type once digested from the user.
2. We can definitely do better at type-hinting function arguments more generally, and return types more specifically.
3. If we want to make type changes to `QPROGRAM`/`Executor`, it's going to take some work. Fun work, however :). I played around with making `QPROGRAM` a `typing.TypeVar`, and it broke everything. A lot of things will need to change in conjuction with that, since `QPROGRAM` being a union means mypy's type checks are pretty loose. We'd have to be much more careful in tracking types in our [conversions](https://github.com/unitaryfund/mitiq/blob/ec5d2584e5d9bd193c4ca9bd7322272b47100e84/mitiq/interface/conversions.py) files.

### Reviewer Note

It's probably easiest to review this commit by commit since some changes modify multiple files.

### DIFF
--- a/mitiq/benchmarks/randomized_benchmarking.py
+++ b/mitiq/benchmarks/randomized_benchmarking.py
@@ -53,10 +53,8 @@ def generate_rb_circuits(
 
     if n_qubits == 1:
         c1 = cliffords.c1_in_xy
-        cfd_mat_1q = cast(
-            npt.NDArray[np.complex64],
-            [_gate_seq_to_mats(gates) for gates in c1],
-        )
+        cfd_mat_1q = np.array([_gate_seq_to_mats(gates) for gates in c1])
+
         circuits = [
             _random_single_q_clifford(qubits[0], num_cliffords, c1, cfd_mat_1q)
             for _ in range(trials)
--- a/mitiq/benchmarks/randomized_benchmarking.py
+++ b/mitiq/benchmarks/randomized_benchmarking.py
@@ -4,10 +4,9 @@
 # LICENSE file in the root directory of this source tree.
 
 """Functions for generating randomized benchmarking circuits."""
-from typing import List, Optional, cast
+from typing import List, Optional
 
 import numpy as np
-import numpy.typing as npt
 
 from cirq.experiments.qubit_characterizations import (
     _single_qubit_cliffords,
--- a/mitiq/benchmarks/randomized_benchmarking.py
+++ b/mitiq/benchmarks/randomized_benchmarking.py
@@ -52,7 +52,9 @@ def generate_rb_circuits(
 
     if n_qubits == 1:
         c1 = cliffords.c1_in_xy
-        cfd_mat_1q = np.array([_gate_seq_to_mats(gates) for gates in c1])
+        cfd_mat_1q = np.array(
+            [_gate_seq_to_mats(gates) for gates in c1], dtype=np.complex64
+        )
 
         circuits = [
             _random_single_q_clifford(qubits[0], num_cliffords, c1, cfd_mat_1q)
--- a/mitiq/cdr/clifford_training_data.py
+++ b/mitiq/cdr/clifford_training_data.py
@@ -135,7 +135,7 @@ def _map_to_near_clifford(
     )
 
     # Replace selected operations.
-    clifford_ops: Sequence[cirq.ops.Operation] = _replace(
+    clifford_ops = _replace(
         [non_clifford_ops[i] for i in indices_of_selected_ops],
         method_replace,
         sigma_replace,
@@ -144,9 +144,7 @@ def _map_to_near_clifford(
 
     # Return sequence of (near) Clifford operations.
     return [
-        cast(List[cirq.ops.Operation], clifford_ops).pop(0)
-        if i in indices_of_selected_ops
-        else op
+        clifford_ops.pop(0) if i in indices_of_selected_ops else op
         for (i, op) in enumerate(non_clifford_ops)
     ]
 
@@ -209,7 +207,7 @@ def _replace(
     method: str = "uniform",
     sigma: float = 1.0,
     random_state: Optional[np.random.RandomState] = None,
-) -> Sequence[cirq.ops.Operation]:
+) -> List[cirq.ops.Operation]:
     """Function that takes the non-Clifford angles and replacement and
     selection specifications, returning the projected angles according to a
     specific method.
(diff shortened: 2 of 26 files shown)


mitiq/cdr/clifford_training_data.py
```diff
@@ -171,8 +169,7 @@ def _select(
                       method_select = 'gaussian'.
         random_state: Random state for sampling.
     """
-    if random_state is None:
-        random_state = np.random  # type: ignore
```
andreamari, line 175: The reason we used `random_state = np.random` instead of a new instance  `random_state = np.RandomState()` is that the user can set a global seed  `np.random.seed(123)` with a single instruction e.g. at the beginning of a notebook and it should propagate to all Mitiq code.
Cirq uses a similar [approach](https://github.com/quantumlib/Cirq/blob/f2cd706ecd308b8088f7452e02e6f24090087e86/cirq-core/cirq/value/random_state.py#L40).

Actually I am not sure if this is good or bad practice. Just wanted to explain the motivation.

mitiq/cdr/clifford_training_data.py
```diff
@@ -171,8 +169,7 @@ def _select(
...
+    random_state = random_state or np.random.RandomState()
```
andreamari, line 172: Took me a while to get this shortcut :-) Nice!
Unfortunately, there is a problem if the user sets `0` as a seed. In that case `random_state` is considered as `False` and is unexpectedly replaced by a new random state.

mitiq/cdr/clifford_training_data.py
natestemen: Thanks for adding some background here (and the link to Cirq), it's super helpful! I agree we should definitely tread lightly around changing RandomState related things, but I'm not sure that [`np.random.seed`](https://numpy.org/doc/stable/reference/random/generated/numpy.random.seed.html) works in that way. From my understanding, this function only effects the _next_ call of a numpy random function, but not subsequent ones. So if a user sets a seed at the top of a file running Mitiq, I would not expect completely reproducible outputs.
```py
>>> import numpy as np
>>> np.random.seed(42)
>>> np.random.randint(500)
102
>>> np.random.seed(42)
>>> np.random.randint(500)
102
>>> np.random.randint(500)
435
```

andreamari: The seed actually affects subsequent functions (the internal random states evolves in a reproducible way):

```python
>>> import numpy as np

>>> np.random.seed(42)
>>> np.random.randint(500)
102
>>> np.random.randint(500)
435
>>> np.random.seed(42)
>>> np.random.randint(500)
102
>>> np.random.randint(500)
435
```

So, everything which depends on `np.random` becomes deterministically reproducible.

andreamari (approved):
Apart from the `np.random.RandomState() ` vs `np.random` problem discussed in the previous comments, all the rest LGTM

mitiq/cdr/clifford_training_data.py
Aaron-Robertson: I tried `random_state: Union[np.random.RandomState, np.random.Generator] = np.random.default_rng(),` and all tests pass. Can remove the None checks and most Optionals too, eg:
```
def _select(
    non_clifford_ops: Sequence[cirq.ops.Operation],
    fraction_non_clifford: float,
    method: str = "uniform",
    sigma: Optional[float] = 1.0,
    random_state: Union[np.random.RandomState, np.random.Generator] = np.random.default_rng(),
) -> List[int]:
...
    num_non_cliff = len(non_clifford_ops)
    num_to_replace = int(round(fraction_non_clifford * num_non_cliff))
```
Happy to push if we like it and you're low on bandwidth @natestemen.

mitiq/interface/conversions.py
```diff
@@ -248,8 +245,6 @@ def new_scaling_function(
             from pyquil import Program
             from pyquil.quilbase import Declare, Measurement
 
-            circuit = cast(Program, circuit)
-
```
Aaron-Robertson, line 252: I'm getting circuits typed as any after this. I think it's good practice to do what we do in the `qiskit` pass above: `circuit = circuit.copy()`?

mitiq/cdr/clifford_training_data.py
Aaron-Robertson: @natestemen, as discussed made a branch to test out these changes. They're [here](https://github.com/unitaryfund/mitiq/tree/ar-clean-types) when you get a chance to take a look!

natestemen: From my testing the `np.random.default_rng` function does not respect if the user has set a seed using `np.random.seed`. So I'll revert my changes here back to what we had previously.

mitiq/interface/conversions.py
natestemen: I added the `cast` call back in [`3af0b0f`](https://github.com/unitaryfund/mitiq/pull/1825/commits/3af0b0f3e69a16053fa6539ca9d8e2d5a07906e9), but I think since we don't need to modify the program for a pyquil program, we do not need to copy the circuit as we do with qiskit.

mitiq/benchmarks/randomized_benchmarking.py
```diff
@@ -53,10 +52,8 @@ def generate_rb_circuits(
 
     if n_qubits == 1:
         c1 = cliffords.c1_in_xy
-        cfd_mat_1q = cast(
-            npt.NDArray[np.complex64],
-            [_gate_seq_to_mats(gates) for gates in c1],
-        )
+        cfd_mat_1q = np.array([_gate_seq_to_mats(gates) for gates in c1])
```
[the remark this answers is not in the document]
natestemen: Done in [`0a9ec5a`](https://github.com/unitaryfund/mitiq/pull/1825/commits/0a9ec5abe46aef3f6c13fb03dbcf641cc59409e9).

andreamari (approved):
Nice to see so many `# type ignore` are gone! 
Thanks Nate.

PR #1831, opened 2023-05-12
Improve twirling test

andreamari:
This PR makes a twirling test more deterministic.
It should avoid random failures like [this](https://github.com/unitaryfund/mitiq/actions/runs/4958936031/jobs/8872423020#step:5:3813).

Specifically, the new tests compares the number of gates instead of comparing the number of moments.


natestemen (approved):
Nice fix. Thank you!

PR #1834, opened 2023-05-15
add unitaryHACK banner to docs

natestemen:
## Description

Adds a simple banner to the top of all doc pages. Example:
Screenshot 2023-05-15 at 1 00 04 PM

In terms of making this viewable when visiting https://mitiq.readthedocs.io/, we have two options (that I know of)

1. Do a patch release (0.26.1) so that the banner shows up on "stable".
2. Change the default version on RTD to be "latest" (now it's "stable") until the official 0.27.0 release at the start of the hackathon.

I have a preference for 2 as it's much simpler. WDYT?

### DIFF
--- a/docs/source/conf.py
+++ b/docs/source/conf.py
@@ -311,6 +311,12 @@ html_favicon = "img/mitiq.ico"
 # These files are copied directly to the root of the documentation.
 html_extra_path = ["robots.txt"]
 
+html_theme_options = {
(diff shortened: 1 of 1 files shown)


Misty-W (approved):

Misty-W:
> I have a preference for 2 as it's much simpler. WDYT?

I prefer 2 for the same reason.

PR #1839, opened 2023-05-19
Quantum Many Body Scars (close #1797)

### LINKED ISSUE #1797: Tutorial : Use ZNE to simulate quantum many body scars (QMBS)
Proposed solution to close this issue is in #1773 

https://arxiv.org/abs/2203.08291

ZNE (dynamical decoupling + readout error mitigation technique) is used to mitigate results in Fig 3.

DHuybrechts:
## Description

I made an example/ tutorial for the Issue #1797. At this moment it is in a Jupyter notebook format, as I did not get the markdown format working yet. Since I expect that some changes may need to be made to the notebook, I figured posting it here already for feedback would optimise the iteration process before it is accepted. My apologies if I missed any requirements, I am fairly new to this submitting process. I would be happy to update the file/ submission if that would be the case.

---

### License


nathanshammah:
Thanks @DHuybrechts, looks great, I am going to start reviewing this next week. Can you please add the Jupyter notebook to the examples file index?
https://github.com/unitaryfund/mitiq/blob/master/docs/source/examples/examples.md
Then, the docs should build showing the example [here](https://mitiq--1839.org.readthedocs.build/en/1839/) (link that is present in the GH Actions above on this page, under "docs/readthedocs.org:mitiq")

DHuybrechts:
Thanks @nathanshammah, I added the notebook to the examples file and also corrected some minor things. I am looking forward to your feedback next week!

PeterPOrth:
Looks great! A couple of (optional) suggestions:
1. including error bars in the final plot. There are different sources of error: 1. shot noise, 2. gate noise, 3. extrapolation uncertainty. It would be useful to include a proper discussion of errors and uncertainties. One simple way would be to run each folded circuit n times and then feed in the n results into the extrapolation routine (as mean and standard deviation). The extrapolation will then output also an uncertainty based on the standard deviation of the input data. The n times would in practice be run with different Pauli twirlings, but that's not needed on simulators. 
2. describe choice of ZNE hyperparameters: choice of fitting factory, noise folding factors, local random versus global gate folding. Explicitly state what are the gates that are being folded. It is not RZZ here as the RZZ is explicitly given in terms of CNOT and RZ.

Other comments that are likely beyond the tutorial and are just general comments: 
3. Beyond the notebook, but generally interesting questions: Impact of the choice of extrapolation function/fitting factory on results in the figure.
4. One comment about what makes this model special (with regards to QEM techniques). One thing that is special about this problem is that the scarred dynamics occurs mostly in the so-called Fibonacci Hilbert subspace, which is spanned by all Z-basis bitstrings with no conscutive 1s in the bitstring. The time evolved wavefunction lies exactly in this subspace for the exact evolution, but Trotterization only breaks this very weakly (see Fig.10 in (https://arxiv.org/abs/2203.08291) ). This can be used for QEM using postselection: discard all measurement results that violated the Fibonacci rule (ie bitstring with two 1s next to each other). This is not the focus of the tutorial though, and it is just a comment.

andreamari (commented):
Thanks @DHuybrechts for this great tutorial!
I have two main suggestions:

1. It is not clear to me what exactly a "scar" is in this context. I mean, what is a scar? What is the motivation for choosing that particular type of Hamiltonian?  One or two sentences giving some physical context/motivation would be great.

2. For Mitiq examples we use `.md` files instead of `.ipynb` files. The advantage of `.md` files is that they are essentially text files and so they are very light and easier to version control.  You can have a look at the Mitiq contribution [guide](https://mitiq.readthedocs.io/en/stable/contributing_docs.html#adding-code-examples) about this. If switching to `.md` format is a blocker for you, we can make an exception and temporarely accept the `ipynb` format.  


A final remark. @PeterPOrth  gave a nice suggestion about running the simulation several times to extract some statistics. I agree that it would be great, but conditioned on a reasonable total execution time. To have a figure of merit, you can consider a budget of say 3 minutes (max 5 minutes) for executing the example. All the code of our documentation is re-executed at each build and so we try keeping the execution time of examples as short as possible.

Congrats again for the nice work.

nathanshammah (changes requested):
Thanks again @DHuybrechts, my additional comments are just cosmetic: 
- **Thumbnail**. Please add a thumbnail, for the example, so that it shows up on the examples gallery page. It could be the nice plot at the end. 
  - You need to add the image here, https://github.com/unitaryfund/mitiq/tree/master/docs/source/_thumbnails
  -   then list it here in the `conf.py` file that builds the documentation, https://github.com/unitaryfund/mitiq/blob/master/docs/source/conf.py#L337
- **References**. Please cite the reference by adding it to the main references of the full documentation. 
  - You'll need to add the `.bib` record to the [`refs.bib` file](https://github.com/unitaryfund/mitiq/blob/master/docs/source/refs.bib).
  - I suggest you list explicitly in text of the notebook the title of the paper, and then add the reference using the format ``` {cite}`FirstAuthor_Year_JRN` ```, see [here](https://raw.githubusercontent.com/unitaryfund/mitiq/master/docs/source/examples/molecular_hydrogen.md) for reference "``` {cite}`OMalley_2016_PRX` ```".

DHuybrechts:
Thank you very much @PeterPOrth @andreamari and @nathanshammah for taking the time to go through this notebook tutorial and providing these useful suggestions!

@PeterPOrth thank you for the detailed comments. I am going to try and implement them. Even if they don't make it to the tutorial notebook (we'll see how long the simulation takes), looking into them will definitely help me get a better grasp of these error mitigation schemes/ techniques.

@andreamari I will convert the notebook to a .md file in the next upload. As noted above we'll have to see how long the simulation takes but it would indeed be quite nice to have error bars.

@nathanshammah I'll include the thumbnail and the references!

Hopefully, I'll be able to give an update at the end of the weekend!

nathanshammah:
@DHuybrechts just to confirm that, about error bars, you could show the standard deviation, based on number of repetitions of the experiments. If this takes too long, we could do without.

DHuybrechts:
@andreamari @nathanshammah I have added everyones suggestions and believe the tutorial is up for another (possibly final) review. Let me know if I need to clarify anything in more detail in the tutorial. 

As a disclaimer, I am trying to get the `make install` command working, so I could not locally test if the docs build correctly. My apologies if anything fails on that end!

docs/source/refs.bib
```diff
@@ -298,6 +337,14 @@ @article{GoogleQuantum_2021_nature
 # Letter I
 # Letter J
 
+@book{James_2021_statlearning,
+  author    = {James, Gareth and Witten, Daniela and Hastie, Trevor and Tibshirani, Robert},
+  title     = {An Introduction to Statistical Learning with Applications in RAn Introduction to Statistical Learning with Applications in R},
+  year      = {2021},
```
nathanshammah, line 343:
```suggestion
  year      = {2021},
  publisher = {Springer},
```

nathanshammah (approved):
This is great, thanks @DHuybrechts!

PR #1840, opened 2023-05-19
Update calibration tutorial

Misty-W:
## Description

---

### License

### DIFF
--- a/docs/source/examples/calibration-tutorial.md
+++ b/docs/source/examples/calibration-tutorial.md
@@ -16,9 +16,8 @@ kernelspec:
 
 <img src="../_thumbnails/calibration.png" width="400">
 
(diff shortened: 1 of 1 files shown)


Misty-W:
Note: I've returned to using RB circuits and FakeJakarta because I get a consistent improvement (15/15 times), whereas with GHZ I saw strange results where all the calibration runs had a red "X", even though the final result with calibration was better than the default result.

andreamari (approved):
Great to see it works now, LGTM!

Maybe `fold_all` can be removed since, for odd scale factors (1, 3, 5), it should produce the same deterministic effect of the default method (`fold_gates_at_random`). 
If I am wrong, no problem with keeping `fold_all`.

PR #1841, opened 2023-05-19
Add mypy to style guidelines

purva-thakre:
Noticed `mypy` was missing from the[ style guidelines in docs.](https://mitiq.readthedocs.io/en/stable/contributing.html#style-guidelines)


purva-thakre:
This is ready. Not sure why `codecov` is failing.

natestemen (approved):
Good thinking adding this. Always good to have our processes documented thoroughly.

PR #1845, opened 2023-05-23
Adding PEC as technique supported by calibration

### LINKED ISSUE #1737: Adding PEC as technique supported by calibration
## Issue Description
Right now calibration supports ZNE as error mitigation technique. It would be great to also support PEC. 

## Proposed Solution
1. One option would require user input about the noise channel model. This could include a call of:
- mitiq.pec.representations.optimal.**find_optimal_representation**(ideal_operation, noisy_operations)
2. Another option would not require input about noise channel model. These could be functions that could be called: 
- mitiq.pec.representations.depolarizing.**represent_operation_with_global_depolarizing_noise**(ideal_operation, noise_level)
- mitiq.pec.representations.depolarizing.**represent_operation_with_local_depolarizing_noise**(ideal_operation, noise_level)
- mitiq.pec.representations.biased_noise.**represent_operation_with_local_biased_noise**(ideal_operation, epsilon, eta)

## Additional References
https://mitiq.readthedocs.io/en/latest/apidoc.html#mitiq.pec.representations.depolarizing.represent_operation_with_global_depolarizing_noise
https://mitiq.readthedocs.io/en/stable/guide/calibrators.html

Misty-W:
## Description

Prototype for PEC calibration. 

Fixes #1737

1. Add simple `PECSettings` object with 5 different noise levels of depolarizing noise, representing CNOT and CZ gates only. 
2. Update `settings.Strategy.mitigation_function` to generate representations from relevant `PECSettings`.
3. Update `representations` functions, e.g. `represent_operation_with_local_depolarizing_noise` to take optional `is_qubit_dependent` argument.
4. Update qubit transformation condition in `utils._equal` to check if the number of qubits in both circuits is the same before performing the qubit mapping. If they don't, the circuits are unequal --> already return `False` and move on to the next iteration without throwing an unnecessary error.
5. Add PEC calibration tests.
6. Update user guide to mention `PECSettings` (more details to come in future issues/PRs).

---

### License

### DIFF
--- a/mitiq/calibration/settings.py
+++ b/mitiq/calibration/settings.py
@@ -3,13 +3,14 @@
 # This source code is licensed under the GPL license (v3) found in the
 # LICENSE file in the root directory of this source tree.
 
+from copy import deepcopy
 from dataclasses import dataclass, asdict
 from functools import partial
 from typing import Any, Callable, cast, List, Dict
 from enum import Enum, auto
 
 import networkx as nx
-import cirq
+from cirq import Circuit, LineQubit, measure, CNOT, CZ
 
 from mitiq import QPROGRAM
 from mitiq.interface import convert_from_mitiq
@@ -65,7 +66,7 @@ class BenchmarkProblem:
     """
 
     id: int
-    circuit: cirq.Circuit
+    circuit: Circuit
     type: str
     ideal_distribution: Dict[str, float]
 
@@ -87,7 +88,7 @@ class BenchmarkProblem:
             The converted circuit with final measurements.
         """
         circuit = self.circuit.copy()
-        circuit.append(cirq.measure(circuit.all_qubits()))
+        circuit.append(measure(circuit.all_qubits()))
         return convert_from_mitiq(circuit, circuit_type)
 
     @property
@@ -143,19 +144,24 @@ class Strategy:
     @property
     def mitigation_function(self) -> Callable[..., float]:
         if self.technique is MitigationTechnique.PEC:
-            rep_function = self.technique_params.pop("representation_function")
-            operations = self.technique_params.pop("operations")
-            is_qubit_dependent = self.technique_params.pop(
-                "is_qubit_dependent", True
-            )
-            noise_level = self.technique_params.pop("noise_level")
+            params_copy = deepcopy(self.technique_params)
+            rep_function = params_copy.pop("representation_function")
+            operations = params_copy.pop("operations")
+            noise_level = params_copy.pop("noise_level")
+            is_qubit_dependent = params_copy.pop("is_qubit_dependent", True)
+            representations = [
+                rep_function(
+                    Circuit(op(*LineQubit.range(2))),
+                    noise_level,
+                    is_qubit_dependent,
+                )
+                for op in operations
+            ]
+
             return partial(
                 self.technique.mitigation_function,
-                [
-                    rep_function(op, noise_level, is_qubit_dependent)
-                    for op in operations
-                ],
-                **self.technique_params,
+                representations=representations,
+                **params_copy,
             )
         else:
             return partial(
@@ -291,7 +297,7 @@ class Settings:
                     f"but got {circuit_type}."
                 )
 
-            circuit = cast(cirq.Circuit, circuit)
+            circuit = cast(Circuit, circuit)
             problem = BenchmarkProblem(
                 id=i,
                 circuit=circuit,
@@ -418,45 +424,50 @@ PECSettings = Settings(
             "representation_function": (
                 represent_operation_with_local_depolarizing_noise
             ),
-            "operations": [cirq.CNOT, cirq.CZ],
+            "operations": [CNOT, CZ],
             "is_qubit_dependent": False,
             "noise_level": 0.001,
+            "num_samples": 50,
         },
         {
             "technique": "pec",
             "representation_function": (
                 represent_operation_with_local_depolarizing_noise
             ),
-            "operations": [cirq.CNOT, cirq.CZ],
+            "operations": [CNOT, CZ],
             "is_qubit_dependent": False,
             "noise_level": 0.005,
+            "num_samples": 50,
         },
         {
             "technique": "pec",
             "representation_function": (
                 represent_operation_with_local_depolarizing_noise
             ),
-            "operations": [cirq.CNOT, cirq.CZ],
+            "operations": [CNOT, CZ],
             "is_qubit_dependent": False,
             "noise_level": 0.01,
+            "num_samples": 50,
         },
         {
             "technique": "pec",
             "representation_function": (
                 represent_operation_with_local_depolarizing_noise
             ),
-            "operations": [cirq.CNOT, cirq.CZ],
+            "operations": [CNOT, CZ],
             "is_qubit_dependent": False,
             "noise_level": 0.015,
+            "num_samples": 50,
         },
         {
             "technique": "pec",
             "representation_function": (
                 represent_operation_with_local_depolarizing_noise
             ),
-            "operations": [cirq.CNOT, cirq.CZ],
+            "operations": [CNOT, CZ],
             "is_qubit_dependent": False,
             "noise_level": 0.02,
+            "num_samples": 50,
         },
     ],
 )
--- a/mitiq/calibration/settings.py
+++ b/mitiq/calibration/settings.py
@@ -3,14 +3,13 @@
 # This source code is licensed under the GPL license (v3) found in the
 # LICENSE file in the root directory of this source tree.
 
-from copy import deepcopy
 from dataclasses import dataclass, asdict
 from functools import partial
-from typing import Any, Callable, cast, List, Dict
+from typing import Any, Callable, cast, List, Dict, Union
 from enum import Enum, auto
 
 import networkx as nx
-from cirq import Circuit, LineQubit, measure, CNOT, CZ
+import cirq
 
 from mitiq import QPROGRAM
 from mitiq.interface import convert_from_mitiq
@@ -25,6 +24,7 @@ from mitiq.pec import execute_with_pec
 from mitiq.pec.representations import (
     represent_operation_with_local_depolarizing_noise,
 )
+from mitiq.pec.types.types import OperationRepresentation
 from mitiq.raw import execute
 from mitiq.zne import execute_with_zne
 from mitiq.zne.inference import LinearFactory, RichardsonFactory
@@ -52,6 +52,12 @@ class MitigationTechnique(Enum):
             return execute
 
 
+calibration_supported_techniques = {
+    "ZNE": MitigationTechnique.ZNE,
+    "PEC": MitigationTechnique.PEC,
+}
+
+
 @dataclass
 class BenchmarkProblem:
     """A dataclass containing information for instances of problems that will
@@ -66,7 +72,7 @@ class BenchmarkProblem:
     """
 
     id: int
-    circuit: Circuit
+    circuit: cirq.Circuit
     type: str
     ideal_distribution: Dict[str, float]
 
@@ -88,7 +94,7 @@ class BenchmarkProblem:
             The converted circuit with final measurements.
         """
         circuit = self.circuit.copy()
-        circuit.append(measure(circuit.all_qubits()))
+        circuit.append(cirq.measure(circuit.all_qubits()))
         return convert_from_mitiq(circuit, circuit_type)
 
     @property
@@ -142,14 +148,13 @@ class Strategy:
     technique_params: Dict[str, Any]
 
     @property
-    def mitigation_function(self) -> Callable[..., float]:
+    def representations(self) -> Union[List[OperationRepresentation], None]:
         if self.technique is MitigationTechnique.PEC:
-            params_copy = deepcopy(self.technique_params)
-            rep_function = params_copy.pop("representation_function")
-            operations = params_copy.pop("operations")
-            noise_level = params_copy.pop("noise_level")
-            is_qubit_dependent = params_copy.pop("is_qubit_dependent", True)
-            representations = [
+            rep_function = self.technique_params["representation_function"]
+            operations = self.technique_params["operations"]
+            noise_level = self.technique_params["noise_level"]
+            is_qubit_dependent = self.technique_params["is_qubit_dependent"]
+            return [
                 rep_function(
                     op,
                     noise_level,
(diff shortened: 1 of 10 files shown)


nathanshammah:
This is great, @Misty-W! I wonder if there is a way to make the ZNESettings and PECSettings show up in the API doc, as with the `ZNESettings = Settings(...)` definition they don't, as they are just an instantiation of `Settings`. However, maybe that's good as it is. 

Looking forward to seeing this in Mitiq!

Misty-W:
> I wonder if there is a way to make the ZNESettings and PECSettings show up in the API doc, as with the `ZNESettings = Settings(...)` definition they don't, as they are just an instantiation of `Settings`. However, maybe that's good as it is.

I thought about trying to include `ZNESettings` and `PECSettings` in the API doc but decided it could be confusing, because they are instantiations, and the user should either use them as-is or create a new `Settings` object.

andreamari (commented):
I think `num_samples=50` is too low, I guess the corresponding results will be subject to large non-deterministic statistical noise.

What about  `num_samples=200` ?  If tests are too slow, we can always reduce the strategies. In the worst case we would keep just `noise_level=0.001` and ` noise_level=0.01`.
In the worst case, we could also reduce the number of benchmark circuits.

In summary, it is better to test less things but in a more stable way, than many things with large of statistical noise.

Misty-W:
> What about `num_samples=200` ? If tests are too slow, we can always reduce the strategies. In the worst case we would keep just `noise_level=0.001` and ` noise_level=0.01`. In the worst case, we could also reduce the number of benchmark circuits.

Thanks @andreamari, that makes sense.

mitiq/pec/representations/biased_noise.py
```diff
@@ -57,6 +58,10 @@ def represent_operation_with_local_biased_noise(
             channels with :math:`\eta = 0` describing a fully depolarizing
             channel and :math:`\eta = \infty` describing a fully dephasing
             channel.
+        is_qubit_dependent: If True, the representation corresponds to the
+            operation on the specific qubits defined in `ideal`. If False, the
+            representation is valid for the same gate even if acting on
+            different qubits from those specified in `ideal`.
```
andreamari, line 64:
```suggestion
            different qubits from those specified in `ideal_operation`.
```

mitiq/calibration/settings.py
```diff
@@ -139,9 +143,30 @@ class Strategy:
...
-        )
+        if self.technique is MitigationTechnique.PEC:
+            params_copy = deepcopy(self.technique_params)
+            rep_function = params_copy.pop("representation_function")
+            operations = params_copy.pop("operations")
+            noise_level = params_copy.pop("noise_level")
+            is_qubit_dependent = params_copy.pop("is_qubit_dependent", True)
+            representations = [
+                rep_function(
+                    Circuit(op(*LineQubit.range(2))),
```
andreamari, line 154: Here we hard code `LineQubit.range(2)`, this is fine only in the case in which `is_qubit_depedent` is `False`.

Maybe, instead of passing gates $[CNOT, CZ]$ to the `operations` argument, we could pass operations (gates on some qubits).

In this way we can get rid of `LineQubit.range(2)`, here.

mitiq/calibration/tests/test_calibration.py
```diff
@@ -119,6 +120,27 @@ def test_ZNE_workflow():
     assert isinstance(cal.best_strategy(), Strategy)
 
 
+def pec_execute(circuit, noise_level=0.01):
```
andreamari, line 123: Can we rename this to `damping_execute`?  The term "pec" in the name is a bit confusing. 
The difference with the existing `execute` is that it simulates amplitude damping noise instead of depolarizing. 

Actually, given that we assume depolarizing noise in `PECSettings`, why do we need to simulate amplitude damping noise? Can't we just use the existing `execute` function?

mitiq/calibration/tests/test_settings.py
```diff
@@ -128,17 +128,27 @@ def test_make_strategies_invalid_technique():
...
     assert all(
         s in repr_string
         for s in ("type", "ideal_distribution", "num_qubits", "circuit_depth")
     )
-
     assert len(circuits) == 4
     assert len(strategies) == 2 * 2 * 2
 
 
+def test_PECSettings():
```
andreamari, line 140: This is a good test, but it only tests Cirq circuits.
We should test that everything works also with a different frontend. 

For example, this is what we did for ZNE:

```python
@pytest.mark.parametrize("circuit_type", SUPPORTED_PROGRAM_TYPES.keys())
def test_ZNE_workflow_multi_platform(circuit_type):
...
```

mitiq/pec/representations/biased_noise.py
```diff
@@ -57,6 +58,10 @@ def represent_operation_with_local_biased_noise(
             channels with :math:`\eta = 0` describing a fully depolarizing
             channel and :math:`\eta = \infty` describing a fully dephasing
             channel.
+        is_qubit_dependent: If True, the representation corresponds to the
+            operation on the specific qubits defined in `ideal`. If False, the
```
andreamari, line 62:
```suggestion
            operation on the specific qubits defined in `ideal_operation`. If False, the
```

mitiq/pec/representations/damping.py
```diff
@@ -42,6 +43,10 @@ def _represent_operation_with_amplitude_damping_noise(
     Args:
         ideal_operation: The ideal operation (as a QPROGRAM) to represent.
         noise_level: The noise level of each amplitude damping channel.
+        is_qubit_dependent: If True, the representation corresponds to the
+            operation on the specific qubits defined in `ideal`. If False, the
```
andreamari, line 47:
```suggestion
            operation on the specific qubits defined in `ideal_operation`. If False, the
```

mitiq/pec/representations/damping.py
```diff
@@ -42,6 +43,10 @@ def _represent_operation_with_amplitude_damping_noise(
...
+            representation is valid for the same gate even if acting on
+            different qubits from those specified in `ideal`.
```
andreamari, line 49:
```suggestion
            different qubits from those specified in `ideal_operation`.
```

mitiq/pec/representations/depolarizing.py
```diff
@@ -76,6 +78,10 @@ def represent_operation_with_global_depolarizing_noise(
     Args:
         ideal_operation: The ideal operation (as a QPROGRAM) to represent.
         noise_level: The noise level (as a float) of the depolarizing channel.
+        is_qubit_dependent: If True, the representation corresponds to the
+            operation on the specific qubits defined in `ideal`. If False, the
```
andreamari, line 82:
```suggestion
            operation on the specific qubits defined in `ideal_operation`. If False, the
```

mitiq/pec/representations/depolarizing.py
```diff
@@ -76,6 +78,10 @@ def represent_operation_with_global_depolarizing_noise(
...
+            representation is valid for the same gate even if acting on
+            different qubits from those specified in `ideal`.
```
andreamari, line 84:
```suggestion
            different qubits from those specified in `ideal_operation`.
```

mitiq/pec/representations/depolarizing.py
```diff
@@ -173,7 +183,10 @@ def represent_operation_with_local_depolarizing_noise(
     Args:
         ideal_operation: The ideal operation (as a QPROGRAM) to represent.
         noise_level: The noise level of each depolarizing channel.
-
+        is_qubit_dependent: If True, the representation corresponds to the
+            operation on the specific qubits defined in `ideal`. If False, the
```
andreamari, line 187:
```suggestion
            operation on the specific qubits defined in `ideal_operation`. If False, the
```

mitiq/pec/representations/depolarizing.py
```diff
@@ -173,7 +183,10 @@ def represent_operation_with_local_depolarizing_noise(
...
+            representation is valid for the same gate even if acting on
+            different qubits from those specified in `ideal`.
```
andreamari, line 189:
```suggestion
            different qubits from those specified in `ideal_operation`.
```

mitiq/pec/representations/optimal.py
```diff
@@ -114,6 +115,10 @@ def find_optimal_representation(
             of the represented operation.
         initial_guess: Optional initial guess for the coefficients
             :math:`\{ \eta_\alpha \}`.
+        is_qubit_dependent: If True, the representation corresponds to the
+            operation on the specific qubits defined in `ideal`. If False, the
```
andreamari, line 119:
```suggestion
            operation on the specific qubits defined in `ideal_operation`. If False, the
```

mitiq/pec/representations/optimal.py
```diff
@@ -114,6 +115,10 @@ def find_optimal_representation(
...
+            representation is valid for the same gate even if acting on
+            different qubits from those specified in `ideal`.
```
andreamari, line 121:
```suggestion
            different qubits from those specified in `ideal_operation`.
```

andreamari (commented):
Thanks @Misty-W! 
I added some comments below directly in the code. 
Moreover, I have a more general comment: what happens to the logging in the case of PEC?
I mean, what is the equivalent of this the following messages:
image

Fine to address this logging aspect in this PR or, if we want to keep it for later, I think it  requires a new issue such that we don't forget about it.

mitiq/pec/representations/biased_noise.py
andreamari, line 64:
```suggestion
            different qubits from those specified in `ideal_operation`.
```

mitiq/calibration/tests/test_settings.py
Misty-W: I've added a similar test for PEC, seems to work.

mitiq/calibration/settings.py
Misty-W: Good catch, I've fixed that.

mitiq/calibration/tests/test_calibration.py
Misty-W: I've fixed the naming to indicate the types of noise used. ZNE tests were already set up with amplitude damping noise, but we cannot "represent with amplitude damping" on non-Cirq frontends for PEC, so I have to stick with depolarizing.

mitiq/calibration/settings.py
```diff
@@ -371,4 +396,56 @@ def make_strategies(self) -> List[Strategy]:
...
+                represent_operation_with_local_depolarizing_noise
+            ),
+            "operations": [
+                Circuit(CNOT(*LineQubit.range(2))),
+                Circuit(CZ(*LineQubit.range(2))),
+            ],
+            "is_qubit_dependent": False,
+            "noise_level": 0.001,
+            "num_samples": 200,
+        },
```
natestemen, line 434: Do you think we should validate each instance that comes in through the strategy list? Probably a bigger change than we'd want to make here, but curious on your thoughts.

natestemen (commented):
Awesome work Misty! This is a pretty big change which touches quite a lot of files, so nice work making the change in a compact, and understandable way.

As Andrea noted, the logging will need some changes, but IMO that could be done in another PR. I fear it might not be easy, especially if we generalize to the case of mixing PEC and ZNE tests (which I think we should do). We'll also need to make some changes to the `Strategy.to_dict` function, since it assumes ZNE as a default strategy. I think we could include that in this PR. WDYT?

Have you tried creating a settings object which mixes ZNE and PEC strategies?

Misty-W:
> Fine to address this logging aspect in this PR or, if we want to keep it for later, I think it requires a new issue such that we don't forget about it.

Thanks @andreamari, I think as Nate pointed out, the logging needs more thought and maybe discussion before implementing, so I've created a separate issue, #1861 .

mitiq/calibration/settings.py
```diff
@@ -139,9 +143,30 @@ class Strategy:
...
+            noise_level = params_copy.pop("noise_level")
+            is_qubit_dependent = params_copy.pop("is_qubit_dependent", True)
+            representations = [
+                rep_function(
+                    op,
+                    noise_level,
+                    is_qubit_dependent,
+                )
+                for op in operations
+            ]
```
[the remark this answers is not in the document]
Misty-W: Sounds like a good idea! It might eliminate the need for copying the parameters.

mitiq/calibration/settings.py
Misty-W: Do we do that for ZNE?

mitiq/pec/representations/biased_noise.py
```diff
@@ -57,6 +58,11 @@ def represent_operation_with_local_biased_noise(
             channels with :math:`\eta = 0` describing a fully depolarizing
             channel and :math:`\eta = \infty` describing a fully dephasing
             channel.
+        is_qubit_dependent: If True, the representation corresponds to the
+            operation on the specific qubits defined in `ideal_operation`.
+            If False, the representation is valid for the same gate even if
+            acting on different qubits from those specified in
+            `ideal_operation`.
```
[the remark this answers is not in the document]
Misty-W: Yes, that's the general idea. We need to be able to specify whether the representations we want from the "represent_with..." functions are qubit dependent or qubit independent. For a fast workflow / prototype, we want qubit independent representations. For refinement, we may want the representations to be qubit dependent.

mitiq/calibration/settings.py
natestemen: Nope!

Misty-W: Then I suggest to address in a future PR, issue #1862 created.

mitiq/calibration/settings.py
```diff
@@ -3,13 +3,14 @@
...
 
+from copy import deepcopy
 from dataclasses import dataclass, asdict
 from functools import partial
 from typing import Any, Callable, cast, List, Dict
 from enum import Enum, auto
 
 import networkx as nx
-import cirq
+from cirq import Circuit, LineQubit, measure, CNOT, CZ
```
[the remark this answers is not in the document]
Misty-W: I had changed it because I find the `cirq.` ... to be harder to read in a chain of Cirq functions, but I agree with your concern about ambiguity, so I've changed back to importing `cirq` as a top-level object.

mitiq/calibration/settings.py
```diff
@@ -139,9 +143,30 @@ class Strategy:
 
     @property
     def mitigation_function(self) -> Callable[..., float]:
-        return partial(
-            self.technique.mitigation_function, **self.technique_params
-        )
+        if self.technique is MitigationTechnique.PEC:
+            params_copy = deepcopy(self.technique_params)
```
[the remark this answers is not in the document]
Misty-W: We don't need a copy now that `pop` is no longer used.

Misty-W:
Thanks for the suggestions @natestemen, I think this PR is looking cleaner now.

>  We'll also need to make some changes to the `Strategy.to_dict` function, since it assumes ZNE as a default strategy. I think we could include that in this PR. WDYT?
Done ✅ .
> Have you tried creating a settings object which mixes ZNE and PEC strategies?
No, but that's good to try in a future improvement.

andreamari (approved):
Thanks the changes. LGTM!

Misty-W:
thanks @purva-thakre! I remembered seeing the Codecov issue before, must've been the PR you linked.

PR #1848, opened 2023-05-26
Pauli Twirling (PT) User Guide

### LINKED ISSUE #1833: Add Pauli twirling in User Guide
## Issue Description
Add Pauli twirling in User Guide like what is done for other techniques. Files can be considerably shorter for Pauli twirling, compared to more elaborate techniques.

Aaron-Robertson:
## Description

Fixes #1833 

To-Dos before merging

- [x] Replace the png with svg - center all text properly : svg's are not rendered correctly depending on the choice of browser. In safari, the text in the SVG was not aligned correctly. Find which changes to make before replacing png. 

---

### License


Misty-W:
hi @Aaron-Robertson, if you add `pt.md` to the `toctree` in `guide.md`, that should fix at least some of the errors.

purva-thakre:
@Aaron-Robertson Your docs build might be failing because you did not separate the 2 cited bib refs by a comma.

docs/source/guide/pt-4-low-level.md
```diff
@@ -0,0 +1,118 @@
...
+---
+
+# What happens when I use PT?
+
+The workflow of Pauli Twirling (PT) in Mitiq is represented in the figure below.
+
+```{figure} ../img/pt_workflow.svg
+---
+width: 700px
+name: pt-workflow-overview
```
purva-thakre, line 21: I think this might have to be renamed to something like `pt-workflow-overview2` because `pt-workflow-overview` is used in the main page `pt.md`.

Aaron-Robertson: @purva-thakre That's by design, since we're using the same image for both places in the docs. Take a look [here](https://mitiq--1848.org.readthedocs.build/en/1848/guide/pt-4-low-level.html) and we can see the code execution errors.

That said, I keep finding inaccuracies (both in the code and docs/theory), so this one isn't quite ready for review yet (when I marked it as so late last night I was apparently rather hopeful 🤣)!

Aaron-Robertson: Thanks @purva-thakre, should be all set now!

docs/source/guide/pt.md
```diff
@@ -0,0 +1,34 @@
+# Pauli Twirling
+
+Pauli Twirling (PT) is an error mitigation technique in which,
+in the Mitiq implementation, particularly noisy operations (e.g. CZ and CNOT)
```
Misty-W, line 4: Suggest to simplify this wording a bit for readability.
```suggestion
In the Mitiq implementation of the Pauli Twirling (PT)  error mitigation technique,
particularly noisy operations (e.g. CZ and CNOT)
```

docs/source/guide/pt-1-intro.md
```diff
@@ -0,0 +1,104 @@
...
+```{code-cell} ipython3
+import mitiq
+
+mitiq.SUPPORTED_PROGRAM_TYPES.keys()
+```
+
+
+## Problem setup
+We first define the circuit of interest. In this example, the circuit has a
+slack window with a length of 4 (in the sense that 4 single-qubit gates can fit in that window).
```
Misty-W, line 27: The circuit is of depth 3, shouldn't the slack length be "2"?
```suggestion
slack window with a length of 2 (in the sense that 2 single-qubit gates can fit in that window).
```

docs/source/guide/pt-1-intro.md
```diff
@@ -0,0 +1,104 @@
...
+    circuit=circuit,
+    executor=execute,
+)
+```
+
+```{code-cell} ipython3
+print(f"Error with mitigation (PT): {abs(ideal_value - mitigated_result) :.3}")
+```
+
+Here we observe that the application of PT reduces the estimation error when compared
```
Misty-W, line 89: Unfortunately this doesn't seem to be the case in https://mitiq--1848.org.readthedocs.build/en/1848/guide/pt-1-intro.html .

At least part of the problem appears to be that there is no noise applied to obtain the "noisy" result.

docs/source/guide/pt-1-intro.md
```diff
@@ -0,0 +1,104 @@
...
+to the unmitigated result.
+
+```{admonition} Note:
+PT is designed to mitigate the simple Markovian noise simulated in this example,
+but PT can still have a non-trivial effect on the final error, but it is not
```
Misty-W, line 94:
```suggestion
but PT can still have a non-trivial effect on the final error, although it is not
```

docs/source/guide/pt-4-low-level.md
```diff
@@ -0,0 +1,126 @@
...
+
+- The user provides a `QPROGRAM`, (i.e. a quantum circuit defined via any of the supported [frontends](frontends-backends.md)).
+- Mitiq modifies the input circuit with the insertion of PT gates on noisy operations.
+- The modified circuit is executed via a user-defined [Executor](executors.md).
+- The error mitigated expectation value is returned to the user.
+
+With respect to the workflows of other error-mitigation techniques (e.g. [ZNE](zne-4-low-level.md) or [PEC](pec-4-low-level.md)),
+PT involves the generation of a _single_ randomly-modified circuit, and then averages over its executions.
+For this reason, there is no need for a complex final inference step, which is necessary for other 
+techniques, and so this average is instead trivial for PT.
```
Misty-W, line 34:
```suggestion
For this reason, there is no need for a complex final inference step (unlike in other 
techniques), and so this average is instead trivial for PT.
```

docs/source/guide/pt-4-low-level.md
```diff
@@ -0,0 +1,126 @@
...
+
+As a final remark, we stress that the low-level procedure that we have shown is exactly what {func}`.execute_with_pt()` does behind the scenes.
+Let's verify this fact: 
+
+```{code-cell} ipython3
+np.isclose(
+  pt.execute_with_pt(circuit, executor),
+  average,
+)
+```
```
Misty-W, line 126: Unfortunately, the results displayed don't match the text- `np.isclose` asserts `False`, see https://mitiq--1848.org.readthedocs.build/en/1848/guide/pt-4-low-level.html .

Misty-W (commented):
Nice to see this filled out, @Aaron-Robertson !

My main concern is that the error-mitigated results displayed do not match the descriptions in the text, i.e. showing a worse result in the intro, and the low-level results not closely matching the results from `execute_w_pt`.

The rest are minor wording suggestions.

andreamari:
@Aaron-Robertson, thanks this is a lot of useful documentation!
This is not a proper review, but I'd like to give some general comments:

1. I think we shouldn't claim that PT is an error _mitigation_ technique. By design, PT is not expected to reduce noise. The aim of PT is to transform any noise channel into a Pauli noise channel. So PT is more a _noise transformation_ technique. From a code point of view, the workflow is similar to other techniques but the physical effect is different. Typically, PT is used as a pre-processing step before applying  other mitigation techniques. This aspect is perhaps the most important thing we should say in the docs about PT, otherwise users will try to apply it in the wrong way.

2. I wouldn't mention the subtle aspects about Markovian vs non-Markovian noise. Those aspects are very important for DDD (a technique originally designed for non-Markovian noise) but are much less important for PT which has a different goal.

3. While DDD is expected to work even with a single circuit (without averaging over different DDD sequences), for PT averaging over many circuits is essential otherwise the noise is not mapped to a Pauli channel.  So, if possible, we should change the workflow figures and the associated discussion accordingly. The input circuit is first mapped to many twirled circuits (step 1) --> execution ---> The results are averaged (step 2).

Aaron-Robertson:
Thanks @andreamari @Misty-W!

> * I think we shouldn't claim that PT is an error _mitigation_ technique. By design, PT is not expected to reduce noise. The aim of PT is to transform any noise channel into a Pauli noise channel. So PT is more a _noise transformation_ technique. From a code point of view, the workflow is similar to other techniques but the physical effect is different. Typically, PT is used as a pre-processing step before applying  other mitigation techniques. This aspect is perhaps the most important thing we should say in the docs about PT, otherwise users will try to apply it in the wrong way.
> *

Great point. I tried to capture aspects of it, but I'll be more explicit in calling it a noise transformation and/or tailoring technique. In this way, would it make sense to remove the expectation value examples from the PT section? Or simply state that the effect on the expectation is as we would expect.

andreamari:
> Great point. I tried to capture aspects of it, but I'll be more explicit in calling it a noise transformation and/or tailoring technique. In this way, would it make sense to remove the expectation value examples from the PT section? Or simply state that the effect on the expectation value is as we would expect?

We have the function `execute_with_pt` in Mitiq and so it is fine show it in the docs and stating that we don't expect a better accuracy when applying PT alone. 
Alternatively we could directly show something like PT+ZNE, which is the expected reduce errors. I originally thought to keep the PT+XXX case as a future Mitiq example, but it could also go in the main docs if you think if fits well.

Any option is good, as long as we don't give the wrong impression that PT mitigates noise.

purva-thakre:
If the goal is to show pauli twirling works to tailor the noise, one option could be to look at the PTM superoperator before & after. The discussion could be about how the off-diagonal terms are gone after pauli twirling.

purva-thakre:
If we are moving past calculating an expectation value for pauli twirling, should we change `execute_with_pauli_twirling` to `tailor_noise_with_pauli_twirling`?

Misty-W:
> @Misty-W @Aaron-Robertson If we are moving past calculating an expectation value for pauli twirling, should we change `execute_with_pauli_twirling` to `tailor_circuit_with_pauli_twirling`?

I vote "yes"!

purva-thakre:
@Aaron-Robertson Do you have any local changes for this PR? We changed the output of the twirling functions in #1977. I can take over from here and finish this PR as well.

Misty-W:
hi @purva-thakre and @Aaron-Robertson, I'm guessing this PR will probably need to go in the next release. Importantly, we've made good progress toward the pre-requisite of checking the PTM, and we can build up incrementally from there.

purva-thakre:
> hi @purva-thakre and @Aaron-Robertson, I'm guessing this PR will probably need to go in the next release. Importantly, we've made good progress toward the pre-requisite of checking the PTM, and we can build up incrementally from there.

Yep, the deadline for this one needs to be stretched to the next milestone. There were a few things outside the Pauli Twirling module that needed to be dealt with before tackling this PR.

purva-thakre:
@Misty-W @Aaron-Robertson @andreamari @nathanshammah @natestemen Do either of you have access to the original images used in this PR? I need to make changes to them as the functions only tailor the noise with the new changes. 

If not, it's ok. I can create my own.

natestemen:
> Do either of you have access to the original images used in this PR?

It looks like an svg, so if you load it into inkscape (or similar), I believe you should be able to modify it?

purva-thakre:
Never edited an svg before. I can give it a try. Thanks!

docs/source/guide/pt-4-low-level.md
```diff
@@ -0,0 +1,128 @@
...
+from typing import cast
+
+average = cast(float, np.average(expvals))
+print(average)
+```
+
+Keep in mind, ths code is for illustration and that the noise level, type of noise (here amplitude damping), and the observable need to be adapted to the specific experiment.
+
+
+On a real backend, they have a different sensitivity to noise. The core idea of the PT technique is that,
```
purva-thakre, line 118: If one were to demonstrate an example of PT + some other mitigation technique, is it preferable to 

1. introduce a separate tutorial?
2. Or demonstrate it in this section to keep it consistent with what @Aaron-Robertson originally intended?

docs/source/guide/pt-3-options.md
```diff
@@ -0,0 +1,59 @@
...
+```
+
+```{code-cell} ipython3
+from mitiq import pt
+twirl_CZ_only = pt.twirl_CZ_gates(
+    circuit = circuit,
+    num_circuits = 1,
+)
+
+twirl_CZ_only = pt.twirl_CNOT_gates(
```
purva-thakre, line 49:
```suggestion
twirl_CNOT_only = pt.twirl_CNOT_gates(
```

docs/source/guide/pt-4-low-level.md
```diff
@@ -0,0 +1,128 @@
...
+demonstration, we compare the off-diagonal terms in the Pauli Transfer Matrix (PTM)
+{cite}`Hashim_2021_PRX` of the original input circuit, the noisy input circuit as well
+as the averaged PTM of the different possible Pauli twirled circuits. A detailed discussion
+on the PTM is available in [theory section of PT](pt-5-theory.md).
+
+Now, instead of using the ideal input circuit, we use PT on the noisy circuit with the
+default option for the number of twirled circuits. 
+
+```{code-cell} ipython3
+from mitiq import pt
```
purva-thakre, line 93:
```suggestion
from mitiq import pt
from cirq import amplitude_damp
from mitiq.interface import convert_to_mitiq


noise_level = 0.1
mitiq_circuit, _ = convert_to_mitiq(circuit)
noisy_circuit = mitiq_circuit.with_noise(amplitude_damp(gamma=noise_level))
```

docs/source/guide/pt-1-intro.md
```diff
@@ -0,0 +1,101 @@
...
+
+```{code-cell} ipython3
+for i in range(len(tailored_circuit_ideal)):
+    print(f"Example Pauli Twirled Circuit {i}: \n {tailored_circuit_ideal[i]}, \n")
+```
+
+Here, we observe that the application of the Pauli Twirling technique has kept the logic
+of the input circuit unchanged while introducing a combination of gates to twilred outputs.
+
+## Combine PT with ZNE
```
purva-thakre, line 94: Move Aaron's code here where the ideal expectation value is compared to the mitigated one.

nathanshammah:
Apparently the svg file for the PT workflow shows differently in some browsers, and in others it looks fine. If possible, it would be nice to fix this so that the text does not go beyond the image. Using `png` format could also be fine.

mitiq/pt/pt_utils.py
```diff
@@ -0,0 +1,92 @@
...
+    num_qubits = len(circuit.all_qubits())
+    try:
+        superop = cirq.kraus_to_superoperator(cirq.kraus(circuit))
+    except TypeError:
+        kraus_list = []
+        for i in range(len(circuit)):
+            kraus_of_op = cirq.kraus(circuit[i])
+            kraus_list.append(kraus_of_op)
+        all_kraus = list(itertools.chain(*kraus_list))
+        superop = cirq.kraus_to_superoperator(all_kraus)
```
purva-thakre, line 59:
```suggestion
    except TypeError:
        kraus_list = []
        for i in range(len(circuit)):
            kraus_of_op = cirq.kraus(circuit[i])
            kraus_list.append(kraus_of_op)
        all_kraus = list(itertools.chain(*kraus_list))
        superop = cirq.kraus_to_superoperator(all_kraus)
```
@Misty-W @andreamari Would either of you mind looking at these lines for me? I am trying to get the superoperator of a noisy circuit or a twirled version of the noisy circuit. Unfortunately, there is an error that I can't find. 

The norm of this superoperator is not 4 (as expected) for a two-qubit circuit with a CNOT gate and depolarizing noise, twirled versions of the noiseless circuit or twirled versions of the circuit with depolarized noise. 

I believe this is screwing up my averaged over PTM off-diagonal terms. I do think for the noiseless circuit, the off-diagonal terms should become zero after twirling different versions of the input circuit. 

I know the tests are failing right now because I have not added any tests for the new functions. I also know I have skipped type annotations in the defined functions. Going to leave this for later.

purva-thakre:
@andreamari @Misty-W This might be a very dumb question but I wanted to check my understanding of how Pauli Twirling should work. 

If I Pauli Twirl the following circuit without any noise inserted into the circuit, should the off-diagonal terms in the PTM disappear?
```python
from cirq import LineQubit, Circuit, CNOT, depolarize

a, b  = LineQubit.range(2)
circuit = Circuit(CNOT.on(a, b))

print(circuit)
``` 
```
0: ───@───
      │
1: ───X───
```

image

The plot of all the PTM array elements before PT (labeled as 'initial' above) and after PT is the same. 

Whereas I expected the PTM after PT to be a diagonal matrix as shown on the right. You can ignore the values for the 'expected' plot as I chose a random diagonal matrix to make my point.

purva-thakre:
@Misty-W Some of the off-diagonal terms don't go to zero even after using PT. I do see their magnitude decrease but regardless of the noise level or noise type, some terms are consistently stuck at being 0.25 when we want them to be closer to 0. Would you like to have a quick meeting to discuss where to go from here? FWIW I do have a possible solution discussed before the last textblock. 

BTW I know my code lacks proper tests and annotations etc. I will work on this later in the week. I spent the weekend trying to figure out if I had an error somewhere. I don't. The PTM function's output satisfies all of the general properties I could find. Something related to how the CNOT propagates errors, I believe. 

Here's my code for showing some terms will not go to zero:

```python
from cirq import LineQubit, Circuit, CZ, CNOT, depolarize

a, b  = LineQubit.range(2)
circuit = Circuit(CNOT.on(a, b))

print(circuit)
```
```
0: ───@───
      │
1: ───X───
```
```python
noisy_circuit = circuit.with_noise(depolarize(p = 0.15))
print(noisy_circuit)
```
```
0: ───@───D(0.15)[cirq.VirtualTag()]───
      │
1: ───X───D(0.15)[cirq.VirtualTag()]───
```
```python
from mitiq.pt.pt_utils import (pauli_transfer_matrix, twirled_pauli_transfer_matrix)

ideal_ptm = pauli_transfer_matrix(circuit)
noisy_ptm = pauli_transfer_matrix(noisy_circuit)
twirled_result = pt.pauli_twirl_circuit(circuit=noisy_circuit, num_circuits = 1000)
twirled_ptm = twirled_pauli_transfer_matrix(twirled_result)

# find the non-zero off-diagonal elements and compare to the noisy ptm
for i in range(16):
    for j in range(16):
        if i != j:
            if not np.isclose(twirled_ptm[i, j], 0.0):
                print((i, j), twirled_ptm[i, j], noisy_ptm[i,j], "\n")
```
Regardless of the noise level `p`, the following elements in the array are stuck to a magnitude of 0.25 after twirling. Before twirling, the magnitude of these terms was 0.5. 

I guess I could make an argument that twirled PTM elements are closer to 0.0 on the number line compared to noisy PTM elements. For example, -0.25 is closer to 0.0 compared to -0.5
```
(2, 14) (-0.25+0j) (-0.5+0j) 

(3, 15) (0.25+0j) (0.5+0j) 

(4, 5) (0.25+0j) (0.5+0j) 

(5, 4) (0.25+0j) (0.5+0j) 

(6, 11) (-0.25+0j) (-0.5+0j) 

(7, 10) (-0.25+0j) (-0.5+0j) 

(8, 9) (-0.25+0j) (-0.5+0j) 

(9, 8) (-0.25+0j) (-0.5+0j) 

(10, 7) (-0.25+0j) (-0.5+0j) 

(11, 6) (-0.25+0j) (-0.5+0j) 

(14, 2) (-0.25+0j) (-0.5+0j) 

(15, 3) (0.25+0j) (0.4999999999999999+0j) 
```

mitiq/pt/pt_utils.py
```diff
@@ -0,0 +1,124 @@
...
+    # need to think more on why I need to use below if-else loop
+    # some bug? Or missing something conceptually?
+    # only the layers inserted in the twirling process matter below
+    if ptm_matrix[0, 0] != 1.0:
+        num_layers = len(circuit)
+        ptm_matrix = ptm_matrix / num_layers
+    else:
+        ptm_matrix = ptm_matrix
+
+    return ptm_matrix
```
purva-thakre, line 105: The bug referenced in another comment is related to these lines.

purva-thakre:
I suggest we merge this once all the tests pass. I have created `Notes` to make a comment along the lines of some things are coming in the future. 

I think I need some more time on the PTM issue because there's either a bug in my code or my understanding of how Pauli Twirling is supposed to work is flawed. For some reason, I am still stuck on thinking the `Easy` gates in randomized compiling help twirl the errors on the `Hard` gate set. In our case, we are twirling the `hard` gate set on its own. Either way, I need to work out the math first.

purva-thakre:
@Misty-W This is ready to be merged on a tentative basis. There are warnings on every page of the user guide that the Pauli Twirling module is currently WIP.

Misty-W (approved):
Approved, with plan to update in the next milestone! Thanks for picking this up @purva-thakre!

PR #1850, opened 2023-05-27
Add function to register user defined Mitiq converters

### LINKED ISSUE #1377: Custom package and circuit interface
Pre-Request Checklist
---------------------

Issue Description
-----------------

Currently, if we want to interface mitiq with quantum packages that not listed in ``interfaces`` submodule, we can take the following workaround:

```python
import mypackage

mycircuit = mypackage.Circuit()

def to_cirq(mycircuit):
    # ...
    return cirq.Circuit

cirqcircuit = to_cirq(mycircuit)

def from_cirq(cirqcircuit):
    # ...
    return mypackage.Circuit()

def executor(cirqcircuit):
    mycircuit = from_cirq(cirqcircuit)
    return mypackage.run(mycircuit)
 

zne.execute_with_zne(cirqcircuit, executor)
```

However, this workflow is not straightforward and error-prone. Since we have defined ``from_cirq`` and ``to_cirq`` anyway, a better integration would be a hook to register these two methods to ``convert_to_mitiq`` and ``convert_from_mitiq``. Then the above work flow will be simplified to 

```python

# after to and from mitiq registeration

mycircuit = mypackage.Circuit()

def executor(mycircuit):
    return mypackage.run(mycircuit)
 

zne.execute_with_zne(mycircuit, executor)

```

Namely, just native user experience as the currently supported quantum software packages.

Proposed Solution
-----------------

For now, the extension on ``convert_to_mitiq`` and ``convert_from_mitiq`` is impossible since the two function are hand coded if else branch with no user customized fallback. Therefore, a possible API for the hook and the corresponding implementation may be as follows:

```python
# the proposed API
mitiq.interfaces.register_to_mitiq(package_name="mypackage", convert_function=to_cirq)
mitiq.interfaces.register_from_mitiq(package_name="mypackage", convert_function=from_cirq)

# the impl for convert_to_mitiq
def register_to_mitiq(package_name, convert_function):
    global register_dict
    register_dict[package_name] = convert_function

def convert_to_mitiq(circuit):
    if 
    elif
    # ...
    else:
        for package_name in register_dict:
             if package_name in package:
                conversion_function = register_dict[package_name]
                break
        else:
             raise UnsupportedCircuitError
    
```

Additional References
---------------------

Aaron-Robertson:
## Description

Fixes #1377

---

### License

### DIFF
--- a/mitiq/interface/conversions.py
+++ b/mitiq/interface/conversions.py
@@ -19,18 +19,20 @@ class UnsupportedCircuitError(Exception):
 class CircuitConversionError(Exception):
     pass
 
+
 register_dict: Dict[Tuple[str, str], Callable[[Any], Circuit]] = {}
 
+
 def register_mitiq_converter(
-    package_name: str, 
-    direction: str, 
-    convert_function: Callable[[Any], Circuit]
-):
+    package_name: str,
+    direction: str,
+    convert_function: Callable[[Any], Circuit],
+) -> None:
+    global register_dict
     if direction not in ["to", "from"]:
         raise ValueError("Invalid direction. Expected 'to' or 'from'.")
     register_dict[(package_name, direction)] = convert_function
-    global register_dict
-    register_dict[package_name] = convert_function
+
 
 def convert_to_mitiq(circuit: QPROGRAM) -> Tuple[Circuit, str]:
     """Converts any valid input circuit to a mitiq circuit.
@@ -80,17 +82,19 @@ def convert_to_mitiq(circuit: QPROGRAM) -> Tuple[Circuit, str]:
 
         def conversion_function(circ: Circuit) -> Circuit:
             return circ
+
     else:
-        for package_name in register_dict:
-             if package_name in package:
-                conversion_function = register_dict[(package_name, "from")]
+        for package_name, _ in register_dict:
+            if package_name in package:
+                conversion_function = register_dict[package_name, "from"]
                 break
         else:
-             raise UnsupportedCircuitError(
-            f"Circuit from module {package} is not supported.\n\n"
-            f"Please define a converter with register_mitiq_converter(),"
-            f"\nor specify a supported Circuit type: \n{SUPPORTED_PROGRAM_TYPES}"
-        )
+            raise UnsupportedCircuitError(
+                f"Circuit from module {package} is not supported.\n\n"
+                f"Please define a converter with register_mitiq_converter(),"
+                f"\nor specify a supported Circuit type:"
+                f"\n{SUPPORTED_PROGRAM_TYPES}"
+            )
 
     try:
         mitiq_circuit = conversion_function(circuit)
@@ -136,17 +140,20 @@ def convert_from_mitiq(circuit: Circuit, conversion_type: str) -> QPROGRAM:
 
         def conversion_function(circ: Circuit) -> Circuit:
             return circ
+
     else:
-        for package_name in register_dict:
-             if package_name == conversion_type:
-                conversion_function = register_dict[(package_name, "from")]
+        for package_name, _ in register_dict:
+            if package_name == conversion_type:
+                conversion_function = register_dict[package_name, "from"]
                 break
         else:
             raise UnsupportedCircuitError(
-            f"Conversion to circuit of type {conversion_type} is unsupported."
-            f"\n\nPlease define a converter with register_mitiq_converter(),"
-            f"\nor specify a supported Circuit type: \n{SUPPORTED_PROGRAM_TYPES}"
-        )
+                f"Conversion to circuit type {conversion_type} is unsupported."
+                f"\n\nPlease register a converter with"
+                f"register_mitiq_converter(),"
+                f"\nor specify a supported Circuit type:"
+                f"\n{SUPPORTED_PROGRAM_TYPES}"
+            )
 
     try:
         converted_circuit = conversion_function(circuit)
--- a/mitiq/interface/conversions.py
+++ b/mitiq/interface/conversions.py
@@ -36,7 +36,9 @@ except NameError:
 def register_mitiq_converter(
     package_name: str,
     direction: str,
-    convert_function: Union[Callable[[Any], Circuit], Callable[[Circuit], Any]],
+    convert_function: Union[
+        Callable[[Any], Circuit], Callable[[Circuit], Any]
+    ],
 ) -> None:
     """Registers converters for unsupported circuit types.
 
--- a/mitiq/interface/conversions.py
+++ b/mitiq/interface/conversions.py
@@ -5,7 +5,7 @@
 
 """Functions for converting to/from Mitiq's internal circuit representation."""
 from functools import wraps
-from typing import Any, Callable, cast, Iterable, Tuple
+from typing import Any, Callable, cast, Iterable, Tuple, Dict
 
 from cirq import Circuit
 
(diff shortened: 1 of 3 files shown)


Aaron-Robertson:
@natestemen @andreamari This PR uncovered a bug in the draft workflow. The oversights have to do with the `mitiq/tests` directory and with pyquil. I opened issue #1851 to track.

Otherwise it's functional so I'm setting as r4r. That said, the typing on `convert_function` is incorrect, and I think it needs to remain incorrect to pass checks until `QPROGRAM` is more generic. Just let me know if there's a better suggestion!

mitiq/interface/conversions.py
```diff
@@ -70,10 +115,18 @@ def conversion_function(circ: Circuit) -> Circuit:
...
+        for package_name, _ in register_from_dict:
+            if package_name in package:
+                input_circuit_type = package_name
+                conversion_function = register_from_dict[package_name, "from"]
+                break
+        else:
+            raise UnsupportedCircuitError(
+                f"Circuit from module {package} is not supported.\n\n"
+                f"Please define a converter with register_mitiq_converter(),"
+                f"\nor specify a supported Circuit type:"
```
Misty-W, line 127: I believe this was a typo?
```suggestion
                f"\or specify a supported Circuit type:"
```

Misty-W (commented):
Great @Aaron-Robertson! 

Can you add a test for the generation of the conversion function in `convert_to_mitiq`?

Codecov is also flagging missing tests for raising an error on an "invalid direction" (not to/from) and for raising an unsupported circuit error when the circuit type hasn't been specified and registered.

Aaron-Robertson:
> Can you add a test for the generation of the conversion function in `convert_to_mitiq`?

@Misty-W ~~I'm not quite sure what you mean here, or at least how this test would differ from the added tests `test_register_from_mitiq()` and `test_register_to_mitiq()`.~~ I found the typo that skipped 177-180, should wrap up here shortly!

mitiq/interface/conversions.py
Aaron-Robertson: This one's a line break. I'll add the space to clarify here and on similar cases nearby.

Misty-W: I see, thanks!

mitiq/interface/conversions.py
```diff
@@ -20,6 +20,51 @@ class CircuitConversionError(Exception):
...
+        convert_function: User specified function to convert to and from an
+            unsupported circuit type. These functions should follow the
+            direction hint above: "from" returns a Mitiq/Cirq circuit,
+            "to" returns a Non-Mitiq circuit.
+    """
+    if direction not in ["to", "from"]:
+        raise ValueError("Invalid direction. Expected 'to' or 'from'.")
+    elif direction == "to":
+        global register_to_dict
+        register_to_dict[(package_name, direction)] = convert_function
```
natestemen, line 62: I don't think the `global` keyword is needed here (or in other places), is it? I think this is only needed when the global variable is immutable.

mitiq/interface/conversions.py
```diff
@@ -20,6 +20,51 @@ class CircuitConversionError(Exception):
     pass
 
 
+register_to_dict: Dict[Tuple[str, str], Callable[[Any], Circuit]]
```
natestemen, line 23: Given we define two dictionaries one for `to` and other for `from`, why do we need to store the direction in the key?

mitiq/interface/conversions.py
```diff
@@ -20,6 +20,51 @@ class CircuitConversionError(Exception):
...
+try:
+    register_to_dict
+except NameError:
+    register_to_dict = {}
+
+register_from_dict: Dict[Tuple[str, str], Callable[[Circuit], Any]]
+try:
+    register_from_dict
+except NameError:
+    register_from_dict = {}
```
natestemen, line 33: Can you explain this code a little bit? I have a feeling it has something to do with not wanting to overwrite an existing conversion dictionary, but I'm not totally sure.

mitiq/tests/test_conversions.py
```diff
@@ -103,6 +131,14 @@ def test_from_mitiq(to_type):
     assert input_type == to_type
 
 
+def test_register_from_mitiq():
+    circuit = qasm_circuit
+    register_mitiq_converter(circuit.__module__, "to", to_qasm)
+    converted_circuit, input_type = convert_to_mitiq(circuit)
+    assert _equal(converted_circuit, cirq_circuit)
+    assert input_type in circuit.__module__
```
natestemen, line 139: I'm confused by this test. If we are using `to_qasm`, shouldn't `converted_circuit` be of type `str`? Should line 137 be `convert_from_mitiq`?

natestemen (commented):
Nice work Aaron! Great to have this in. Was just giving it a quick read through and just had some questions if you wouldn't mind answering. Only one seems to require a potential change (`test_register_from_mitiq`).

natestemen:
This changes seems to be causing some errors in other builds: https://github.com/unitaryfund/mitiq/actions/runs/5125742133/jobs/9219321180#step:5:3905

Misty-W:
> This changes seems to be causing some errors in other builds: https://github.com/unitaryfund/mitiq/actions/runs/5125742133/jobs/9219321180#step:5:3905

I think in that case it should, Mitiq doesn't support QASM directly, so the user needs to define a conversion function or convert to a supported frontend.

natestemen:
But that's exactly what this change (and the test that is failing) is for, no?

Misty-W:
> But that's exactly what this change (and the test that is failing) is for, no?

Yes it is, but the user didn't define or specify a converter, and that's why the test is failing.

natestemen:
Ahhhh we've created **order dependent tests** 😨... We _do_ have a test that registers the `from_mitiq` function (for QASM) which is supposed to go into a global variable that can later be accessed:
https://github.com/unitaryfund/mitiq/blob/c9fdac0eab9d244f5b1588be6549d0c0da629e23/mitiq/tests/test_conversions.py#L100-L105
This test comes before the later test (in terms of lines of code) which is failing, which means we are relying on execution order for the tests to pass. This is a "Known bad practice"™️, so much so that ruby has a function which you can put at the top of your tests called [`i_suck_and_my_tests_are_order_dependent!`](https://www.rubydoc.info/gems/minitest/Minitest%2FTest.i_suck_and_my_tests_are_order_dependent!) 🤣.

I think we can merge these two tests to ensure they are executed together? Any downside to that?

mitiq/interface/conversions.py
Aaron-Robertson: That's correct. Initially the explicit `global` sufficed, but this one became necessary at some point (I can't remember exactly why), so in #1864 I left this `try/catch` but removed `global`.

mitiq/tests/test_conversions.py
Aaron-Robertson: I made more explicit the use of the mock class in the update: `CircuitStr(str)` takes a string, and defines a `__module_="qasm"` property on it`

Aaron-Robertson:
> I think we can merge these two tests to ensure they are executed together? Any downside to that?

It's kind of a weird pattern throughout `test_conversions.py`, since in every `from` conversion test we use both `convert_from_mitiq` and `convert_to_mitiq`. In this case, I forgot to do that-part of how I missed the entanglement! We can certainly use only a single `...from_to...` test (as I do in the update), but if we wanted to follow the pattern in `test_conversions.py` then we might want to keep both (and just re-register the `to` converter in both).

PR #1863, opened 2023-05-30
Improve cost estimation for Calibrator

### LINKED ISSUE #1715: Improve the cost estimation to run a Calibrator
## Issue Description

The Calibrator class as a [get_cost()](https://github.com/unitaryfund/mitiq/blob/ac53f1dd700507a92fbcb0fdbe6757fccefd68a0/mitiq/calibration/calibrator.py#L88) method to estimate the number of circuit executions that are necessary to get all the required data. 

At the moment, the estimate is not correct since it doesn't take into account that each Strategy requires the execution of multiple circuits for a single error-mitigated expectation value (e.g. ZNE requires k circuits where k is the number of noise scale factors, PEC requires `num_samples` circuits, etc.. ).

This issue is about getting a more faithful estimate of the actual cost.

## Proposed Solution

Maybe each Strategy object could have a `num_circuits_required()` method, such that one can simply scale the result in [get_cost()](https://github.com/unitaryfund/mitiq/blob/ac53f1dd700507a92fbcb0fdbe6757fccefd68a0/mitiq/calibration/calibrator.py#L88)
by that number. 

Other ideas are welcome.

YuNariai:
Link to the issue:
Fixes #1715 Improve the cost estimation to run a Calibrator

## Description
1 Add `num_circuits_required()` method to Strategy to count the number of circuits for ZNE and PEC

2 Modify `get_cost()` in calibrator using `num_circuits_required()`

3 Modify `to_dict` in Strategy so that it can actually return scale_factors for ZNE and return num_samples for PEC

---

### License

### DIFF
--- a/mitiq/calibration/calibrator.py
+++ b/mitiq/calibration/calibrator.py
@@ -181,7 +181,9 @@ class Calibrator:
             A summary of the number of circuits to be run.
         """
         num_circuits = len(self.problems)
-        num_options = len(self.settings.technique_params)
+        num_options = 0
+        for strategy in self.settings.make_strategies():
+            num_options += strategy.num_circuit_required()
 
         noisy = num_circuits * num_options
         ideal = 0  # TODO: ideal executor is currently unused
--- a/mitiq/calibration/calibrator.py
+++ b/mitiq/calibration/calibrator.py
@@ -182,7 +182,7 @@ class Calibrator:
         """
         num_circuits = len(self.problems)
         num_options = sum(
-            strategy.num_circuit_required() for strategy in self.strategies
+            strategy.num_circuits_required() for strategy in self.strategies
         )
 
         noisy = num_circuits * num_options
--- a/mitiq/calibration/calibrator.py
+++ b/mitiq/calibration/calibrator.py
@@ -181,9 +181,9 @@ class Calibrator:
             A summary of the number of circuits to be run.
         """
         num_circuits = len(self.problems)
-        num_options = 0
-        for strategy in self.settings.make_strategies():
-            num_options += strategy.num_circuit_required()
+        num_options = sum(
+            strategy.num_circuit_required() for strategy in self.strategies
+        )
 
         noisy = num_circuits * num_options
         ideal = 0  # TODO: ideal executor is currently unused
--- a/mitiq/calibration/settings.py
+++ b/mitiq/calibration/settings.py
@@ -203,9 +203,7 @@ class Strategy:
         if self.technique is MitigationTechnique.ZNE:
             inference_func = self.technique_params["factory"]
             summary["factory"] = inference_func.__class__.__name__
-
             summary["scale_factors"] = inference_func._scale_factors
-
             summary["scale_method"] = self.technique_params[
                 "scale_noise"
             ].__name__
--- a/mitiq/calibration/settings.py
+++ b/mitiq/calibration/settings.py
@@ -245,6 +245,8 @@ class Strategy:
             return len(summary["scale_factors"])
         elif self.technique is MitigationTechnique.PEC:
             return summary["num_samples"]
+        elif self.technique is MitigationTechnique.RAW:
+            return 1
         return 0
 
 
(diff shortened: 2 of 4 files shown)


YuNariai:
The check says I lost the return in this function:
```
def num_circuit_required(self) -> int:
        summary = self.to_dict()
        if self.technique is MitigationTechnique.ZNE:
            return len(summary["scale_factors"])
        # TODO: elif self.technique is MitigationTechnique.PEC:
```
but it does have return, how should I fix this?

mitiq/calibration/settings.py
```diff
@@ -237,6 +240,14 @@ def print_line(self, performance: str, circuit_type: str) -> None:
 
     def __repr__(self) -> str:
         return str(self.to_dict())
+    
+    def num_circuit_required(self) -> int:
+        summary = self.to_dict()
+        if self.technique is MitigationTechnique.ZNE:
+            return len(summary["scale_factors"])
+        elif self.technique is MitigationTechnique.PEC:
+            return summary["num_samples"]
```
natestemen, line 249: To make `mypy` happy, you could add something simple like `return 0` at the end of the function. `mypy` does not know that the only options for `self.technique` are `MitigationTechnique.ZNE` or `MitigationTechnique.PEC`.

```suggestion
    def num_circuit_required(self) -> int:
        summary = self.to_dict()
        if self.technique is MitigationTechnique.ZNE:
            return len(summary["scale_factors"])
        elif self.technique is MitigationTechnique.PEC:
            return summary["num_samples"]
        return 0
```
_**untested!**_

---

BTW, this PR is looking great. I plan to take a deeper look soon. Is it all ready for review?

YuNariai: Yes it's ready!

natestemen:
@YuNariai looks like you'll have to resolve some linting issues. You can check them using `make check-style`. Once those are resolved and `build/validate` is passing, I'll take a look!

mitiq/calibration/calibrator.py
```diff
@@ -181,7 +181,9 @@ def get_cost(self) -> Dict[str, int]:
             A summary of the number of circuits to be run.
         """
         num_circuits = len(self.problems)
-        num_options = len(self.settings.technique_params)
+        num_options = 0
+        for strategy in self.settings.make_strategies():
+            num_options += strategy.num_circuit_required()
```
natestemen, line 186: Strategies get stored in each `Calibrator` instance, so no need to regenerate them here.
```suggestion
        num_options = sum(
            strategy.num_circuit_required() for strategy in self.strategies
        )
```

mitiq/calibration/settings.py
```diff
@@ -203,7 +203,9 @@ def to_dict(self) -> Dict[str, Any]:
         if self.technique is MitigationTechnique.ZNE:
             inference_func = self.technique_params["factory"]
             summary["factory"] = inference_func.__class__.__name__
+
             summary["scale_factors"] = inference_func._scale_factors
+
```
natestemen, line 208: Were these added for readability or were they left in from debugging maybe?

YuNariai:
To make` mypy` happy, we added a `return 0`, but make codecov/patch unhappy...

Misty-W:
> To make` mypy` happy, we added a `return 0`, but make codecov/patch unhappy...

@YuNariai, to satisfy codecov you can add a simple test that checks if the function returns 0 when the appropriate condition is met.

mitiq/calibration/tests/test_settings.py
```diff
@@ -304,3 +305,12 @@ def test_to_dict():
...
     }
+
+
+def test_num_circuits_required():
+    undefine_strategy = Strategy(
+        id=1,
+        technique=MitigationTechnique.RAW,
+        technique_params={},
+    )
+    assert undefine_strategy.num_circuits_required() == 0
```
natestemen, line 316: Good thinking with this test. It made me realize that we should actually have another case in `num_circuits_required` where `RAW` should return 1 since raw execution just means execute the circuit.

If you add that as a case in `num_circuits_required` you can change this test so that it passes, and please ensure to use a specific name for this test, for example `test_num_circuits_required_raw_execution`. If you do this, you'll have to add another test (maybe `test_num_circuits_unknown_method`) to assert num circuits is 0.

Sorry for the back and forth! Once this change is made, it's ready to go!

natestemen (approved):
Awesome work! Technically `return None` is still not tested, but IMO it's completely fine.

PR #1864, opened 2023-05-30
Clean up global, isolate tests, and fix mock module

Aaron-Robertson:
## Description

Fixes #1850

---

### License

### DIFF
--- a/mitiq/interface/conversions.py
+++ b/mitiq/interface/conversions.py
@@ -20,17 +20,17 @@ class CircuitConversionError(Exception):
     pass
 
 
-from_mitiq_dict: Dict[str, Callable[[Circuit], Any]]
+FROM_MITIQ_DICT: Dict[str, Callable[[Circuit], Any]]
 try:
-    from_mitiq_dict
+    FROM_MITIQ_DICT
 except NameError:
-    from_mitiq_dict = {}
+    FROM_MITIQ_DICT = {}
 
-to_mitiq_dict: Dict[str, Callable[[Any], Circuit]]
+TO_MITIQ_DICT: Dict[str, Callable[[Any], Circuit]]
 try:
-    to_mitiq_dict
+    TO_MITIQ_DICT
 except NameError:
-    to_mitiq_dict = {}
+    TO_MITIQ_DICT = {}
 
 
 def register_mitiq_converters(
@@ -53,8 +53,8 @@ def register_mitiq_converters(
             unsupported circuit type. This function returns a Mitiq/Cirq
             circuit.
     """
-    from_mitiq_dict[package_name] = convert_to_function
-    to_mitiq_dict[package_name] = convert_from_function
+    FROM_MITIQ_DICT[package_name] = convert_to_function
+    TO_MITIQ_DICT[package_name] = convert_from_function
 
 
 def convert_to_mitiq(circuit: QPROGRAM) -> Tuple[Circuit, str]:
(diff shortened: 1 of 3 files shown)


mitiq/interface/conversions.py
```diff
@@ -115,10 +113,10 @@ def conversion_function(circ: Circuit) -> Circuit:
             return circ
 
     else:
-        for package_name, _ in register_from_dict:
+        for package_name in register_from_dict:
             if package_name in package:
                 input_circuit_type = package_name
-                conversion_function = register_from_dict[package_name, "from"]
+                conversion_function = register_from_dict[package_name]
                 break
```
natestemen, line 120: With the new form of `register_from_dict`, I think we can avoid the for loop:
```py
conversion_function = register_from_dict.get(package_name)
input_circuit_type = package_name
if conversion_function is None:
    raise UnsupportedCircuitError(...)
```

natestemen (commented):
Thanks for this quick fix! I like the changes you made. While we are making some changes, what do you think about changing the API for `register_mitiq_converter` to take both the `to` and `from` functions in one call. I can't imagine many instances where one would want to register the conversions only in one direction. Definitely not saying we have to make this change, but curious on your thoughts!

mitiq/interface/conversions.py
```diff
@@ -108,25 +99,24 @@ def convert_to_mitiq(circuit: QPROGRAM) -> Tuple[Circuit, str]:
 
         input_circuit_type = "pennylane"
         conversion_function = from_pennylane
+
+    elif register_from_dict.get(package):
+        input_circuit_type = package
+        conversion_function = register_from_dict.get(package)
```
natestemen, line 105: Oh shoot I've never had a chance to actually use the [walrus operator](https://realpython.com/python-walrus-operator/) (`:=`), but I think this might be a good use case.
```suggestion
    elif (conversion_function := register_from_dict.get(package)):
        input_circuit_type = package
```

Aaron-Robertson: @natestemen tests pass with the walrus, fun stuff! But I'm not sure how to resolve the type errors now, so if we could brainstorm for a few that'd be great!

mitiq/tests/test_conversions.py
```diff
@@ -97,13 +92,18 @@ def test_to_mitiq(circuit):
     assert input_type in circuit.__module__
 
 
-def test_register_to_mitiq():
-    circuit = qasm_circuit
-    register_mitiq_converter(circuit.__module__, "from", from_qasm)
-    converted_circuit, input_type = convert_to_mitiq(circuit)
-    assert _equal(converted_circuit, cirq_circuit)
-    assert input_type in circuit.__module__
+def test_register_from_to_mitiq(qasm_str=qasm_str, cirq_circuit=cirq_circuit):
```
natestemen, line 95: Could remove these since they are global variables.

```suggestion
def test_register_from_to_mitiq():
```

mitiq/interface/conversions.py
```diff
@@ -20,25 +20,24 @@ class CircuitConversionError(Exception):
     pass
 
 
-register_to_dict: Dict[Tuple[str, str], Callable[[Any], Circuit]]
+from_mitiq_dict: Dict[str, Callable[[Circuit], Any]]
```
natestemen, line 23: Since `from_mitiq_dict` is global, should we define it as `FROM_MITIQ_DICT`? I like the new name BTW!.

natestemen (approved):
Nice, and quick, work! Left two optional comments. Happy to go forward how you see fit.

🚢

Misty-W (approved):
Nice fixes @Aaron-Robertson! 

It's not blocking for me, but any idea why line 171 is flagged as not being covered by the tests?
I understand `codecov/project` has been behaving strangely lately, but this seems to be affecting `codecov/patch`.

Aaron-Robertson:
> It's not blocking for me, but any idea why line 171 is flagged as not being covered by the tests?
> I understand `codecov/project` has been behaving strangely lately, but this seems to be affecting `codecov/patch`.

I think it was never covered? Since this PR affects that function it got flagged, but otherwise it was merged and so might never get flagged as a "newly not covered" line.

Misty-W:
Great, thanks @Aaron-Robertson!

PR #1873, opened 2023-06-14
Add Calibration logging for PEC

### LINKED ISSUE #1861: Logs for PEC Calibration
What happens to the logging in the case of PEC- what is the equivalent of this the following messages:
image

_Originally posted by @andreamari in https://github.com/unitaryfund/mitiq/pull/1845#pullrequestreview-1450484565_

Misty-W:
## Description

Add Calibration logging for PEC. 

Fixes #1861 

A mixture of ZNE and PEC strategies is also supported in a simple way, as the appropriate header is printed initially and when the current technique is different from the previous technique. 

---

### License

### DIFF
--- a/mitiq/calibration/calibrator.py
+++ b/mitiq/calibration/calibrator.py
@@ -33,9 +33,9 @@ ZNE_TABLE_HEADER_STR = (
 )
 
 PEC_TABLE_HEADER_STR = (
-    "| performance | circuit | method | "
-    "representation function | operations to mitigate | noise level | "
-    " qubit dependent |\n"
+    "| performance | circuit | method "
+    "| representation function | operations to mitigate | noise level "
+    "| qubit dependent |\n"
     "| ----------- | ------- | ------ | ----------------------- "
     "| ---------------------- | ----------- | --------------- | "
 )
@@ -231,8 +231,10 @@ class Calibrator:
                 if (
                     log
                     and count
-                    and (strategy.technique
-                    is not self.strategies[count - 1].technique)
+                    and (
+                        strategy.technique
+                        is not self.strategies[count - 1].technique
+                    )
                 ):
                     if strategy.technique is MitigationTechnique.ZNE:
                         print(ZNE_TABLE_HEADER_STR)
--- a/mitiq/calibration/calibrator.py
+++ b/mitiq/calibration/calibrator.py
@@ -34,10 +34,10 @@ TABLE_HEADER_STR_ZNE = (
 
 TABLE_HEADER_STR_PEC = (
     "| performance | circuit | method | "
-    "representation function | noise level | noise bias factor "
-    "| qubit dependent |\n"
-    "| ----------- | ------- | ------ | ----------------------- | ----------- "
-    "| ----------------- | --------------- | "
+    "representation function | operations to mitigate | noise level | "
+    " qubit dependent |\n"
+    "| ----------- | ------- | ------ | ----------------------- "
+    "| ---------------------- | ----------- | --------------- | "
 )
 
 
--- a/mitiq/calibration/calibrator.py
+++ b/mitiq/calibration/calibrator.py
@@ -21,14 +21,23 @@ from mitiq.calibration.settings import (
     ZNESettings,
     Strategy,
     BenchmarkProblem,
+    MitigationTechnique,
 )
 from mitiq.interface import convert_from_mitiq
 
-TABLE_HEADER_STR = (
+TABLE_HEADER_STR_ZNE = (
     "| performance | circuit | method | extrapolation | scale factors "
     "| scale_method         |\n"
     "| ----------- | ------- | ------ | ------------- | ------------- "
     "| -------------------- |"
+)
+
+TABLE_HEADER_STR_PEC = (
+    "| performance | circuit | method | "
+    "representation function | noise level | noise bias factor "
+    "| qubit dependent |\n"
+    "| ----------- | ------- | ------ | ----------------------- | ----------- "
+    "| ----------------- | --------------- | "
 )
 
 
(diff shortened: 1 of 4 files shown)


Misty-W:
@natestemen, thought I'd clarify that this PR is ready for review now (after a few false starts, oops!)

mitiq/calibration/settings.py
```diff
@@ -220,21 +221,33 @@ def to_dict(self) -> Dict[str, Any]:
...
+            row = (
+                performance,
+                circuit_type,
+                self.technique.name,
+                summary["representation_function"],
+                str(summary["operations"]),
+                str(summary["noise_level"]),
             )
-        )
+            print(row)
```
natestemen, line 250: When I run this with the `light_pec_settings` object defined in the tests I get
```
| performance | circuit | method | representation function | operations to mitigate | noise level |  qubit dependent |
| ----------- | ------- | ------ | ----------------------- | ---------------------- | ----------- | --------------- | 
('✅', 'mirror', 'PEC', <function represent_operation_with_local_depolarizing_noise at 0x2880f0940>, '[cirq.Circuit([\n    cirq.Moment(\n        cirq.CNOT(cirq.LineQubit(0), cirq.LineQubit(1)),\n    ),\n]), cirq.Circuit([\n    cirq.Moment(\n        cirq.CZ(cirq.LineQubit(0), cirq.LineQubit(1)),\n    ),\n])]', '0.001')
('❌', 'mirror', 'PEC', <function represent_operation_with_local_depolarizing_noise at 0x2880f0940>, '[cirq.Circuit([\n    cirq.Moment(\n        cirq.CNOT(cirq.LineQubit(0), cirq.LineQubit(1)),\n    ),\n]), cirq.Circuit([\n    cirq.Moment(\n        cirq.CZ(cirq.LineQubit(0), cirq.LineQubit(1)),\n    ),\n])]', '0.001')
```

I think we want to use the [`format`](https://docs.python.org/3/library/stdtypes.html#str.format) function like we do with ZNE above. We'll also want to make sure to use the function name so we don't print extraneous info like the memory address. Similarly, do we want to print the operations? Or is there a shorter way to represent them? I don't think putting them in a table like this is very readable.

mitiq/calibration/calibrator.py
```diff
@@ -21,16 +21,25 @@
     ZNESettings,
     Strategy,
     BenchmarkProblem,
+    MitigationTechnique,
 )
 from mitiq.interface import convert_from_mitiq
 
-TABLE_HEADER_STR = (
+TABLE_HEADER_STR_ZNE = (
```
natestemen, line 28: I think the method should come first in the variable name now that there are two headers. So something like `ZNE_TABLE_HEADER`, WDYT?

mitiq/calibration/calibrator.py
```diff
@@ -210,12 +222,22 @@ def run(self, log: bool = False) -> None:
...
                     warnings.simplefilter("ignore", UserWarning)
                     mitigated_value = strategy.mitigation_function(
                         circuit, expval_executor
                     )
+                if (
+                    log
+                    and count
+                    and strategy.technique
+                    is not self.strategies[count - 1].technique
+                ):
```
natestemen, line 236: I think for readability the last condition
```py
strategy.technique is not self.strategies[count - 1].techique
```
should be in parenthesis. Technically, I think it's not totally needed based on pythons [operator precedence](https://docs.python.org/3/reference/expressions.html#operator-precedence) (`is not` binds more tightly than `and` if I'm reading this correctly), but IMO it reads a little easier. Not totally necessary, but WDYT?

mitiq/calibration/calibrator.py
```diff
@@ -210,12 +222,22 @@ def run(self, log: bool = False) -> None:
...
+                    if strategy.technique is MitigationTechnique.ZNE:
+                        print(TABLE_HEADER_STR_ZNE)
+                    elif strategy.technique is MitigationTechnique.PEC:
+                        print(TABLE_HEADER_STR_PEC)
```
natestemen, line 240: I think this is necessary for the moment, but I don't love how we now have to print table headers in two places (in the code). I feel like it's going to need to be refactored at some point to make all this work a little cleaner, but I'm not sure what that is now. Any thoughts?

natestemen (changes requested):
Nice work, and good tests. Biggest request is getting the lines to print in a formatted way similar to ZNE, but most other stuff looks good.

mitiq/calibration/calibrator.py
Misty-W: That makes sense, I've added the parentheses.

mitiq/calibration/calibrator.py
Misty-W: Sounds good, I've updated the names accordingly.

mitiq/calibration/settings.py
Misty-W: Good catch, thanks for pointing this out! 
I fixed the function name printing. 
As for the operations, I think users would want to track which operations are represented in the PEC calibration runs, even if that causes the table to be a bit bulky, but now the info is printed more compactly as operations (as intended).

mitiq/calibration/calibrator.py
Misty-W: I tried removing the `if count` condition and the code for initial printing of the header, which I thought would be a more succinct equivalent, but then the first header wasn't printed so I had to revert back.

natestemen (changes requested):
When running an example using the `light_pec_settings` from the tests it's printing out double headers.

Screenshot 2023-07-12 at 2 46 46 PM

The function name is also overflowing it's column. Do you think continuing to use a table here is the right choice if you want to print out the operations? I don't see a way that will fit in a text table.

Misty-W:
> When running an example using the `light_pec_settings` from the tests it's printing out double headers.
> 
> Screenshot 2023-07-12 at 2 46 46 PM

Thanks for pointing this out! Not sure how I missed this behavior before, but it turned out that the count determining whether to print the header "initially" was resetting with every "problem" instance.

Misty-W:
> The function name is also overflowing it's column. Do you think continuing to use a table here is the right choice if you want to print out the operations? I don't see a way that will fit in a text table.

I propose to keep the table and remove the operations, since all 2-qubit op's are represented by default now when running PEC calibration.

natestemen (approved):
Nice work Misty! Thanks for working to ensure everything fits into the table nicely.

Issue #1876, opened 2023-06-14
REM: execute_with_rem raises UnboundLocalError when called with Executor


d3rhans:
## Issue Description
`rem.execute_with_rem` expects an `executor` of type `Union[Executor, Callable[[QPROGRAM], MeasurementResult]]` as parameter. When called with some `Callable`, everything works as expected, when called with an `Executor` the function raises an `UnboundLocalError` exception.

## How to Reproduce
Minimal example to reproduce:

```py
import cirq
import numpy as np
from mitiq import rem, Executor, MeasurementResult, Observable, PauliString

def execute(circuit) -> MeasurementResult:
    noisy_circuit = circuit.with_noise(cirq.bit_flip(p=0.02))
    simulator = cirq.Simulator(seed=0)
    result = simulator.run(noisy_circuit, repetitions=100)
    
    bitstrings = np.column_stack(list(result.measurements.values()))
    return MeasurementResult(result=bitstrings)

qubits = cirq.LineQubit.range(2)
circuit = cirq.Circuit([cirq.X.on_each(*qubits)])

observable = Observable(PauliString("ZI"), PauliString("IZ"))
inverse_confusion_matrix = rem.generate_inverse_confusion_matrix(2, 0.02, 0.02)
executor = Executor(execute)

mitigated_result = rem.execute_with_rem(
    circuit,
    executor,
    observable,
    inverse_confusion_matrix=inverse_confusion_matrix,
)
```
When passing the `execute` function directly, everything works as expected. The error is caused in `mitiq/rem/rem.py` lines 41-46.

```py
    if not isinstance(executor, Executor):
        executor_obj = Executor(executor)

    executor_with_rem = mitigate_executor(
        executor_obj, inverse_confusion_matrix=inverse_confusion_matrix
    )
```
When an `Executor` object is passed, `executor_obj` never gets initialized. For comparison, `mitigate_executor` got this case covered:

```py
    if not isinstance(executor, Executor):
        executor_obj = Executor(executor)
    else:
        executor_obj = deepcopy(executor)
```

## Expected behavior

Both, passing an `Executor` and a `Callable` should work.

---

<details>
<summary>Environment Context</summary>

```
Mitiq: A Python toolkit for implementing error mitigation on quantum computers
==============================================================================
Authored by: Mitiq team, 2020 & later (https://github.com/unitaryfund/mitiq)

Mitiq Version:	0.26.0

Core Dependencies
-----------------
Cirq Version:	1.1.0
[...]

PR #1877, opened 2023-06-15
Ensure `execute_with_rem` works with Executor object

natestemen:
## Description

This PR adds support for using an `Executor` object with `execute_with_rem`. In addition we've removed a call to deepcopy which we think is unnecessary. We also ensured that tests cover both cases of using an `Executor` and a user defined python function.

fixes https://github.com/unitaryfund/mitiq/issues/1876

### DIFF
--- a/mitiq/rem/rem.py
+++ b/mitiq/rem/rem.py
@@ -40,6 +40,8 @@ def execute_with_rem(
     """
     if not isinstance(executor, Executor):
         executor_obj = Executor(executor)
+    else:
+        executor_obj = executor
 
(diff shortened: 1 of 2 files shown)


andreamari (approved):
Thanks for the quick fix, LGTM! 

PS: I think I added `deepcopy` for some reason that I now don't remember. Anyway I agree with removing `deepcopy` since, the expected behavior is that the `Executor` should mutate in order to record executed circuits, raw results, etc..   If we see in the future unexpected effects due to mutating executors, we'll try to fix them in a different way.

Issue #1846, opened 2023-05-24
Maxcut-demo tutorial triggers warning or freeze #48


nathanshammah:
## Issue Description
Locally executing `make docs` in a `conda` environment on a freshly installed development version of `mitiq` (`make install`),  prompts debugging warning or freeze at the build of documentation.

```~/github/mitiq/docs/source/examples/maxcut-demo.md: Executing notebook using local CWD [mystnb]
0.00s - Debugger warning: It seems that frozen modules are being used, which may
0.00s - make the debugger miss breakpoints. Please pass -Xfrozen_modules=off
0.00s - to python to disable frozen modules.
0.00s - Note: Debugging will proceed. Set PYDEVD_DISABLE_FILE_VALIDATION=1 to disable this validation.
```

purva-thakre:
I think this might be a duplicate of #1826 

https://github.com/unitaryfund/mitiq/actions/runs/4921156704/jobs/8790708057?pr=1745#step:6:144

![image](https://github.com/unitaryfund/mitiq/assets/66048318/aa8b8c9e-e5cc-4b71-9f94-eb53184d8765)

purva-thakre:
Refer to andrea's comment [here](https://github.com/unitaryfund/mitiq/pull/1879#issuecomment-1607134050) for disabling validation in a user's device. 

Might need to update the docs for how to enable `PYDEVD_DISABLE_FILE_VALIDATION=1` in `make docs` if the user wishes to disable the warnings.

natestemen:
Not able to reproduce at this time using `rm -rf docs/build && make docs`. Maybe we no longer need `PYDEVD_DISABLE_FILE_VALIDATION=1` either?

natestemen:
Ahhh okay this seems to be a python 3.11 performance improvement (using frozen modules). Some more information can be found here: https://stackoverflow.com/questions/75114841/

PR #1879, opened 2023-06-20
Doc build debugger warning

### LINKED ISSUE #1826: Debugger warning about frozen modules in docs build
The following warning occurs at every notebook execution in the docs build:

> 0.00s - Debugger warning: It seems that frozen modules are being used, which may
> 0.00s - make the debugger miss breakpoints. Please pass -Xfrozen_modules=off
> 0.00s - to python to disable frozen modules.
> 0.00s - Note: Debugging will proceed. Set PYDEVD_DISABLE_FILE_VALIDATION=1 to disable this validation.

See https://github.com/unitaryfund/mitiq/actions/runs/4921156704/jobs/8790708057?pr=1745 for an example.

See https://stackoverflow.com/questions/75114841/debugger-warning-from-ipython-frozen-modules for solution options.

purva-thakre:
Fixes #1826

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -61,6 +61,7 @@ jobs:
       - name: Build and test Sphinx docs
         run: |
           export BQSKIT_DOC_CHECK_OVERRIDE=1
+          export PYDEVD_DISABLE_FILE_VALIDATION=1
           make docs
           make doctest
           make linkcheck


purva-thakre:
This one is ready to be merged.

andreamari:
Hi @purva-thakre, your solution LGTM! 
However, I don't think this will fix https://github.com/unitaryfund/mitiq/issues/1846 since the proposed fix only acts during the Github workflow and not in a local environment.

So I would approve this PR after changing the current description from:
Fixes https://github.com/unitaryfund/mitiq/issues/1826
Fixes https://github.com/unitaryfund/mitiq/issues/1846

To just:
Fixes https://github.com/unitaryfund/mitiq/issues/1826

Alternately, we could consider inserting `PYDEVD_DISABLE_FILE_VALIDATION=1` in a different place, e.g. in the definition of the  `make docs` command. But I am not sure about the drawbacks of switching off warnings in the local computer users without their explicit consent.

purva-thakre:
@andreamari Good catch!

Let's keep this issue specific to workflows then. One possible solution for #1846 would be to update the docs with instructions on how to disable the warnings if the user wishes to do so.

andreamari (approved):
LGTM!

PR #1899, opened 2023-07-07
[Classical Shadow] Implementation of classical shadows in mitiq

Min-Li:
## Description

---
This is the prototype of classical shadow implementation in Mitiq, w/o calibration.  In this pull request:
 
    - `mitiq/shadows/__init__.py`  initiation file
    - `mitiq/shadows/classical_postprocessing.py` classical shadow postprocessing functions.
    - `mitiq/shadows/quantum_processing.py` perform z-basis measurements.
    - `mitiq/shadows/shadows.py`  a main function of classical shadow has been written.
    - `mitiq/shadows/shadows_example.ipynb` example notebook for this module.
    - `mitiq/shadows/shadows_utils.py` utility functions, including bounds, fidelity, etc.
    - `mitiq/shadows/test/` including test functions for each file, labeled by file names.
---

### License

- [x] I license this contribution under the terms of the GNU GPL, version 3 and grant Unitary Fund the right to provide additional permissions as described in section 7 of the GNU GPL, version 3.

Before opening the PR, please ensure you have completed the following where appropriate.
- [x] I added unit tests for new code.
- [x] I used [type hints](https://www.python.org/dev/peps/pep-0484/) in function signatures.
- [x] I used [Google-style](https://google.github.io/styleguide/pyguide.html#383-functions-and-methods) docstrings for functions.
- [ ] I [updated the documentation](../blob/master/docs/CONTRIBUTING_DOCS.md) where relevant.
- [ ] Added myself / the copyright holder to the AUTHORS file

PR #1905, opened 2023-07-09
[Executor] Adjust executor.py for computational basis measurements

Min-Li:
## Description

Adjust executor.py for computational basis measurements, from #1899

---

### License

### DIFF
--- a/mitiq/executor/executor.py
+++ b/mitiq/executor/executor.py
@@ -194,15 +194,17 @@ class Executor:
             ]
 
         elif self._executor_return_type in MeasurementResultLike:
-            observable = cast(Observable, observable)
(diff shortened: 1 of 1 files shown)


andreamari (commented):
Hi @Min-Li, is this change to the `Executor` class necessary?
The `Executor.evaluate()` method is designed to get expectation values.
The `Executor.run()` method is designed to get bitstrings.

Can you try to use `Executor.run()` ? (you may need to manually apply measurement gates to circuits if they are not present).

If `Executor.run()` works, I would suggest to use that and close this PR.

Min-Li:
@andreamari  Thanks for your suggestion, it works!

PR #1903, opened 2023-07-09
[Classical Shadows 1] Rotation gate functions

Min-Li:
## Description
Rotation gates and test for #1899 ( fixed #1878)

---

### License

- [x] I license this contribution under the terms of the GNU GPL, version 3 and grant Unitary Fund the right to provide additional permissions as described in section 7 of the GNU GPL, version 3.

Before opening the PR, please ensure you have completed the following where appropriate.
- [x] I added unit tests for new code.
- [x] I used [type hints](https://www.python.org/dev/peps/pep-0484/) in function signatures.
- [x] I used [Google-style](https://google.github.io/styleguide/pyguide.html#383-functions-and-methods) docstrings for functions.
- [ ] I [updated the documentation](../blob/master/docs/CONTRIBUTING_DOCS.md) where relevant.
- [ ] Added myself / the copyright holder to the AUTHORS file

PR #1906, opened 2023-07-09
[Classical Shadows] quantum processing and test

Min-Li:
## Description

z basis measurement and test from #1899

---

### License


mitiq/shadows/quantum_processing.py
```diff
@@ -0,0 +1,183 @@
+from typing import Tuple, Callable, Dict, Any, List, Union
+
+import cirq
+import numpy as np
+from numpy.typing import NDArray
+from qiskit_aer import Aer
+from tqdm.auto import tqdm
```
natestemen, line 7: I haven't had a chance to look at this PR in full yet, but if you want to add a new library as a dependency, you'll have to add it to our `requirements.txt` or `dev_requirements.txt` file so this information will trickle down to users when doing installs.

docs/source/apidoc.md
```diff
@@ -134,6 +134,16 @@ See Ref. :cite:`Czarnik_2021_Quantum` for more details on these methods.
    :members:
 ```
 
+## Classical Shadows
+
+### Rotation Gates
+
```
nathanshammah, line 140:
```suggestion
```

docs/source/apidoc.md
```diff
@@ -134,6 +134,16 @@ See Ref. :cite:`Czarnik_2021_Quantum` for more details on these methods.
    :members:
 ```
 
+## Classical Shadows
+
```
nathanshammah, line 138: Please add a one-line description of the package, like it's done for the other techniques. You could include references too.

nathanshammah (changes requested):
Thanks @Min-Li! I made some comments related to the api-doc documentation. I assume that the comments made by other reviewers in #1903 are addressed here in the code base.  

What about calling this module `quantum_process`? It's shorter.

mitiq/shadows/quantum_processing.py
```diff
@@ -0,0 +1,181 @@
...
+        rotated_circuit = circuit.copy()
+        for i, pauli in enumerate(pauli_string):
+            qubit = qubits[i]
+            # Pauli X measurement is equivalent to H plus a Z measurement
+            if pauli == "X":
+                rotated_circuit.append(cirq.H(qubit))
+            # Pauli X measurement is equivalent to S^-1*H plus a Z measurement
+            elif pauli == "Y":
+                rotated_circuit.append(cirq.S(qubit) ** -1)
+                rotated_circuit.append(cirq.H(qubit))
```
natestemen, line 76: Could we use something similar to the `PAULI_MAP` dictionary you defined in [#1908](https://github.com/unitaryfund/mitiq/pull/1908/files#diff-0fdf554fbe706eabb33bdc94f895a9b0d3be06b830c272a06dd9f5eb4d6d54b6R11) to simplify this a bit?

```suggestion
        for qubit, pauli in zip(qubits, pauli_string):
            rotated_circuit.append(PAULI_MAP[pauli].on(qubit)
```

mitiq/shadows/quantum_processing.py
```diff
@@ -0,0 +1,181 @@
...
+         The list of circuits with rotation and measurement gates appended.
+    """
+    qubits = list(circuit.all_qubits())
+    num_qubits = len(qubits)
+    rotated_circuits = []
+    for pauli_string in pauli_strings:
+        assert len(pauli_string) == num_qubits, (
+            f"Pauli string must be same length as number of qubits, "
+            f"got {len(pauli_string)} and {num_qubits}"
+        )
```
natestemen, line 66: We generally raise some sort of error when doing validation instead of using `assert` statements, so I'd recommend the same for here _if_ you think the validation is need. I don't think it's totally necessary.

mitiq/shadows/quantum_processing.py
```diff
@@ -0,0 +1,181 @@
...
+         outcomes: Tuple of two numpy arrays. The first array contains
+            measurement outcomes (-1, 1) while the second array contains the
+            index for the sampled Pauli's (0,1,2=X,Y,Z). Each row of the arrays
+            corresponds to a distinct snapshot or sample while each column
+            corresponds to a different qubit.
+    """
+
+    # Generate random Pauli unitaries
+    qubits = list(circuit.all_qubits())
+    num_qubits = len(qubits)
```
natestemen, line 122:
```suggestion
    num_qubits = len(circuit.all_qubits())
```

mitiq/shadows/quantum_processing.py
```diff
@@ -0,0 +1,181 @@
...
+    )
+    # Attach measurement gates to the circuit
+    rotated_circuits = tqdm(
+        get_rotated_circuits(
+            circuit,
+            pauli_strings,
+        ),
+        desc="Measurement",
+        leave=False,
+    )
```
natestemen, line 134: Does this take a long time? Wondering if `tqdm` is needed here.

mitiq/shadows/quantum_processing.py
```diff
@@ -0,0 +1,181 @@
...
+            sampling_function(rotated_circuit, **sampling_function_config)
+            for rotated_circuit in rotated_circuits
+        ]
+
+    # Transform the outcomes into a numpy array 0 -> 1, 1 -> -1.
+    shadow_outcomes = []
+    for result in results:
+        bitstring = list(result.get_counts().keys())[0]
+        outcome = [1 - int(i) * 2 for i in bitstring]
+        shadow_outcomes.append(outcome)
```
natestemen, line 175: Can you explain what you want to do here? Are you trying to get the bitstring which was measured the most during the sampling? Are you trying to always get the bitstring from the `00...0` state?

mitiq/shadows/test/test_quantum_processing.py
```diff
@@ -0,0 +1,255 @@
...
+
+    # Check that the function returned a list of the correct length
+    assert isinstance(pauli_strings, list)
+    assert len(pauli_strings) == num_strings
+
+    # Check that each Pauli string is the correct length and contains only
+
+    for s in pauli_strings:
+        assert len(s) == num_qubits
+        assert set(s).issubset({"X", "Y", "Z"})
```
natestemen, line 42: Nice test, and use of `issubset`!

mitiq/shadows/test/test_quantum_processing.py
```diff
@@ -0,0 +1,255 @@
...
+    circuit_1.append(cirq.H(qubits[1]))
+    circuit_1.append(cirq.measure(*qubits))
+    circuit_2 = circuit.copy()
+    circuit_2.append(cirq.S(qubits[0]) ** -1)
+    circuit_2.append(cirq.H(qubits[0]))
+    circuit_2.append(cirq.measure(*qubits))
+    assert rotated_circuits == [
+        circuit_1,
+        circuit_2,
+    ], f"Expected {rotated_circuits}, got {[circuit_1, circuit_2]}"
```
natestemen, line 71: Forgo second argument in `assert` statement.

mitiq/shadows/test/test_quantum_processing.py
```diff
@@ -0,0 +1,255 @@
...
+    for rc in rotated_circuits:
+        assert isinstance(rc, cirq.Circuit)
+
+
+def test_generate_random_pauli_strings_time() -> None:
```
natestemen, line 76:
```suggestion
def test_generate_random_pauli_strings_time():
```

mitiq/shadows/test/test_quantum_processing.py
```diff
@@ -0,0 +1,255 @@
...
+    """
+    Test if the execution time of generate_random_pauli_strings linearly with
+    the number of Pauli strings.
```
natestemen, line 79: Is this what you mean?
```suggestion
    Test if the execution time of generate_random_pauli_strings scales linearly with
    the number of Pauli strings.
```

mitiq/shadows/test/test_quantum_processing.py
```diff
@@ -0,0 +1,255 @@
...
+    """
+    # Define the number of qubits
+    num_qubits: int = 3
+    times: list = []
```
natestemen, line 83: No need for type hints in test files, and no need for comments when code is very simple like this.

```suggestion
    num_qubits = 3
    times = []
```

mitiq/shadows/test/test_quantum_processing.py
```diff
@@ -0,0 +1,255 @@
...
+    num_strings = [100, 200, 300, 400, 500]
+    for n in num_strings:
+        # Measure the execution time for generating random Pauli strings
+        start_time = time.time()
+        generate_random_pauli_strings(num_qubits, n)
+        times.append(time.time() - start_time)
+    for i in range(1, len(times)):
+        assert times[i] / times[i - 1] == pytest.approx(
+            num_strings[i] / num_strings[i - 1], rel=1
+        )
```
natestemen, line 93: Cool test! TIL `pytest.approx`, that seems handy.

I'm not sure we really need to test this, however. Are you concerned that this function could be too slow for actual use? (_same comment for next test_)

mitiq/shadows/test/test_quantum_processing.py
```diff
@@ -0,0 +1,255 @@
...
+
+
+n_total_measurements = 10
+
+
+def simple_test_circuit(qubits):
+    circuit = cirq.Circuit()
+    num_qubits = len(qubits)
+    for i, qubit in enumerate(qubits):
+        circuit.append(cirq.H(qubit))
```
natestemen, line 122: I think this should work as well
```suggestion
    circuit.append(cirq.H.on_each(*qubits))
```

mitiq/shadows/test/test_quantum_processing.py
```diff
@@ -0,0 +1,255 @@
...
+@pytest.fixture
+def sampling_function(request):
+    return request.param
+
+
+@pytest.mark.parametrize(
+    "sampling_function",
+    ["cirq", "qiskit"],
+    indirect=True,
+)
```
natestemen, line 138: Do we need the pytest fixture here? I may not be understanding why it's needed, but if it is, could you give a quick explanation? I haven't had to use these much.

```suggestion
@pytest.mark.parametrize(
    "sampling_function",
    ["cirq", "qiskit"],
)
```

natestemen (changes requested):
Good work Min. Impressive stuff.

mitiq/shadows/quantum_processing.py
Min-Li: have put tqdm and Aer as optional imports.
P.S. qiskit_aer is included in qiskit installation, which is already in dev_requirements.txt
https://github.com/Min-Li/mitiq/blob/990145c9b8c591a53f64b2aff8699a90f6f2cfc1/mitiq/shadows/quantum_processing.py#L7-L14

mitiq/shadows/quantum_processing.py
Min-Li: The `pauli=="Y"` case is not suitable for this style, since it needs two `append` operations in a sequence.

mitiq/shadows/quantum_processing.py
Min-Li: thanks for your suggestion!

mitiq/shadows/quantum_processing.py
natestemen: You could define a custom gate (the product of $S^\dagger$ and $H$ in this case) using cirq's [`MatrixGate`](https://quantumai.google/reference/python/cirq/MatrixGate). Just thinking of ways to remove redundancy since we have very similar code in another PR.

mitiq/shadows/quantum_processing.py
Min-Li: yes, it takes a long time when measurements get larger.  I think it's important to keep track of the time before killing the program. But I have modified it to an optional one.

mitiq/shadows/quantum_processing.py
Min-Li: yes, we need eigenvalue, 1, -1 instead of state vector |0>, |1>,

mitiq/shadows/quantum_processing.py
Min-Li: Matrix gate will get this more complicated. A good thing about this format is that when the Pauli is `Z`, no gate is needed to be added.

mitiq/shadows/test/test_quantum_processing.py
Min-Li: thanks!

mitiq/shadows/test/test_quantum_processing.py
Min-Li: nope, just test this function is valid as expected.  Misty suggested adding this kind of test last time we met.

docs/source/apidoc.md
Min-Li: sorry, I didn't find any descriptions on the apidoc for any packages (i.e., headers of `##`), so I find no examples there. But I do see some one-line descriptions under `###` headers. I'll add necessary descriptions to the header of files.

mitiq/shadows/test/test_quantum_processing.py
natestemen: @Misty-W can you comment on why you think this kind of test is needed? Do we have other tests that test performance like this, or do you think we should have more?

Misty-W: Good catch @natestemen ! I don’t remember suggesting testing execution time- maybe it was a misunderstanding. If I mentioned execution time I meant that in total all shadows tests should add at most ~2-3 minutes to the test time on CI.

mitiq/shadows/quantum_processing.py
```diff
@@ -0,0 +1,199 @@
...
+        if add_measurements:
+            # append measurement gates
+            rotated_circuit.append(cirq.measure(*qubits))
+        # append to list of circuits
+        rotated_circuits.append(rotated_circuit)
+    return rotated_circuits
+
+
+# Stage 1 of Classical Shadows: Measurements
+def get_z_basis_measurement(
```
andreamari, line 102: I think this name does not reflect what the function does. Better names could be:

- `measure_shadows`

- `make_shadow_measurements`

- `rotate_and_measure`

- `quantum_processing`

mitiq/shadows/quantum_processing.py
```diff
@@ -0,0 +1,199 @@
...
+    circuit: cirq.Circuit,
+    n_total_measurements: int,
+    sampling_function: Union[str, Callable[..., MeasurementResult]] = "cirq",
+    sampling_function_config: Dict[str, Any] = {},
+) -> Tuple[NDArray[Any], NDArray[Any]]:
+    r"""Given a circuit, perform z-basis measurements on the circuit and return
+    the outcomes in terms of a string, which represents for z-basis measurement
+    outcomes $$1:=\{1,0\}$$, $$-1:=\{0,1\}$$.
```
andreamari, line 110: This docstring does not correspond to what the function actually does.
The function measures random Pauli observables. For each random Pauli string, the input circuit is rotated in the associated basis and measured by calling the input `executor`.

mitiq/shadows/quantum_processing.py
```diff
@@ -0,0 +1,199 @@
...
+        # append to list of circuits
+        rotated_circuits.append(rotated_circuit)
+    return rotated_circuits
+
+
+# Stage 1 of Classical Shadows: Measurements
+def get_z_basis_measurement(
+    circuit: cirq.Circuit,
+    n_total_measurements: int,
+    sampling_function: Union[str, Callable[..., MeasurementResult]] = "cirq",
```
andreamari, line 105: For consistency with all Mitiq techniques, I suggest to call this function `executor` and to avoid having a default noiseless executor.   A design principle of Mitiq is that users are supposed to write down their own `Executor` objects or `execute` functions before using Mitiq. In all Mitiq techniques, the definition of the `executor` argument should not be internally given by Mitiq, but externally given by users.

```suggestion
    executor: Union[Executor, Callable[[QPROGRAM], MeasurementResult]]
```
or `executor: Union[Executor, Callable[[cirq.Circuit], MeasurementResult]]` if we only consider Cirq circuits.

mitiq/shadows/quantum_processing.py
```diff
@@ -0,0 +1,199 @@
...
+    sampling_function_config: Dict[str, Any] = {},
```
andreamari, line 106: Based on the previous suggestion of having user's defined executors, if possible, it would be better to remove the `sampling_function_config` argument. 
The user can set options of executors (i.e. arguments that are not `circuit`) in this way:

```python
from functools import partial

executor = partial(execute_with_options, option_a=..., option_b= ....)
```

mitiq/shadows/quantum_processing.py
```diff
@@ -0,0 +1,199 @@
...
+
+    Returns:
+         outcomes (array): Tuple of two numpy arrays. The first array
+      contains measurement outcomes (-1, 1) while the second array contains the
+      index for the sampled Pauli's (0,1,2=X,Y,Z). Each row of the arrays
+      corresponds to a distinct snapshot or sample while each column
+      corresponds to a different qubit.
+    """
+
+    # Generate random Pauli unitaries
```
andreamari, line 133:
```suggestion
```

mitiq/shadows/quantum_processing.py
```diff
@@ -0,0 +1,199 @@
...
+    num_qubits = len(circuit.all_qubits())
+    pauli_strings = generate_random_pauli_strings(
+        num_qubits, n_total_measurements
+    )
+    # Attach measurement gates to the circuit
```
andreamari, line 138:
```suggestion
    # Rotate and attach measurement gates to the circuit
```

mitiq/shadows/quantum_processing.py
```diff
@@ -0,0 +1,199 @@
...
+
+    # Transform the outcomes into a numpy array 0 -> 1, 1 -> -1.
+    shadow_outcomes = []
+    for result in results:
+        bitstring = list(result.get_counts().keys())[0]
+        outcome = [1 - int(i) * 2 for i in bitstring]
+        shadow_outcomes.append(outcome)
+
+    # output computational basis outcomes |b>
+    # and the random unitaries in {X,Y,Z}.
```
andreamari, line 196:
```suggestion
```

andreamari (changes requested):

mitiq/shadows/quantum_processing.py
```diff
@@ -0,0 +1,199 @@
...
+         sampling_function_config (Optional[Dict[str, Any]]): Configuration
+            for the sampling function. Defaults to {}. If sampling_function is
+            None, then this argument is ignored.
+
+    Returns:
+         outcomes (array): Tuple of two numpy arrays. The first array
+      contains measurement outcomes (-1, 1) while the second array contains the
+      index for the sampled Pauli's (0,1,2=X,Y,Z). Each row of the arrays
+      corresponds to a distinct snapshot or sample while each column
+      corresponds to a different qubit.
```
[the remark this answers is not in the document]
Min-Li: Thanks, will be done in the new version.

mitiq/shadows/quantum_processing.py
Min-Li: Thanks! I'll call it random_pauli_measurement, how does this sound?

mitiq/shadows/quantum_processing.py
Min-Li: Thanks! I'll call it random_pauli_measurement, how does this sound?

mitiq/shadows/quantum_processing.py
Min-Li: sure, change to this: Given a circuit, perform random Pauli measurements on the circuit and
    return outcomes. The outcomes are represented as a string where a
    z-basis measurement outcome of :math:`|0\rangle` corresponds to 1, and
    :math:`|1\rangle` corresponds to -1.

mitiq/shadows/quantum_processing.py
andreamari: sounds good!

mitiq/shadows/quantum_processing.py
andreamari: good!

mitiq/shadows/quantum_processing.py
Min-Li: have been updated in the new version.

mitiq/shadows/quantum_processing.py
Min-Li: Now no default executor has been defined, and I have defined examples in the test file. Please check how this looks.

Min-Li:
@nathanshammah Thanks for your comments, I addressed most of them, for the optional, I prefer it stays as it is :)
@andreamari Thanks for your comments, please check out the new version.

mitiq/shadows/quantum_processing.py
natestemen: Sorry Min, I need a bit more explanation. I'm asking you to explain the code.

As of python 3.7, dictionaries keep their "insertion ordering". So when getting the `0`th element of `list(result.get_counts().keys())`, are you expecting to get a particular bitstring? A random one? More explanation would be helpful.

mitiq/shadows/test/test_quantum_processing.py
natestemen: I see you used a lot of these. I think they're unnecessary since `pytest` does a good job of explaining failures. Do you have an argument for keeping them?

mitiq/shadows/test/test_quantum_processing.py
natestemen: Generally tests do not need to be typehinted. Please remove.

mitiq/shadows/test/test_quantum_processing.py
natestemen: Looks like this change needs to be made again in the test file.

mitiq/shadows/test/test_quantum_processing.py
natestemen: @Min-Li please make sure to comment if you're going to resolve without making the change (which is okay, but we want to be on the same page!).

mitiq/shadows/quantum_processing.py
natestemen: The `assert` statement is still here. Please remove it, or if you think it's necessary, elevate it to an error.

mitiq/shadows/quantum_processing.py
natestemen: Okay, in that case, the `zip` function is still useful in this code. You can use that to clean up the loop / remove the need for `enumerate`.

mitiq/shadows/quantum_processing.py
```diff
@@ -0,0 +1,148 @@
...
+    return outcomes. The outcomes are represented as a string where a
+    z-basis measurement outcome of :math:`|0\rangle` corresponds to 1, and
+    :math:`|1\rangle` corresponds to -1.
+
+    Args:
+        circuit: Cirq circuit.
+        n_total_measurements: number of snapshots.
+        executor: Executor to use. If None, then the
+            default sampling function for the backend is used. The
+            callable is used as the executor.
```
andreamari, line 105:
```suggestion
        executor: Executor to use. The
            callable is used as the executor.
```
Since we removed default executors.

andreamari (commented):
Thanks for having applied all my suggestions! 
All looks very good from my side! However there are several comments and suggestions (by Nate) that haven't been considered.

Sometimes this can happen on GitHub when one clicks on "Resolve" my mistake and afterwards forget about comments because they become hidden. I would suggest to click on "Show resolved" in all comments to double-check if they are actually resolved.

mitiq/shadows/quantum_processing.py
andreamari: good!

mitiq/shadows/quantum_processing.py
Min-Li: I got some errors, and I find that assert is useful at least to me.  I prefer to keep it if it is not necessary to delate it.

Min-Li:
@andreamari Thanks, I have replied to Nate's comments.

Min-Li:
Accidentally closed the pull request. Now reopen it.

mitiq/shadows/test/test_quantum_processing.py
natestemen: This has not been addressed.

mitiq/shadows/test/test_quantum_processing.py
natestemen: In that case, Min, do you think this test is necessary?

natestemen (commented):
@Min-Li it looks like a few of my previous comments were resolved without a change / comment. I unresolved a few of my previous comments. Please make sure to leave a comment if you are not going to make the requested change before resolving.

mitiq/shadows/quantum_processing.py
Min-Li: Hi Nate, thanks for the question. I understand that the line `bitstring = list(result.get_counts().keys())[0]` may cause some confusion. Let me explain it in detail below.
In clasical shadow, for each sampled random Pauli string (say "YXZ"), we only perform one-shot measurement. Hence, the `results.get_counts()` always has one key (which is the measurement outcome), and its value is always 1. Hence, to obtain the one-shot measurement outcome for this Pauli string, we just need to take the only key in `results.get_counts()` and convert it from string to integers, e.g., `"010" --> "[1, -1, 1"]`, where the bitstring "010" represents the state `|0>|1>|0>` and classical shadow needs their eigen-value representation (i.e., `"0" -> 1` and `"1" -> -1`).

Here is the printed results for `pauli_strings[0]`, `results[0]`, `results[0].get_counts())`:
```
YZX
MeasurementResult: {'nqubits': 3, 'qubit_indices': (0, 1, 2), 'shots': 1, 'counts': {'010': 1}}
{'010': 1}
```

Min-Li:
@natestemen Hi Nate, I resolved some conversations without leaving comments since I already address your comments in some previous commits. I will add comments before resolving conversations later.

nathanshammah (approved):
Thank @Min-Li, LGTM, for the api-doc part.

mitiq/shadows/quantum_processing.py
```diff
@@ -0,0 +1,139 @@
...
+    Given a circuit, perform random Pauli measurements on the circuit and
+    return outcomes. The outcomes are represented as a string where a
+    z-basis measurement outcome of :math:`|0\rangle` corresponds to 1, and
+    :math:`|1\rangle` corresponds to -1.
+
+    Args:
+        circuit: Cirq circuit.
+        n_total_measurements: number of snapshots.
+        executor: Executor to use. The
+            callable is used as the executor.
```
andreamari, line 96:
```suggestion
        executor: A callable which runs a circuit and returns a single bitstring.
            If more bitstrings are returned, only one of them is actually used.
```

Users of this function don't know that the executor must return a single shot.
This comment is to inform users about this fact.

andreamari (approved):
Hi @Min-Li, good form my side. But please let's wait for Nate's approval before merging this PR. 

I also left a suggestion to explain that the executor is expected to return a single shot.

mitiq/shadows/quantum_processing.py
andreamari: Thanks Nate for noticing this aspect and thanks Min for explaining that executors are expected to be called with a single shot. I had completely missed this fact.

I added a [suggestion](https://github.com/unitaryfund/mitiq/pull/1906#discussion_r1265164249) to explain this assumption in the function docstring.

mitiq/shadows/quantum_processing.py
Min-Li: thanks, I'll adopt it

mitiq/shadows/test/test_quantum_processing.py
Min-Li: I have replied last time for this, but not sure why this comment does not show here... I learned this feature from chatGPT, but you would get error if change to
# @pytest.mark.parametrize(
#     "executor",
#     [cirq_executor, qiskit_executor],
#     indirect=True,
# )
@pytest.mark.parametrize("n_qubits", [1, 2, 5],
                         [cirq_executor, qiskit_executor],)
 In test_random_pauli_measurement_output_dimensions: indirect fixture '<function cirq_executor at 0x288f6eef0>' doesn't exist
!!!!!!!!!!!!!!!!!!!!!!!!! Interrupted: 1 error during collection !!!!!!!!!!!!!!!!!!!!!!!!!

mitiq/shadows/test/test_quantum_processing.py
Min-Li: Thanks for both of your comments, I think last time when we mention this, it was a test that check this function is indeed a linear function as anticipated.

mitiq/shadows/quantum_processing.py
Min-Li: Thanks, Andrea, I have adopted this!

mitiq/shadows/test/test_quantum_processing.py
natestemen: Since we don't have any other tests that test performance like this, I'd vote to remove it (especially since this is a pretty simple function), but okay with me if you want to keep it.

mitiq/shadows/test/test_quantum_processing.py
Min-Li:
```
# test different generators
# @pytest.fixture
# def executor(request):
#     return request.param


@pytest.mark.parametrize(
    "executor",
    [cirq_executor, qiskit_executor],
)
```

mitiq/shadows/quantum_processing.py
```diff
@@ -0,0 +1,140 @@
...
+    return outcomes. The outcomes are represented as a string where a
+    z-basis measurement outcome of :math:`|0\rangle` corresponds to 1, and
+    :math:`|1\rangle` corresponds to -1.
+
+    Args:
+        circuit: Cirq circuit.
+        n_total_measurements: number of snapshots.
+        executor: A callable which runs a circuit and returns a single
+            bitstring. If more bitstrings are returned, only one of them
+            is actually used.
```
natestemen, line 97: What do you think about adding a warning here in the API doc?
```suggestion
        executor: A callable which runs a circuit and returns a single
            bitstring.

    Warning:
        If your ``executor`` returns statistics for more than a single shot
        (i.e. returns counts for multiple bitstrings), a **single bitstring**
        will be chosen at random (from your distribution).
```

mitiq/shadows/quantum_processing.py
```diff
@@ -0,0 +1,140 @@
...
+            leave=False,
+        )
+    results = [
+        executor(rotated_circuit) for rotated_circuit in rotated_circuits
+    ]
+
+    # Transform the outcomes into a numpy array 0 -> 1, 1 -> -1.
+    shadow_outcomes = []
+    for result in results:
+        bitstring = list(result.get_counts().keys())[0]
```
natestemen, line 133: Should we throw an error here if `len(result.get_counts().keys()) > 1`?

mitiq/shadows/test/test_quantum_processing.py
```diff
@@ -0,0 +1,230 @@
...
+            circuit,
+            100,  # number of total measurements
+            executor=executor,
+        )
+
+        times.append(time.time() - start_time)
+    for i in range(1, len(times)):
+        log_ratio_times = np.log(times[i] / times[i - 1])
+        log_ratio_qubits = np.log(n_qubits[i] / n_qubits[i - 1])
+        assert log_ratio_times == pytest.approx(log_ratio_qubits, rel=5)
```
natestemen, line 230: Could we remove this test? It's quite slow, and it's not testing the correctness of the code. I ran `pytest mitiq/shadows --durations=15` to get the tests ordered by speed in descending order.
```
==================================================================== slowest 15 durations =============================================
13.96s call     mitiq/shadows/test/test_quantum_processing.py::test_random_pauli_measurement_time_power_growth[qiskit_executor]
3.96s call     mitiq/shadows/test/test_quantum_processing.py::test_random_pauli_measurement_time_growth[qiskit_executor]
1.14s call     mitiq/shadows/test/test_quantum_processing.py::test_random_pauli_measurement_time_power_growth[cirq_executor]
0.40s call     mitiq/shadows/test/test_quantum_processing.py::test_generate_random_pauli_strings_time
0.30s call     mitiq/shadows/test/test_quantum_processing.py::test_random_pauli_measurement_output_types[qiskit_executor-2]
0.30s call     mitiq/shadows/test/test_quantum_processing.py::test_random_pauli_measurement_output_dimensions[qiskit_executor-5]
0.30s call     mitiq/shadows/test/test_quantum_processing.py::test_random_pauli_measurement_output_dimensions[qiskit_executor-2]
0.29s call     mitiq/shadows/test/test_quantum_processing.py::test_random_pauli_measurement_no_errors[qiskit_executor-5]
0.29s call     mitiq/shadows/test/test_quantum_processing.py::test_random_pauli_measurement_output_types[qiskit_executor-5]
0.28s call     mitiq/shadows/test/test_quantum_processing.py::test_random_pauli_measurement_no_errors[qiskit_executor-1]
0.27s call     mitiq/shadows/test/test_quantum_processing.py::test_random_pauli_measurement_no_errors[qiskit_executor-2]
0.20s call     mitiq/shadows/test/test_quantum_processing.py::test_random_pauli_measurement_output_types[qiskit_executor-1]
0.20s call     mitiq/shadows/test/test_quantum_processing.py::test_random_pauli_measurement_output_dimensions[qiskit_executor-1]
0.15s call     mitiq/shadows/test/test_quantum_processing.py::test_random_pauli_measurement_time_growth[cirq_executor]
0.01s call     mitiq/shadows/test/test_quantum_processing.py::test_random_pauli_measurement_output_dimensions[cirq_executor-5]
```

natestemen (commented):
Nice work Min. I think this is pretty much ready to go from my POV.

Please make sure to add `tqdm` to our `dev_requirements.txt` file.

Misty-W:
> Nice work Min. I think this is pretty much ready to go from my POV.
> 
> Please make sure to add `tqdm` to our `dev_requirements.txt` file.

@natestemen, @Min-Li, I hesitate to add a new dependency to Mitiq. My understanding is that `tqdm` is a status bar, please correct me if I'm missing something. Is there a particular reason why a status bar is a "must have" for Classical Shadows?

mitiq/shadows/quantum_processing.py
andreamari: Makes sense to me. It is quite easy to forget about the single-shot assumption. 

Note that, if we throw an error, the previously suggested warning should be changed into something like this:
```
Warning:
    The ``executor`` must return a ``MeasuremenrResult``
    for a single shot (i.e. a single bitstring).
```

nathanshammah:
Wanted just to comment that the errors in CI in build-docs occurred in the `bqskit` tutorial. As they are unrelated to these changes, they can be disregarded.

natestemen (approved):
Thanks for your patience Min. This looks good! 🚢

mitiq/shadows/quantum_processing.py
```diff
@@ -0,0 +1,150 @@
...
+    Args:
+        circuit: Cirq circuit.
+        n_total_measurements: number of snapshots.
+        executor: A callable which runs a circuit and returns a single
+            bitstring.
+
+    Warning:
+        If your ``executor`` returns statistics for more than a single shot
+        (i.e. returns counts for multiple bitstrings), a **single bitstring**
+        will be chosen at random (from your distribution).
```
andreamari, line 101:
```suggestion
    Warning:
        The ``executor`` must return a ``MeasuremenrResult``
        for a single shot (i.e. a single bitstring).
```

andreamari (approved):
After updating the warning in the function docstring consistently with the new `ValueError`, this looks good to me!

Min-Li:
@Misty-W Hey Misty, thanks for your comment
,"My understanding is that `tqdm` is a status bar, please correct me if I'm missing something. " yes, you are right, 
"Is there a particular reason why a status bar is a "must have" for Classical Shadows?" This measurement may contain more than the number of hundred thousand measurements, depending on the number of qubits in the circuit, error rates, failer rate, and observables. This could be quite long progress, and without giving any hints of the progress bar, I am concerned about the user might kill the ongoing progress.  
But I have made the `tqdm` an optional package, even if the user doesn't have it, it would be definitely okay. So we do not need to add it as a dependency.

PR #1907, opened 2023-07-09
[Classical Shadows 1] utils for shadows and unit test

Min-Li:
## Description

utils for shadows and unit test from #1899

---

### License


natestemen:
It looks like some of the tests are failing on python 3.8 due to a [failure in subscripting `np.ndarray`](https://github.com/unitaryfund/mitiq/actions/runs/5513969866/jobs/10052705378#step:5:3887). As in the other PR (https://github.com/unitaryfund/mitiq/pull/1908#discussion_r1260342542) we should try to use `numpy.typing` for type hints.

mitiq/shadows/shadows_utils.py
```diff
@@ -0,0 +1,99 @@
...
+            / 2 ** int(np.log2(cirq.unitary(opt).shape[0])),
+            ord=np.inf,
+        )
+        ** 2
+    )
+    N = 34 * max(shadow_norm(o) for o in observables) / error**2
+    return int(np.ceil(N * K)), int(K)
+
+
+def operator_2_norm(R: NDArray[Any]) -> float:
```
andreamari, line 66: Is this equivalent to the [numpy function](https://numpy.org/doc/stable/reference/generated/numpy.linalg.norm.html#numpy-linalg-norm`
) `numpy.linalg.norm(R, order=2)` ?

If so, we can use that and avoid defining a new function.

andreamari (commented):
Minor comments/suggestions

nathanshammah (approved):
Thanks @Min-Li, looks good to me, besides @andreamari's comments.

mitiq/shadows/shadows_utils.py
```diff
@@ -0,0 +1,99 @@
...
+
+"""Defines utility functions for classical shadows protocol."""
+from typing import Tuple, List, Any
+from numpy.typing import NDArray
+import cirq
+import numpy as np
+from cirq.ops.pauli_string import PauliString
+
+
+def min_n_total_measurements(epsilon: float, num_qubits: int) -> int:
```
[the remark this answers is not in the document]
Min-Li: Ah, this `min_n_total_measurements` refers to the minimum number of classical shadows required for state reconstruction. It only takes one variable epsilon, the error rate of state reconstruction. The `calculated_shadow_bound` function takes two variables, error rate, and accuracy, and it calculates both the minimum number of classical shadows required and the number of groups "K" into which we need to split the shadows for achieving the desired accuracy in operator expectation value estimation.

mitiq/shadows/shadows_utils.py
Min-Li: Thanks for bring it up, it works!

mitiq/shadows/shadows_utils.py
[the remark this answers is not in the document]
andreamari: Thanks for explaining this. Could you also add this explanation in the docstrings of the two functions?
For example you could say:

- _This function returns the minimum number of classical shadows required for state reconstruction._

- _This function returns the minimum number of classical shadows required and the number of groups "K" into which we need to split the shadows for achieving the desired accuracy in operator expectation value estimation._

Optionally, you could rename the two functions accordingly:

- `min_n_total_measurements` -> `sampling_cost_for_tomography`
- `calculate_shadow_bound` -> `sampling_cost_for_estimation`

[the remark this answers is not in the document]
Min-Li: Thanks for your suggestion, Andrea!  I have applied your suggestion to the doc string!  However, I'm not sure whether the name should be changed to `sampling_cost_for_*` since the sampling cost seems not clear for what it was used for, both of them are a lower bound on measurements/snapshots for a particular target... I'll come up with some other name if this name looks confusing to you.

mitiq/shadows/shadows_utils.py
```diff
@@ -0,0 +1,84 @@
...
+    Returns:
+        An integer that gives the number of snapshots required to satisfy the
+        shadow bound.
+    """
+    return int(34 * (4**num_qubits) * epsilon ** (-2))
+
+
+def calculate_shadow_bound(
+    error: float,
+    observables: List[PauliString],  # type: ignore
```
natestemen, line 33:
```suggestion
    observables: List[PauliString[Any]],
```

mitiq/shadows/shadows_utils.py
```diff
@@ -0,0 +1,84 @@
...
+    This function returns the minimum number of classical shadows required and
+    the number of groups "k" into which we need to split the shadows for
+    achieving the desired accuracy and failure rate in operator expectation
+    value estimation.
+
+    Args:
+        error: The error on the estimator.
+        observables: List of cirq.PauliString corresponding to the
+        observables we intend to measure.
+        failure_rate: Rate of failure for the bound to hold.
```
natestemen, line 46:
```suggestion
        error: The error on the estimator.
        observables: List of cirq.PauliString corresponding to the
            observables we intend to measure.
        failure_rate: Rate of failure for the bound to hold.
```

mitiq/shadows/shadows_utils.py
```diff
@@ -0,0 +1,84 @@
...
+
+    Returns:
+        An integer that gives the number of snapshots required to satisfy
+        the shadow bound and the chunk size required attaining the specified
+        failure rate.
```
natestemen, line 51:
```suggestion
        Integers quantifying the number of snapshots required to satisfy
        the shadow bound and the chunk size required to attain the specified
        failure rate.
```

mitiq/shadows/shadows_utils.py
```diff
@@ -0,0 +1,84 @@
...
+
+    shadow_norm = (
+        lambda opt: np.linalg.norm(
+            cirq.unitary(opt)
+            - np.trace(cirq.unitary(opt))
+            / 2 ** int(np.log2(cirq.unitary(opt).shape[0])),
+            ord=np.inf,
+        )
+        ** 2
+    )
```
natestemen, line 64: I think it makes sense to have this be a standalone function in this file. It would also allow you to define a variable `unitary = cirq.unitary(opt)` which, from my testing, speeds up the tests in this PR by 20%.

mitiq/shadows/shadows_utils.py
```diff
@@ -0,0 +1,84 @@
...
+    Args:
+        state_vector: The vector whose norm we want to calculate.
+        rho: The operator whose norm we want to calculate.
+
+    Returns:
+        Scalar corresponding to the fidelity.
+    """
+    return float(
+        np.reshape(state_vector.conj().T @ rho @ state_vector, -1).real
+    )
```
natestemen, line 84:
```suggestion
    return np.reshape(state_vector.conj().T @ rho @ state_vector, -1).real[0]
```

mitiq/shadows/shadows_utils.py
```diff
@@ -0,0 +1,84 @@
...
+        )
+        ** 2
+    )
+    N = 34 * max(shadow_norm(o) for o in observables) / error**2
+    return int(np.ceil(N * K)), int(K)
+
+
+def fidelity(
+    state_vector: NDArray[Any],
+    rho: NDArray[Any],
```
natestemen, line 71:
```suggestion
    state_vector: NDArray[np.complex64],
    rho: NDArray[np.complex64],
```

natestemen (approved):
Nice work. Just some small clean up items from me.

mitiq/shadows/shadows_utils.py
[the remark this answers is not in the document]
andreamari: That would be helpful. Sorry for me complaining often about names, it seems a silly aspect but good names are very  important for making open source code easy to understand and to maintain.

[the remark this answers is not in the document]
Min-Li: I see, thanks for explain Andrea! How about 
get_shadow_bound_tomography
get_shadow_bound_operator

or

n_measurements_tomography_bound
n_measurements_opratorexpectation_bound

Since both of them are in fact bounds of number of measurements for achieving the aimed error rate and failure rate.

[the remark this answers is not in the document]
natestemen: [Naming things is notoriously hard](https://www.martinfowler.com/bliki/TwoHardThings.html), but as Andrea said, extremely important!

---

IMO if an int is being returned, having `bound` at the **end** of the function name makes much more sense than `tomography`/`operator` as I would expect those to return different things.

mitiq/shadows/shadows_utils.py
Min-Li: thanks, I have moved it out!

andreamari (approved):
From my side, LGTM! Thanks Min!

nathanshammah (approved):

mitiq/shadows/shadows_utils.py
```diff
@@ -0,0 +1,157 @@
...
+    """
+    norm = (
+        np.linalg.norm(
+            cirq.unitary(opt)
+            - np.trace(cirq.unitary(opt))
+            / 2 ** int(np.log2(cirq.unitary(opt).shape[0])),
+            ord=np.inf,
+        )
+        ** 2
+    )
```
natestemen, line 49: defining `unitary = cirq.unitary(opt)` will speed this up a bit so it doesn't have to be computed 3 times. If speed is not an issue (even with large `opt`s, feel free to ignore), but it would [DRY](https://en.wikipedia.org/wiki/Don%27t_repeat_yourself) the code up a bit.

mitiq/shadows/test/test_shadows_utils.py
```diff
@@ -0,0 +1,72 @@
+import cirq
+import numpy as np
+import mitiq
+import pytest
+from mitiq.shadows.shadows_utils import (
+    n_measurements_tomography_bound,
+    n_measurements_opts_expectation_bound,
+    fidelity,
+    transform_to_cirq_paulistring,
+)
```
natestemen, line 10: Best to group the `mitiq` imports.

mitiq/shadows/shadows_utils.py
```diff
@@ -0,0 +1,157 @@
...
+                continue
+            qubit_pauli_map[cirq.LineQubit(i)] = pauli_map[pauli_name]
+
+        return coefficient, cirq.PauliString(qubit_pauli_map=qubit_pauli_map)
+
+    else:
+        raise ValueError(
+            "pauli_str must be cirq.PauliString, mitiq.PauliString, "
+            "cirq.PauliSum or string."
+        )
```
natestemen, line 157: Raising a `TypeError` would be more appropriate here.

mitiq/shadows/shadows_utils.py
```diff
@@ -0,0 +1,157 @@
...
+
+    Returns:
+        Scalar corresponding to the fidelity.
+    """
+    return np.reshape(state_vector.conj().T @ rho @ state_vector, -1).real[0]
+
+
+def transform_to_cirq_paulistring(
+    pauli_str: Union[str, mitiq.PauliString, cirq.PauliString[Any]]
+) -> Tuple[float, cirq.PauliString[Any]]:
```
natestemen, line 104: Why is this function needed? Is there a reason to convert to cirq `PauliString` objects instead of the mitiq `PauliString`?

natestemen (commented):
Looking good! A few minor questions/comments.

mitiq/shadows/shadows_utils.py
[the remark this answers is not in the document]
Min-Li: Thanks!  I will try my best to come up with names that work for all of us!

mitiq/shadows/shadows_utils.py
Min-Li: I have rewritten this to mitiq.PauliString, and this line becomes 
```
opt = obs.matrix()
    norm = (
        np.linalg.norm(
            opt - np.trace(opt) / 2 ** int(np.log2(opt.shape[0])),
            ord=np.inf,
        )
        ** 2
    )
```

mitiq/shadows/test/test_shadows_utils.py
Min-Li: sorry, I don't quite understand.  do you mean `from mitiq.shadows.shadows_utils import *`?

mitiq/shadows/shadows_utils.py
Min-Li: have deleted this function.

mitiq/shadows/shadows_utils.py
Min-Li: have deleted this function

mitiq/shadows/test/test_shadows_utils.py
natestemen: Group imports from the same library into adjacent lines. Typically in the order of
1. Standard library imports
2. Third party imports
3. First party imports

So this example would look like this:
```py
import pytest

import cirq
import numpy as np

import mitiq
from mitiq.shadows.shadows_utils import (
    n_measurements_tomography_bound,
    n_measurements_opts_expectation_bound,
    fidelity,
    transform_to_cirq_paulistring,
)
```

mitiq/shadows/test/test_shadows_utils.py
natestemen, line 1: Looks like you added the copyright for this file in [`f1837f6`](https://github.com/unitaryfund/mitiq/pull/1907/commits/f1837f6e8c47d7284916f34037cbcde47df895a6), and then deleted it in [`8bf579c`](https://github.com/unitaryfund/mitiq/pull/1907/commits/8bf579cc02df730c00894cee73c34be294f12803). Please add back

mitiq/shadows/shadows_utils.py
```diff
@@ -0,0 +1,101 @@
...
+    """
+    return int(34 * (4**num_qubits) * epsilon ** (-2))
+
+
+def local_clifford_shadow_norm(obs: mitiq.PauliString) -> float:
+    """
+    Calculate shadow norm of an operator with random unitary sampled from local
+    Clifford group.
+    Args:
+        opt: a self-adjoint operator.
```
natestemen, line 40: Needs updating now that this function takes a mitiq `PauliString`.

natestemen (approved):

mitiq/shadows/shadows_utils.py
natestemen: Completed in [a6c5793](https://github.com/unitaryfund/mitiq/pull/1907/commits/a6c579348fe6c55b0291dce80a3012d0106c64a2).

PR #1908, opened 2023-07-09
[Classical Shadows 1] classical postprocessing

Min-Li:
## Description
Calculate snapshots, state reconstruction,  estimate expectation values functions, and unit test from #1899

---

### License


mitiq/shadows/post_processing.py
```diff
@@ -0,0 +1,141 @@
+from typing import Tuple, List, Union, Any
+
+import numpy as np
+
+PAULI_MAP = {"X": 0, "Y": 1, "Z": 2}
+
+
+def snapshot_state(
+    b_list: List[Union[int, float]], u_list: List[str]
+) -> np.ndarray[Any, Any]:
```
natestemen, line 10: A few comments here:

1. Are there better variable names we could use for `b_list` and `u_list`? Always helpful for readability/maintainability if we know what the variable refers to without looking at examples/the actual code.
2. `int` is a subtype of `float`, so `List[float]` should suffice for `b_list`.
3. I think the `np.ndarray[Any, Any]` means a numpy array of any size with any datatype. Is that right? We usually use the `numpy.typing` library for type-hinting numpy related functions, and in this case I think we should use [`numpy.typing.NDArray`](https://numpy.org/devdocs/reference/typing.html#numpy.typing.NDArray) with maybe something like `npt.NDArray[np.complex64]`. See other examples from our codebase [here](https://github.com/search?q=repo%3Aunitaryfund%2Fmitiq%20npt.NDArray&type=code).

mitiq/shadows/test/test_post_processing.py
```diff
@@ -0,0 +1,168 @@
...
+    expectation_estimation_shadow,
+)
+
+
+def test_snapshot_state():
+    b_list = [1, -1]
+    u_list = ["X", "Y"]
+    expected_result = np.array(
+        # here put the expected result based on the
+        # specific b_list and u_list inputs
```
natestemen, line 16: Unnecessary comment.

mitiq/shadows/test/test_post_processing.py
```diff
@@ -0,0 +1,168 @@
...
+    assert isinstance(
+        result, np.ndarray
+    ), f"Expected a numpy array, got {type(result)}"
+    assert result.shape == (2 ** len(b_list), 2 ** len(b_list),), (
+        f"Expected shape {(2 ** len(b_list), 2 ** len(b_list))}, "
+        f"got {result.shape}"
+    )
+    assert np.allclose(
+        result, expected_result
+    ), f"Expected {expected_result}, got {result}"
```
natestemen, line 34: We can remove the second "argument" to the `assert` call. Pytest gives us this information for free!

mitiq/shadows/post_processing.py
```diff
@@ -0,0 +1,141 @@
...
+    """
+    Implement a single snapshot state reconstruction,
+
+    Args:
+        b_list (array): The list of classical outcomes for the snapshot.
+        u_list (array): Indices for the applied Pauli measurement.
```
natestemen, line 16: Two things:

1. No need to add the type in the argument description. In fact it looks like it's losing the type-hint information when the docs are building the [API-doc](https://mitiq--1908.org.readthedocs.build/en/1908/apidoc.html#mitiq.shadows.post_processing.snapshot_state).
2. The description for `u_list` seems wrong. These aren't indices, but rather which Pauli measurement was performed.

mitiq/shadows/post_processing.py
```diff
@@ -0,0 +1,141 @@
...
+
+    Returns:
+        Numpy array with the reconstructed snapshot.
```
natestemen, line 19: Maybe something like this? WDYT?
```suggestion
    Returns:
        Reconstructed quantum state in density operator form.
```

mitiq/shadows/post_processing.py
```diff
@@ -0,0 +1,141 @@
...
+
+    # computational basis states, e.g.|0>=(1,0)
+    zero_state = np.array([[1, 0], [0, 0]])
+    one_state = np.array([[0, 0], [0, 1]])
+
+    # local unitary that applied to the computational basis states, e.g.
+    # $$Z-$$basis measurement,
+    # which equivalent to a random Pauli measurement, i.e. for each qubit,
+    # we randomly decide to measure the Pauli operators.
+    phase_z = np.array([[1, 0], [0, -1j]], dtype=np.complex128)
```
natestemen, line 31: I don't think the `dtype` should be needed here. Is it?
```suggestion
    phase_z = np.array([[1, 0], [0, -1j]])
```

mitiq/shadows/post_processing.py
```diff
@@ -0,0 +1,141 @@
...
+    hadamard = np.array([[1, 1], [1, -1]]) / np.sqrt(2)
+    identity = np.eye(2)
+
+    # act $$Ad_{U^\dagger}$$ on the computational basis states, where $$U$$ is
+    # Define ensemble which the random unitary sample from.
```
natestemen, line 36: incomplete comment.

mitiq/shadows/post_processing.py
```diff
@@ -0,0 +1,141 @@
...
+    unitaries = [hadamard, hadamard @ phase_z, identity]
+
+    # reconstructing a single snapshot state by applying Eq. (S44)
```
natestemen, line 39: If you're going to add an equation reference, make sure the paper URL is somewhere nearby. Could even go in the docstring.

mitiq/shadows/post_processing.py
```diff
@@ -0,0 +1,141 @@
+from typing import Tuple, List, Union, Any
+
+import numpy as np
+
+PAULI_MAP = {"X": 0, "Y": 1, "Z": 2}
```
natestemen, line 5: Something like this could simplify some code below. WDYT?
```suggestion
PAULI_MAP = {"X": hadamard, "Y": hadamard @ phase_z, "Z": identity}
```

mitiq/shadows/post_processing.py
```diff
@@ -0,0 +1,141 @@
...
+
+    # act $$Ad_{U^\dagger}$$ on the computational basis states, where $$U$$ is
+    # Define ensemble which the random unitary sample from.
+    unitaries = [hadamard, hadamard @ phase_z, identity]
+
+    # reconstructing a single snapshot state by applying Eq. (S44)
+    rho_snapshot = np.array([1], dtype=np.complex128)
+    for i in range(num_qubits):
+        state = zero_state if b_list[i] == 1 else one_state
+        U = unitaries[PAULI_MAP[u_list[i]]]
```
natestemen, line 43: We can get rid of the index variable with [`zip`](https://docs.python.org/3.8/library/functions.html#zip)!
```suggestion
    for outcome, measurement in zip(b_list, u_list):
        state = zero_state if outcome == 1 else one_state
        U = unitaries[PAULI_MAP[measurement]]
```

mitiq/shadows/post_processing.py
```diff
@@ -0,0 +1,141 @@
...
+    )
+    for i in range(num_snapshots):
+        shadow_rho += snapshot_state(b_lists[i], u_lists[i])
+
+    return shadow_rho / num_snapshots
+
+
+def expectation_estimation_shadow(
+    measurement_outcomes: Tuple[np.ndarray[Any, Any], np.ndarray[Any, Any]],
+    observable: Any,
```
natestemen, line 82: Missing meaningful type. There is one in the docstring, but if we are looking to use an observable here, maybe we should try to use Mitiq's [observable](https://mitiq.readthedocs.io/en/stable/guide/observables.html) class.

mitiq/shadows/post_processing.py
```diff
@@ -0,0 +1,141 @@
...
+        `shadow_measure_with_executor`.
+        observable (cirq.PauliString): Single cirq observable consisting of
+        single Pauli operators e.g. cirq.X(0) * cirq.Y(1).
+        k (int): number of splits in the median of means estimator. k * N = R,
+        where R is the total number of measurements,
+         N is the number of snapshots.
+
+    Returns:
+        Scalar corresponding to the estimate of the observable.
+    """
```
natestemen, line 106: Make sure to check how the docs look [[link](https://mitiq--1908.org.readthedocs.build/en/1908/apidoc.html#mitiq.shadows.post_processing.snapshot_state)] when building them. Right now it needs some work to be formatted correctly.

mitiq/shadows/test/test_post_processing.py
```diff
@@ -0,0 +1,168 @@
...
+                [
+                    0.5 + 0.0j,
+                    -0.5 + 0.0j,
+                    0.5 + 0.0j,
+                    0.0 + 1.5j,
+                    0.5 - 0.5j,
+                    0.0 + 0.0j,
+                    0.0 + 0.0j,
+                    0.0 + 0.0j,
+                ],
```
natestemen, line 56: How was this test data computed?

natestemen (changes requested):
Great start Min! Thanks for adding some nice tests, it's helpful both to ensure correctness, and for me to see how things are being used.

I've added a bunch of minor comments, but I think all stuff that should be pretty small changes. No major changes needed I don't think.

mitiq/shadows/post_processing.py
Min-Li: I changed to NDArray. But NDArray[np.complex_] typing somehow breaks the classical shadow results. Changing to NDArray[Any] solves the problem. I can not figure out what happened, you are very welcome if you want to look into this problem, but please also check the changing type would not destroy the result in the notebook in #1899.

mitiq/shadows/test/test_post_processing.py
Min-Li: computed by hand

mitiq/shadows/post_processing.py
Min-Li: I already removed all complex128 in previous commits.

mitiq/shadows/post_processing.py
Min-Li: This is resolved in my previous commit.

mitiq/shadows/post_processing.py
Min-Li: Thanks! I've resolved this issue in previous commits.

mitiq/shadows/post_processing.py
Min-Li: I have deleted this.

mitiq/shadows/post_processing.py
natestemen: Thanks Min for addressing this! In the future linking to the commit where the change was resolved is always super helpful!

mitiq/shadows/post_processing.py
natestemen: Remember to update the `b_list` and `u_list` variable names. Maybe `bits` and `ops` respectively? I'm not totally sure what the best names are, I'll leave that up to you.

Sure enough naming things is [one of the two hardest things in computer science](https://www.martinfowler.com/bliki/TwoHardThings.html).

mitiq/shadows/post_processing.py
natestemen: What is now line 43 still contains a reference to a numbered equation.

mitiq/shadows/post_processing.py
natestemen: Does this work with Mitiq's observable class?

mitiq/shadows/classical_postprocessing.py
```diff
@@ -0,0 +1,136 @@
...
+    Args:
+        b_list: The list of classical outcomes for the snapshot. Here,
+            b = 1 corresponds to :math:`|0\rangle`, and
+            b = -1 corresponds to :math:`|1\rangle`.
+
+        u_list: Array of ("X", "Y", "Z") for the applied
+            Pauli measurement on each qubit.
+
+    Returns:
+        Reconstructed snapshot in terms of nparray.
```
natestemen, line 33:
```suggestion
    Returns:
        Reconstructed snapshot state as density matrix.
```

mitiq/shadows/classical_postprocessing.py
```diff
@@ -0,0 +1,136 @@
...
+    """
+    num_snapshots, num_qubits = measurement_outcomes[0].shape
+
+    # classical values
+    b_lists, u_lists = measurement_outcomes
+
+    # Averaging over snapshot states.
+    shadow_rho = np.zeros((2**num_qubits, 2**num_qubits), dtype=complex)
+    for i in range(num_snapshots):
+        shadow_rho += snapshot_state(b_lists[i], u_lists[i])
```
natestemen, line 76: Not sure if `bits` and `op` are good variable names, but WDYT of this?
```suggestion
    shadow_rho = sum(
        snapshot_state(bits, op) for bits, op in zip(b_lists, u_lists)
    )
```

mitiq/shadows/test/test_classical_postprocessing.py
```diff
@@ -0,0 +1,198 @@
...
+    )
+    k_shadows = 1
+
+    # Act
+    result = expectation_estimation_shadow(
+        measurement_outcomes, observable, k_shadows
+    )
+
+    # Assert
+    # All means should be 0 as there are no matching indices.
```
natestemen, line 197: I don't understand these comments. Are they maybe left over from something else? If not, can you clarify them.

natestemen (changes requested):

mitiq/shadows/classical_postprocessing.py
```diff
@@ -0,0 +1,136 @@
...
+
+    return shadow_rho / num_snapshots
+
+
+def expectation_estimation_shadow(
+    measurement_outcomes: Tuple[NDArray[Any], NDArray[np.str0]],
+    observable: cirq.PauliString,  # type: ignore
```
andreamari, line 83: I have three comments about the `observable` argument:

- In Mitiq we always use the argument `observable` for a `mitiq.Observable` object. Here instead the function expects a `cirq.PauliString`. So, can we rename the the `observable`  argument to `pauli_string` ?

- What about if instead of `cirq.PauliString` we use the `mitiq.PauliString` object. They are very similar objects but `mitiq.PauliString` is looks more familiar to Mitiq users, especially if they typically use non-Cirq frontends like Qiskit.

- Both the `cirq.PauliString`  and `mitiq.PauliString` objects, can have a multiplicative scalar coefficient different from 1 (see e.g. `cirq.PauliString.with_coefficient()` or the mitiq.PauliString.coeff). However this function assumes the coefficient is 1. Perhaps we can just scale the final result by that coefficient.

andreamari (commented):
Thanks @Min-Li !

mitiq/shadows/post_processing.py
Min-Li: Thanks for the suggestions, Nate!  But actually, those names are aligned with the paper.  Users don't need to look at examples/the actual code.  But they need to know what the theories are actually doing, by either reading paper or learning from any talks, all of them using "b" and "u" actually. I could change them to 

`oneshot_computational_basis_measurement_state_list`
`oneshot_random_unitaries_sampled_from_cl2_list`

but I actually don't thank that is necessary. and naming them by simple names such as bits/states or ops/unitaries actually would be more confusing.  But I'll let you decide whether they are useful, or which one you want to choose!

mitiq/shadows/test/test_post_processing.py
Min-Li: removed!

mitiq/shadows/test/test_post_processing.py
Min-Li: removed!

mitiq/shadows/post_processing.py
Min-Li: Thanks! I have correct it, please check the new version!

mitiq/shadows/classical_postprocessing.py
```diff
@@ -0,0 +1,136 @@
...
+
+
+def shadow_state_reconstruction(
+    measurement_outcomes: Tuple[NDArray[Any], NDArray[np.str0]]
+) -> NDArray[Any]:
+    """Reconstruct a state approximation as an average over all snapshots.
+
+    Args:
+        measurement_outcomes: A shadow tuple obtained
+        from `z_basis_measurement`.
```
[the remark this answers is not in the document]
Min-Li: just measurement outcomes/outputs of the functions! tuple of [b_lists,  u_lists]!

mitiq/shadows/classical_postprocessing.py
```diff
@@ -0,0 +1,136 @@
...
+    Args:
+        measurement_outcomes: A shadow tuple obtained from
+            `z_basis_measurement`.
+        observable: Single cirq observable consisting of
+            Pauli operators.
+        k_shadows: number of splits in the median of means estimator.
+
+    Returns:
+        Float corresponding to the estimate of the observable
+        expectation value.
```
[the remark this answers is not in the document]
Min-Li: corrected!

mitiq/shadows/classical_postprocessing.py
Min-Li: actually, they can't be called density matrices, since "density matrices" has to be an element of $B(H)_*^+$ but it is not. You can think the physics reason that it is not a density matrix by it is not obtained by manipulating it in the lab/ by CPTP maps.  It is just a matrix.

mitiq/shadows/classical_postprocessing.py
```diff
@@ -0,0 +1,136 @@
...
+    b_one = np.array([[0.0 + 0.0j, 1.0 + 0.0j]])
+    one_state = b_one.T @ b_one
+
+    # reconstructing a single snapshot state by applying Eq. (S44)
+    rho_snapshot = np.array([1.0 + 0.0j])
+
+    for b, u in zip(b_list, u_list):
+        state = zero_state if b == 1 else one_state
+        U = PAULI_MAP[u]
+        # act $$Ad_{U^\dagger}$$ on the computational basis states
```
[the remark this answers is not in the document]
Min-Li: yes, you are right, it is a definition in group theory.

mitiq/shadows/test/test_classical_postprocessing.py
Min-Li: removed! thanks!

mitiq/shadows/classical_postprocessing.py
```diff
@@ -0,0 +1,136 @@
...
+    """
+
+    # z-basis measurement outcomes
+    b_zero = np.array([[1.0 + 0.0j, 0.0 + 0.0j]])
+    zero_state = b_zero.T @ b_zero
+
+    b_one = np.array([[0.0 + 0.0j, 1.0 + 0.0j]])
+    one_state = b_one.T @ b_one
+
+    # reconstructing a single snapshot state by applying Eq. (S44)
```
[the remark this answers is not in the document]
Min-Li: removed!

mitiq/shadows/classical_postprocessing.py
```diff
@@ -0,0 +1,136 @@
...
+from numpy.typing import NDArray
+
+# local unitary that applied to the qubits
+phase_z = cirq.S._unitary_().conj()
+hadamard = cirq.H._unitary_()
+identity = cirq.I._unitary_()
+PAULI_MAP = {"X": hadamard, "Y": hadamard @ phase_z, "Z": identity}
+
+
+def snapshot_state(b_list: List[float], u_list: List[str]) -> NDArray[Any]:
```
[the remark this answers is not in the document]
Min-Li: Yes, you are right! Thanks!

mitiq/shadows/classical_postprocessing.py
```diff
@@ -0,0 +1,136 @@
...
+        Float corresponding to the estimate of the observable
+        expectation value.
+    """
+    # target observable
+    target_obs, target_locs = [], []
+    for qubit, pauli in observable.items():
+        target_obs.append(str(pauli))
+        target_locs.append(int(qubit))
+
+    # classical values stored in classical computer
```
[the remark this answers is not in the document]
Min-Li: delated!

mitiq/shadows/classical_postprocessing.py
```diff
@@ -0,0 +1,136 @@
...
+    # loop over the splits of the shadow:
+    for i in range(0, n_total_measurements, n_total_measurements // k_shadows):
+        # assign the splits temporarily
+        b_lists_k, u_lists_k = (
+            b_lists[i : i + n_total_measurements // k_shadows],
+            u_lists[i : i + n_total_measurements // k_shadows],
+        )
+        # number of measurements/shadows in each split
+        n_group_measurements = len(b_lists_k)
+        # find the exact matches for the observable of
```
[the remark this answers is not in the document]
Min-Li: Yes, we can indeed write in this way. However, this new inner-loop `for b_list, u_list in zip(b_lists_k, u_lists_k)` may incur some compute overhead since Python's for loop is known to be slow; while my current implementation is faster thanks to the fast `numpy.prod` operation (though slightly difficult to understand).
My implementation follows [PennyLane's tutorial](https://pennylane.ai/qml/demos/tutorial_classical_shadows#estimating-pauli-observables-with-classical-shadows) (correcting some of their incorrect implementations).

mitiq/shadows/classical_postprocessing.py
Min-Li: Thanks! I just changed it to 
```python
    shadow_rho = np.sum(
        [
            classical_snapshot(b_list, u_list)
            for b_list, u_list in zip(b_lists, u_lists)
        ],
        axis=0,
    )
```
in a new commit

mitiq/shadows/classical_postprocessing.py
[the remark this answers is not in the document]
Min-Li: That is equivalent to `np.string_`. I've changed it to `np.string_` in a new commit.

Min-Li: Thanks for the suggestion! 
I've made 'expectation_estimation_shadow' compatible with str, mitiq.PauliString, and cirq.PauliString. and rename the argument "observable" as "pauli_str". 

The function`transform_to_cirq_paulistring` is implemented in #1907 , and note the utils have been merged to this PR

andreamari: Thanks! What about my third point?  What happens if `pauli_str` has a scalar coefficient different from 1, e.g. 
` 0.5 * X(1)` ?

Min-Li: it could work now.

mitiq/shadows/classical_postprocessing.py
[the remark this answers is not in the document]
Misty-W: Then I would remove the word "shadow" here.

mitiq/shadows/shadows_utils.py
```diff
@@ -101,52 +101,57 @@ def fidelity(
...
+        or string.
 
     Returns:
       A tuple containing the coefficient and a cirq.PauliString.
     """
     if isinstance(pauli_str, cirq.PauliString):
         # Extract coefficient and return cirq.PauliString without coefficient
-        coeff = pauli_str.coefficient.real
-        new_pauli_str = cirq.PauliString(
+        coeff = np.real(pauli_str.coefficient)  # type: ignore
```
Misty-W, line 117: Since you're in effect casting to a real value, I'd expect this `# type: ignore` is no longer needed.

mitiq/shadows/classical_postprocessing.py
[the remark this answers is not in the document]
natestemen: Ah okay I referred back to my Lie groups/algebras text and **you** are correct in the definition of $\mathrm{Ad}_X$.

mitiq/shadows/classical_postprocessing.py
natestemen: Ah okay, thanks for the explanation! Let's just say "matrix" then.

---

I take $B(H)_\ast^+$ to mean bounded linear operators on a Hilbert space $H$ that are positive definite ($+$), but what does the $\ast$ mean?

mitiq/shadows/classical_postprocessing.py
natestemen: Hmm, doesn't adding `axis=0` change the effect of this sum, from the original `for` loop?

mitiq/shadows/shadows_utils.py
Min-Li: it will rise error when checking 
```
mitiq/shadows/shadows_utils.py:121: error: Argument 1 to "real" has incompatible type "Union[complex, float, Complex, Expr]"; expected "_SupportsReal[Any]"  [arg-type]
Found 1 error in 1 file (checked 78 source files)
```
The checking seems quite restrictive.  I need to keep this...

mitiq/shadows/classical_postprocessing.py
Min-Li: oh, sorry for the jargon on the notations. You can understand it as kind of a dual space of B(H), 
this notation means predual, or trace class operators, where the definition could be found here: https://en.wikipedia.org/wiki/Trace_class#As_the_predual_of_bounded_operators

mitiq/shadows/classical_postprocessing.py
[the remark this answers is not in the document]
Min-Li: I have added acknowledgments in the new version, thanks!

mitiq/shadows/classical_postprocessing.py
Min-Li: I'll get errors as your suggested code. I think it's because if you delete axis=0 from np.sum(), numpy will sum all the elements in the array, resulting in a single scalar value, rather than an array.

mitiq/shadows/shadows_utils.py
```diff
@@ -0,0 +1,161 @@
...
+    Args:
+      pauli_str: A mitiq.PauliString, cirq.PauliString, cirq.PauliSum
+        or string.
+
+    Returns:
+      A tuple containing the coefficient and a cirq.PauliString.
+    """
+    if isinstance(pauli_str, cirq.PauliString):
+        # Extract coefficient and return cirq.PauliString without coefficient
+        coeff = np.real(pauli_str.coefficient)  # type: ignore
```
andreamari, line 121: Coefficients can be complex and the user is not aware that we'll take the real part.
Is it necessary to take the real part? Can't we keep a complex coefficient instead?

With the current code we get:
`Paulistring(cirq.X, coefficient=1.0j) --(this function)-->Paulistring(cirq.X, coefficient=0.0) --> 0`

mitiq/shadows/shadows_utils.py
```diff
@@ -0,0 +1,161 @@
...
+        return coeff, new_pauli_str
+
+    elif isinstance(pauli_str, mitiq.PauliString):
+        qubit_pauli_map = {
+            cirq.LineQubit(i): gate
+            for i, gate in enumerate(pauli_str._pauli.values())
+        }
+        return np.real(pauli_str.coeff), cirq.PauliString(
+            qubit_pauli_map=qubit_pauli_map
+        )
```
andreamari, line 134:
```suggestion
    elif isinstance(pauli_str, mitiq.PauliString):
        return pauli_str._pauli
```
`mitiq.PauliString._pauli` is a `cirq.PauliString` so maybe we can just return that.

mitiq/shadows/shadows_utils.py
```diff
@@ -0,0 +1,161 @@
...
+    elif isinstance(pauli_str, str):
```
andreamari, line 135: Do you ever use this `str` case in the classical shadows code?
If not we can drop the case of a `str` input.

mitiq/shadows/classical_postprocessing.py
```diff
@@ -0,0 +1,151 @@
...
+
+    # loop over the splits of the shadow:
+    for i in range(0, n_total_measurements, n_total_measurements // k_shadows):
+        # assign the splits temporarily
+        b_lists_k, u_lists_k = (
+            b_lists[i : i + n_total_measurements // k_shadows],
+            u_lists[i : i + n_total_measurements // k_shadows],
+        )
+        # number of measurements/shadows in each split
+        n_group_measurements = len(b_lists_k)
```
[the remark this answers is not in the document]
Min-Li: Thanks! I changed it to
```python
    group_idxes = np.array_split(np.arange(n_total_measurements), k_shadows)
    # loop over the splits of the shadow:
    for idxes in group_idxes:
        b_lists_k = b_lists[idxes]
        u_lists_k = u_lists[idxes]
```

mitiq/shadows/shadows_utils.py
```diff
@@ -0,0 +1,161 @@
...
+) -> Tuple[float, cirq.PauliString[Any]]:
+    """Transforms mitiq.PauliString, cirq.PauliString, cirq.PauliSum or string
+    to a cirq.PauliString class.
+
+    Args:
+      pauli_str: A mitiq.PauliString, cirq.PauliString, cirq.PauliSum
+        or string.
+
+    Returns:
+      A tuple containing the coefficient and a cirq.PauliString.
```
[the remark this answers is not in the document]
Min-Li: removed.

mitiq/shadows/shadows_utils.py
Min-Li: I was taught that the so-called observable should be a self-adjoint operator. But I have changed this to allow complex coeff.

mitiq/shadows/shadows_utils.py
Min-Li: removed the whole function now.

mitiq/shadows/shadows_utils.py
Min-Li: have removed this option

mitiq/shadows/shadows_utils.py
Min-Li: don't worry, I have removed this function now! Thanks anyway, Misty!

mitiq/shadows/classical_postprocessing.py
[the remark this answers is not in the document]
andreamari: Perfect! Thanks

mitiq/shadows/shadows_utils.py
andreamari: Sometimes it can be useful to consider operators with complex coefficients even if they are not  "observables" in the physical sense. This is why a `cirq.PauliString` (or `mitiq.PauliString`) object  can be defined with a complex coefficient. 

Personally, I also think this aspect is a bit confusing but unless we revisit how the classes `cirq.PauliString` or `mitiq.PauliString` are defined, we need to assume the possibility of complex coefficients.

andreamari (approved):
Thanks Min, looks good to me!

mitiq/shadows/classical_postprocessing.py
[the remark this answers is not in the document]
Min-Li: Thank you!

mitiq/shadows/shadows_utils.py
Min-Li: Ahh, I see, thanks a lot for your explanation, Andrea!

mitiq/shadows/classical_postprocessing.py
```diff
@@ -0,0 +1,142 @@
...
+        Reconstructed snapshot in terms of nparray, which is not a physical
+        state.
+    """
+
+    # z-basis measurement outcomes
+    b_zero = np.array([[1.0 + 0.0j, 0.0 + 0.0j]])
+    zero_state = b_zero.T @ b_zero
+
+    b_one = np.array([[0.0 + 0.0j, 1.0 + 0.0j]])
+    one_state = b_one.T @ b_one
```
natestemen, line 47: I think this is a bit more readable, WDYT?
```suggestion
    zero_state = np.array([[1, 0], [0, 0]], dtype=complex)
    one_state = np.array([[0, 0], [0, 1]], dtype=complex)
```

docs/source/apidoc.md
```diff
@@ -136,20 +136,19 @@ See Ref. :cite:`Czarnik_2021_Quantum` for more details on these methods.
...
 ```
 
-## Digital Dynamical Decoupling
-
-### Digital Dynamical Decoupling (High-Level Tools)
+### Quantum Processing
 ```{eval-rst}
-.. automodule:: mitiq.ddd.ddd
+.. automodule:: mitiq.shadows.quantum_processing
    :members:
```
natestemen, line 148: Make sure not to remove the DDD sections of the API-doc here!

mitiq/shadows/classical_postprocessing.py
natestemen: Actually not sure if `dtype=complex` is completely needed since below we are taking the tensor product with complex matrices.

natestemen:
RTD is failing, but failure is unrelated to these changes.

natestemen (approved):

Issue #1667, opened 2023-01-10
Loosen core dependency requirements


natestemen:
# Current status of dependency management

As it stands, we restrict our mandatory requirements to a "compatible release" using `~=`.
https://github.com/unitaryfund/mitiq/blob/a2964908db14df6e8838ee00a20d7414fedc9d13/requirements.txt#L1-L3
For the first line, the `~=` syntax means `>=1.24.1` _and_ `==1.24.*`.

Because we use dependabot to regularly open PRs to update our requirements files, these versions are often the latest possible version, and as a result, we are _exactly_ pinning versions as there are no versions more recent than what we specify.

## Tradeoffs

Strict(er) pinning is good for reproducibility. It ensures that when a user is downloading mitiq, they have the versions of dependencies that we use for testing mitiq during development. This hopefully reduces the number of spurious bug reports, and otherwise impossible to pin down problems that can arise when someone experiences a problem related to using a collection of dependency versions with which we have not tested. On the other hand, strict dependency resolution can make it harder for users to use mitiq if they have a specific version/range of versions they would like to use for a specific dependency. Requiring strict versions of dependencies means we are forcing external developers to also be strict with their dependencies.

Loose(r) pinning mostly benefits users who want to get spun up with mitiq as quickly as possible. Allowing for a range of versions means less time dealing with dependency conflicts (we all know how fun these can be to resolve). The biggest negative here is that it potentially opens the door for more complicated, system-dependent bugs. Theoretically, this should not happen if there is a perfect maintenance of allowed version range, but it can be very hard to know how exactly what are valid bounds for a specific dependency. E.g. if we depend on **A** and **B**, and **B** depends on **A>1.2.2**, we should take that into account in our dependency range. This is something a dependency resolver does automatically, but it would become partially our burden.

## Other projects
- [cirq](https://github.com/quantumlib/Cirq/blob/25ecba92f478e1ee4cd3cd36822f417e7aab4c5a/cirq-core/requirements.txt) - loose specification
- [pennylane](https://github.com/PennyLaneAI/pennylane/blob/359130c783b48fea04bac38c5602df2cc3aa401d/requirements.txt) loose(r) specification
- [qiskit-terra](https://github.com/Qiskit/qiskit-terra/blob/main/requirements.txt) loose specification
- [qiskit-ibmq-provider](https://github.com/Qiskit/qiskit-ibmq-provider/blob/master/requirements.txt) loose specification
[...]

PR #1917, opened 2023-07-15
loosen remaining core dependency versions

natestemen:
## Description

In https://github.com/unitaryfund/mitiq/pull/1672 we loosened the version requirement for `numpy` and haven't had any complaints from users in regards to irreproducible bugs/inconsistencies.

We could do each one of these dependencies at a time, but considering how little we use `scipy`, I think this should be fine to go out as one. Happy to change if others have different opinions.

Fixes https://github.com/unitaryfund/mitiq/issues/1667


andreamari (approved):
LGTM! 
Thanks for catching this old issue. I almost forgot about it.

requirements.txt
```diff
@@ -1,3 +1,3 @@
 numpy>=1.22.0
-scipy~=1.10.1
-cirq~=1.1.0
+scipy>=1.5.0,<=1.10.0
```
andreamari, line 2: What is the reason for the upper bound `<=1.10.0` for `scipy`? And why is it smaller than `1.10.1` (current setting)?

andreamari (approved):
Thanks for catching this old issue!
See minor comment about the scipy version upper bound.
After clarifying this small aspect, LGTM!

requirements.txt
natestemen: Good point. I think I was just testing stuff and used 1.10 for simplicity. I've bumped to the lastest working version for us in [`46b0237`](https://github.com/unitaryfund/mitiq/pull/1917/commits/46b02375640c4ed47b4a791525a49eed52c2532e).

PR #1919, opened 2023-07-17
Fix docstring of initialized_depolarized_noise

### LINKED ISSUE #1918: 'noise_level' of 'initialized_depolarizing_noise' correct?
For this noise model:

from mitiq.interface.mitiq_qiskit.qiskit_utils import initialized_depolarizing_noise
noise_model = initialized_depolarizing_noise(noise_level=0.001)

The annotation of this function says:

Args:
        noise_level: The noise strength as a float, e.g., 0.01 is 0.1%.

Why 0.1% instead of 1%? Commonly, 0.01 is used to present 1%.

andreamari:
Fixes #1918

### License


natestemen (approved):
Thanks for the quick fix Andrea!

PR #1921, opened 2023-07-18
[Classical Shadows 4] Main function

### LINKED ISSUE #1952: Robust shadow estimation
## Issue Description

Add functions and tests implementing the Robust Shadow Estimation technique in Mitiq.

## Additional References

See [RFC for Robust shadow estimation](https://docs.google.com/document/d/1B5FnqQDvoRYap5fGPqzcbp-RXIrUFjbBcLiWIUrLmuA/edit#heading=h.4wr9qx2m833b) for details.

Min-Li:
Fixes #1952 

## Description

- `mitiq/shadows/shadows.py`  a main function of classical shadow has been written.
- `mitiq/shadows/_init_.py` initial functions for shadows.
---

### License


mitiq/shadows/shadows.py
```diff
@@ -0,0 +1,130 @@
...
+    shadow_state_reconstruction,
+    expectation_estimation_shadow,
+)
+from mitiq.shadows.shadows_utils import (
+    n_measurements_tomography_bound,
+    n_measurements_opts_expectation_bound,
+)
+
+
+def execute_with_shadows(
```
andreamari, line 26: All `execute_with_xxx` functions in Mitiq have a similar signature, they take as input a `circuit`, an `executor`, an optional single `mitiq.Observable` (a linear combination of `PauliString`s) and they produce as output a single expectation value. The motivation for such a fixed structure is to easily switch between different error mitigation techniques, or to easily combine multiple techniques.

This function instead takes as input a list of many `mitiq.PauliString` objects, corresponding to many Pauli "observables", and returns a dictionary with all the results.

I agree that, due to the structure of the classical shadow algorithm, having many Pauli strings and therefore many expectation values is essential. However I wonder if we should use a different name for this function. E.g. something like `run_shadow_protocol()`.

[**Optional**]
If you think it can be useful, we could also add a new function `execute_with_shadows()`, which takes as input a single `mitiq.Observable` (which is a linear combination of many pauli strings) and evaluates its single expectation value by internally calling the general `run_shadow_protocol()` function and by taking a linear combination of all the individual expectation values associated to each pauli string.
I don't know if having this new function is important at this stage. It could become more useful when we'll have robust shadows (with error mitigation).

mitiq/shadows/shadows.py
```diff
@@ -0,0 +1,130 @@
...
+        A dictionary containing the shadow outcomes, the Pauli strings, and
+        either the estimated density matrix or the estimated expectation
+        values of the observables.
+    """
+
+    qubits = list(circuit.all_qubits())
+    num_qubits = len(qubits)
+
+    if observables is None:
+        assert (
```
andreamari, line 71: Instead of using `assert`, it is better to use `ValueError`, `TypeError`, or `RuntimeError`, etc... 

If I am not wrong, assert statement should be used for statements that are _always_ `True`, unless there is a bug in the code. This is why they are typically used in tests.

For errors that are _possible_ even with a correct code without bugs (due to mistakes of the user) rising Python errors is the best practice.

andreamari (commented):
Thanks Min!

Misty-W:
hi @Min-Li, are any of the changes on this PR still needed?

mitiq/shadows/shadows.py
```diff
@@ -0,0 +1,198 @@
+# Copyright (C) Unitary Fund
+#
+# This source code is licensed under the GPL license (v3) found in the
+# LICENSE file in the root directory of this source tree.
+"""Classical shadow estimation for quantum circuits. Based on the paper"""
```
natestemen, line 5: Add reference to paper, or remove second sentence.

docs/source/apidoc.md
```diff
@@ -149,7 +147,13 @@ See Ref. :cite:`Czarnik_2021_Quantum` for more details on these methods.
...
 ```
 
-### Utility Functions
+### Classical Post-Processing
+```{eval-rst}
+.. automodule:: mitiq.shadows.classical_postprocessing
+   :members:
+```
+
+### Classical Shadow Utility Functions
```
natestemen, line 156:
```suggestion
### Utility Functions
```
Since this is under the "Classical Shadows" level-2 header, I don't think we need to repeat that here.

docs/source/apidoc.md
```diff
@@ -133,13 +133,11 @@ See Ref. :cite:`Czarnik_2021_Quantum` for more details on these methods.
 .. automodule:: mitiq.interface.mitiq_qiskit.qiskit_utils
    :members:
 ```
-
-
```
natestemen, line 137: Please add these back to maintain spacing in the `apidoc.md` file.

docs/source/apidoc.md
```diff
@@ -133,13 +133,11 @@ See Ref. :cite:`Czarnik_2021_Quantum` for more details on these methods.
...
 ## Classical Shadows
 
-### Classical Post-Processing
+### Classical Shadows (High-Level Tools)
```
natestemen, line 138: What if we remove this header and just have these functions directly under the "Classical Shadows" header? WDYT?

mitiq/shadows/shadows.py
```diff
@@ -0,0 +1,198 @@
...
+    k_calibration: int = 1,
+    locality: Optional[int] = None,
+    zero_state_shadow_outcomes: Optional[Tuple[List[str], List[str]]] = None,
+    qubits: Optional[List[cirq.Qid]] = None,
+    executor: Optional[Callable[[cirq.Circuit], MeasurementResult]] = None,
+    num_total_measurements_calibration: Optional[int] = 20000,
+) -> Dict[str, complex]:
+    r"""
+    This function returns the dictionary of the median of means estimation
+    of Pauli fidelities: {:math:`\{'b':f_{b}\}_{b\in\{0,1\}^n}`}.
```
natestemen, line 42: Can you clarify in the docstring if this dictionary will contain an entry for every $b\in\{0, 1\}^n$, or just a subset.

mitiq/shadows/shadows.py
```diff
@@ -0,0 +1,198 @@
...
+
+    Args:
+        k_calibration: Number of groups of "median of means" used for
+            calibration.
```
natestemen, line 46: What does does the "median of means" mean here? I don't see any collection of means we could be taking a median of, so it's a little unclear.

mitiq/shadows/shadows.py
```diff
@@ -0,0 +1,198 @@
...
+        locality: The locality of the operator, whose expectation value is
+            going to be estimated by the classical shadow. e.g. if operator is
+            Ising model Hamiltonian with nearist neighbour interacting, then
+            locality = 2.
+        zero_state_shadow_outcomes: The output of function
+            `shadow_quantum_processing` of zero calibrate state.
```
natestemen, line 52: If you want to reference a function, and have a link show up in the API-doc you can use the syntax `` :func:`function_name` `` to do so. See an example [here](https://github.com/unitaryfund/mitiq/blob/891684b29519afc37801c1a563913345e13365a8/mitiq/pec/representations/depolarizing.py#L180).

Could be useful here to link to `shadow_quantum_processing`.

mitiq/shadows/shadows.py
```diff
@@ -0,0 +1,198 @@
...
+        qubits: The qubits to measure, needs to spercify when the
+            `zero_state_shadow_outcomes` is None.
+        executor: The function to use to do quantum measurement, must be same
+            as executor in `shadow_quantum_processing`. Needs to spercify when
+            the `zero_state_shadow_outcomes` is None.
+        num_total_measurements_calibration: Number of shots per group of
+            "median of means" used for calibration. Needs to spercify when
+            the `zero_state_shadow_outcomes` is None.
```
natestemen, line 60: `spercify -> specify`

mitiq/shadows/shadows.py
```diff
@@ -0,0 +1,198 @@
...
+def shadow_quantum_processing(
+    circuit: cirq.Circuit,
+    executor: Callable[[cirq.Circuit], MeasurementResult],
+    num_total_measurements_shadow: int,
+    random_seed: Optional[int] = None,
+    qubits: Optional[List[cirq.Qid]] = None,
+) -> Tuple[List[str], List[str]]:
+    r"""
+    Executes a circuit with classical shadows. This function can be used for
+    state reconstruction or expectation value estimation of observables.
```
natestemen, line 101: This description for `shadow_quantum_processing` is the same as for `classical_post_processing`. I don't think that's intended, and I don't think this function is actually executing the classical shadows technique.

mitiq/shadows/shadows.py
```diff
@@ -0,0 +1,198 @@
...
+    """
+    output: Dict[str, Union[float, NDArray[Any]]] = {}
+    if state_reconstruction:
+        reconstructed_state = shadow_state_reconstruction(
+            shadow_outcomes, rshadows, f_est=calibration_results
+        )
+        output["reconstructed_state"] = reconstructed_state  # type: ignore
+    elif (
+        observables is not None
+    ):  # Estimation expectation value of observables
```
natestemen, line 185: I think the comment can go on a newline to avoid breaking this statement over multiple lines (if you think it's necessary). I'm not sure it's helping me understand the code here, so if there's something you're trying to hint at, would you mind explaining a little more?

mitiq/shadows/test/test_shadows.py
```diff
@@ -0,0 +1,117 @@
...
+    sample_bitstrings as cirq_sample_bitstrings,
+)
+from mitiq.shadows.shadows import (
+    pauli_twirling_calibrate,
+    shadow_quantum_processing,
+    classical_post_processing,
+)
+
+# define a fully entangled state
+# number of qubits in the circuit
```
natestemen, line 22:
```suggestion
```

mitiq/shadows/test/test_shadows.py
```diff
@@ -0,0 +1,117 @@
...
+num_qubits: int = 2
+# qubits in the circuit prepared in the $|0\rangle$ state
```
natestemen, line 24:
```suggestion
```

mitiq/shadows/test/test_shadows.py
```diff
@@ -0,0 +1,117 @@
...
+qubits = cirq.LineQubit.range(num_qubits)
+circuit = cirq.Circuit([cirq.H(q) for q in qubits])
+circuit.append(cirq.CNOT(qubits[0], qubits[1]))
+# Define list of observables
```
natestemen, line 28:
```suggestion
```

natestemen:
Please update the description of this PR so that it reflects it's current state. A good description will have a summary of what the PR contains, and explains how it fits into the bigger picture of your other code.

mitiq/shadows/shadows.py
Min-Li: thanks, have deleted it.

mitiq/shadows/shadows.py
Min-Li: The number of $f_b$ is related to the number of irreps in the PTM rep of the twirling group, and related to the locality of the interested operator, I added a reference for clarification.

mitiq/shadows/shadows.py
Min-Li: I don't think I could make it clear locally.  I have added references, which I believe users should know what they are calculating. The more clear reference has been given in the "What's the theory behind" part, as well as in the rshadows_tutorial.md notebook.

mitiq/shadows/shadows.py
Min-Li: thanks! will add it!

mitiq/shadows/shadows.py
Min-Li: thanks for checking spelling!

mitiq/shadows/shadows.py
Min-Li: I'll remove it. Thanks

mitiq/shadows/shadows.py
Min-Li: Thanks for checking this. I have updated the doc string.

mitiq/shadows/shadows.py
```diff
@@ -0,0 +1,199 @@
...
+    In the notation of arXiv:2011.09636, this function estimates the
+    coefficient :math:`f_b`, which are expansion coefficients of the twirled
+    channel :math:`\mathcal{M}=\sum_b f_b\Pi_b`.
+
+    Args:
+        k_calibration: Number of groups of "median of means" used to solve for
+            Pauli fidelity.
+        locality: The locality of the operator, whose expectation value is
+            going to be estimated by the classical shadow. e.g. if operator is
+            Ising model Hamiltonian with nearist neighbour interaction, then
```
andreamari, line 46:
```suggestion
            Ising model Hamiltonian with nearest neighbour interaction, then
```

mitiq/shadows/shadows.py
```diff
@@ -0,0 +1,199 @@
...
+        zero_circuit = cirq.Circuit()
+        # perform random Pauli measurement one the calibration circuit
+        calibration_measurement_outcomes = random_pauli_measurement(
+            zero_circuit,
+            n_total_measurements=num_total_measurements_calibration,
+            executor=executor,
+            qubits=qubits,
+        )
+    else:
+        calibration_measurement_outcomes = zero_state_shadow_outcomes
```
andreamari, line 82: according to codecove, this case (`zero_state_shadow_outcomes` note None) is not tested. Could you test it somewhere in the test file?

mitiq/shadows/shadows.py
```diff
@@ -0,0 +1,199 @@
...
+        executor=executor,
+        qubits=qubits,
+    )
+    return output
+
+
+def classical_post_processing(
+    shadow_outcomes: Tuple[List[str], List[str]],
+    rshadows: bool = False,
+    calibration_results: Optional[Dict[str, float]] = None,
```
andreamari, line 143:
```
rshadows: bool = False,
calibration_results: Optional[Dict[str, float]] = None,
```
Do we need the the `rshadows` argument?  Can we deduce if we want robust or non-robust shadow estimation from `calibration_results` beeing `not None` or `None`.

If you agree, we can simply remove the `rshadows` argument. 
Otherwise, if `rshadows` cannot be removed, I would rename `rshadows`-> `robust` or `use_calibration`.

mitiq/shadows/shadows.py
```diff
@@ -0,0 +1,199 @@
...
+    r"""
+    This function returns the dictionary of the median of means estimation
+    of Pauli fidelities: :math:`\{`"b": :math:`f_{b}\}_{b\in\{0,1\}^n}`.
+    The number of :math:`f_b` is :math:`2^n`, or :math:`\sum_{i=1}^d C_n^i` if
+    the locality :math:`d` is given.
+
+    In the notation of arXiv:2011.09636, this function estimates the
+    coefficient :math:`f_b`, which are expansion coefficients of the twirled
+    channel :math:`\mathcal{M}=\sum_b f_b\Pi_b`.
+
```
andreamari, line 40:
```suggestion
    In the notation of arXiv:2011.09636, this function estimates the
    coefficient :math:`f_b`, which are expansion coefficients of the twirled
    channel :math:`\mathcal{M}=\sum_b f_b\Pi_b`.

    In practice, the output of this function can be used as calibration data
    for performing the classical shadows protocol in a way which is more
    robust to noise.
```

mitiq/shadows/shadows.py
```diff
@@ -0,0 +1,199 @@
...
+    r"""
+    This function returns the bit strings and Pauli strings corresponding to
+    the executor measurement outcomes for a given circuit, rotated by unitaries
+    randomly sampled from a fixed unitary ensemble :math:`\mathcal{G}`.
+
+    In the current implementation, the unitaries are sampled from the local
+    Clifford group for :math:`n` qubits, i.e.,
+    :math:`\mathcal{G} = \mathrm{Cl}_2^n`.
+
+    Args:
```
andreamari, line 105:
```suggestion
    :math:`\mathcal{G} = \mathrm{Cl}_2^n`.

    In practice, the output of this functions provides the raw experimental
    data necessary to perform the classical shadows protocol.

    Args:
```

andreamari (commented):
Thanks @Min-Li ! 
Looks good, here are my comments.

mitiq/shadows/shadows.py
Min-Li: sure, will be add in the next version!

mitiq/shadows/shadows.py
Min-Li: I think it's just a reminder to users whether they'll use the pauli_twilling_calibration. This bool argument has been used as an `if` in the post-processing state.  I prefer to keep this argument. I'll change it to use_calibration.

mitiq/shadows/shadows.py
```diff
@@ -0,0 +1,206 @@
...
+    """
+    if zero_state_shadow_outcomes is None:
+        if qubits is None:
+            raise ValueError("qubits should not be None.")
+        if executor is None:
+            raise ValueError("executor should not be None.")
+        if num_total_measurements_calibration is None:
+            raise ValueError(
+                "num_total_measurements_calibration should not be None."
+            )
```
natestemen, line 74: Wondering if these should be `TypeError`s?

mitiq/shadows/shadows.py
Misty-W:
```suggestion
            Ising model Hamiltonian with nearest neighbor interaction, then
```

mitiq/shadows/shadows.py
```diff
@@ -0,0 +1,206 @@
...
+        1st qubit, local Y-basis measurement the 2ed qubit in the circuit, etc.
+    """
+    if random_seed is not None:
+        np.random.seed(random_seed)
+    r"""
+    Shadow stage 1: Sample random unitary form
+    :math:`\mathcal{g}\subset \mathrm{U}(2^n)` and perform computational
+    basis measurement. In the current state, we have implemented
+    local Pauli measurement, i.e. :math:`\mathcal{g} = \mathrm{Cl}_2^n`.
+    """
```
natestemen, line 136: This comment won't be rendered anywhere in the API-doc/elsewhere. Do you want to include it in the docstring, or is it just for us to provide context?

If you do want to put it in the docstring, you can use something like
```
Additional Information:
    Shadow stage 1: ...
```

mitiq/shadows/shadows.py
Misty-W, line 112:
```suggestion
    :math:`\mathcal{G} = \mathrm{Cl}_2^n`.
    In practice, the output of this functions provides the raw experimental
    data necessary to perform the classical shadows protocol.
    
    Args:
```

Misty-W: Re-iterating Andrea's comment which was marked "outdated" after later changes.

natestemen (approved):
LGTM.

I'd just ask that you remove some of the comments (mostly in the test file) that don't necessarily add any explanation. E.g. I think the following is unnecessary.

```python
# Check that the result is a dictionary
assert isinstance(result, dict)
```

The python code is already really readable, and the comment just duplicates that statement! Happy for you to decide what is and is not necessary.

mitiq/shadows/shadows.py
Min-Li: Thanks! I have checked this through!

mitiq/shadows/shadows.py
```diff
@@ -0,0 +1,208 @@
...
+        num_total_measurements_calibration: Number of shots per group of
+            "median of means" used for calibration. Needs to specify when
+            the `zero_state_shadow_outcomes` is None.
+
+    Returns:
+        A dictionary containing the calibration outcomes.
+    """
+    if zero_state_shadow_outcomes is None:
+        if qubits is None:
+            raise TypeError("qubits should not be None.")
```
Misty-W, line 68: I would make the error message more like, "qubits must be specified when zero_state_shadow_outcomes is None."

Misty-W:
Almost there, @Min-Li! 😅 

Codecov is flagging missing tests for the lines where an error is raised for missing input.

mitiq/shadows/shadows.py
Min-Li: thanks, changed!

mitiq/shadows/shadows.py
Min-Li: btw, I have removed 
```
# Check that the result is a dictionary
assert isinstance(result, dict)

```
as you suggested!

mitiq/shadows/shadows.py
Min-Li: added, thanks.

mitiq/shadows/shadows.py
```diff
@@ -0,0 +1,208 @@
...
+        if executor is None:
+            raise TypeError("executor should not be None.")
```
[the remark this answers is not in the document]
Min-Li: Thanks, Will be changed in the next version!

mitiq/shadows/shadows.py
```diff
@@ -0,0 +1,208 @@
...
+        if num_total_measurements_calibration is None:
+            raise TypeError(
+                "num_total_measurements_calibration should not be None."
```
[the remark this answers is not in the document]
Min-Li: Thanks, Will be changed in the next version!

Misty-W (approved):
LGTM, thanks @Min-Li!

mitiq/shadows/shadows.py
natestemen: BTW in https://github.com/unitaryfund/mitiq/pull/1921#pullrequestreview-1577943726 I did not recommend you remove the `assert` statement, only the comments that duplicate the code.

andreamari (approved):

Issue #1920, opened 2023-07-18
Bqskit example in docs is failing


andreamari:
The docs build is failing due to a `bqskit` import problem.

[link to build error](https://github.com/unitaryfund/mitiq/actions/runs/5586535180/jobs/10210681228#step:6:265)

[link to example with error in docs](https://mitiq.readthedocs.io/en/latest/examples/bqskit.html)

PR #1927, opened 2023-07-20
Temporary fix for documentation problem

andreamari:
This is a temporary workaround for #1920.

Not a definite solution but at least we can clean our CI until we figure out a better solution.


natestemen (approved):
Good thinking Andrea, and thanks for fixing the broken link as well.

PR #1947, opened 2023-08-01
Use `pytest` before `make test` in dev docs

natestemen:
## Description

As part of the mitiq [retrospective](https://miro.com/app/board/uXjVMzS2q8k=/) we discussed how running the tests can be very slow, and perhaps pushes people away from running tests before committing code (leading to CI failing). This PR adds a suggestion to use `pytest` directly on the command line first to test your changes in a more localized way before moving to `make test`.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -12,7 +12,7 @@ The rest of this document describes the technical details of getting set up to d
 
 ## Development environment
 
-1. Ensure you have python 3.8 or greater installed. If not, you can find the downloads [here](https://www.python.org/downloads/). 
+1. Ensure you have python 3.8 or greater installed. If not, you can find the downloads [here](https://www.python.org/downloads/).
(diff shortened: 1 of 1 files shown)


CONTRIBUTING.md
```diff
@@ -31,21 +31,23 @@ What follows are recommendations/requirements to keep in mind while contributing
...
+Use `pytest mitiq/$MODULE` to run the tests for the module you are working on.
+Once they pass, you can run the entire test suite (excluding those that require the pyQuil QVM) by running the following command.
 
-For example, to run all the tests that do not require a pyQuil QVM, run
 ```bash
 make test
 ```
-This is typically suitable for most development tasks and is the easiest, and most common way to test.
+
+This can often be slow, however, so testing your changes iteravely using `pytest` is often faster when doing development.
```
purva-thakre, line 50:
```suggestion
This can often be slow, however, so testing your changes iteratively using `pytest` is often faster when doing development. Inside the Mitiq repository, navigate to the `tests` folder of interest (`cd mitiq/mitiq/some_module/tests`).  

If there are multiple test files in this folder & you want to run all the tests, `(path to some tests folder) pytest` will do the job. But if you want to run the tests in only one specific file `(path to some tests folder) pytest test_some_feature.py` will work. 
```

natestemen: I think this is explained on lines 42/43. Do you think we need this as well?

Misty-W (approved):
LGTM!

PR #1902, opened 2023-07-09
Implements high-level functions for QSE

bubakazouba:
## Description

Closes #1812

Adds execute_with_qse, qse_decorator and mitigate_executor to Quantum Subspace Expansion.

---

### License

- [X] I license this contribution under the terms of the GNU GPL, version 3 and grant Unitary Fund the right to provide additional permissions as described in section 7 of the GNU GPL, version 3.

Before opening the PR, please ensure you have completed the following where appropriate.
- [X] I added unit tests for new code.
- [X] I used [type hints](https://www.python.org/dev/peps/pep-0484/) in function signatures.
- [X] I used [Google-style](https://google.github.io/styleguide/pyguide.html#383-functions-and-methods) docstrings for functions.
- [X] I [updated the documentation](../blob/master/docs/CONTRIBUTING_DOCS.md) where relevant.
- [X] Added myself / the copyright holder to the AUTHORS file

PR #1959, opened 2023-08-09
Add `isort` to organize dependency imports

natestemen:
## Description

This PR adds [`isort`](https://pycqa.github.io/isort/index.html) as a dependency. `isort` is short for "import sort" and sorts dependencies (both alphabetically in each import statement, and groups them according to standard library -> 3rd party -> 1st party imports) in each python file. In addition this PR adds `isort` to run both as part of CI via the makefile and adds config for it to run properly in our `pyproject.toml` file.

### Motivation

I've made some version of [this](https://github.com/unitaryfund/mitiq/pull/1902#discussion_r1260587256) comment more than twice now, so easiest just to automate away the conversation!

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -9,6 +9,7 @@ check-all: check-format check-style check-types
 
 .PHONY: check-format
 check-format:
+	isort --check mitiq
 	black --check --diff mitiq
 
 .PHONY: check-style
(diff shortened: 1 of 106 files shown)


andreamari (approved):
Cool!

Misty-W (approved):

PR #1963, opened 2023-08-15
Pauli Transfer Matrix for Pauli Twirling

### LINKED ISSUE #1914: PT+ ZNE Tutorial and Pauli Transfer Matrix (PTM) in Cirq
Add tutorial combining ZNE with PT and show this via PTM. 

To show that pauli twirling has worked and the noise has been tailored, PTM superoperator could be used. 

If twirling is done perfectly, the off-diagonal terms in this superoperator should be suppressed. An example of this is in Sec VII of the supplemental material of https://arxiv.org/abs/2010.00215

before:
image

after:

image

This superoperator is defined in Qiskit : https://qiskit.org/documentation/stubs/qiskit.quantum_info.PTM.html

Use PTM in cirq to show `execute_with_pauli_twirling` has worked as expected.

purva-thakre:
Fixes #1914

- [x] Tests for the private functions converting from/to a square matrix to/from column vector: square matrix is a column vector, a column vector is a square matrix. 
- [x] Tests for the private function `_pauli_vectorized_list`: the size of the output list should be a power of 4^num_qubits, size of each matrix in the output array should be a power of 2^num_qubits
- [x] Main function for the PTM matrix
- [x] Test for the main function: TBD
- [ ] Use Pauli twirling on a circuit of CNOT/CZ & show the off-diagonal terms in the PTM matrix have disappeared after pauli twirling. 
- [ ] Formatting, type annotation, update docstrings etc.

### DIFF
--- a/mitiq/pt/tests/test_utils.py
+++ b/mitiq/pt/tests/test_utils.py
@@ -51,9 +51,25 @@ def test_size_of_pauli_vec():
 
 def test_ptm_matrix_size():
     """Checks the output of the PTM matrix function."""
+    # one qubit circuit
     q0 = cirq.LineQubit(0)
     circuit = cirq.Circuit()
     circuit.append(cirq.Rz(rads=0.25 * np.pi)(q0))
     expected_ptm_matrix = ptm_matrix(circuit, 1)
     assert expected_ptm_matrix.shape == (4**1, 4**1)
-    assert np.isclose(expected_ptm_matrix[0, 1], 0.0, rtol=1e-05)
+    expected_array = np.array(
+        [
+            [1.0, 0.0, 0.0, 0.0],
+            [0.0, 0.7071, -0.7071, 0.0],
+            [0.0, 0.7071, 0.7071, 0.0],
+            [0.0, 0.0, 0.0, 1.0],
+        ]
+    )
+    np.allclose(expected_ptm_matrix, expected_array, atol=1e-08)
+
+    # two qubit circuit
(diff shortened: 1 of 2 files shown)


mitiq/pt/tests/test_utils.py
```diff
@@ -0,0 +1,59 @@
...
+
+
+def test_ptm_matrix_size():
+    """Checks the output of the PTM matrix function."""
+    q0 = cirq.LineQubit(0)
+    circuit = cirq.Circuit()
+    circuit.append(cirq.Rz(rads=0.25 * np.pi)(q0))
+    expected_ptm_matrix = ptm_matrix(circuit, 1)
+    assert expected_ptm_matrix.shape == (4**1, 4**1)
+    assert np.isclose(expected_ptm_matrix[0, 1], 0.0, rtol=1e-05)
```
purva-thakre, line 59: Change this line to assert the output array is close to the expected array instead of comparing just one element.
```
    assert np.isclose(expected_ptm_matrix[0, 1], 0.0, rtol=1e-05)
```

purva-thakre: Also, include a 2 qubit circuit in this test to check the output is as expected. 

Will be required for twirling a CNOT gate.

purva-thakre:
Closing this for now because functions defined in this PR already exist as stated in #1981. If there's a need to further define functions for Pauli Twirling, I need to create a branch from #1987

PR #1967, opened 2023-08-18
ZNE Stim tutorial

Misty-W:
Create a Mitiq tutorial that runs on the Stim stabilizer simulator backend. 

Similar workflow to that followed in https://arxiv.org/abs/2304.14985 -  simulate logical RB circuits, model errors as single-qubit bit flip and phase flip errors occurring with probability equal to the logical error rate, which is calculated from code distance according to an empirical formula from https://clelandlab.uchicago.edu/pdf/fowler_et_al_surface_code_submit_3po.pdf. 

### License


docs/source/examples/zne_logical_rb_cirq_stim.md
```diff
@@ -0,0 +1,302 @@
...
+```
+
+## Plot results
+
++++
+
+Finally, plot the mean unmitigated and ZNE-mitigated expectation values, averaged over the trial circuits, at every code distance simulated. The error bars indicate the standard deviation.
+
+```{code-cell} ipython3
+:tags: [remove-output]
```
natestemen, line 244: I think we can use `skip-execution` here if we aren't going to show output, right?

docs/source/examples/zne_logical_rb_cirq_stim.md
```diff
@@ -0,0 +1,302 @@
...
+---
+
+name: zne-stim-plot
+---
+Plot of the unmitigated and ZNE-mitigated expectation values obtained from executing the logical RB circuits.
+```
+
+We can see from the above plot that the ZNE-mitigated expectation values are closer to the ideal value of 1.0 at every code distance simulated. 
+The effect is more pronounced at lower code distances, which correspond to a higher logical error rate, whereas by $d = 21$ both the mitigated and unmitigated expectation values approach 1.0. 
+We can also think of the results in terms of an effective code distance, where e.g. .
```
natestemen, line 295: uh oh something got dropped here!

natestemen (commented):
Nice work Misty! Really great to see mitiq working with other libraries.

1. I think it would help simplify some of the later code if we used Mitiq's register to/from functionality (see comment for more details)
2. I'm missing where the error correction is happening. Could you clarify (to me, doesn't necessarily have to be in the notebook, unless you think it should) where this is happening?
3. I missed why we need to apply the extrapolation/scaling by hand instead of using one of the functions to handle this. Can you add some details as to why this was (again comment here should be fine so I understand).

docs/source/_thumbnails/zne_stim.png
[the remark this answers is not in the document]
Misty-W: Good idea, making a combined one :)

dev_requirements.txt
```diff
@@ -32,3 +32,5 @@ pyscf==2.2.1; sys_platform != 'win32'
 openfermion==1.5.1; sys_platform != 'win32'
 openfermionpyscf==0.5; sys_platform != 'win32'
 bqskit[ext]==1.0.4
+stim==v1.11.0
```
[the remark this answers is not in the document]
Misty-W: oops, that was an unedited copy/paste.

docs/source/examples/zne_logical_rb_cirq_stim.md
```diff
@@ -0,0 +1,302 @@
...
+:tags: [remove-cell]
+
+# hidden cell for faster docs build
+trials = 5
+n_cliffords = 10
+p_err = 0.006  # generate meaningful plot at lower depth for diagnostics
+cirq_circuits = mitiq.benchmarks.generate_rb_circuits(
+    n_qubits, num_cliffords, trials
+)
+```
```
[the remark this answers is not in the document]
Misty-W: Update just after writing this comment: I should've used `hide-cell` instead of `remove-cell`, so I'm fixing that now.

docs/source/examples/zne_logical_rb_cirq_stim.md
```diff
@@ -0,0 +1,302 @@
...
+  name: python3
+---
+
+# ZNE on Stim backend with Cirq: Logical randomized benchmarking circuits
+
++++
+
+This tutorial demonstrates a method of combining quantum error mitigation (QEM) and quantum error correction (QEC), with the goal of reducing the effective logical error rate of the computation. 
+While QEM techniques such as zero noise extrapolation (ZNE) and probabilistic error cancellation are typically thought of as belonging to the NISQ regime, recently it has been shown that they can also benefit applications within the fault-tolerant regime {cite}`Piveteau_2021_PRL, Suzuki_2022_PRX, Wahl_2023_arXiv_ds_zne`.
+In this example, we will apply ZNE with noise scaling by global (circuit-level) unitary folding on logical randomized benchmarking (RB) circuits.
```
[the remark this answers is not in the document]
Misty-W: Removed "circuit-level" and added folding link.

docs/source/examples/zne_logical_rb_cirq_stim.md
```diff
@@ -0,0 +1,302 @@
...
+This tutorial also introduces the use of Mitiq's ZNE functions with Cirq as the frontend and the [Stim](https://github.com/quantumlib/Stim) stabilizer simulator as the backend.
+
++++
+
+## Setup
+
++++
+
+Since Stim is not a Mitiq-supported frontend, and the [stimcirq](https://pypi.org/project/stimcirq/) library has a conversion utility for Cirq and Stim circuits, we will use Cirq as the frontend for this example.
```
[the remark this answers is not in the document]
Misty-W: Per discussion in last week's pairing call, this tutorial won't use `register_mitiq_converters`  because top-level Mitiq functions are not used.

docs/source/examples/zne_logical_rb_cirq_stim.md
```diff
@@ -0,0 +1,302 @@
...
+        idle = False
+        for q in qubits:
+            if not m.operates_on_single_qubit(q):
+                idle = True
+                filled_circuit.append(m.with_operations(cirq.I(q)))
+                break
+        if not idle:
+            filled_circuit.append(m)
+
+    return filled_circuit
```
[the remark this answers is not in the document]
Misty-W: Changed function name to `fill_circuit_with_identities` and use `expand_to`.

docs/source/examples/zne_logical_rb_cirq_stim.md
```diff
@@ -0,0 +1,302 @@
...
+```{code-cell} ipython3
+def add_noise_to_stim_circuit(circuit, p_err, p_th, d):
+    noisy = stim.Circuit()
+    for instruction in circuit:
+        noisy.append(instruction)
+        if instruction.name != "TICK":
+            noisy.append(
+                "X_ERROR",
+                instruction.targets_copy(),
+                gen_noise_model(p_err, p_th, d),
```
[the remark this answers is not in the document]
Misty-W: Nice catch, moved it out of the loop.

docs/source/examples/zne_logical_rb_cirq_stim.md
```diff
@@ -0,0 +1,302 @@
...
+        [r[0].astype(int) + r[1].astype(int) for r in result]
+    )
+    return 1 - counts / shots
+```
+
+We will simulate the RB circuits at six different code distances and corresponding logical error rates.
+The distance of the error correcting code parameterizes the size of the physical data block per logical qubit and is directly related to the number of errors the code can correct.
+
+```{code-cell} ipython3
+d_array = np.linspace(21, 11, 6, dtype=int)
```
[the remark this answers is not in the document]
Misty-W: Went with option 2, so the user can easily modify the tutorial.

docs/source/examples/zne_logical_rb_cirq_stim.md
Misty-W: Changed my mind and removed the half-sentence about effective code distance.

docs/source/examples/zne_logical_rb_cirq_stim.md
```diff
@@ -0,0 +1,302 @@
...
+    filled_circuit.append(cirq.measure(filled_circuit.all_qubits()))
+    filled_circuits.append(filled_circuit)
+```
+
+## Noise model
+
++++
+
+The noise is modeled as single-qubit $X$ and $Z$ errors, with probability $p_L$ given by an empirical formula from Ref. {cite}`Fowler_2012_PRA`: 
+$$\mathcal{P}_L\cong 0.03 (p/p_{th})^{(d + 1)/2}$$.
```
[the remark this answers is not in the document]
Misty-W: Fixed while addressing other comments.

docs/source/examples/zne_logical_rb_cirq_stim.md
```diff
@@ -0,0 +1,302 @@
...
+n_qubits = 2
+p_err = 0.004
+p_th = 0.009
+base_shots = 10**4
+device_size = 1500
+```
+
+Randomized benchmarking is typically performed at the physical level, but it can be applied at the logical level as shown in Ref {cite}`Combes_2017_arXiv_logical_rb`.
+The modularity of logical RB circuits allows for parallelization of circuit executions, as permitted by the size of the device and number of physical qubits required for the logical circuit.
+We use a built-in Mitiq function, `mitiq.benchmarks.generate_rb_circuits`, to generate the RB circuits,
```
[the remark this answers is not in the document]
Misty-W: API-doc link added.

docs/source/examples/zne_logical_rb_cirq_stim.md
```diff
@@ -0,0 +1,302 @@
...
++++
+
+The noise is modeled as single-qubit $X$ and $Z$ errors, with probability $p_L$ given by an empirical formula from Ref. {cite}`Fowler_2012_PRA`: 
+$$\mathcal{P}_L\cong 0.03 (p/p_{th})^{(d + 1)/2}$$.
+
+```{code-cell} ipython3
+def gen_noise_model(p_err, p_th, distance):
+    """Create sweepable Pauli noise model."""
+    logical_err = 0.03 * (p_err / p_th) ** int((distance + 1) / 2)
+    return logical_err
```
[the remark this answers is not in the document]
Misty-W: Name changed to `calculate_logical_error_rate`.

docs/source/examples/zne_logical_rb_cirq_stim.md
Misty-W: We keep the `remove-output` tag to test that the full example runs w/o errors, while only showing the plot from the long circuits because that's where a fast stabilizer simulator like Stim is needed.

docs/source/examples/zne_logical_rb_cirq_stim.md
```diff
@@ -0,0 +1,302 @@
...
+```
+
+We will simulate the RB circuits at six different code distances and corresponding logical error rates.
+The distance of the error correcting code parameterizes the size of the physical data block per logical qubit and is directly related to the number of errors the code can correct.
+
+```{code-cell} ipython3
+d_array = np.linspace(21, 11, 6, dtype=int)
+```
+
+Given a fixed (serial) sampling budget, here `base_shots = 10,000`, at code distances below the maximum available on the device, we can parallelize executions of the modular logical RB circuits and thereby take additional samples.
```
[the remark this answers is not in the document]
Misty-W: Changed wording to "we can parallelize executions of each logical RB circuit and thereby take additional samples simultaneously".

docs/source/examples/zne_logical_rb_cirq_stim.md
```diff
@@ -0,0 +1,302 @@
...
+    logical_err = 0.03 * (p_err / p_th) ** int((distance + 1) / 2)
+    return logical_err
+```
+
+```{code-cell} ipython3
+def add_noise_to_stim_circuit(circuit, p_err, p_th, d):
+    noisy = stim.Circuit()
+    for instruction in circuit:
+        noisy.append(instruction)
+        if instruction.name != "TICK":
```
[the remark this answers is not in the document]
Misty-W: Added a comment `# append X and Z errors after every operation, ignoring timing ticks` above `if instruction.name != "TICK":`

Misty-W:
> 1. I think it would help simplify some of the later code if we used Mitiq's register to/from functionality (see comment for more details)

 Replied to comment.
> 2. I'm missing where the error correction is happening. Could you clarify (to me, doesn't necessarily have to be in the notebook, unless you think it should) where this is happening?

Error correction is not simulated directly but assumed in the noise model of X and Z errors with probability equal to the logical error rate calculated from the empirical formula in the text.
> 3. I missed why we need to apply the extrapolation/scaling by hand instead of using one of the functions to handle this. Can you add some details as to why this was (again comment here should be fine so I understand).

Every time we call `execute_with_zne`, the circuit is folded per the supplied scale factors. Even at the reduced (hidden) Clifford depth of 10, folding takes too long to be re-done at every code distance for every RB circuit instance, so instead we fold the circuits once and execute them at each noise level, before performing extrapolation on the resulting expectation values. This is explained briefly in the tutorial text w/o calling much attention to this compromise.

Misty-W:
Commit history got a bit weird after merging in `master`, so I'll do a rebase once the changes are approved.

natestemen:
> Commit history got a bit weird after merging in `master`, so I'll do a rebase once the changes are approved.

Would you mind doing the rebase ahead of a review? There are a lot of changes in the [diff](https://github.com/unitaryfund/mitiq/pull/1967/files) that are not apart of this PR and I don't want to skip over anything accidentally.

nathanshammah:
Thanks Misty. All looks great to me including the stim logo in the thumbnail. It would be nice as Nate says to see the rebase with cleaner edits. One suggestion: You can consider citing more explicitly in a last section (like "## Additional material" or something like it, the [research folder](https://github.com/unitaryfund/research/tree/main/ds_zne)).

Misty-W:
Unfortunately, rebasing has only made the situation worse. I tried `git reset --hard` but it didn't help.

docs/source/_thumbnails/zne_stim.png
natestemen, line 1: I think this should be removed now?

docs/source/conf.py
```diff
@@ -159,9 +159,9 @@
     r"https://doi.org/10.1088/1367-2630/12/7/075008",
     r"https://doi.org/10.7566/JPSJ.90.032001",
     r"https://doi.org/10.7566/jpsj.90.032001",
-    r"https://dl.acm.org/doi/10.1145/3466752.3480059",
     r"https://doi.org/10.1145/3466752.3480059",
-    r"https://doi.org/10.7566/jpsj.90.032001",
+    r"https://dl.acm.org/doi/abs/10.1145/3503222.3507739",
+    r"https://www.sciencedirect.com/science/article/abs/pii/S037704271930344"
```
natestemen, line 164: Are these changes because there were some more links broken?

docs/source/examples/zne_logical_rb_cirq_stim.md
```diff
@@ -0,0 +1,303 @@
...
+
+Since Stim is not a Mitiq-supported frontend, and the [stimcirq](https://pypi.org/project/stimcirq/) library has a conversion utility for Cirq and Stim circuits, we will use Cirq as the frontend for this example.
+
+```{code-cell} ipython3
+import cirq
+import matplotlib as mpl
+import numpy as np
+import stim
+import stimcirq
+import mitiq
```
natestemen, line 37: Just to group related dependencies.
```suggestion
import matplotlib as mpl
import numpy as np

import stim
import stimcirq
import cirq

import mitiq
```

docs/source/examples/zne_logical_rb_cirq_stim.md
```diff
@@ -0,0 +1,303 @@
...
+```
+
+In this example, we assume a physical error rate of 0.004 and a threshold of 0.009.
+Since the physical error rate is below the threshold, the computation is operating in the fault-tolerant regime.
```
natestemen, line 41: Question just for my understanding, and doesn't need to go in the tutorial:

If the physical error rate is above the threshold can you still combine QEM with QEC to get something useful?

docs/source/examples/zne_logical_rb_cirq_stim.md
```diff
@@ -0,0 +1,303 @@
...
+```
+
+## Noise model
+
++++
+
+The noise is modeled as single-qubit $X$ and $Z$ errors, with probability $p_L$ given by an empirical formula from Ref. {cite}`Fowler_2012_PRA`: 
+
+$$
+p_L\cong 0.03(p/p_{th})^{(d+1)/2}
```
natestemen, line 103: Since this is already in display mode, might as well do a full standing fraction. Also making $\mathrm{th}$ romanized for consistency with original source.
```suggestion
p_L\cong 0.03(\frac{p}{p_\mathrm{th}})^{(d+1)/2}
```

docs/source/examples/zne_logical_rb_cirq_stim.md
```diff
@@ -0,0 +1,303 @@
...
+                l_err,
+            )
+    return noisy
+```
+
+## Execute circuits
+
++++
+
+The Stim executor accepts a Stim circuit, to which it applies the noise model, and then compiles and executes.
```
natestemen, line 139:
```suggestion
The Stim executor defined below accepts a Stim circuit, to which it applies the noise model, and then compiles and executes.
```

docs/source/examples/zne_logical_rb_cirq_stim.md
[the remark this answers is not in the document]
natestemen: I think it would be great to add the description you gave here (maybe slightly modified) to the tutorial.

docs/source/examples/zne_logical_rb_cirq_stim.md
```diff
@@ -0,0 +1,303 @@
...
+
+To minimize execution time, we convert each trial circuit only once, outside of the execution loop.
+
+```{code-cell} ipython3
+noisy_results = np.zeros((trials, len(code_distances)))
+stim_circuits = [
+    stimcirq.cirq_circuit_to_stim_circuit(c) for c in filled_circuits
+]
+
+for di in range(len(code_distances)):
```
natestemen, line 180: This might simplify some of the arguments to `stim_executor`.
```suggestion
for i, code_distance in enumerate(code_distances):
```

docs/source/examples/zne_logical_rb_cirq_stim.md
```diff
@@ -0,0 +1,303 @@
...
+        noisy_results[t, di] = stim_executor(
+            stim_circuits[t],
+            p_err,
+            p_th,
+            code_distances[di],
+            scale_shots(device_size, code_distances[di], 4 * base_shots, n_qubits),
+        )
+```
+
+Next we apply ZNE, with the noise scaling method of global unitary folding and scale factors {1, 3, 5, 7}, to each trial circuit and at each code distance simulated in the unmitigated case.
```
natestemen, line 191:
```suggestion
Next we apply ZNE with the noise scaling method of global unitary folding and scale factors `[1, 3, 5, 7]`, to each trial circuit and at each code distance simulated in the unmitigated case.
```

docs/source/examples/zne_logical_rb_cirq_stim.md
```diff
@@ -0,0 +1,303 @@
...
+The effect is more pronounced at lower code distances, which correspond to a higher logical error rate, whereas by $d = 21$ both the mitigated and unmitigated expectation values approach $1.0$. 
+
+```{note}
+Not all logical circuits can be folded, even at the circuit level.
+One alternative noise scaling method for logical qubits is scaling the code distance, which is referred to as distance-scaled ZNE or DS-ZNE {cite}`Wahl_2023_arXiv_ds_zne`.
+```
+
+## Additional information on DS-ZNE
+
+Modeling the noise as in Eq. (1), assuming the computation is operating in the fault tolerant regime, we see that logical error rate decreases as code distance increases.
```
natestemen, line 301: There is a way to get these reference to be automatically generated by using the syntax described [here](https://mystmd.org/guide/math#math-directives). Since there is only one equation here I don't think it's needed, but wanted to flag for future use.

natestemen (approved):
Awesome tutorial! A few small comments from me.

docs/source/examples/zne_logical_rb_cirq_stim.md
Misty-W: Generally not a good idea to implement QEC if p_err > p_th because the encoding will, if anything, worsen the errors instead of correct them. We don't have QEM techniques that would change that, other than noise suppression techniques, but in that case one could maybe argue that p_err could be taken as the noise-suppressed error rate.

docs/source/conf.py
Misty-W: At one point yes, but I didn't mean for any changes beyond what Purva fixed in her PR. Changing it back to match `master` now.

docs/source/examples/zne_logical_rb_cirq_stim.md
Misty-W: Suggestion applied, thanks Nate!

docs/source/_thumbnails/zne_stim.png
Misty-W: Yes, removed. Thanks for the reminder :)

docs/source/examples/zne_logical_rb_cirq_stim.md
```diff
@@ -0,0 +1,307 @@
...
+
+for i, code_distance in enumerate(code_distances):
+    for t in range(trials):
+        noisy_results[t, i] = stim_executor(
+            stim_circuits[t],
+            p_err,
+            p_th,
+            code_distances[i],
+            scale_shots(device_size, code_distances[i], 4 * base_shots, n_qubits),
+        )
```
natestemen, line 192: You've already got the `code_distance` from the for loop. No need to fetch it again! You might be able to do something similar with the inner loop going over `stim_circuits` as opposed to `range(trials)`, but I'm not confident `len(stim_circuits) = trials`.
```suggestion
for i, code_distance in enumerate(code_distances):
    for t in range(trials):
        noisy_results[t, i] = stim_executor(
            stim_circuits[t],
            p_err,
            p_th,
            code_distance,
            scale_shots(device_size, code_distance, 4 * base_shots, n_qubits),
        )
```

natestemen: I think this comment applies where you converted another for loop to use `enumerate` as well.

Misty-W: Good catch, and indeed `trials` is the same as `len(stim_circuits)`, so I simplified that as well.

PR #1977, opened 2023-08-28
Change the main function in Pauli Twirling

purva-thakre:
As discussed in the user guide PR.

https://github.com/unitaryfund/mitiq/pull/1848#issuecomment-1693881459

### DIFF
--- a/mitiq/pt/pt.py
+++ b/mitiq/pt/pt.py
@@ -50,11 +50,14 @@ CZ_twirling_gates = [
 ]
 
 
-def tailor_circuit_with_pauli_twirling(
+def pauli_twirl_circuit(
     circuit: QPROGRAM,
     num_circuits: int = 10,
 ) -> List[QPROGRAM]:
     """Return the Pauli twirled versions of the input circuit.
+
+    Only the $CZ$ and $CNOT$ gates in an input circuit
+    are Pauli twirled as specified in :cite:`saki2023hypothesis`.
 
     Args:
         circuit: The input circuit to execute with twirling.
--- a/mitiq/pt/pt.py
+++ b/mitiq/pt/pt.py
@@ -54,10 +54,11 @@ def pauli_twirl_circuit(
     circuit: QPROGRAM,
     num_circuits: int = 10,
 ) -> List[QPROGRAM]:
-    """Return the Pauli twirled versions of the input circuit.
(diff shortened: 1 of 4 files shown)


mitiq/pt/pt.py
```diff
@@ -51,38 +50,25 @@
...
     circuit: QPROGRAM,
-    executor: Union[Executor, Callable[[QPROGRAM], QuantumResult]],
-    observable: Optional[Observable] = None,
-    *,
     num_circuits: int = 10,
-) -> float:
-    """Estimates the expectation value of the input circuit by averaging
-    expectation values obtained from Pauli twirled circuits.
+) -> List[QPROGRAM]:
+    """Return the Pauli twirled versions of the input circuit.
```
natestemen, line 57: Can we add a short note here that only $\mathrm{CZ}$ and $\mathrm{CNOT}$ gates are twirled?

mitiq/pt/pt.py
```diff
@@ -51,38 +50,25 @@
 ]
 
 
-def execute_with_pauli_twirling(
+def tailor_circuit_with_pauli_twirling(
```
natestemen, line 53: Not opposed to this naming, but what do you think of `pauli_twirl_circuit`? Only reason I'm thinking to change it is because the main action (verb) of this function is twirl, not tailor.

natestemen (approved):
Thanks for clarifying this! Just two minor comments.

mitiq/pt/pt.py
```diff
@@ -51,38 +50,28 @@
...
-    observable: Optional[Observable] = None,
-    *,
     num_circuits: int = 10,
-) -> float:
-    """Estimates the expectation value of the input circuit by averaging
-    expectation values obtained from Pauli twirled circuits.
+) -> List[QPROGRAM]:
+    """Return the Pauli twirled versions of the input circuit.
+
+    Only the $CZ$ and $CNOT$ gates in an input circuit
```
natestemen, line 59: Minor nitpick:
```suggestion
    Only the $\mathrm{CZ}$ and $\mathrm{CNOT}$ gates in an input circuit
```

_This line may be too long now, if you agree to the change._

purva-thakre: Good catch!

PR #1989, opened 2023-09-01
Move functions to utils

### LINKED ISSUE #1981: Some utils are common to multiple techniques
For example, 

- Kronecker product is repeated in [shadows](https://github.com/unitaryfund/mitiq/blob/415c6fa62327b3cd29abec1149d0042b93fa6b53/mitiq/shadows/shadows_utils.py#L29) and [pec](https://github.com/unitaryfund/mitiq/blob/master/mitiq/pec/channels.py#L89)
- Matrix to vector form and vector to matrix form will be repeated in [pauli_twirling](#1963 ) and [pec](https://github.com/unitaryfund/mitiq/blob/master/mitiq/pec/channels.py#L106).
- PTM might be repeated in [shadows](https://github.com/unitaryfund/mitiq/blob/415c6fa62327b3cd29abec1149d0042b93fa6b53/mitiq/shadows/shadows_utils.py#L43) and #1963. However, the PTM method in shadows seems to be using a different definition than the one in #1963. 

Do we want to move functions like these to a folder called `useful_operations` along with descriptive docstrings of what they do instead of repeating them? These could be further distinguished into two categories: `state_operations` and `channel_operations`. The first step to this would be figuring out all the operations that are repeatedly defined across different techniques. 

Open to other possible solutions.

purva-thakre:
Fixes #1981


mitiq/__init__.py
```diff
@@ -13,6 +13,9 @@
     QuantumResult,
 )
 
+from mitiq.utils import (
+    tensor_product, matrix_to_vector, _safe_sqrt, vector_to_matrix)
+
```
andreamari, line 18: What about if we remove this? The reason is to keep the root of `mitiq` free from very specific and technical functions.

We can still access utility functions with `from mitiq.utils import xxxx` instead of using the shortcut `from mitiq import xxxx` .

Misty-W:
@andreamari and @purva-thakre, wanted to check if this PR is likely to make Friday afternoon's release. Looks to be in good shape, but let me know if there's any way I can help.

purva-thakre:
> @andreamari and @purva-thakre, wanted to check if this PR is likely to make Friday afternoon's release. Looks to be in good shape, but let me know if there's any way I can help.

@Misty-W Yes, this should be done by tomorrow night. I still need to move a few things.

purva-thakre:
This should be ready to merge if all tests pass. 

I have moved 2 functions previously labeled[ `tensor_product` ](https://github.com/unitaryfund/mitiq/pull/1989/files#diff-c946d3a8144f9476f8391c254b489e09b5d3252e9c9eafa608d5560178bae9efR277)and [`kronecker_product`](https://github.com/unitaryfund/mitiq/pull/1989/files#diff-c946d3a8144f9476f8391c254b489e09b5d3252e9c9eafa608d5560178bae9efR327) but their outputs are not consistent. 
One outputs the tensor product matrix as expected. The other one's output is a list of tensor products (I think). 

Still need to figure out a way to use a common `tensor_product` function to replace both. Code for both functions is almost the same except for their input & output types. Leaving this unresolved for now. To be fixed in a new PR for the PTM matrix for reasons discussed below.

The function labeled `operator_ptm_vector_rep` should be a 4 x 4 matrix for a 1-qubit channel. It is not so [in a unit test](https://github.com/unitaryfund/mitiq/pull/1989/files#diff-79483cb81cc52d9e4d9bd7db40fee904ed003695e6379a0a338e755685d8f11cR425). Need to change the name for this function and also add a new PTM matrix function.

andreamari (approved):
Thanks @purva-thakre! It's already a good improvement, so worth merging now.

About the future improvements that you mentioned:

- the two tensor product functions can easily become single one. They only differ in the way of passing the input matrices: `funcion_1(matrix_a, matrix_b, matrix_c) `  vs `function_2([matrix_a, matrix_b, matrix_c])`. So we can keep one of the two by alwyas calling it in corresponding way.
- The output shape of `operator_ptm_vector_rep` is correct. It is just the name which is misleading. It is expected to output a _vector_ in the vector space associated to the PTM representation. Basically such a vector is the object on which a PTM can be applied.  So we may need a new function  for mapping an operator to its PTM.

purva-thakre:
@andreamari Thanks!

- Ok about keeping one of the two. I pushed this change to the future because I wanted to understand the differences in how the inputs for both functions are unpacked. Very interesting!
- Can I re-label `operator_ptm_vector_rep`  function as `vector_for_action_of_ptm`? I will change the docstring to be more detailed as well. I spent quite some time being confused about this function.

andreamari:
> * Can I re-label `operator_ptm_vector_rep`  function as `vec_of_operator_for_action_of_ptm`? I will change the docstring to be more detailed as well. I spent quite some time being confused about this function.

Good idea to rename the function. Your proposal is a bit long. Something shorter would be better e.g.  `operator_to_pauli_basis` , or, `expand_in_pauli_basis" , etc..

PR #2028, opened 2023-09-26
New QEM benchmarking method: "rotated" RB circuits

### LINKED ISSUE #2005: Introduce new benchmarks with ideal expectation values different from 1
## Issue Description

Most examples and most benchmarks in Mitiq are based on benchmark circuits whose ideal expectation value is 1.
For example:
 https://github.com/unitaryfund/mitiq/blob/9a13b3ec476cd2cd978e572d12081af4d73536cb/mitiq/benchmarks/randomized_benchmarking.py#L23

Typically we test how good a QEM technique is, by checking how  close a mitigated expectation value is to 1 (ideal result) compared to the noisy unmitigated result.

This is very convenient and intuitive. However, it is also useful to have more general benchmarks in which the expectation value can be any number in a continuous interval e.g.  

$$ E_{\rm ideal} \in  [-1, 1] $$

This would enable testing QEM techniques in more general scenarios, closer to real-world applications in which expectation values can take arbitrary values.

## Proposed Solution

A possible way to achieve this is to introduce a new function for generating "rotated" RB circuits in which we 
insert a $R_z(\theta)$ rotation in the middle of an RB circuit.
Something like:

$$ C(\theta) =    G_n \dots G_{n/2 +1} R_z(\theta)G_{n/2} \dots G_2 G_1 $$

where $G_j$ are Clifford elements or Clifford gates.

This should generate expectation values which are sinusoidal functions of $\theta$, so something varying in with a continuous range of ideal expectation values.

At the same time since, up to factors of 2, we have $R_z(\theta) =cos(\theta) I +  i \ sin(\theta) Z$,  the rotated Clifford circuit $C(\theta)$ can be written as a  linear combination of just two Clifford circuits and, therefore, it is still easy to classically simulate.

**We are open to other proposals**, please share alternative ideas in the comments below this issue.

## Additional References

For similar tricks, but used in a different context, see  https://arxiv.org/abs/2010.02538

If large scale near Clifford circuits are necessary, [Qrack](https://github.com/unitaryfund/qrack) can be used as a fast classical simulator.

Misty-W:
## Description

Fixes #2005.

Add a function `generate_rotated_rb_circuits` which generates a list of randomized benchmarking circuits, each with a random $R_z$ rotation inserted in the middle. 

This benchmarking method enables testing QEM techniques in more general scenarios, closer to real-world applications in which expectation values can take arbitrary values.

Rotated randomized bencmarking circuits are randomized benchmarking circuits in which an $R_z(\theta)$ rotation is inserted in the middle, such that:

$$ C(\theta) =    G_n \dots G_{n/2 +1} R_z(\theta)G_{n/2} \dots G_2 G_1 $$

where $G_j$ are Clifford elements or Clifford gates.

The circuits generate expectation values which are sinusoidal functions of $\theta$, which in the ideal (noiseless) case vary in a continuous interval of $E_{\rm ideal} \in  [-1, 1]$.

Since (up to factors of 2) we have $R_z(\theta) =cos(\theta) I +  i \ sin(\theta) Z$, the rotated Clifford circuit $C(\theta)$ can be written as a linear combination of just two Clifford circuits and, therefore, it is still easy to classically simulate.

---

### License


natestemen (commented):
Nice, _quick_, work! Looking forward to using this to do some more testing! I've got a few higher level comments before diving into the details.

## Math

The math from the docstring isn't being displayed properly. You can either use double backslashes in place of the existings `\`s, or you can use the following syntax.
```
:math:`equation`
```
for inline math, and
```
..math::
    equation
```
for display math. An example of these in use can be found here: https://github.com/unitaryfund/mitiq/blob/b6cb5c1342652b64156259ae2d795c048af645b9/mitiq/pec/channels.py#L94-L104

## API

The other thing that I could imagine we might want to use this function for is testing EM methods as a function of this $\theta$ parameter. Instead of using a seed to set the angle, what if we let the user pass a value for `\theta`, falling back on random value when it's not passed. That would allow us to write code such as
```py
for theta in np.linspace(0, 2*np.pi, 50):
    circuit = generate_rotated_rb_circuits(...) # passing theta in here
    # calculate circuits noisy, ideal, and mitigated expval
    # plot the error as a function of theta
```
Does this seem like something that would be interesting?

## Combining this with `generate_rb_circuits`

Did you consider merging this function with the [existing function](https://mitiq.readthedocs.io/en/stable/apidoc.html#mitiq.benchmarks.randomized_benchmarking.generate_rb_circuits)? I could imagine we could allow that function to take a parameter `theta` that would rotate the circuit midway through if it was passed in. Otherwise, returning the "standard" RB circuit. We probably want to avoid overloading functions with too many options, but this may be okay? Not suggesting we have to go this route, just curious if you considered it.

---

### Benchmarking utilities consistency

This PR made me realize that our utilities to generate benchmarking circuits all return a single `QPROGRAM` except for RB circuits (and now rotated RB circuits) where we return `List[QPROGRAM]`. @andreamari is there a reason for this? I feel like we should try to be consistent here. This is not necessarily something we should address in this PR, but maybe it's something we want to open an issue for if we want to address it?

Misty-W:
Thanks @natestemen! All good points to consider. 

I've fixed the math rendering.

As for combining the functions, I thought about it, but decided against it because when we insert the rotation, it's no longer really an RB circuit anymore, but a linear combination of Clifford groups. 

As for setting $\theta$, it should be randomly sampled and generally not specified by the user. The seed is for troubleshooting purposes or other cases where we want to generate the same circuit repeatedly (e.g. in a tutorial where we want consistent results).

mitiq/benchmarks/rotated_randomized_benchmarking.py
```diff
@@ -0,0 +1,80 @@
...
+
+    Returns:
+        A list of rotated randomized benchmarking circuits.
+    """
+
+    circuits = generate_rb_circuits(n_qubits, num_cliffords, 2 * trials)
+    rotated_circuits = []
+
+    for circ in circuits:
+        rotated_circ, _ = convert_to_mitiq(circ)
```
andreamari, line 69: This is probably unnecessary, since `circ` is a Cirq circuit.

andreamari (commented):
Thanks @Misty-W for the clean and simple solution! 
I have only one doubt: are you sure about adding a layer many rotations (one for each qubit)?

Initially I was thinking of a single rotation gate applied to just one qubit, possibly a qubit approximately in the middle of the quantum register. If I am not wrong, this should also work and should produce (with very high probability) an expectation value which is exactly given by $E(\theta)=cos(\theta)$.

With a layer of many rotations, if I am not wrong, the expectation value will contain a product of many sinusoidal functions. So $E(\theta)$ will be more complicated.

Mine are just untested guesses. So, a practical strategy is probably to just plot $E(\theta)$ and check which of the two solutions looks better.

nathanshammah:
Great work @Misty-W! @Misty-W @andreamari what do you think of adding this functionality to the main RB function, e.g., as a `theta` optional parameter? Is it too different of a functionality from the standard RB tests?

mitiq/benchmarks/rotated_randomized_benchmarking.py
Misty-W: I had used `convert_to_mitiq` to satisfy `mypy`, but after discussing with Nate we think it's better to `cast` the output of `generate_rb_circuits` to `cirq.Circuit`.

Misty-W:
thanks for the suggestions @andreamari ! 

I've changed the function to only insert an $R_z$ gate  on the first qubit, in (or near) the middle of the circuit.

Here are some plots of the output

andreamari:
> Here are some plots of the output- seems to be what we want in terms of the rotation, but any idea why they are so "spiky"?
> 

**About different plots with different simulation methods:**
In the first plot you computed the square of the wave function amplitude, instead of the square of the absolute value. This is probably the reason why the two plots look very different.

**About the spikes:** 
I think they are the consequence of two facts:

- For each `theta` a completely new random RB circuit is generated. This explains the non-continuous jumps. Adding the possibility of setting the same seed should help.

- Since the circuit is small (2 qubits), there is a non-negligible probability that the rotation has no effect on the expectation value. This only happens in the particular cases in which 
$$C_1 C_2... C_{d/2} Z C_{d/2 + 1}  ... C_d = P$$  where P is a Pauli operator diagonal in the Z basis, in those cases the expectation value is always 1 for any theta. This is very unlikely when the number of qubits is large, but it can happen quite often for 2 qubits.

At the moment it's not easy to increase the number of qubits of RB circuits (this could be a new interesting issue), so my suggestions to avoid strange effects are:

- Add the possibility of setting a seed for the RB circuit. This is always a good thing to have.
- [Alternative A] If theta!=0,  discard a pathological RB circuit if the expectation value is equal to 1 and regenerate a new random RB circuit.  
- [Alternative B] In addition to the circuit also return the ideal expectation value. In this way, the user can easily detect and discard pathological cases.

You may have better solutions. I think anything is fine as long as the user is aware of the expected behavior and doesn't get surprised by unexpected results (as we did! :smiley:  ).

Misty-W:
@andreamari, since the Cirq random Clifford functions don't accept a random seed, I went with [Alternative A]: 

> If theta!=0, discard a pathological RB circuit if the expectation value is equal to 1 and regenerate a new random RB circuit.

Misty-W:
RTD problem is unrelated to this PR- seems GitHub automatically shows a failure when RTD takes longer than 1 hour to finish building, even though it built successfully.

andreamari (commented):
Hi Misty, 

could you test if the new function produces a smooth Cos(theta) plot?
As you tried in [this message](https://github.com/unitaryfund/mitiq/pull/2028#issuecomment-1739899331).

I see some problems with the current code:

- Main problem: it seems that we replace a pathological circuit, i.e. the case with E(theta)=1, with a new random RB circuit but **without** the intermediate RZ gate. We should instead add the RZ gate otherwise E(theta) would still be = to 1, so still pathological. 
- Second problem: the current code doesn't check if the second attempt is pathological as well. 

Maybe a good pseudo-workflow is the following:

1) Generate a RB circuit (`rb_circuit`) and save a copy of it (`rb_copy`).
2) Insert a Z gate (equivalent to RZ(\pi) ) in the middle of the circuit.
3) If the expectation value is close to 1, we have a pathological circuit, and we restart from step 1. Else, the case is not pathological, and we can go to next step.
4) Insert the R_Z(theta) rotation in  `rb_copy ` and return it. 

Whatever solution you decide to take, the final goal is having a smooth cos(theta) plot for the expectation value.

Misty-W:
hi @andreamari,
 
> could you test if the new function produces a smooth Cos(theta) plot? As you tried in [this message](https://github.com/unitaryfund/mitiq/pull/2028#issuecomment-1739899331).

Yes, I'll regenerate the plots with the new code. 

> I see some problems with the current code:
> 
> * Main problem: it seems that we replace a pathological circuit, i.e. the case with E(theta)=1, with a new random RB circuit but **without** the intermediate RZ gate. We should instead add the RZ gate otherwise E(theta) would still be = to 1, so still pathological.
> * Second problem: the current code doesn't check if the second attempt is pathological as well.

I'm using `generate_rotated_rb_circuits` recursively, so the intermediate RZ gate is inserted and the second attempt is also checked.

andreamari:
> I'm using `generate_rotated_rb_circuits` recursively, so the intermediate RZ gate is inserted and the second attempt is also checked. If I had used `generate_rb_circuits` the above problems would have materialized.

Ah, I missed the "rotated" word. Now I see! 
However, the second recursive call to  `generate_rotated_rb_circuits` is missing the argument `theta`.

Note also that you may enter an infinite recursion when theta is a multiple of 2pi.

Misty-W:
In the end, I decided to use a seed. It lets the user generate a set of rotated RB circuits with different values of theta but otherwise the same circuit, which is useful for testing and troubleshooting. Once the user's workflow is set up, then they can randomly choose or sweep over `theta` and plot the error rate rather than the raw expectation value. 

 To do this efficiently I had to use some code from the Cirq functions `_random_single_qubit_clifford` and `random_two_qubit_clifford`. I have credited the developers in a similar manner to what is done in the `shadows` module.

Misty-W:
Plotting the expectation values as a function of theta at three seed values:

```
import cirq
import numpy as np
import matplotlib.pyplot as plt

from mitiq.benchmarks.rotated_randomized_benchmarking import (
    generate_rotated_rb_circuits,
)

thetas = np.linspace(0, 2 * np.pi, 100)
seeds = (1, 2, 3)
results = []
for seed in seeds:
  zero_probs = []
  for theta in thetas:
      circuit = generate_rotated_rb_circuits(
              n_qubits=2, num_cliffords=5, theta=theta, seed=seed,
            )[0]
      zero_probs.append(cirq.DensityMatrixSimulator()
          .simulate(circuit)
          .final_density_matrix[0, 0]
          .real
      )
  results.append(zero_probs)
plt.plot(thetas, results[0], label="seed = 1")
plt.plot(thetas, results[1], label="seed = 2")
plt.plot(thetas, results[2], label="seed = 3")
plt.legend()
plt.xlabel(r"$\theta$")
plt.ylabel("E")
plt.show()
```

image

mitiq/benchmarks/rotated_randomized_benchmarking.py
```diff
@@ -0,0 +1,80 @@
...
+    that:
+
+    .. math::
+        C(\theta) =    G_n \dots G_{n/2 +1} R_z(\theta)G_{n/2} \dots G_2 G_1
+
+    where :math:`G_j` are Clifford elements or Clifford gates.
+
+    The circuits generate expectation values which are sinusoidal functions of
+    :math:`\theta`, which in the ideal (noiseless) case vary in a continuous
+    interval of :math:`E_{\rm ideal} \in  [-1, 1]`.
```
andreamari, line 42: To avoid unexpected results, I would say something similar to this:

```suggestion
    For most values of the seed, the probability of the zero state is a
    sinusoidal function of :math:`\theta`. For some values of the seed
    the probability of the zero state is 1 for all :math:`\theta`.
```

mitiq/benchmarks/randomized_benchmarking.py
```diff
@@ -47,35 +55,56 @@ def generate_rb_circuits(
             "Only generates RB circuits on one or two "
             f"qubits not {n_qubits}."
         )
-    qubits = LineQubit.range(n_qubits)
+    qubits = cirq.LineQubit.range(n_qubits)
     cliffords = _single_qubit_cliffords()
-
+    np.random.seed(seed)
```
andreamari, line 60: directly seeding `np.random` is dangerous since this adds a global seed to the full NumPy module and would affect all other Mitiq functions based on numpy.

The trick is to define a local random state as done, e.g., in [this function](https://github.com/unitaryfund/mitiq/blob/0c0b7a0fe73f582aa1974c4ab3c49b9da0b5753e/mitiq/observable/observable.py#L99).

andreamari (approved):
Cool! Very nice to have the `seed` option.
This LGTM, once the comments below are addressed.

mitiq/benchmarks/randomized_benchmarking.py
Misty-W: thanks for the suggestion @andreamari, I've updated the function accordingly.

Misty-W:
thanks for the helpful comments @andreamari, @natestemen, and @nathanshammah! 

I think I've addressed everything, so once the tests pass here I'll merge this.

mitiq/benchmarks/randomized_benchmarking.py
```diff
@@ -38,6 +45,7 @@ def generate_rb_circuits(
             returned circuits. See the keys of
             ``mitiq.SUPPORTED_PROGRAM_TYPES`` for options. If ``None``, the
             returned circuits have type ``cirq.Circuit``.
+        seed: A seed for generating radomzed benchmarking circuits.
```
natestemen, line 48:
```suggestion
        seed: A seed for generating radomized benchmarking circuits.
```

PR #2038, opened 2023-10-04
Include calculated error in calibrator logs

### LINKED ISSUE #2014: Include calculated error in calibrator logs
Along with having the good/bad indicators, it would be nice to have a column which included the actual error.

Idea by @andreamari

kozhukalov:
## Description

Calibration log table is extended with errors calculated
during calibration and sorted so that best
techniques are on the top of the table.

Fixes #2014

---

### License

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -53,3 +53,4 @@ Abdulrahman Sahmoud
 Mohammad Zuhair Khan
 Brian Goldsmith
 Farzad Kianvash
+Vladimir Kozhukalov
--- a/mitiq/calibration/calibrator.py
+++ b/mitiq/calibration/calibrator.py
@@ -6,6 +6,7 @@
 import warnings
 from itertools import product
 from typing import (
+    Any,
     Callable,
     Dict,
     Iterator,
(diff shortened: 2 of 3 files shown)


natestemen (commented):
Really nice work Vladimir! Thanks for adding this!

I think we should add the errors by default as part of the table instead of adding the additional `log_errors` option. IMO if someone is asking for logs, lets provide everything!

kozhukalov:
Ok, I'll make it default. 

The only concern here is backward compatibility. Some people probably try to parse the performance log with their automation tools and we can break their pipelines. However I am new in Mitiq. Not sure how important it is.

kozhukalov:
I am not sure what is wrong with the docs/readthedocs.org:mitiq check. It says `Read the Docs build failed!` but the build looks successful. @natestemen can you please help me understand what I should fix to make it green?

purva-thakre:
@kozhukalov This is a bug in RTD. When there are multiple PRs running checks at the same time, RTD fails to build the documentation in all of them. 

I do not know a way to re-trigger this build besides making a dummy commit.

kozhukalov:
@purva-thakre The issue is that even if this RTD job is green it the becomes red later when another PR runs RTD job. So I think it should not be voting.

natestemen (approved):
Awesome work with this! Thanks for the nice simple upgrade. :shipit:

PR #2064, opened 2023-10-19
More general conversion decorator and fix conversion bug in PEC

### LINKED ISSUE #2061: `pec.sample_circuit` drops idle qubits
## Issue Description

Circuits returned by `pec.sample_circuit` do not contain idle qubits present in the original circuit, and also have different register names.

## How to Reproduce

```python
import qiskit
from mitiq import pec

# Circuit.
qreg = qiskit.QuantumRegister(3, name="Q")
circuit = qiskit.QuantumCircuit(qreg)
circuit.h(qreg[0])
circuit.cnot(qreg[0], qreg[1])
circuit.measure_active()
print(circuit)
#            ┌───┐      ░ ┌─┐   
#       Q_0: ┤ H ├──■───░─┤M├───
#            └───┘┌─┴─┐ ░ └╥┘┌─┐
#       Q_1: ─────┤ X ├─░──╫─┤M├
#                 └───┘ ░  ║ └╥┘
#       Q_2: ───────────░──╫──╫─
#                       ░  ║  ║ 
# measure: 2/══════════════╩══╩═
#                          0  1 

# Representation.
cnot_circuit = qiskit.QuantumCircuit(qreg)
cnot_circuit.cnot(qreg[0], qreg[1])
rep = pec.represent_operation_with_global_depolarizing_noise(cnot_circuit, noise_level=0.01)

# Sample.
circuits, *_ = pec.sample_circuit(circuit, [rep])
for circuit in circuits:
    print(circuit)
#                    ┌───┐     ┌─┐   
#               q_0: ┤ H ├──■──┤M├───
#                    └───┘┌─┴─┐└╥┘┌─┐
#               q_1: ─────┤ X ├─╫─┤M├
#                         └───┘ ║ └╥┘
# m_reg_measure_0: 1/═══════════╩══╬═
#                               0  ║ 
# m_reg_measure_1: 1/══════════════╩═
#                                  0 
```

## Expected behavior

The circuit(s) returned by `pec.sample_circuit` should have the same registers and qubits as the input circuit: here, one register named "Q", the qubit "Q_2" present, and one measurement register named "measure" with two bits.

---

<details>
<summary>Environment Context</summary>

Use the `about()` function to summarize information on operating system, python version and dependencies.
```python
from mitiq import about
about()
```

```text
Mitiq: A Python toolkit for implementing error mitigation on quantum computers
==============================================================================
Authored by: Mitiq team, 2020 & later (https://github.com/unitaryfund/mitiq)

Mitiq Version:	0.30.0

Core Dependencies
-----------------
Cirq Version:	1.2.0
NumPy Version:	1.26.1
SciPy Version:	1.11.2

Optional Dependencies
---------------------
PyQuil Version:	3.2.1
Qiskit Version:	0.44.2
Braket Version:	Not installed

Python Version:	3.10.13
Platform Info:	Linux (x86_64)
```

Additional Python Environment Details (`pip freeze` or `conda list`):

```
Copy and paste the output of `pip freeze` or `conda list` here.
```

</details>

andreamari:
Fixes #2061  (conversions bug in PEC)

### License

### DIFF
--- a/mitiq/interface/conversions.py
+++ b/mitiq/interface/conversions.py
@@ -5,7 +5,7 @@
 
 """Functions for converting to/from Mitiq's internal circuit representation."""
 from functools import wraps
-from typing import Any, Callable, Dict, Iterable, Tuple, cast
+from typing import Any, Callable, Dict, Iterable, Optional, Tuple, cast
 
 import cirq
 
@@ -259,17 +259,20 @@ def atomic_one_to_many_converter(
 
 
 def accept_qprogram_and_validate(
-    cirq_circuit_modifier: Callable[..., Any]
+    cirq_circuit_modifier: Callable[..., Any],
+    one_to_many: Optional[bool] = False,
 ) -> Callable[..., Any]:
     """This decorator performs two functions:
 
-        1. Transforms a function of signature (cirq.Circuit -> Any) to
-        (QPROGRAM -> Any)
+        1. Transforms a function of signature (cirq.Circuit -> cirq.Circuit)
+        to (QPROGRAM -> QPROGRAM).
         2. Validates the incoming QPROGRAM instance to ensure Mitiq's error
         mitigation techniques can be applied to it.
 
     Args:
-        cirq_circuit_modifier: The function to scale.
+        cirq_circuit_modifier: The function modifying a Cirq circuit.
+        one_to_many: If True, cirq_circuit_modifier is expected to return
+            a sequence of Cirq circuits instead of a single Cirq circuit.
 
     Returns:
         The transformed function which can take any QPROGRAM, and performs
@@ -277,9 +280,7 @@ def accept_qprogram_and_validate(
     """
 
     @wraps(cirq_circuit_modifier)
-    def new_scaling_function(
-        circuit: QPROGRAM, *args: Any, **kwargs: Any
-    ) -> QPROGRAM:
+    def new_function(circuit: QPROGRAM, *args: Any, **kwargs: Any) -> QPROGRAM:
         # Pre atomic conversion
         if "qiskit" in circuit.__module__:
             from qiskit.transpiler.passes import RemoveBarriers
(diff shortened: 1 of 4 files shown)


mitiq/interface/conversions.py
```diff
@@ -296,68 +297,83 @@ def new_scaling_function(
...
+
+                out_circuit.remove_final_measurements()
+                _transform_registers(out_circuit, new_qregs=circuit.qregs)
+                _remove_identity_from_idle(out_circuit, idle_qubits)
+                if circuit.cregs and not out_circuit.cregs:
+                    out_circuit.add_register(*circuit.cregs)
+
+                for q, c in _measurement_order(circuit):
+                    out_circuit.measure(q, c)
+            circuits_to_return.append(out_circuit)
```
andreamari, line 368: FYI: this block is unchanged apart from re-naming a variable. For some reason Git Hub diff shows it as new code block (green) but it's actually not new.

mitiq/interface/conversions.py
```diff
@@ -259,27 +259,28 @@ def qprogram_modifier(
...
-    cirq_circuit_modifier: Callable[..., Any]
+    cirq_circuit_modifier: Callable[..., Any],
+    one_to_many: Optional[bool] = False,
 ) -> Callable[..., Any]:
     """This decorator performs two functions:
 
-        1. Transforms a function of signature (cirq.Circuit -> Any) to
-        (QPROGRAM -> Any)
+        1. Transforms a function of signature (cirq.Circuit -> cirq.Circuit)
+        to (QPROGRAM -> QPROGRAM).
```
natestemen, line 268: Prior to this change, this docstring was more general than what was actually supported. Is that right? I.e. this isn't because a change you made in the code, but rather bringing the docstring in line with the existing functionality.

mitiq/interface/conversions.py
```diff
@@ -259,27 +259,28 @@ def qprogram_modifier(
...
         2. Validates the incoming QPROGRAM instance to ensure Mitiq's error
         mitigation techniques can be applied to it.
 
     Args:
-        cirq_circuit_modifier: The function to scale.
+        cirq_circuit_modifier: The function modifying a Cirq circuit.
+        one_to_many: If True, cirq_circuit_modifier is expected to return
+            a sequence of Cirq circuits instead of a single Cirq circuit.
```
natestemen, line 275:
```suggestion
        cirq_circuit_modifier: The function modifying a Cirq circuit.
        one_to_many: If True, ``cirq_circuit_modifier`` is expected to return
            a sequence of Cirq circuits instead of a single Cirq circuit.
```
**Warning**: this may make `flake8` unhappy.

mitiq/interface/conversions.py
natestemen: I think GitHub is displaying it this way as not only has the variable been renamed, but the code has been added inside an `if` block (hence indenting all of this code by an additional level).

mitiq/interface/conversions.py
```diff
@@ -296,68 +297,83 @@ def new_scaling_function(
...
+                _remove_identity_from_idle(out_circuit, idle_qubits)
+                if circuit.cregs and not out_circuit.cregs:
+                    out_circuit.add_register(*circuit.cregs)
+
+                for q, c in _measurement_order(circuit):
+                    out_circuit.measure(q, c)
+            circuits_to_return.append(out_circuit)
+
+        if not one_to_many:
+            assert len(circuits_to_return) == 1
```
natestemen, line 371: I don't think we normally have `assert` statements in our non-test code. I'm not opposed to it, but wanted to flag this in case it was left over from testing.

mitiq/pec/sampling.py
```diff
@@ -121,13 +123,14 @@ def sample_circuit(
             input circuit. If a representation cannot be found for an operation
             in the circuit, a ValueError is raised.
         random_state: Seed for sampling.
-        num_samples: The number of samples.
+        num_samples: The number of samples,
+        extra_data: Mutable dictionary to save extra data beyond output
+            circuits. After the function call it will contain ``signs`` (the
+            signs associated to sampled_circuits) and ``norm`` (the one-norm
+            of the full circuit representation).
```
natestemen, line 130: This is an interesting way that I've never considered to give users additional info from a function. I'm accustomed to having a return type of something like `Union[float, Tuple[float, Dict[str, int]]]` where `float` is returned by default, and `Tuple[float, Dict[str, int]]` is returned if some flag is passed as `True` (or similar).

Interesting idea!

What — in your mind — are the tradeoffs of using this method, over the more "traditional" method of returning extra data if the user asks?

natestemen (commented):
Nice changes Andrea. I like the way you've added the `accept_qprogram_and_validate` function into the PEC workflow.

I've left a few minor commments/questions, but overall I'm still wondering _where exactly were we dropping the extra (unused) qubits_? Would you mind elaborating on that a little?

mitiq/interface/conversions.py
andreamari: Exactly!

mitiq/pec/sampling.py
andreamari: I have seen this approach in other contexts. The first time I have seen this was on C++ OpenCV functions :smile:    see e.g. the output of the [blur](https://docs.opencv.org/3.4/d4/d86/group__imgproc__filter.html#ga8c45db9afe636703801b0b2e440fce37)() function.

Personally, I find it a bit counter-intuitive but there are cases in which it can be useful (my debatable opinions):
 
- For logging information on a mutable "recording" variable passing through multiple functions. I think it's handy to just pass to the function the previous logged data and ask, "please continue appending info to this object".
- If the mutable argument is optional, it simplifies the structure of the returned type. Avoiding complicated things like: 
```
if return_all: 
    return y, data 
else 
    return y
```
- In cases in which the function return type cannot be changed, in order to interfacing well with other functions or decorators.  This is exactly the case of this PR. The conversion decorator expects a Cirq function returning `List[cirq.Circuit]` but we also wanted to output additional data: `signs` and `norm`. So using the `extra_data` argument in this case was quite helpful for making the function compatible with the decorator.


Having said that, I don't know it this is a good practice.

mitiq/interface/conversions.py
andreamari: Once I tried reading something online about `assert` vs `exceptions`. 
If I understand well, `assert` can be in the code to "assert" things which, unless we are missing something, must be always true. In practice, there shouldn't be a simple way of triggering that error. But since you are not 100% sure, you add the assert statement as a double check.

Exceptions instead are designed for "triggerable" errors caused by using a function in the wrong way, typically by a user of the code. So the error condition can be both true or false, depending on good or bad usage of the function.

Again, I am not sure if this makes sense. But it sounds reasonable.

natestemen: Ah okay that's a good distinction to make. Thanks for flushing that out. Happy to keep it (and this practice moving forward).

mitiq/pec/sampling.py
natestemen: Thanks Andrea for the thoughts. I agree that in certain scenarios this pattern seems acceptable (maybe even good!). If for some reason down the line we run into problems, we can always modify it.

natestemen (approved):
:shipit:

PR #2066, opened 2023-10-23
Support python 3.12 and drop 3.9

natestemen:
## Description

Add support for python 3.12, **drop support for python 3.9**.

Fixes #2395
Fixes https://github.com/unitaryfund/mitiq/issues/2215

PR #2068, opened 2023-10-23
Drop support for python 3.8

natestemen:
## Description

Does what is says on the 🥫.

This is needed over https://github.com/unitaryfund/mitiq/pull/2066 as #2066 includes upgrading to 3.12 which it looks like we cannot do just yet due to some dependencies not supporting 3.12.


andreamari (approved):
Thanks!

PR #2069, opened 2023-10-24
Clean up GitHub CI

natestemen:
## Description

This PR contains three changes:

1. Removal of the `draft` option from the build trigger
2. Removed (in 9f99c82a623a246b75bdf9709aa066360b7dd40d) the rest of the code from the "draft workflow" which was partially removed in https://github.com/unitaryfund/mitiq/pull/2019.
3. Consolidation of the build matrix so that all linux/mac/windows builds occur in the same matrix.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -7,7 +7,6 @@ on:
       - reopened
       - synchronize
       - ready_for_review
-      - draft
     branches:
       - master
   push:
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -86,7 +86,6 @@ jobs:
           pytest mitiq/tests/test_without_third_party_packages.py
 
   test-linux:
-    if: github.event.pull_request.draft == false
     strategy:
       fail-fast: false
       matrix:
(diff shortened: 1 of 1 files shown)


.github/workflows/build.yml
```diff
@@ -109,98 +107,15 @@ jobs:
...
           make install requirements
       - name: Run the quilc & qvm Docker images
+        if: ${{ matrix.os == 'ubuntu-latest' }}
         run: |
           docker run --rm -idt -p 5000:5000 rigetti/qvm -S
           docker run --rm -idt -p 5555:5555 rigetti/quilc -R
-      - name: Run all the unit tests & generate coverage report with pytest
-        run: make test-all
+      - name: Run all the unit tests
+        run: ${{ matrix.os == 'ubuntu-latest' && 'make test-all' || 'make test' }}
```
andreamari, line 115: Is it possible to put an if condition `matrix.os == 'ubuntu-latest'` directly inside a `run` command ?
It seems it works but just double checking.

andreamari (approved):
Thanks for the cleaning! Approved, but also left a small question.

.github/workflows/build.yml
natestemen: Yeah! The syntax `A && B || C` is a ternary operator which runs `B` if `A` is true, and otherwise `C`. An example of this can be found on the [GitHub docs](https://docs.github.com/en/actions/learn-github-actions/expressions#example).

I'd prefer to use an `if`/`else` block as I think it's more readable, but currently there is [no way](https://stackoverflow.com/questions/60916931/github-action-does-the-if-have-an-else) to do an `else` using GitHub actions.

andreamari: Thanks for the explanation!

PR #1970, opened 2023-08-18
Remove binder links

andreamari:
This PR removes the unstable/not-working mybinder badge, a few links pointing to it and a bot posting it to each new PR.

Based on discussion and decision taken in #1724. 
Fixes #1724.

Related to #1820.

---

### License

- [ x] I license this contribution under the terms of the GNU GPL, version 3 and grant Unitary Fund the right to provide additional permissions as described in section 7 of the GNU GPL, version 3.

PR #2071, opened 2023-10-25
Remove "binder" remnants

natestemen:
## Description

Since we no longer use/advertise the use of mybinder (#1970), we can remove the `binder` directory. Support for binder was originally added in https://github.com/unitaryfund/mitiq/pull/841.

### DIFF
--- a/binder/requirements.txt
+++ b/binder/requirements.txt
@@ -1,1 +1,0 @@
-.[development]


Misty-W (approved):
Thanks @natestemen ! Looks fine to merge once the docs and RTD finish.

natestemen:
RTD finished and did indeed build (if got the ❌ because the known problem about the builds taking a long time). Going ahead with merging.

PR #2099, opened 2023-11-28
add support for single circuit on exeuctor run method

### LINKED ISSUE #2098: Add support for single QPROGRAM to `Executor.run()`
## Issue Description

Currently, the `run()` method of the `Executor` object only accepts lists of `QPROGRAM`s. When a single `QPROGRAM` is passed, it raises an error which does not directly tell the user what is the problem:

``` python
TypeError: bad input to run() function;circuits and schedules cannot be mixed in a single run
```

Letting the user pass a single `QPROGRAM` is a subtle but useful feature, particularly if one does not want to go deep in the code to see what is happening.

## Proposed Solution

Change:

``` python
def run(
        self,
        circuits: Sequence[QPROGRAM],
        ...
```

for 

``` python
def run(
        self,
        circuits:Union[QPROGRAM, Sequence[QPROGRAM]],
        ...
```

and  add:

``` python
 if not isinstance(circuits, List):
            circuits = [circuits]
```

EmilianoG-byte:
## Description

This PR adds the option to pass a single `QPROGRAM` to the `Executor.run()` method by converting into a list when necessary.

Closes #2098 

---

### License

### DIFF
--- a/mitiq/executor/executor.py
+++ b/mitiq/executor/executor.py
@@ -211,7 +211,7 @@ class Executor:
 
     def run(
         self,
-        circuits: Sequence[QPROGRAM],
+        circuits: Union[QPROGRAM, Sequence[QPROGRAM]],
         force_run_all: bool = True,
         **kwargs: Any,
     ) -> Sequence[QuantumResult]:
@@ -219,11 +219,14 @@ class Executor:
         the executor.
 
         Args:
-            circuits: Sequence of circuits to execute using the executor.
+            circuits: Circuit or sequence thereof to execute with the executor.
             force_run_all: If True, force every circuit in the input sequence
                 to be executed (if some are identical). Else, detects identical
                 circuits and runs a minimal set.
         """
+        if not isinstance(circuits, Sequence):
+            circuits = [circuits]
+
         start_result_index = len(self._quantum_results)
 
         if force_run_all:


EmilianoG-byte:
Hi 😄! This is the implementation I suggested to close the issue #2099. 

I have not added my name to the AUTHORS because I could not find such file (is it the changelog?) 

The Greetings test is failing, is that customary for first PR's?

EmilianoG-byte:
Also, I found an issue when using the `qiskit-aer` version from the development requirements (0.13.1), it throws an error regarding the `DiagonalGate`. I could solve this by downgrading to `0.12.2`.

purva-thakre:
@EmilianoG-byte https://github.com/unitaryfund/mitiq/blob/3457774734f9759c2a828d84ad5af3801eab23e1/AUTHORS

EmilianoG-byte:
> @EmilianoG-byte https://github.com/unitaryfund/mitiq/blob/3457774734f9759c2a828d84ad5af3801eab23e1/AUTHORS

Thank you! I have no idea how I missed it. I could not find a pattern on the order, so I added myself at the end. Hopefully this is okay!

purva-thakre:
> Also, I found an issue when using the `qiskit-aer` version from the development requirements (0.13.1), it throws an error regarding the `DiagonalGate`. I could solve this by downgrading to `0.12.2`.

@EmilianoG-byte Since you have figured out the issue for the build failures, can you make changes to the following line in your branch of this PR?

https://github.com/unitaryfund/mitiq/blob/3457774734f9759c2a828d84ad5af3801eab23e1/dev_requirements.txt#L3

EmilianoG-byte:
> @EmilianoG-byte Since you have figured out the issue for the build failures, can you make changes to the following line in your branch of this PR?
> 
> https://github.com/unitaryfund/mitiq/blob/3457774734f9759c2a828d84ad5af3801eab23e1/dev_requirements.txt#L3

Done! Now all checks are passing. Was the problem with the Aer version happening before?

purva-thakre:
>Done! Now all checks are passing. 

Yay!

>Was the problem with the Aer version happening before?

I don't think so. I noticed this when I updated [my fork](https://github.com/purva-thakre/mitiq/commits/master). Looks like some major changes happened from when `qiskit-aer` was updated to this week. There were no failures when it was upgraded by dependabot almost a month ago 29adcad9f6486cde03b4ae71b5998e1faca962ea .

EmilianoG-byte:
> Thanks @EmilianoG-byte and thanks @purva-thakre for reviewing this!
> 
> Downgrading qiskit is not optimal since in principle we want to remain compatible with the latest version. However it seems a pragmatic solution which is easy to revert as soon as compatibility is restored.

Thank you for merging it Andrea! Glad my solution was suitable 😄

PR #2102, opened 2023-11-30
Add support for qibo circuits

francescsabater:
Add support for Qibo circuits. Create the needed functions for the conversions between Qibo and Cirq and their respective tests. Update the documentation. Also, fix a bug in identity_insertion that did not allow to implement identity scaling with non-cirq circuits:  

Add Qibo to dev requirments 

Add to_qibo and from_qibo to conversions. 

Add Qibo to QPROGRAM

Create init file in mitiq_qibo.  

Create Qibo conversions file. 

Create tests file for Qibo conversions.  

Update test_pec.

Update frontends-backends.md

Add example qibo-noisy-simulation

Solve bug in identity insertion. 

I license this contribution under the terms of the GNU GPL, version 3 and grant Unitary Fund the right to provide additional permissions as described in section 7 of the GNU GPL, version 3.
 
Authored by Francesc Sabater <francescsabater11@gmail.com>.
This contribution is part of the internship of the Master In Quantum Science and Technology of Barcelona carried out at the Barcelona Supercomputing Center under the supervision of Alba Cervera-Lierta.


andreamari:
Thanks @francescsabater , having Qibo circuits is an important improvement!
This PR looks great and well designed, however it is almost 700 new lines of code and I wonder if it would be possible to split it in 2 PRs. 

E.g. first PR for conversions functions and second PR for documentation/example.

This would help us to make the review process more efficient.

francescsabater:
I have removed from this pull all the changes and additions that affected the documentation, I hope this will make the review process easier for you as @andreamari suggested. The changes in documentation may be added in a future pull. As for making the pull even shorter, it is complicated since all the code is necessary for the mitiq functions to accept qibo circuits.

Misty-W (commented):
Nice work, overall this is in good shape. Before final review, please rebase and fix conflicts, run `make format` for formatting fixes, and address the questions and suggestions in the inline comments.

Thanks for this great contribution to Mitiq!

mitiq/interface/mitiq_qibo/conversions.py
```diff
@@ -0,0 +1,390 @@
...
+        gates.CNOT(q0,q1), 
+        gates.U1(q1,-theta/2),
+        gates.CNOT(q0,q1), 
+        gates.U1(q1,theta/2),
+        gates.H(q1),
+    ]
+    return decomp_gate
+
+
+def iswap_decomp(gate: gates.iSWAP) -> list:
```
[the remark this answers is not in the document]
francescsabater: True, yet when you try to convert a circuit containing an iswap gate, cirq doesn't seem to know how to go from iswap in qasm to cirq. The following error (if the iswap decomposition is deleted) arises if one tries to convert an iswap from qibo to cirq: 

qibo_circuit = Circuit(2) 
qibo_circuit.add(gates.iSWAP(0,1))
converted_from_qibo=from_qibo(qibo_circuit)

cirq.contrib.qasm_import.exception.QasmException: Unknown gate "iswap" at line 5

mitiq/interface/mitiq_qibo/conversions.py
```diff
@@ -0,0 +1,390 @@
...
+    """Returns a QiboCircuit circuit equivalent to the input QiboCircuit
+    with all unknown by cirq gates decomposed.
+
+    Args:
+        qibo_circuit: QiboCircuit to decompose.
+
+    Returns:
+        Decomposed QiboCircuit
+    """
+    decomp_circuit = qibo_circuit.__class__(qibo_circuit.nqubits)
```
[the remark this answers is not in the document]
francescsabater: There was no particular reason, I have changed the code for your suggestion which is much simpler and more understandable than the code before.

mitiq/zne/scaling/identity_insertion.py
```diff
@@ -67,7 +68,7 @@ def _calculate_id_layers(
         )
         return (num_uniform_layers, num_partial_layers)
 
-
+@accept_qprogram_and_validate
```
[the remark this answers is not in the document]
francescsabater: Yes, if you wanted to use the indentity insertion method in ZNE for any type of circuit other than Cirq, (Qiskit for example) an error occurred as the function did not convert the circuit to cirq to modify the circuit.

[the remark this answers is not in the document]
Misty-W: awesome, thanks!

mitiq/interface/mitiq_qibo/conversions.py
```diff
@@ -0,0 +1,390 @@
...
+        circuit: Cirq circuit to convert to a QiboCircuit.
+
+    Returns:
+        QiboCircuit object equivalent to the input Mitiq circuit.
+    """
+    qasm = cirq_to_qasm(circuit)
+    nqubits, gate_list = QiboCircuit._parse_qasm(qasm)
+    
+    if not isinstance(nqubits, int):
+        raise UnsupportedCirqCircuitError(
```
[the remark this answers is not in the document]
Misty-W: Makes sense, thanks!

mitiq/interface/mitiq_qibo/conversions.py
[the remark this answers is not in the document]
Misty-W: Makes sense, thanks!

Misty-W:
Looks like there is a `scipy` dependency conflict causing `validate` to fail. Can you loosen the Qibo version requirement to something that accepts `scipy <1.10.0`? If so I think that should resolve it.

Also there are some failing tests that need to be fixed.

Misty-W:
thanks for your patience @francescsabater! 

For this PR we would like for you to lower the version of Qibo to v0.1.11 and remove the gates not compatible with v0.1.11.

Hopefully OpenFermion will release a new version soon that’s compatible with a later version of scipy.

Misty-W:
hi @francescsabater, great progress here! 

Codecov is flagging that the lines validating the input aren't tested, and it would be good to make sure they're covered. You might be able to use `unittest.mock` for that purpose.

`mypy` is also flagging some missing and inconsistent type hints. Let me know if you need help resolving them.

I'm not immediately sure why the windows tests are failing when Linux and mac are passing. I'll bring this up to the group tomorrow to see if they have any ideas.

bdg221:
The failures from the Windows tests are a result of using different versions of Qibo. In the dev_requirements.txt the following is set: `qibo~=0.1.11`

The gates.QASM_GATES and gates.PARAMETRIZED_GATES were removed in [version 0.1.13](https://github.com/qiboteam/qibo/releases/tag/v0.1.13) (The change is labeled as "Add gate labels".) 

For some reason, Ubuntu and MacOS tests chose to use `qibo-0.1.11` ([see the "Install Python dependencies" section](https://github.com/qiboteam/qibo/releases/tag/v0.1.13) of the test) `INFO: pip is looking at multiple versions of qibo to determine which version is compatible with other requirements. This could take a while.)` 

The Windows tests on the other hand does not show the same pip message and only checks and uses [qibo-0.1.16] (https://github.com/unitaryfund/mitiq/actions/runs/7458145597/job/20414102643?pr=2102#step:4:956).

We would still need to look into why pip behaved differently, but this should at least allow the issue to be corrected.

purva-thakre:
@bdg221 What if we temporarily pin the version as `qibo==0.1.11`?

Using `~=` will get `pip` stuck in finding a better version through the dependency resolver. https://stackoverflow.com/a/67374648

Misty-W:
good find @bdg221! 

I agree with @purva-thakre to pin the version- `qibo==0.1.11` looks to be the one.

dev_requirements.txt
```diff
@@ -6,6 +6,8 @@ pyquil~=3.5.4
 pennylane-qiskit~=0.33.1
 pennylane~=0.33.1
 amazon-braket-sdk~=1.65.1
+qibo~=0.1.11
```
Misty-W, line 9: @francescsabater this should fix the problem with the Windows tests, as it seems the Windows dependency resolver behaves differently from Mac and Linux.

```suggestion
qibo==0.1.11
```

francescsabater: Yes, problem fixed! Thanks  @Misty-W  and @bdg221 @purva-thakre !!!

Misty-W (commented):
hi @francescsabater, I was able to resolve some `mypy` errors with these suggestions.

mitiq/interface/mitiq_qibo/conversions.py
```diff
@@ -0,0 +1,529 @@
...
+                qubit_list.append(qubits[qubit])
+            if gatename == "sx": 
+                gate_list.append(("RX", list(qubit_list),  [pi/2]))
+            elif gatename == "sxdg": 
+                gate_list.append(("RX", list(qubit_list),  [-pi/2]))
+            else: 
+                gate_list.append((gates.QASM_GATES[gatename], list(qubit_list), params))
+
+    # Create measurement gate qubit lists from registers
+    for i, (gatename, register, _) in enumerate(gate_list):
```
Misty-W, line 522: Seems that `register` here is a `List[int]` but seems we want a string to index `registers` in line 524.

mitiq/interface/mitiq_qibo/conversions.py
```diff
@@ -0,0 +1,529 @@
...
+        line for line in qasm_code.split("\n") if line and line[:2] != "//" and line[1:3] != "//"
+    ).split(";") if line and "//" not in line)
+
+    if next(lines) != "OPENQASM 2.0":
+        raise_error(ValueError, "QASM code should start with 'OPENQASM 2.0'.")
+
+    qubits: Dict[Tuple[str, int], int] = {}  
+    cregs_size = {}  
+    registers: Dict[str, Optional[Dict[int, int]]] = {}
+    gate_list = []
```
Misty-W, line 399: Not sure if this is the best solution, but since `mypy` is asking for type hints on the others, we could resolve another error with this type hint. @natestemen what do you think about this?
```suggestion
    gate_list: List[Tuple[str, List[int], Optional[List[float]]]] = []
```

mitiq/interface/mitiq_qibo/conversions.py
```diff
@@ -0,0 +1,529 @@
...
+                        denominator = 1
+                        if "/" in p:
+                            p, denominator = p.split("/")
+                        if "pi" in p:
+                            from functools import reduce
+                            from operator import mul
+
+                            s = p.replace("pi", str(pi)).split("*")
+                            p = reduce(mul, [float(j) for j in s], 1)
+                        params[i] = float(p)
```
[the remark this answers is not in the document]
francescsabater: Params is initally a list of strings mainly numbers, for example params=['0']. 
In the end of the function we append the params to gate list*, but this params should be a list of floats, params=[ 0 ]. 
In the case of having a pi in params we have initially params=['pi*-0.5']
After the conversion (entering an if to deal with 'pi') we have params=[1.5707963267948966] where the number is type float. 
If this is causing problems to mypy because of the change of type in the list we could define another list of floats and add to this list the floats without changing the types in the initial params list. 
*Line where we append params to gate list
gate_list.append((gates.QASM_GATES[gatename], list(qubit_list), params))

mitiq/interface/mitiq_qibo/conversions.py
francescsabater: I've realised that in the case of having a measurement instead of an usual gate register is indeed a string. I've changed the order in wich we assign register to gate list at first becuase I was assigning register to the second index when it should be third. Then, the third gate_list index can be not only List[Float] or None but also str (when having a measurment).

mitiq/interface/mitiq_qibo/conversions.py
francescsabater: Yes! This solved some errors with the correct type indicator for gate_list: 
`gate_list: List[Tuple[str, List[int], Union[List[float], str, None]]] = []`
Without indicating the type mypy assumes the type of the list third argument after it's first assignment in the code wich might be either List[float],str or None, when trying to assign another type later on in the code mypy raised an error. Pre-assigning the type of gate_list solves this error.

mitiq/interface/mitiq_qibo/conversions.py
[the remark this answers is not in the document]
francescsabater: `gate_list.append((gates.QASM_GATES[gatename], list(qubit_list), params))`
This line was causing an error because, as you pointed out, params was initially set to a list of strings but converted into a list of floats mid-code. To solve the error mypy was rasing I've implemented the following: 
`float_params = None if params is None else [float(element) for element in params]
gate_list.append((str(gates.QASM_GATES[gatename]), list(qubit_list), float_params))`

mitiq/interface/mitiq_qibo/conversions.py
```diff
@@ -0,0 +1,531 @@
...
+            raise UnsupportedQiboCircuitError(
+                f"OpenQASM does not support capital letters in register names but {reg_name} was used."
+            )
+    for gate in qibo_circuit.queue:
+        if isinstance(gate, gates.M):
+            continue
+        if gate.is_controlled_by:
+            raise UnsupportedQiboCircuitError(
+                "OpenQASM does not support multi-controlled gates."
+            )
```
natestemen, line 312: From checking out the code of `controlled_by` (which `is_controlled_by` refers to), it seems like you can create a singly-controlled gate using `controlled_by`. Does OpenQASM support arbitrary singly-controlled gates?

mitiq/interface/mitiq_qibo/conversions.py
```diff
@@ -0,0 +1,531 @@
...
+
+
+def read_args(args: str) ->  Generator[Tuple[str, int], None, None]:
+        _args = iter(re.split(r"[\[\],]", args))
+        for name in _args:
+            if name:
+                index = next(_args)
+                if not index.isdigit():
+                    raise_error(ValueError, "Invalid QASM qubit arguments:", args)
+                yield name, int(index)
```
natestemen, line 369: This function needs a docstring.

mitiq/interface/mitiq_qibo/conversions.py
```diff
@@ -0,0 +1,531 @@
...
+        gate_list: List that specifies the gates of the circuit.
+            Contains tuples of the form
+            (Qibo gate name, qubit IDs, optional additional parameter).
+            The additional parameter is the ``register_name`` for
+            measurement gates or ``theta`` for parametrized gates.
+    """
+     
+    lines = (line for line in "".join(
+        line for line in qasm_code.split("\n") if line and line[:2] != "//" and line[1:3] != "//"
+    ).split(";") if line and "//" not in line)
```
natestemen, line 390: I think this would be more readable if it was written as regular for loop. It's quite hard to understand what it's doing (and if all of the checks are necessary) as it stands.

mitiq/interface/mitiq_qibo/conversions.py
```diff
@@ -0,0 +1,531 @@
...
+    """
+    
+    try:  
+        qasm = cirq_to_qasm(circuit)
+        nqubits, gate_list = _parse_qasm_modified(qasm)
+        
+    except ValueError: 
+        decomp_circuit = Circuit(decompose(circuit)) 
+        qasm = cirq_to_qasm(decomp_circuit)
+        nqubits, gate_list = _parse_qasm_modified(qasm)
```
natestemen, line 341: Could we use the [`qibo.models.circuit.Circuit.from_qasm`](https://github.com/qiboteam/qibo/blob/7379f86437b448cbd5e93abd8129d0fa1b022ddb/src/qibo/models/circuit.py#L1047)
 function here? It would allow us to remove a lot of the bespoke code you wrote later in the file.


```suggestion
    try:  
        qasm = cirq_to_qasm(circuit)
    except ValueError: 
        decomp_circuit = Circuit(decompose(circuit)) 
        qasm = cirq_to_qasm(decomp_circuit)
    return QiboCircuit.from_qasm(qasm)
```

natestemen (changes requested):
Nice work here Francesc! My biggest question is whether we can use the [`from_qasm`](https://github.com/qiboteam/qibo/blob/7379f86437b448cbd5e93abd8129d0fa1b022ddb/src/qibo/models/circuit.py#L1047) function built into Qibo to cut down on the amount of code in this PR. It will also mean fewer mypy errors!

After addressing some of the comments I made, make sure to run `black mitiq` and `isort mitiq` in the root of the mitiq project directory. This will format the code so we don't get stuck on CI on any formatting issues.

natestemen:
I've gotten all the issues resolved by using `qibo.models.circuit.Circuit.from_qasm` on this commit here: https://github.com/unitaryfund/mitiq/commit/708d7eb9711d206321767270fec7380c4f3154f7. Feel free to [cherry-pick](https://git-scm.com/docs/git-cherry-pick) it onto your branch. If you do, most of the comments I made above won't need to be addressed since they apply to code that is removed in 708d7eb9711d206321767270fec7380c4f3154f7.

mitiq/interface/mitiq_qibo/conversions.py
natestemen: I tested this change and it works!

francescsabater: The issue with using `qibo.models.circuit.Circuit.from_qasm` with Qibo version 0.1.11 is that, in this version, the gates `sx` and` sxdg `are not defined. This causes the benchmark tests, specifically `test_mirror_circuits.py` and `test_mirror_qv_circuits.py`, to fail. This is why I had to include a modified version of the` from_qasm` function from Qibo in the code, which has led to the necessity of adding a significant amount of code.

Misty-W:
@francescsabater [we recently heard](https://github.com/quantumlib/OpenFermion/pull/864#issuecomment-1917715173) that [OpenFermion](https://github.com/quantumlib/OpenFermion) will do a release by the end of the week.

Once the latest version of OpenFermion is out, we can then upgrade it and scipy (which OpenFermion was blocking). With the latest version of scipy, we should be able to use the latest version of Qibo.

mitiq/interface/mitiq_qibo/conversions.py
natestemen: Ahhh okay I was failing to have run the benchmark tests. Sorry about the false start there then! Hopefully we can get the latest openfermion as you've been trying and can go from there.

nathanshammah:
I don't know if it helps, but on [Stack Overflow](https://stackoverflow.com/questions/16584552/how-to-state-in-requirements-txt-a-direct-github-source) there is discussion on it.

dev_requirements.txt
```diff
@@ -30,7 +31,7 @@ nbsphinx==0.9.1
 matplotlib==3.8.1
 pandas==2.1.3
 pyscf==2.5.0; sys_platform != 'win32'
-openfermion==1.6.0; sys_platform != 'win32'
+-e git://github.com/quantumlib/OpenFermion.git
```
Misty-W, line 34: Great news, OpenFermion patch released!
```suggestion
openfermion==1.6.1; sys_platform != 'win32'
```

bdg221:
The build / validate is complaining about a mypy error. This error is coming from code in mitiq/rem/inverse_confusion_matrix.py which has NOT been modified from this PR.

The dependabot PR for moving from OpenFermion 1.6.0 to 1.6.1 shows the same error:
https://github.com/unitaryfund/mitiq/actions/runs/7839567108/job/21392797830?pr=2182

Misty-W:
> The build / validate is complaining about a mypy error. This error is coming from code in mitiq/rem/inverse_confusion_matrix.py which has NOT been modified from this PR.
> 
> The dependabot PR for moving from OpenFermion 1.6.0 to 1.6.1 shows the same error: https://github.com/unitaryfund/mitiq/actions/runs/7839567108/job/21392797830?pr=2182

@francescsabater this issue is fixed on `master`, updating your branch from `master` should fix it.

mitiq/interface/mitiq_qibo/conversions.py
Misty-W: It appears so: https://openqasm.com/language/gates.html .

mitiq/interface/mitiq_qibo/conversions.py
Misty-W: Also fixed by using latest version of qibo, QASM parsing removed.

Misty-W (approved):
LGTM @francescsabater, thank you for staying with it through all the back and forth!

@natestemen I think the changes you requested are addressed but wanted to give you a chance to take a look before merging.

mitiq/interface/mitiq_qibo/conversions.py
```diff
@@ -0,0 +1,531 @@
...
+
+from mitiq.interface.mitiq_qiskit import from_qasm as cirq_from_qasm
+from mitiq.interface.mitiq_qiskit import to_qasm as cirq_to_qasm
+
+
+def crx_decomp(gate: gates.CRX) -> List[Gate]:
+    """Decomposes CRX gate to Cirq known gates.
+
+    Args:
+        qibo_gate: CRX gate to decompose.
```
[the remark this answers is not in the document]
natestemen: Minor thing here, but the argument name should match what is described in the docstring. So both should be `gate` or `qibo_gate`, whichever you decide. Looks like you made this change in the other functions, but missed this one!

mitiq/interface/mitiq_qibo/conversions.py
natestemen: @francescsabater can you comment on why this error here is needed? From the looks of the link Misty provided it looks like multi-controlled gates should be possible. Whether they are actually implemented is another story, but it would be good to know what the situation is.

mitiq/interface/mitiq_qibo/conversions.py
```diff
@@ -0,0 +1,394 @@
...
+            for line in qasm.split("\n")
+            if not line.strip().startswith("//")
+        ]
+    )
+    # Remove problematic spaces in gate arguments
+    qasm = "\n".join([line.replace(", ", ",") for line in qasm.split("\n")])
+    # Remove in line problematic comments
+    lines = qasm.split("\n")
+    clean_lines = []
+    for line in lines:
```
natestemen, line 388: I think we can avoid converting between strings and lists a few times here by keeping things as a list after the first conversion. WDYT? (code not tested and formatting may be off!)

```suggestion
    # Remove problematic comment lines in the qasm code
    qasm = [
        line.replace(", ", ",") # remove spaces in gate arguments
        for line in qasm.split("\n")
        if not line.strip().startswith("//")
    ]
    
    clean_lines = []
    for line in qasm:
```

natestemen (approved):
Awesome work Francesc. Apologies for the slow turnaround time on this, but I think it's best that we got it done the "right" way. Thank you for your patience.

Just left a few minor things/questions, but good to go once those are addressed/answered!

Misty-W:
image
@natestemen @francescsabater

mitiq/interface/mitiq_qibo/conversions.py
francescsabater: The thing with multi-controlled gates is that in the` QiboCircuit.to_qasm()` function from Qibo, if the gate is controlled by a similar error is raised. They do not comment much on why they raise the error though.
https://github.com/qiboteam/qibo/blob/fb55eee2eca731a10fa97a6d2195ee1fa65f5b9a/src/qibo/models/circuit.py#L1160

francescsabater: I've managed to avoid some of the unnecessary back-and-forth conversions between lists and strings, but I had to edit the suggested code a bit since mypy was complaining about the inconsistent type of `qasm`.

Misty-W:
@francescsabater you may merge when you’re ready, or if the button is grayed out, let me know and I’ll do it.
Awesome work, excited to connect Mitiq and Qibo!

nathanshammah:
Great to have this in, thanks for the persistence @francescsabater and all who helped in the review, @Misty-W @natestemen @bdg221.

PR #2116, opened 2023-12-08
2115 pauli twirling callibration of expectation estimation shadow needs continue

### LINKED ISSUE #2115: pauli twirling callibration of expectation_estimation_shadow needs continue
In shadows/classical_postprocessing.py the expectation_estimation_shadow code includes a section that uses pauli_twirling_calibration. If the f_val returned however is None, then the value, 0.0, is added to the means List. However, the code will continue to execute and a second (invalid) value, np.sum(product) / len(idxes), is added to the means List. Only a single mean should be added to the means List object per split (each loop of the for loop.) 

for idxes in group_idxes:
        if len(
            np.nonzero(np.all(u_lists_shadow[idxes] == target_obs, axis=1))[0]
        ):
            product = (-1) ** np.sum(
                b_lists_shadow[
                np.nonzero(
                    np.all(u_lists_shadow[idxes] == target_obs, axis=1)
                )
                ].astype(int),
                axis=1,
            )

            if pauli_twirling_calibration:
                if f_est is None:
                    raise ValueError(
                        "estimation of Pauli fidelity must be provided for"
                        "Pauli twirling calibration."
                    )

                b = create_string(num_qubits, target_locs)
                f_val = f_est.get(b, None)
                if f_val is None:
                    means.append(0.0)
                else:
                    # product becomes an array of snapshots expectation values
                    # witch satisfy condition (1) and (2)
                    product = (1.0 / f_val) * product
            else:
                product = 3 ** (len(target_locs)) * product

            # append the mean of the product in each split
            means.append(np.sum(product) / len(idxes))
        else:
            means.append(0.0)

The solution would be to add "continue" after "means.append(0.0) to have the code properly move on to the next split.

bdg221:
fixes #2115 

Added `continue` to the case where f_val from the pauli twirling calibration is None.

 This now prevents an additional mean value being incorrectly included for a particular split.

### License


mitiq/shadows/classical_postprocessing.py
```diff
@@ -368,6 +368,7 @@ def expectation_estimation_shadow(
                 f_val = f_est.get(b, None)
                 if f_val is None:
                     means.append(0.0)
+                    continue
```
andreamari, line 371:
```suggestion
                    product = 0.0
```
Is this equivalent? If so, I would suggest using `product = 0.0` since it makes the workflow more linear and easier to debug in the future. 

If my suggestion doesn't work, I am fine with `continue`.

andreamari (approved):

mitiq/shadows/classical_postprocessing.py
bdg221: That makes sense. Should a similar thing be done with the "else" on line 374 (when there aren't any matches between the measurement_outcome and the pauli_str expected outcomes.) That else would also include a product = 0.0 and at the end of the for loop do a single means.append(np.sum(product) / len(idxes))?

andreamari (approved):
Thanks @bdg221 !

PR #2118, opened 2023-12-08
Add randomized clifford+T benchmarking circuits.

### LINKED ISSUE #1230: Add a function to generate random Clifford-plus-T circuits.
Pre-Request Checklist
---------------------

Issue Description
-----------------

Currently we have several functions that crate benchmark circuits, i.e., circuits that can be used to test error mitigation techniques.
The existing functions are:
```
generate_rb_circuits
generate_mirror_circuit
generate_ghz_circuit
```

Feature Request
-----------------------

All current benchmarks are based on Clifford circuits. It would be nice to have a new function called e.g.:
`generate_random_clifford_t_circuit()`. 
The output should be a random non-Clifford circuit composed of Clifford gates (H, S, CNOT or CZ), and T gates.  

Note: this issue is very similar to #1229 .   The main difference is the specific gate set and the possibility of controlling the number
of T gates.

Nice-to-have features
-----------------------------

- It would be nice to have the possibility of setting the following numbers:
     - The number of single-qubit clifford gates.
     - The number of two-qubit Clifford gates (e.g. how many CNOTs)
     - The number of T gates (typically called t-count). 
     
If having all the three options is not practical, at least we should have the possibility of setting the number of T gates.

Optional feature
---------------------
- It would be great (but not crucial) if the circuits generate states with a sufficient level of randomness. E.g. if they can form a 2-design. - Optionally, we could return the list of heavy bitstrings as discussed for the generic random circuit function proposed in #1229.

FarLab:
## Description

This PR adds the feature requested here https://github.com/unitaryfund/mitiq/issues/1230.

closes #1230 

---

### License

### DIFF
--- a/docs/source/apidoc.md
+++ b/docs/source/apidoc.md
@@ -12,24 +12,28 @@
 ### Classical Shadows
 
 #### Classical Shadows (High-Level Tools)
+
 ```{eval-rst}
 .. automodule:: mitiq.shadows.shadows
    :members:
 ```
 
 #### Quantum Processing
+
 ```{eval-rst}
 .. automodule:: mitiq.shadows.quantum_processing
    :members:
 ```
 
 #### Classical Post-Processing
+
 ```{eval-rst}
 .. automodule:: mitiq.shadows.classical_postprocessing
    :members:
 ```
 
 #### Utility Functions
+
 ```{eval-rst}
 .. automodule:: mitiq.shadows.shadows_utils
    :members:
@@ -38,18 +42,21 @@
 ### Clifford Data Regression
 
 #### Clifford Data Regression (High-Level Tools)
+
 ```{eval-rst}
 .. automodule:: mitiq.cdr.cdr
    :members:
 ```
 
 #### Clifford Training Data
+
 ```{eval-rst}
 .. automodule:: mitiq.cdr.clifford_training_data
    :members:
 ```
 
 #### Data Regression
+
 ```{eval-rst}
 .. automodule:: mitiq.cdr.data_regression
    :members:
@@ -60,25 +67,28 @@ See Ref. {cite}`Czarnik_2021_Quantum` for more details on these methods.
 ### Digital Dynamical Decoupling
 
 #### Digital Dynamical Decoupling (High-Level Tools)
+
 ```{eval-rst}
 .. automodule:: mitiq.ddd.ddd
    :members:
 ```
 
 #### Insertion
+
 ```{eval-rst}
 .. automodule:: mitiq.ddd.insertion
    :members:
 ```
 
 #### Rules
+
 ```{eval-rst}
 .. automodule:: mitiq.ddd.rules.rules
    :members:
 ```
 
-
 ### Pauli Twirling
+
 ```{eval-rst}
 .. automodule:: mitiq.pt.pt
    :members:
@@ -87,12 +97,14 @@ See Ref. {cite}`Czarnik_2021_Quantum` for more details on these methods.
 ### Probabilistic Error Cancellation
 
 #### Probabilistic Error Cancellation (High-Level Tools)
+
 ```{eval-rst}
 .. automodule:: mitiq.pec.pec
    :members:
 ```
 
 #### Quasi-Probability Representations
+
 ```{eval-rst}
 .. automodule:: mitiq.pec.representations.optimal
    :members:
@@ -105,6 +117,7 @@ See Ref. {cite}`Czarnik_2021_Quantum` for more details on these methods.
 ```
 
 #### Learning-based PEC
+
 ```{eval-rst}
 .. automodule:: mitiq.pec.representations.biased_noise
    :members:
@@ -114,37 +127,44 @@ See Ref. {cite}`Czarnik_2021_Quantum` for more details on these methods.
 ```
 
 #### Sampling from a Noisy Decomposition of an Ideal Operation
+
 ```{eval-rst}
 .. automodule:: mitiq.pec.sampling
    :members:
 ```
 
 #### Probabilistic Error Cancellation Types
+
 ```{eval-rst}
 .. automodule:: mitiq.pec.types.types
    :members:
 ```
 
 #### Utilities for Quantum Channels
+
 ```{eval-rst}
 .. automodule:: mitiq.pec.channels
    :members:
 ```
 
 ### Quantum Subspace Expansion
+
 ```{eval-rst}
 .. automodule:: mitiq.qse.qse
    :members:
 ```
+
 ### Readout-Error Mitigation
 
 #### Postselection
+
 ```{eval-rst}
 .. automodule:: mitiq.rem.post_select
    :members:
 ```
 
 #### REM Technique
+
 ```{eval-rst}
 .. automodule:: mitiq.rem.rem
    :members:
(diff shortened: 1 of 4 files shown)


andreamari:
The function `compute_eavy_bitstrings` is already in `quantum_volume_circuits.py`. So, if necessary, we can use that function instead of re-defining it in the new file. 

Moreover, even if it was originally suggested in the issue, now I wonder if we should simply return the circuit without the associated heavy bitstrings. Heavy bitstrings are important for quantum volume circuits and I am not sure if the are important in this case. Moreover the user can always call the function `compute_heavy_bitstrings()` by themselves, if necessary.

So, I would suggest to simply don't consider heavy bitstrings in this new benchmark. 

But please wait for a second opinion by @natestemen about this. I don't want to make you remove some code that other people think it is instead useful. 

Further unrelated comment: please add the new function in the API docs in this file `https://github.com/unitaryfund/mitiq/blob/master/docs/source/apidoc.md` .  You can look for the **Benchmarks** section in that file and see how other similar functions are listed.

FarLab:
Thanks Andrea for your comments. I will probably remove the heavy bitstring function since it already exists. I will also add the function in the API docs.

mitiq/benchmarks/tests/test_randomized_clifford_t_circuit.py
```diff
@@ -0,0 +1,60 @@
...
+
+import pytest
+
+from mitiq import SUPPORTED_PROGRAM_TYPES
+from mitiq.benchmarks.randomized_clifford_t_circuit import (
+    generate_random_clifford_t_circuit,
+)
+
+
+def test_generate_model_circuit_with_seed():
```
natestemen, line 22: I think this test name is copied from another file, and I know we use it in other places, but I think we can make it a little more descriptive of what the actual intention is. Maybe something like `test_seed_circuit_equality`? I don't love that name either, but I'm sure we can come up something better than "model circuit".

mitiq/benchmarks/tests/test_randomized_clifford_t_circuit.py
```diff
@@ -0,0 +1,60 @@
...
+# This source code is licensed under the GPL license (v3) found in the
+# LICENSE file in the root directory of this source tree.
+
+"""Tests for quantum volume circuits. The Cirq functions that do the main work
+are tested here:
+cirq-core/cirq/contrib/quantum_volume/quantum_volume_test.py
+
+Tests below check that generate_quantum_volume_circuit() works as a wrapper and
+fits with Mitiq's interface.
+"""
```
natestemen, line 12: comment needs removal, or updating.

mitiq/benchmarks/tests/test_randomized_clifford_t_circuit.py
```diff
@@ -0,0 +1,60 @@
...
+
+@pytest.mark.parametrize("return_type", SUPPORTED_PROGRAM_TYPES.keys())
+def test_conversion(return_type):
+    circuit = generate_random_clifford_t_circuit(
+        num_qubits=2,
+        num_oneq_cliffords=2,
+        num_twoq_cliffords=2,
+        num_t_gates=2,
+        return_type=return_type,
+        seed=3,
```
natestemen, line 58: Let's remove this line just so it's running the test with random circuit each time.
```suggestion
```

natestemen (changes requested):
Really nice work here Farrokh! A great first addition to Mitiq.

I've left a few small comments, but my biggest is a small concern that in order to appease mypy, we've made the code quite a bit more complicated by shifting everything to manipulating indices. I feel like a simpler approach would be to cast all the objects that get passed to `np.random.choice` as `np.ndarray`s. WDYT?

mitiq/benchmarks/randomized_clifford_t_circuit.py
```diff
@@ -0,0 +1,86 @@
...
+        num_t_gates: Number of T gates to be used.
+        seed: Seed for generating random circuit.
+        return_type: String which specifies the type of the returned
+            circuits. See the keys of ``mitiq.SUPPORTED_PROGRAM_TYPES``
+            for options. If ``None``, the returned circuits have type
+            ``cirq.Circuit``.
+
+    Returns:
+        A quantum circuit acting on ``num_qubits`` qubits.
+        A list of the heavy bitstrings for the returned circuit.
```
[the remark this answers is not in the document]
FarLab: good catch, will remove!

mitiq/benchmarks/randomized_clifford_t_circuit.py
```diff
@@ -0,0 +1,86 @@
...
+    rnd_state = np.random.RandomState(seed)
+
+    oneq_cliffords = [cirq.S, cirq.H]
+    twoq_cliffords = [cirq.CNOT, cirq.CZ]
+    oneq_idx_list = [
+        rnd_state.choice(list(range(len(oneq_cliffords)))) for _ in range(3)
+    ]
+    twoq_idx_list = [
+        rnd_state.choice(list(range(len(twoq_cliffords)))) for _ in range(3)
+    ]
```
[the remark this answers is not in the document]
FarLab: yes, you are right, that should not be 3 (which was probably hardcoded for some reason)

mitiq/benchmarks/tests/test_randomized_clifford_t_circuit.py
FarLab: done

mitiq/benchmarks/tests/test_randomized_clifford_t_circuit.py
FarLab: test_seed_circuit_equality sounds good to me. we can change it if we find a better name :)

mitiq/benchmarks/tests/test_randomized_clifford_t_circuit.py
FarLab: agree

mitiq/benchmarks/randomized_clifford_t_circuit.py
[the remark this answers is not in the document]
FarLab: I like this solution more that casting the objects that go into np.random.choice as np.ndarray. also, I think the 2 should be len(oneq_cliffords) and len(twoq_cliffords), right?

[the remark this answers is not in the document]
natestemen: That's right!

mitiq/benchmarks/randomized_clifford_t_circuit.py
```diff
@@ -0,0 +1,82 @@
+# Copyright (C) Unitary Fund
+#
+# This source code is licensed under the GPL license (v3) found in the
+# LICENSE file in the root directory of this source tree.
+
+"""Functions for generating rotated randomized benchmarking circuits."""
```
natestemen, line 6: Needs update.

natestemen (approved):
One small comment needs updating, and we'll need to add a line into [`mitiq/benchmarks/__init__.py`](https://github.com/unitaryfund/mitiq/blob/master/mitiq/benchmarks/__init__.py) to make these functions importable like
```py
from mitiq.benchmarks import generate_random_clifford_t_circuit
```


Once those two things are done, we are good to go!

mitiq/benchmarks/randomized_clifford_t_circuit.py
FarLab: good catch, I will update these two last points now

mitiq/benchmarks/__init__.py
```diff
@@ -14,3 +14,4 @@
 )
 from mitiq.benchmarks.w_state_circuits import generate_w_circuit
 from mitiq.benchmarks.qpe_circuits import generate_qpe_circuit
+from mitiq.benchmarks import generate_random_clifford_t_circuit
```
natestemen, line 17: We need to add the full path here so that when importing this from `mitiq.benchmarks` is possible. I think it's going to look something like this (there might be a formatting change, I'm not 100% sure).
```suggestion
from mitiq.benchmarks.randomized_clifford_t_circuit import (
    generate_random_clifford_t_circuit,
)
```

docs/source/apidoc.md
```diff
@@ -241,8 +267,15 @@ See Ref. {cite}`Czarnik_2021_Quantum` for more details on these methods.
    :members:
 ```
 
+#### Randomized Clifford+T benchmarking Circuits
```
natestemen, line 270:
```suggestion
#### Randomized Clifford+T Circuits
```

mitiq/benchmarks/__init__.py
FarLab: thanks, als checked the formatting and it didn't complain :)

natestemen (approved):
Nice work Farrokh! Can't wait to do some more testing with these circuits!

🚢

Issue #2080, opened 2023-10-30
Speed up Mitiq tests (3rd iteration)


Misty-W:
This is a continuation of the progress made on speeding up Mitiq's tests in milestone 0.31. 
previous issues: https://github.com/unitaryfund/mitiq/issues/1427, https://github.com/unitaryfund/mitiq/issues/1998

previous PRs speeding up tests: https://github.com/unitaryfund/mitiq/pull/1996, https://github.com/unitaryfund/mitiq/pull/2047

## Proposed Solution

Remove redundant tests, and look for other opportunities to reduce test time.

PR #2126, opened 2023-12-14
Speed up tests using mocks

natestemen:
## Description

Speed up the tests round three. This time mostly using the mocking approach in order to cut down on the expensive computations that are performed. I also cleaned up some of the adjacent code while trying to understand the QSE tests.

The following table is the runtime of `make test` before and after the changes included in this PR.

| before | after |
| ------ | ----- |
| 5:31   | 3:39  |

fixes https://github.com/unitaryfund/mitiq/issues/2080

### DIFF
--- a/mitiq/qse/qse_utils.py
+++ b/mitiq/qse/qse_utils.py
@@ -74,15 +74,14 @@ def get_expectation_value_for_observable(
         )[0]
         return (pauli_expectation_cache[cache_key] * pauli_string.coeff).real
 
-    expectation_value = 0.0
-    if isinstance(observable, PauliString):
-        pauli_string = observable
-        expectation_value += get_expectation_value_for_one_pauli(pauli_string)
-    elif isinstance(observable, Observable):
-        for pauli_string in observable.paulis:
-            expectation_value += get_expectation_value_for_one_pauli(
-                pauli_string
-            )
+    paulis = (
+        [observable]
+        if isinstance(observable, PauliString)
+        else observable.paulis
+    )
+    expectation_value = sum(
+        get_expectation_value_for_one_pauli(pauli) for pauli in paulis
+    )
     return expectation_value
 
 
--- a/mitiq/qse/qse_utils.py
+++ b/mitiq/qse/qse_utils.py
@@ -5,7 +5,8 @@
 
 """Functions for computing the projector for subspace expansion."""
 
-from typing import Callable, Dict, List, Sequence, Union
+from typing import Callable, Dict, Sequence, Union, Optional
+from itertools import product
 
 import numpy as np
 import numpy.typing as npt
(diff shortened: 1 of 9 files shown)


mitiq/qse/qse_utils.py
```diff
@@ -53,82 +54,58 @@ def get_expectation_value_for_observable(
...
+    return expectation_value
 
 
 def _compute_overlap_matrix(
     circuit: QPROGRAM,
     executor: Union[Executor, Callable[[QPROGRAM], QuantumResult]],
     check_operators: Sequence[PauliString],
-    pauli_string_to_expectation_cache: Dict[PauliString, complex] = {},
+    pauli_expectation_cache: Dict[PauliString, complex] = {},
+    code_hamiltonian: Optional[Observable] = None,
```
andreamari, line 94: Is this new argument here in order to get rid of `_compute_hamiltonian_overlap_matrix` ?
Seems good, just trying to understand.

mitiq/qse/tests/test_qse.py
```diff
@@ -109,44 +109,27 @@ def test_mitigate_executor_batched():
...
+        observable,
+        {},
     )
-    def decorated_executor(circuit: QPROGRAM) -> np.ndarray:
-        return compute_density_matrix(
-            convert_to_mitiq(circuit)[0],
-            noise_model_function=cirq.depolarize,
-            noise_level=(0.01,),
-        )
+    assert mock_execute_with_qse.call_count == num_circuits
```
andreamari, line 125: Do you know how `mock_execute_with_qse.call_count` "knows" that it should return `num_circuits`?
All seems to work. I am just curious how `unittest.mock.patch` is able to mock functions without explicitly setting return values.

andreamari (approved):
Good speed improvement! 
To me it seems that mocking is used in a way which is not dangerous (e.g. by implicitly testing less real code). For LGTM

mitiq/qse/qse_utils.py
natestemen: Yup that's exactly right! We previously had

1. `_compute_hamiltonian_overlap_matrix`, and
2. `_compute_overlap_matrix`

which were largely the same code, except for one line. I combined the two functions in order to cut down on repeated code.

mitiq/qse/tests/test_qse.py
natestemen: I was also surprised you don't have to set a return value, but it's nice not having to set some random value for it since it's meaningless (in this instance).

As for how it actually works, [`unittest.mock.patch`](https://docs.python.org/3/library/unittest.mock.html#unittest.mock.patch) creates a [`MagicMock`](https://docs.python.org/3/library/unittest.mock.html#unittest.mock.MagicMock) instance for the `execute_with_qse` function. Under the hood my guess is that the `MagicMock` class looks something like this:

```py
class MagicMock:
    def __init__(self):
        self.call_count = 0
    
    def __call__(self):
        self.call_count += 1
```

PR #2128, opened 2023-12-15
Compute super operator gate by gate in find_optimal_representation

andreamari:
This is a small update to the `find_optimal_representation` function of the PEC module.

It should be equivalent to the current code (if all tests pass), but it is going to help if we want to support Reset gates, since in that case cirq.kraus(circuit) would give an error, but computing it gate-by-gate works.

### License

### DIFF
--- a/mitiq/pec/representations/optimal.py
+++ b/mitiq/pec/representations/optimal.py
@@ -5,6 +5,7 @@
 
 """Functions for finding optimal representations given a noisy basis."""
 
+import functools
 from typing import List, Optional, cast
 
 import numpy as np
@@ -124,9 +125,15 @@ def find_optimal_representation(
     Returns: The optimal OperationRepresentation.
     """
     ideal_cirq_circuit, _ = convert_to_mitiq(ideal_operation)
-    ideal_matrix = kraus_to_super(
-        cast(List[npt.NDArray[np.complex64]], kraus(ideal_cirq_circuit))
-    )
+
+    ops = list(ideal_cirq_circuit.all_operations())
+    super_ops = [
+        kraus_to_super(cast(List[npt.NDArray[np.complex64]], kraus(op)))
+        for op in ops
+    ]
+
+    # Compute super operator of the full ideal_circuit
+    ideal_matrix = functools.reduce(lambda a, b: a @ b, super_ops)
 
     try:
         basis_matrices = [


mitiq/pec/representations/optimal.py
```diff
@@ -124,9 +125,15 @@ def find_optimal_representation(
...
     ideal_cirq_circuit, _ = convert_to_mitiq(ideal_operation)
-    ideal_matrix = kraus_to_super(
-        cast(List[npt.NDArray[np.complex64]], kraus(ideal_cirq_circuit))
-    )
+
+    ops = list(ideal_cirq_circuit.all_operations())
+    super_ops = [
+        kraus_to_super(cast(List[npt.NDArray[np.complex64]], kraus(op)))
+        for op in ops
+    ]
```
natestemen, line 133: Two suggestions put into one:

1. No need to cast `all_operations()` to a list since we are just going to iterate over it.
2. `super_ops` does not need to be a list either since we are going to iterate over it.
```suggestion
    super_ops = (
        kraus_to_super(cast(List[npt.NDArray[np.complex64]], kraus(op)))
        for op in ideal_cirq_circuit.all_operations()
    )
```

natestemen (approved):
Nice little change! Can you give a little more detail as to why this works with reset gates and the previous implementation did not?

mitiq/pec/representations/optimal.py
natestemen: Oh I also just found [`cirq.operation_to_superoperator`](https://quantumai.google/reference/python/cirq/operation_to_superoperator). It looks like under the hood it does all the Kraus stuff for us.

andreamari:
> Nice little change! Can you give a little more detail as to why this works with reset gates and the previous implementation did not?

Because for some cirq-related reason `cirq.kraus(circuit_with_a_reset_gate)` fails to return the kraus operators of the full circuit. The error is something like "kraus method not implemented for the circuit". On the other hand  `cirq.kraus(reset_operation)` works! So by computing super-operators operation-by-operation and taking their product we avoid the problem.

mitiq/pec/representations/optimal.py
andreamari: Nice! I'll try that next week and maybe add some tests with a reset gate. I don't think this PR is essential for the release and we can proably keep it out. But if you want, you can merge it in the current form and I'll make the updates in a second step.

natestemen:
> Because for some cirq-related reason cirq.kraus(circuit_with_a_reset_gate) fails to return the kraus operators of the full circuit. The error is something like "kraus method not implemented for the circuit". On the other hand cirq.kraus(reset_operation) works! So by computing super-operators operation-by-operation and taking their product we avoid the problem.

Huh, that kind of seems like a bug to be honest!

PR #2136, opened 2023-12-20
Revert recent commits on classical shadows to fix tutorials in documentation

andreamari:
This PR, reverts #2115 and #2113, to recover the original code that used to generate good results in the docs examples.

It also manually re-apply the small changes of #2115, since to me they seem are correct.
So, overall, this PR corresponds to an effective revert of just #2113 .

We can always re-implement something similar to #2113, after we figure out the hidden problem with that.

Meanwhile, reverting to the original code looks like a pragmatic solution.


bdg221:
PR #2137 should resolve the issue without requiring backing out of the other PRs.

andreamari:
Closing since the problem is now directly fixed by #2137.

PR #2137, opened 2023-12-20
Changed to use the split for b_list_shadow

bdg221:
This corrects an issue from PR #2113.

In particular, the issue was that indexes that are relative to a split were used against the full b_lists_shadow object. 

The correction changes it so that the indexes that are relative to a split are used against that split.

After making changes, values through the `expectation_estimation_shadow` function were compared between the previous code and the new code.
image

Also, the results in the final plot are now matching the previous results:
new code:
image

v0.31.0 code:
image

### License

### DIFF
--- a/mitiq/shadows/classical_postprocessing.py
+++ b/mitiq/shadows/classical_postprocessing.py
@@ -343,7 +343,7 @@ def expectation_estimation_shadow(
             np.all(u_lists_shadow[idxes] == target_obs, axis=1)
         )
 
-        if len(matchingIndexes[0]):
+        if len(matching_indexes[0]):
             product = (-1) ** np.sum(
                 b_lists_shadow[idxes][matching_indexes].astype(int),
                 axis=1,
--- a/mitiq/shadows/classical_postprocessing.py
+++ b/mitiq/shadows/classical_postprocessing.py
@@ -343,7 +343,7 @@ def expectation_estimation_shadow(
             np.nonzero(np.all(u_lists_shadow[idxes] == target_obs, axis=1))[0]
         ):
             product = (-1) ** np.sum(
-                b_lists_shadow[
+                b_lists_shadow[idxes][
                     np.nonzero(
                         np.all(u_lists_shadow[idxes] == target_obs, axis=1)
                     )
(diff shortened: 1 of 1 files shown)


bdg221:
@andreamari @Min-Li This should correct the issue as confirmed by the matching values, as well as the matching plots.

mitiq/shadows/classical_postprocessing.py
```diff
@@ -343,7 +343,7 @@ def expectation_estimation_shadow(
             np.nonzero(np.all(u_lists_shadow[idxes] == target_obs, axis=1))[0]
```
andreamari, line 343: Could you assign `np.nonzero(np.all(u_lists_shadow[idxes] == target_obs, axis=1)` to a new variable? 
I think it would make the code easier to understand, especially since the same long expression is used twice.

andreamari (approved):
Great! Thanks for finding out the issue!
I added a small optional suggestion.

mitiq/shadows/classical_postprocessing.py
```diff
@@ -339,15 +339,13 @@ def expectation_estimation_shadow(
 
     # loop over the splits of the shadow:
     for idxes in group_idxes:
-        if len(
-            np.nonzero(np.all(u_lists_shadow[idxes] == target_obs, axis=1))[0]
-        ):
+        matchingIndexes = np.nonzero(
```
andreamari, line 342:
```suggestion
        matching_indexes = np.nonzero(
```

mitiq/shadows/classical_postprocessing.py
```diff
@@ -339,15 +339,13 @@ def expectation_estimation_shadow(
...
+        )
+
+        if len(matchingIndexes[0]):
             product = (-1) ** np.sum(
-                b_lists_shadow[
-                    np.nonzero(
-                        np.all(u_lists_shadow[idxes] == target_obs, axis=1)
-                    )
-                ].astype(int),
+                b_lists_shadow[idxes][matchingIndexes].astype(int),
```
andreamari, line 348:
```suggestion
                b_lists_shadow[idxes][matching_indexes].astype(int),
```

andreamari (approved):

natestemen:
Thanks Brian for the quick fix!

PR #2148, opened 2024-01-05
Attempt 3: Pauli Transfer Matrix

purva-thakre:
## Description

Related to #1914

Previous attempts to add this function were not working as expected. I initially thought this was due to how the matrix to vector function worked or how cirq calculated the superoperator of a noise channel. I was wrong. 

I think this is due to how we expect a Pauli Twirled CNOT gate's noise to be tailored to reduce the noise effects. The discussion below will show that the coherent noise effects do not disappear after twirling. 

## Example CNOT gate circuit

I have manually checked the calculations and the code is correctly calculating the PTM matrix for some gates and error channels. 

I did also double check the results with those published in https://arxiv.org/abs/1905.00450 for a CNOT gate.

Published result             |  What I get
:-------------------------:|:-------------------------:
image  |  

In the small error limit, the PTM of a noisy CNOT circuit is still the PTM of the ideal CNOT gate. This does not change if I Pauli twirl the circuit and average over all the PTMs to find the twirled PTM. The circuit I considered is shown below. 

image

We expect the off-diagonal terms to go to zero in the small error limit. They are still unchanged in the small error limit after twirling.

   before twirling       |  after twirling (averaged over 1000 twirled circuits)
:-------------------------:|:-------------------------:
image |  image

If I remove the small error limit consideration, the PTM does change compared to the ideal PTM but the PTM before and after twirling of the noisy input circuit are once again identical.

image

   before twirling       |  after twirling (averaged over 1000 twirled circuits)
:-------------------------:|:-------------------------:
image  |  image

## Possible Solution 

All of the above does make me think that the current Pauli twirling technique does not tailor noise as expected. The coherent errors are still present after twirling i.e. there's no advantage over Pauli twirling only the CNOT and CZ gates. Thus, using a PTM to show the noise was tailored also fails. 

#### Previous WIP: Other tests to add
Directly taken from Sec 2.1.2 and Sec 2.2 of https://arxiv.org/abs/1509.02921:
- [ ] all entries of the PTM must also be real and in the interval [-1,1]
- [ ] the PTM of a composite map is simply the matrix product of the individual PTMs
- [ ] the first row of the PTM is one and all zeros
- [ ]  the PTM for a unital map must have one and all zeros in its first column

---


mitiq/pt/pt_utils.py
```diff
@@ -0,0 +1,85 @@
...
+
+
+def ptm_matrix(circuit: Circuit, num_qubits: int) -> npt.NDArray[np.complex64]:
+    """Find the Pauli Transfer Matrix (PTM) of a circuit."""
+    superop = cirq.kraus_to_superoperator(cirq.kraus(circuit))
+    vec_pauli = _pauli_vectorized_list(num_qubits)
+    n_qubit_paulis = _n_qubit_paulis(num_qubits)
+    ptm_matrix = np.zeros([4**num_qubits, 4**num_qubits], dtype=complex)
+
+    for i in range(len(vec_pauli)):
```
purva-thakre, line 66: Use `enumerate` instead of `range(len(list))`

https://docs.python.org/3/library/functions.html#enumerate

nathanshammah:
Great @purva-thakre, could you please share a notebook with the code for before/after pt?

purva-thakre:
@nathanshammah Here's a pdf of the notebook and a link to the `ipnyb` in Google drive. 

https://drive.google.com/file/d/1sO_KVsSn5ePqXquVyHzXeXYivLRbH_rY/view?usp=drive_link

The pdf version of the jupyter notebook is weirdly formatted (FYI)
[mitiq_ptm.pdf](https://github.com/unitaryfund/mitiq/files/14196931/mitiq_ptm.pdf)

I think I have a possible solution at the end of the notebook.

purva-thakre:
Follow up to my previous comment: it appears that our choice of layers is indeed causing issues with pauli twirling not working correctly. **I spun up a quick example** based on the notebook example by adding randomly chosen Paulis before the coherent error channel. **We can now observe coherent noise being tailored via pauli twirling. 🎆**

FWIW I am a bit confused on how to choose the additional layers. We might need to agree upon a design first before implementing changes to the pauli twirling module.

I will add the discussion in this comment to the jupyter notebook later. Apologies for the very long comment!

------------------------------------------------------------------------------------------------------------------------------------------------------------------------- 

PTM of noisy circuit             |  PTM of noisy twirled circuit averaged over 100 circuits
:-------------------------:|:-------------------------:
image  |  image

We can now observe the off-diagonal terms reduce when the PTM is calculated over a varied number of Pauli twirled circuits.

PTM of noisy circuit  averaged over 10 circuits           |  PTM of noisy twirled circuit averaged over 100 circuits
:-------------------------:|:-------------------------:
image  |  image

#### Ideal Circuit

```python
import cirq

q0 = cirq.LineQubit(0)
q1 = cirq.LineQubit(1)

circuit = cirq.Circuit()
circuit.append(cirq.CNOT(q0, q1))
circuit
```
image

#### Noisy Circuit
```python
noisy_circuit = circuit + cirq.Circuit([cirq.Rx(rads=(1/12) * np.pi)(q0), cirq.Rx(rads=(1/12) * np.pi)(q1)])
noisy_circuit
```
image

#### PTM of the noisy circuit
```python
import seaborn as sns
import matplotlib.pylab as plt
from mitiq.pt.pt_utils import  _pauli_vectorized_list, _n_qubit_paulis, ptm_matrix

ptmcnot = ptm_matrix(noisy_circuit, 2)
ax = sns.heatmap(ptmcnot.real, linewidth=0.5)
plt.show()
```
See the first column of the first table for the output of this code cell.

#### Create a circuit with an additional layer

Get the pauli twirled versions of the ideal circuit.

```python
from mitiq import pt

twirl_CNOT = pt.pauli_twirl_circuit(circuit=circuit, num_circuits = 10)
twirl_CNOT[0]
```
image

Splice the twirled circuits, insert the chosen error channel and a randomly chosen pauli layer.

```python
import random

# add an extra twirl layer after the twirling operations have been applied to the CNOT
twirling_gates = [cirq.I, cirq.X, cirq.Y, cirq.Z]
twirl_CNOT_additional_twirl_layer = []
for i in range(len(twirl_CNOT)):
    twirl_CNOT_i = twirl_CNOT[i]
    gate_q0 = random.choice(twirling_gates).on(q0)
    gate_q1 = random.choice(twirling_gates).on(q1)
    gate_q2 = random.choice(twirling_gates).on(q0)
    gate_q3 = random.choice(twirling_gates).on(q1)
    # do we need the first layer?
    # first layer of randomly chosen twirling operations before the error channel
    # new_circuit_i = twirl_CNOT_i + cirq.Circuit([gate_q0, gate_q1])
    # add the error channel - if we use the first layer, replace twirl_CNOT_i below with new_circuit_i
    new_circuit_i = twirl_CNOT_i + cirq.Circuit([cirq.Rx(rads=(1/12) * np.pi)(q0), cirq.Rx(rads=(1/12) * np.pi)(q1)])
    # second layer of randomly chosen twirling operations after the error channel
    new_circuit_i = new_circuit_i + cirq.Circuit([gate_q2, gate_q3])
    twirl_CNOT_additional_twirl_layer.append(new_circuit_i)

twirl_CNOT_additional_twirl_layer[0]
```
image

#### Averaged over PTM of the new version of Pauli twirled circuits

```python
twirled_ptm = np.zeros(16)

# average over all the PTMs
for i in range(len(twirl_CNOT_additional_twirl_layer)):
    ptmi = ptm_matrix(twirl_CNOT_additional_twirl_layer[i], 2)
    twirled_ptm = twirled_ptm + ptmi

twirled_ptm2 = (1/len(twirl_CNOT_additional_twirl_layer)) * twirled_ptm

ax = sns.heatmap(twirled_ptm2.real, linewidth=0.5)
plt.show()
```
See the first column of the second table for the output of this code cell.

purva-thakre:
Closing this for now as we decided to define a function where

- the input circuit is twirled first
- we ask the user for which type of noise they want to add to their CNOT/CZ gates
- use a find and replace loop to insert noisy layers after the CNOT and in between the twirled layers. 

Once this is added, then the functions defined in this PR will become useful to show the transformation from a coherent channel to an incoherent channel.

Issue #2129, opened 2023-12-18
Make changes to the functions in `mitiq.shadows`


purva-thakre:
This issue will be used to keep track of all the things that could be changed, removed etc. from `mitiq.shadows`. 

- [ ] A function to calculate `fidelity` is in the utilities. But a similar function also exists in [`cirq.fidelity`](https://quantumai.google/reference/python/cirq/fidelity). 
https://github.com/unitaryfund/mitiq/blob/05be477d2453ce647a32a4df7170426c5ca0b83f/mitiq/shadows/shadows_utils.py#L140

- [ ]

Misty-W:
Partially addressed in 
Refactor classical postprocessing in shadows module #2152 , but more work to be done to close this issue. Remove from 0.34.0.

purva-thakre:
Compared to other techniques, `mitiq.shadows` does not follow the pattern of having a function called `execute_with_shadows`, `shadows.mitigate_executor`, and `shadows.decorator`. 

Whenever we get to the cleanup of this module, we should consider a refactor of the above.

PR #2152, opened 2024-01-09
Refactor classical postprocessing in shadows module

natestemen:
## Description

PR to refactor [`mitiq/shadows/classical_postprocessing.py`](https://github.com/unitaryfund/mitiq/blob/e14a799a90665d12ea004e6342de0c39961a35dc/mitiq/shadows/classical_postprocessing.py).

Addresses some issues raised in #2129 

### Docs

Despite there being tests for the shadows module, as we saw in https://github.com/unitaryfund/mitiq/pull/2113/ the tests do not cover everything and a change can pass tests despite making drastic changes to the results in a tutorial.

| shadow tutorial before | after |
| ---------------------- | ----- |
| Screenshot 2024-01-10 at 4 21 03 PM | Screenshot 2024-01-10 at 4 19 44 PM | 

Note there is a _slight_ difference in the above plot. I'm not totally sure why.

| rshadow tutorial before | after |
| ----------------------- | ----- |
| Screenshot 2024-01-10 at 5 42 45 PM | Screenshot 2024-01-10 at 4 31 58 PM |

All the other plots that I checked look either identical, or very close.

### DIFF
--- a/docs/source/examples/rshadows_tutorial.md
+++ b/docs/source/examples/rshadows_tutorial.md
@@ -249,7 +249,7 @@ plt.xticks(
 )
 
 plt.ylabel("Pauli fidelity")
-plt.legend()
+plt.legend();
 ```
 
 
--- a/docs/source/examples/rshadows_tutorial.md
+++ b/docs/source/examples/rshadows_tutorial.md
@@ -329,14 +329,12 @@ def compare_shadow_methods(
         
     output_shadow = classical_post_processing(
         shadow_outcomes=shadow_measurement_result,
-        use_calibration=False,
         observables=observables,
         k_shadows=k_shadows,
     )
 
     output_shadow_cal = classical_post_processing(
         shadow_outcomes=shadow_measurement_result,
-        use_calibration=True,
         calibration_results=f_est,
         observables=observables,
         k_shadows=k_shadows,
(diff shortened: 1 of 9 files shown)


Misty-W:
@Min-Li, we noticed some slight differences in the plots after the re-factor. Is this a concern, or do you think the module is working fine?

FarLab:
Hi Nate, I had a look at your PR today by going over the commits and it looks good to me! The only concern is the very small difference in the first plots you show, but that might be some randomness that is not fixed.

natestemen:
> only concern is the very small difference in the first plots

Any idea where that change could be coming from? I have not been able to find it.

FarLab:
> > only concern is the very small difference in the first plots
> 
> Any idea where that change could be coming from? I have not been able to find it.

looking at the shadows_tutorial notebook, the seed is fixed, and you didn't change that. So maybe that is not the cause.

FarLab (approved):
Looks good, not much to add besides the insignificant differences in the plot in the tutorial notebook.

Min-Li:
> @Min-Li, we noticed some slight differences in the plots after the re-factor. Is this a concern, or do you think the module is working fine?

Looks good to me, since there are some changes in the code, for me, this slight change of the figure seems to result from adding/removing stages of random sampling (especially in the quantum measurement). But overall, I don't think there are any issues here.

PR #2165, opened 2024-01-26
Reduce doc build time for learning representations

### LINKED ISSUE #2075: Reduce doc build time for Learning Representations
The "PEC with Cirq: Learning representations" example (learning-depolarizing-noise.md) in the documentation takes significant time to build (over 4 minutes.)

In particular, it is the learn_depolarizing_noise_parameter() call that requires the large amount of time to complete. By saving the results of this call to a local file, the time to build this piece of the documentation could be reduced.

Misty-W:
## Description

Fixes #2075 

Switch 

---

### License

### DIFF
--- a/docs/source/examples/learning-depolarizing-noise.md
+++ b/docs/source/examples/learning-depolarizing-noise.md
@@ -163,6 +163,52 @@ Here we use the Nelder-Mead method in [`scipy.optimize.minimize`](https://docs.s
 to find the value of epsilon that minimizes the depolarizing noise loss function.
 
 ```{code-cell} ipython3
+:tags: [remove-cell]
+pec_data = np.loadtxt(
+        os.path.join(
+            "./mitiq/pec/representations/tests/learning_pec_data",
+            f"learning_pec_data_eps_{eps_string}.txt",
+        )
+    )
+
+    [success, epsilon_opt] = learn_depolarizing_noise_parameter(
+        operations_to_learn,
+        circuit,
+        ideal_executor,
+        noisy_executor,
+        num_training_circuits=5,
+        fraction_non_clifford=0.2,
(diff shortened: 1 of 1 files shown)


docs/source/examples/learning-depolarizing-noise.md
```diff
@@ -50,7 +50,7 @@ Here we use a simple Rx-Rz-CNOT circuit, with an (optional) seed for reproducibi
 
 ```{code-cell} ipython3
 circuit = random_x_z_cnot_circuit(
-    LineQubit.range(2), n_moments=10, random_state=np.random.RandomState(1)
+    LineQubit.range(2), n_moments=5, random_state=np.random.RandomState(1)
```
natestemen, line 53: Why is it that this change does not require a change to the data that is loaded in further into the file? Naively, I would have thought if we are changing the circuit, then the PEC data would need to change as well. Does the PEC data consist of representations of the gates, which does not change here?

natestemen (approved):
Nice work! We love fast docs!


The one (nonblocking) question I have is if someone downloads this notebook (as a markdown file) and converts it a python notebook, will it run out of the box, or create an error looking for the PEC data file? Either way, I'm okay to merge. Just something I think we should be aware of/know the answer to.

docs/source/examples/learning-depolarizing-noise.md
Misty-W: The change is to match the circuits used to generate the file. The file was generated separately from this notebook and is already used in `test_learning.py` to make the tests run efficiently.

natestemen: Ahh okay that makes sense! Thanks for the clarification. Can you comment on what's in the data file? (just for the sake of me learning!)

PR #2195, opened 2024-02-19
Add clarification for virtualenv files vs mitiq directory

smburdick:
## Description

Adding a line of clarification to allow make checks to perform inside a virtual environment, specifically that virtualenv files and mitiq must be in different directories. Since the documentation should be beginner friendly, this change is warranted.

---

### License

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -57,3 +57,4 @@ Yash Prabhat
 Vladimir Kozhukalov
 Francesc Sabater
 Emiliano Godinez
+Sam Burdick
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -14,7 +14,7 @@ The rest of this document describes the technical details of getting set up to d
 
 1. Ensure you have python 3.9 or greater installed. If not, you can find the downloads [here](https://www.python.org/downloads/).
 2. Set up a virtual environment to isolate dependencies. This can be done with many different tools including [Virtualenv](https://virtualenv.pypa.io/en/latest/), [Pipenv](https://pypi.org/project/pipenv/), [Poetry](https://python-poetry.org/), and [Anaconda](https://www.anaconda.com/products/distribution). In what follows we will use Anaconda, but if you're familiar with other tools feel free to use those.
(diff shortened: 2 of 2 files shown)


natestemen (commented):
Thanks Sam for the comments here. It's really important that we have clear development documentation.

I'm not totally clear on what you're suggesting, however. Would you mind expanding on your suggestion a little so I can better understand what you're suggesting? Spelling it out with CLI commands can sometimes be helpful.

smburdick:
> Thanks Sam for the comments here. It's really important that we have clear development documentation.
> 
> I'm not totally clear on what you're suggesting, however. Would you mind expanding on your suggestion a little so I can better understand what you're suggesting? Spelling it out with CLI commands can sometimes be helpful.

Hi Nate,

When I went to set up mitiq on my own, I could not run flake due to a max recursion depth error. The root cause turned out to be due to the fact that I set up my virtual environment in the same folder as mitiq. The solution to the problem looks like this:

```
mkdir mitiq-venv && cd mitiq-venv
virtualenv [options]
source bin/activate
cd ..
git clone [mitiq-repo]
cd mitiq-repo # Now I'm ready to run the tests
```

natestemen:
Ahh I see what you mean! Thanks for the extra details. I've actually ran into that before as well since I had a virtual environment in the root of mitiq. Since I started using conda for managing environments this hasn't been a problem, but we should be agnostic to the environment management tool.

That said, I think it actually makes sense to have the virtual environment **inside** the root of the project. That way it's kept track of based on where you are in your file system. I think a potential fix could be to modify the command that is run when `make check-style` is run so that it only runs `flake8` on the `mitiq` directory. What do you think of this plan?

natestemen:
Since https://github.com/unitaryfund/mitiq/pull/2222 is now merged I'm not sure if this behavior still exists. @smburdick would you mind checking if running `make format` or `make check-format` on master (with a virtualenv in the root of the mitiq repo) throws a similar sort of error?

natestemen:
Going to close this now, but if it's still a problem with `ruff` let me know and we can get this fixed!

Issue #2202, opened 2024-02-22
Techniques missing from top level `__init__.py`.


natestemen:
Only a subset of QEM techniques in Mitiq are in the top level `__init__.py` file.
https://github.com/unitaryfund/mitiq/blob/c5afbd4df6850cacc331a1aa1a477265d4bb584a/mitiq/__init__.py#L29

This means to use the newer techniques such as shadows, QSE, and twirling, you must use the full path as demonstrated in the [shadows tutorial](https://mitiq.readthedocs.io/en/stable/examples/shadows_tutorial.html). This also manifested when I was trying to test `mitiq.pt` locally and noticed that when I tried to run `>>> mitiq.pt` I got `AttributeError: module 'mitiq' has no attribute 'pt'`.

We should add all the techniques/modules we have here.

natestemen:
So I'm not entirely sure why I was seeing this error. When taking a look with Misty a day or so after raising this issue, I was not able to reproduce.

We opened #2206 while investigating this as we realized we had some unnecessary imports. Technically, #2206 does not "fix" this issue, as there **isn't** a reproducible issue. Still, good to have redundant code removed so will go ahead with merging that PR. I'm going to close this now just to indicate that technically #2206 does not "fix" this issue.

FarLab:
Thanks for documenting this!

purva-thakre:
Reopening this https://github.com/unitaryfund/mitiq/commit/ae7dad90b659af89d97227a3da37f89e638ae657#r141079046

natestemen:
@purva-thakre can you document a minimally reproducible process for how to recreate this bug?

purva-thakre:
Pinging @jordandsullivan for the workflow on how to recreate this bug as they came across it initially. https://github.com/unitaryfund/mitiq/pull/2298#issuecomment-2060171860

I was trying to understand why it could be so.

jordandsullivan:
This is how you reproduce which modules are throwing this error:
```python
from types import ModuleType
import mitiq 

mod_list = [
    'mitiq.cdr', 
    'mitiq.ddd', 
    'mitiq.pec', 
    'mitiq.pt', 
    'mitiq.qse', 
    'mitiq.rem', 
    'mitiq.zne'
    ]
for module in mod_list:
    try:
        assert isinstance(eval(module), ModuleType)
    except Exception as e:
        print(module, e)
```
which gives the following output:
```python
[...]

PR #2206, opened 2024-02-22
remove redundant imports from top level `__init__.py`

natestemen:
## Description

fixes https://github.com/unitaryfund/mitiq/issues/2202 by removing the lines all together as they are redundant to the `__init__.py` files that live in each of those respective directories.

### DIFF
--- a/mitiq/__init__.py
+++ b/mitiq/__init__.py
@@ -18,15 +18,9 @@ from mitiq.typing import (
 from mitiq.executor import Executor
 from mitiq.observable import PauliString, Observable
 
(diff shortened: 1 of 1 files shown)


FarLab (approved):

natestemen:
Thanks Farrokh! I've added a comment over on the issue to explain what happened here https://github.com/unitaryfund/mitiq/issues/2202#issuecomment-1967384534. Thanks for making sure we have it documented, and thanks for the review!

First PR of milestone 0.35 💪🏻

PR #2212, opened 2024-02-27
Update Qibo demo snippet on v0.34 changelog 

Misty-W:
## Description

Now that Qibo is integrated with Mitiq, we don't need to use `convert_to_mitiq` for Qibo circuits.

---

### License

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,7 +8,6 @@ Qibo is an "end-to-end open source platform for quantum simulation, self-hosted 
 
 ```py
 from qibo import Circuit
-from mitiq.conversions import convert_to_mitiq
 
 circuit = Circuit(2)
 circuit.add(gates.H(0))
@@ -20,7 +19,7 @@ def executor(circuit):
     return circuit.execute()
 
 
-mitigated = mitiq.zne.execute_with_zne(convert_to_mitiq(circuit), executor)
+mitigated = mitiq.zne.execute_with_zne(circuit, executor)
 ```
 
 Thank you to new contributor Francesc Sabater for excellent work integrating Qibo and Mitiq!
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,7 +7,8 @@ Announcing support for [Qibo](https://qibo.science/), a newly integrated fronten
 Qibo is an "end-to-end open source platform for quantum simulation, self-hosted quantum hardware control, calibration and characterization".
 
 ```py
-from qibo import Circuit
+from qibo import Circuit, gates
+import mitiq
 
 circuit = Circuit(2)
 circuit.add(gates.H(0))


natestemen (changes requested):
Thanks for removing this! I'm still getting an error[^1] when trying to run this example:
```
TypeError: unsupported operand type(s) for /: 'QuantumState' and 'int'
```
Is it running on your side? If it's not, it looks like we will need to use a different executor.

[^1]: We need to import `gates` from `qibo` as well as `import mitiq` to make sure it runs, but then after those issues, we still get the type error :/.

Misty-W:
I'm looking at this and realize now there are more problems- the circuit is trivial, no noise applied, and I'm not sure where the executor error is coming from.

cosenal:
Closing this as v0.34 has been released already, and it looks like a working version of the Qibo snippet [made it into the changelog](https://github.com/unitaryfund/mitiq/blob/main/CHANGELOG.md#version-0340).

natestemen:
The code snippet in the changelog is still broken (I don't think it ever worked). Whether it's worth fixing, I'm not sure, but I don't love the fact that we have code in the changelog (which, IMO, is a suggestion for users to run) which does not run.

@Misty-W do you still want to update this?

Misty-W:
> The code snippet in the changelog is still broken (I don't think it ever worked). Whether it's worth fixing, I'm not sure, but I don't love the fact that we have code in the changelog (which, IMO, is a suggestion for users to run) which does not run.
> 
> @Misty-W do you still want to update this?

I’ll remove it.

cosenal:
Sorry, I closed this PR because I thought the snippet from the changelog had been fixed already.

Anyway, I looked at the code of the snippet. It doesn't make sense to me even before executing it, and that's neither in the original form, nor in the form of this PR.

In the original form, we are converting the circuit from Qibo to Cirq, and then using a `circuit.execute()` executor function, which implies that Cirq has an `execute` function, but it doesn't look like that's the case.

On the other hand, in the form of this PR, we are trivially using the Qibo [execute()](https://qibo.science/qibo/stable/_modules/qibo/models/circuit.html#Circuit.execute) function, which does exist, but doesn't even have the signature required, i.e., it returns an object that is not Mitiq's QuantumResult. 

In both cases, I don't see why this should work in the simple form of the snippet, unless we write a non-trivial executor, similar to [the one](https://github.com/unitaryfund/mitiq/blob/832daf992493101d10e35c5d3617c54de513def3/docs/source/examples/qibo-noisy-simulation.md#defining-the-executor) in the «Error mitigation with Qibo» example.

natestemen:
> In both cases, I don't see why this should work in the simple form of the snippet

My guess is that it went untested in https://github.com/unitaryfund/mitiq/pull/2191. Maybe a good practice to help eliminate this in the future is require 2 reviewers for release notes?

PR #2220, opened 2024-03-05
Add qibo example to docs

francescsabater:
Add an example on how to perform ZNE on a Qibo circuit. The example is pretty much the same one as in the Qibo blog post: https://unitary.fund/posts/adding_qibo/
---

### License


natestemen (changes requested):
Thank you so much for adding this Francesc! I wanted to make your life simpler, especially since you've already done a lot for us in getting Qibo supported here, so I went ahead and added two commits to your branch. [One](https://github.com/unitaryfund/mitiq/pull/2220/commits/8cdafdac36c2c48416d2aff06f5dfb792291cd51) is needed to ensure the example shows up on our [example gallery](https://mitiq.readthedocs.io/en/stable/examples/examples.html). [The other](https://github.com/unitaryfund/mitiq/pull/2220/commits/9ffa74fb3f174d90068618beb13c0d320df439dc) is some rewording of stuff to hopefully make it a little more clear. Feel free to modify the wording if you find them unclear!

My main concern here is that currently all the measured expectation values as part of ZNE are showing up as identical! We need to ensure the noisy simulation is actually worsening the result as the circuit is folded.

The last thing that we will need to do in order to get this over the line is add a thumbnail for the tutorial! Take a look at the example gallery for inspiration, but it's up to you how you want to do it. A simple idea could be to put the Qibo and Mitiq logo together. Once you create it, you will need to upload it into the [`docs/source/_thumbnails`](https://github.com/unitaryfund/mitiq/tree/6a30f16edd18d768ac5cbd2ba571ef3e3284f58c/docs/source/_thumbnails) directory, and add a line to the `nbsphinx_thumbnails` object here:
https://github.com/unitaryfund/mitiq/blob/6a30f16edd18d768ac5cbd2ba571ef3e3284f58c/docs/source/conf.py#L337-L370

francescsabater:
Thanks for the commits @natestemen! 

> My main concern here is that currently all the measured expectation values as part of ZNE are showing up as identical! We need to ensure the noisy simulation is actually worsening the result as the circuit is folded.

What do you mean with "all the measured expectation values as part of ZNE are showing up as identical"? I've run the code in the example (just copy paste it) in a jupyter notebook and when plotting the extrapolation the expectation value gets worse when increasing the scale factor: 
Captura de pantalla 2024-03-07 131759

francescsabater:
Changing `counts_0 = result_freq.get(0, 0)`  to  `counts_0 = result_freq.get(0, 0)` should solve the issue. For some reason I was running the code in my compiter with an older version of Qibo where `result_freq.get(0, 0)` worked. In version 0.2.4 the keys of `result_freq` are strings, which is why we were getting 0 as a result since we were obtaining no counts.

natestemen (approved):
Awesome, thanks so much for figuring that out! The plot is displaying correctly and looks great.

Just one, non-blocking question: Is it possible to suppress the output that Qibo outputs when running circuits? E.g. you can see here when we run ZNE it prints out info about the backend that Qibo is using. Can we get rid of that?
Screenshot 2024-03-07 at 1 12 29 PM

francescsabater:
Well seen Nate, maybe a way to reduce these statistical oddities without having to grow the circuit and sticking with the RichardsonFactory is to set `num_to_average=3`  instead of the default one `execute_with_zne`? I've tried and it seems to improve the fit.

natestemen:
Using linear extrapolation should make this more reproducible. Thanks for that! Going to merge so we don't continue going back and forth making smaller and smaller changes :)

Issue #1790, opened 2023-04-17
Use python 3.11 for RTD builds


natestemen:
RTD is having some problems building our documentation on using python 3.11 so our upgrade does not include putting RTD on python 3.11. This is probably something we want to do down the line.

PR #2231, opened 2024-03-14
Use python 3.11 on RTD

natestemen:
## Description

Read The Docs is using python 3.10 to build our documentation. This was originally part of our upgrade to support python 3.11 (https://github.com/unitaryfund/mitiq/pull/1663), but we went forward without it because there was some troubles with python 3.11 on RTD.

This is only one data point, but [`latest`](https://readthedocs.org/projects/mitiq/builds/23767064/) on RTD is building in 2697 seconds, and [this PR](https://readthedocs.org/projects/mitiq/builds/23767071/) is building in 2351 seconds. 5 minute improvement! Thanks python 3.11! Now if we can get our RTD build below 30 min that would be really cool, but something to think about down the line.

fixes https://github.com/unitaryfund/mitiq/issues/1790


cosenal (approved):

cosenal:
Just to add another data point from a local build on my machine 
```make docs-clean  962.92s user 424.57s system 193% cpu 11:57.48 total```

PR #2248, opened 2024-03-27
adding rotated randomized benchmarking circuits to calibrator

FarLab:
## Description

---

### License

- [ ] I license this contribution under the terms of the GNU GPL, version 3 and grant Unitary Fund the right to provide additional permissions as described in section 7 of the GNU GPL, version 3.

Before opening the PR, please ensure you have completed the following where appropriate.

- [ ] I added unit tests for new code.
- [ ] I used [type hints](https://www.python.org/dev/peps/pep-0484/) in function signatures.
- [ ] I used [Google-style](https://google.github.io/styleguide/pyguide.html#383-functions-and-methods) docstrings for functions.
- [ ] I [updated the documentation](../blob/master/docs/CONTRIBUTING_DOCS.md) where relevant.
- [ ] Added myself / the copyright holder to the AUTHORS file

PR #2252, opened 2024-03-29
Exit early when circuit type is not supported

cosenal:
## Description

A small thing I noticed when reviewing #2248. Unless I am missing something, there is no point in generating the circuit for a circuit type we don't support.

`make test` passes successfully.


natestemen (approved):
Good catch, and thanks for the quick patch!

PR #2260, opened 2024-04-02
Remove redundant word in release title

cosenal:
## Description

One of the manual step in the release process I witnessed today was to remove the word 'Release' from the release title, which I agree with @natestemen it is redundant. We may as well remove it formally and save a manual step.


natestemen (approved):
Nice find!

PR #2264, opened 2024-04-03
Pre-commit hook with style checks

cosenal:
## Description

I got annoyed by a PR I opened and failed on Github checks, so I created a git hook that checks code style before each commit.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -48,6 +48,12 @@ linkcheck:
 install:
 	pip install -e .[development]
 
+.PHONY: install-hooks
+install-hooks:
+	@git config --local core.hooksPath .git-hooks/
+	@chmod +x .git-hooks/*
+	@echo "Git hooks installed."
+
 .PHONY: requirements
 requirements: requirements.txt
 	pip install -r requirements.txt
--- a/.git-hooks/pre-commit
+++ b/.git-hooks/pre-commit
@@ -1,0 +1,10 @@
+#!/bin/sh
+
+make check-format && make check-types
+
+if [ $? -ne 0 ]; then
+  echo "Code checks have failed. Fix issues before committing."
+  exit 1
+fi
+
+exit 0
--- a/.git-hooks/pre-commit
+++ b/.git-hooks/pre-commit
@@ -1,6 +1,6 @@
 #!/bin/sh
 
-make check-format && make check-types
+make check-all
 
 if [ $? -ne 0 ]; then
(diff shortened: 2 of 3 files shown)


vprusso (approved):
Nice. LGTM!

Makefile
```diff
@@ -48,6 +48,12 @@ linkcheck:
 install:
 	pip install -e .[development]
 
+.PHONY: install-hooks
+install-hooks:
+	@git config --local core.hooksPath .git-hooks/
+	@chmod +x .git-hooks/*
+	@echo "Git hooks installed."
```
natestemen, line 55: What are the `@`s for?

natestemen (approved):
Awesome! I've always heard of git hooks, but never actually used them. I double checked on this branch that it works and indeed it does:

```
$ # made some formatting change
$ git add .
$ git commit -m "attempt"
ruff check
ruff format --check
Would reformat: mitiq/qse/qse.py
1 file would be reformatted, 121 files already formatted
make: *** [check-format] Error 1
Code checks have failed. Fix issues before committing.
```

🎉

Makefile
cosenal: It runs the command silently, without the command itself being printed. It's useful in front of the `echo` command, otherwise you have 
```
@echo "Git hooks installed."
"Git hooks installed."
```
on the screen, which looks a bit silly. The usage in front of `git` and `chmod` command is debatable, one may argue that obfuscates what is happening.

natestemen: Well TIL! I'm cool with keeping them silent.

jordandsullivan:
I wonder if we might also add `make format` to the git-hooks, since it fixes format issues automatically instead of you needing to do so manually after it checks your code. I know some people may not like the machine changing their code without asking though. Any thoughts @cosenal ?

cosenal:
@jordandsullivan I thought about this when I was implementing the hook. I leaned towards only doing the check and not the format for a few reasons:

1. it mimics what we do in the github workflow
2. it's not too invasive (as you said, some people may not like the machine changing their code)
3. from what I observed, make format (ruff) doesn't always do a good job at fixing all the formatting issue, so manual intervention is required anyway in most cases.

I would keep it the way it is for now, we can always change it in the future. And each of us can always have personal hooks on top of the team-shared one.

jordandsullivan:
Makes sense!

Issue #746, opened 2021-06-22
Better type for `QPROGRAM`


rmlarose:
Currently `QPROGRAM` is either a `typing.Union` of several quantum circuit types, or a single type if only Cirq is installed. It should be consistent in both cases - it was a bit of a pain to write a simple test in #745. Further, the `typing.Union` type is not particularly useful, notably for `isinstance` checks.

## Desired behavior

- `QPROGRAM` should have the same type regardless of how many supported circuit types are installed.
- You should be able to check how many circuit types are in `QPRGORAM`. 
- You should be able to check if a particular circuit type is in `QPROGRAM`.

nathanshammah:
@cosenal great you are working on this, @FarLab I'm adding it to the current milestone.

cosenal:
@rmlarose Sorry for necromancing this after many years :D

I understand how `QPROGRAM` collapses to the type Circuit when no other circuit types are installed. We could add a dummy type in the Union to avoid this.

However, `QPROGRAM` should only be used for type hinting. When do you envision Mitiq users/maintainers to inspect `QPROGRAM` at run-time? The only use cases I can think of are similar to the one in the test you implemented in #745, but those are "meta" use-case that in turn are due to the way we implement those imports.

cosenal:
@natestemen A friendly reminder to explicit your concern about the use of a union type here (I remember you had something to say about this at the last community call).

natestemen:
Thanks for the ping, and apologies for the delay!

My main concern is related to how the definition of `QPROGRAM` affects the capability of defining good typehints for executors. Right now, the simplest thing to do is `Callable[[QPROGRAM], float]`. This allows for somewhat strange executors to pass the typecheck such as the following.
```py
def execute(circuit: Union[qiskit.QuantumCircuit, pyquil.Program]) -> float:
    ...
```

Ideally, we should still be able to use `Callable[[QPROGRAM], float]` since it is succinct and gets the idea across, but only functions which accept a **single** `QPROGRAM` type should pass. This _could_ be achieved with something like this typehint.[^1]
```py
ExecutorLike = Union[
    Callable[[_Circuit], float],
    Callable[[_Program], float],
    Callable[[_QuantumCircuit], float],
    Callable[[_BKCircuit], float],
    Callable[[_QuantumTape], float],
    Callable[[_QiboCircuit], float],
]
```
As to whether it's worth putting time in to achieve this goal, I'm not totally sure.

[...]

PR #745, opened 2021-06-22
Add test without optional 3rd party packages

rmlarose:
Description
-----------

Closes #740 by adding a basic test.

Checklist
-----------

Check off the following once complete (or if not applicable) after opening the PR. The PR will be reviewed once this checklist is complete and all tests are passing.

- [x] I added unit tests for new code.
- [x] I used [type hints](https://www.python.org/dev/peps/pep-0484/) in function signatures.
- [x] I used [Google-style](https://sphinxcontrib-napoleon.readthedocs.io/en/latest/example_google.html) docstrings for functions.
- [x] I [updated the documentation](../blob/master/docs/CONTRIBUTING_DOCS.md) where relevant.

If some items remain, you can mark this a [draft pull request](https://github.blog/2019-02-14-introducing-draft-pull-requests/).

Tips
----

- If the validation check fails:

    1. Run `make check-style` (from the root directory of the repository) and fix any [flake8](http://flake8.pycqa.org) errors.

    2. Run `make format` to format your code with the [black](https://black.readthedocs.io/en/stable/index.html) autoformatter.

  For more information, check the [Mitiq style guidelines](https://mitiq.readthedocs.io/en/stable/contributing.html#style-guidelines).
  
- Write "Fixes #XYZ" in the description if this PR fixes Issue #XYZ.

PR #2270, opened 2024-04-07
Make test for 3rd party imports meaningful again

cosenal:
## Description

(This fix is a prerequisite for working on https://github.com/unitaryfund/mitiq/issues/746.)

A test was created in https://github.com/unitaryfund/mitiq/pull/745 to double check that one can import mitiq without _any_ 3rd party integration package installed. However, this was relying on `pip install .` not installing any integration packages, which is no longer the case, since `cirq` is in the main `requirements.txt` and in turn installs `pyquil`.

In other words, the scenario in which QPROGRAM collapses to the single Circuit type (and not an Union) was never hit. This PR addresses this.

A more elegant fix would imply to only have `cirq-core` in the main requirements, but that requires some deeper redesign of how we integrate 3rd parties packages. 

## Test

I tested this with the awesome [act](https://github.com/nektos/act/) tool (note: it requires Docker) by running

```sh
act push --container-architecture linux/amd64 -j core
```
From the output (in the screenshot), one can see that pyquil is found and uninstalled.

Screenshot 2024-04-07 at 20 48 45

### DIFF
--- a/mitiq/benchmarks/randomized_benchmarking.py
+++ b/mitiq/benchmarks/randomized_benchmarking.py
@@ -3,8 +3,7 @@
 # module.
 # Original authors: Cirq developers: Xiao Mi, Dave Bacon, Craig Gidney,
 # Ping Yeh, Matthew Neely.
-# Code URL = ('https://github.com/quantumlib/Cirq/blob/main/cirq-core/cirq/
-# experiments/qubit_characterizations.py').
+# Code URL = ('https://github.com/quantumlib/Cirq/blob/main/cirq-core/cirq/experiments/qubit_characterizations.py').
 #
 # This source code is licensed under the GPL license (v3) found in the
 # LICENSE file in the root directory of this source tree.
--- a/mitiq/tests/test_without_third_party_packages.py
+++ b/mitiq/tests/test_without_third_party_packages.py
@@ -11,7 +11,9 @@ mitiq.interface.mitiq_[package], where [package] is any supported package that
 interfaces with Mitiq (see mitiq.SUPPORTED_PROGRAM_TYPES).
 """
 
-from abc import ABCMeta
+from typing import Union, get_origin
+
+from cirq import Circuit
 
 
 def test_import():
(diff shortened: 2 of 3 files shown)


mitiq/benchmarks/randomized_benchmarking.py
```diff
@@ -3,8 +3,7 @@
 # module.
 # Original authors: Cirq developers: Xiao Mi, Dave Bacon, Craig Gidney,
 # Ping Yeh, Matthew Neely.
-# Code URL = ('https://github.com/quantumlib/Cirq/blob/main/cirq-core/cirq/
-# experiments/qubit_characterizations.py').
+# Code URL = ('https://github.com/quantumlib/Cirq/blob/main/cirq-core/cirq/experiments/qubit_characterizations.py').
```
cosenal, line 6: Unrelated to the PR's scope; it just annoyed me that I couldn't Ctrl+Click on that link from my IDE 😄

nathanshammah:
Great @cosenal. `act` seems very useful independent on #746: I advise opening an issue and related PR (with info on contributing to docs, etc.) for `act`.

natestemen:
That'll happen, unfortunately. I usually check to ensure the links are **not** truly broken, and if they work, then just make a note of it being the reason for the red X and continue on.

mitiq/tests/test_without_third_party_packages.py
```diff
@@ -20,14 +22,14 @@ def test_import():
     """
     import mitiq
 
-    if isinstance(mitiq.QPROGRAM, ABCMeta):
-        pass  # If only Cirq is installed, QPROGRAM is not a typing.Union.
-    else:
+    if get_origin(mitiq.QPROGRAM) is Union:
```
natestemen, line 25: Nice, this is an easier way to understand that the type of something is a `Union`!

natestemen (approved):
Good catch!

We should merge this, but it does make me wonder as something to consider for the future: can we remove `cirq` as our main dependency and use `cirq-core` instead?

cosenal:
> Good catch!
> 
> We should merge this, but it does make me wonder as something to consider for the future: can we remove `cirq` as our main dependency and use `cirq-core` instead?

That's exactly what I wrote in the last sentence of the description 😬

natestemen:
You're right, sorry!

cosenal:
Docs build fails because of linkcheck getting 403 from iopscience. I inspected those links manually and they work, hence merging.

PR #2271, opened 2024-04-08
Update to codecov-action v4 and use token

cosenal:
## Description

Upload to codecov fails in our pull requests builds, so we try to update to codecov-action v4. This version requires a token even for public repository. The secret token was taken from https://app.codecov.io/gh/unitaryfund/mitiq/settings and then set in https://github.com/unitaryfund/mitiq/settings/secrets/actions.


nathanshammah (approved):
LGTM, thanks @cosenal!

PR #2277, opened 2024-04-10
Retry linkcheck when building docs

cosenal:
## Description

Linkcheck when building docs fails too often on PR build, e.g. [this run](https://github.com/unitaryfund/mitiq/actions/runs/8629414504/job/23653574928?pr=2276).

Let's try retries!

### DIFF
--- a/docs/source/conf.py
+++ b/docs/source/conf.py
@@ -166,6 +166,8 @@ linkcheck_ignore = [
     r"https://arxiv.org/abs/1807.05572",
 ]
 
+linkcheck_retries = 3
+
 
 class ApsStyle(pybtex.style.formatting.unsrt.Style):
     """Style that mimicks APS journals."""


purva-thakre:
@cosenal Sometimes `linkcheck` fails because websites might ask for a check to verify the entity accessing the link is a human. This is typical of `IOP` links from what I have noticed. 

The only solution I have found for this so far is editing the `bibtex` entry to use the `arxiv` URL instead and skip providing a `doi` entry. https://arxiv.org/abs/2103.12680

purva-thakre (approved):

cosenal:
@purva-thakre That makes sense, it's a measure to avoid bot scraping their website. I agree that when it's possible we should use arXiv links instead.

cosenal:
Not very effective, builds are still failing with linkcheck errors :/

purva-thakre:
Maybe we should create an issue to replace all IOP links with arxiv links or make linkcheck ignore the broken link. 

https://github.com/unitaryfund/mitiq/blob/7bfa0fe9b8d292a0fe084047583c26c2374ea13b/docs/source/conf.py#L156

cosenal:
We could also run linkcheck on a scheduled workflow (say once a day), instead of running it on every pull request. What do you think @purva-thakre @natestemen ?

cosenal:
If one opens [this link](https://iopscience.iop.org/article/10.1088/2058-9565/abd5c9) from say, a browser, they can see it actually returns 403 all the time. I guess this is how IOPScience is set up.

purva-thakre:
I wonder if using arxiv links for IOPScience articles would work. arxiv has some instructions for 403: https://arxiv.org/denied.html

For example, the link below is ignored by `linkcheck` but it could be instead changed to `https://export.arxiv.org/abs/1807.05572`
https://github.com/unitaryfund/mitiq/blob/7bfa0fe9b8d292a0fe084047583c26c2374ea13b/docs/source/conf.py#L166

> We could also run linkcheck on a scheduled workflow (say once a day), instead of running it on every pull request.

Running it on a scheduled workflow sounds like a nice idea! We could also run it once a week. I don't think those links will change quickly. Let's wait for Nate's comment on this.

cosenal:
@purva-thakre Nice find! Perhaps we can set a redirect pattern from arxiv.org to export.arxiv.org.

natestemen:
Making a comment over on #2283 since that's open now.

PR #2279, opened 2024-04-10
Use Jupyter cache in gh workflow docs build

cosenal:
## Description

This is an attempt to use [Github Actions Cache](https://github.com/actions/cache) to restore jupyter_cache from one workflow run to the next one. That cache is created by Sphinx Myst-NB because we are setting [this](https://github.com/unitaryfund/mitiq/blob/ffa4b386fd77398e9135766d5eb2509480a72743/docs/source/conf.py#L128) in the configuration.

---

## Test
Unfortunately I can't test this with the awesome`act` tool, because this job starts docker in one of the steps (#dockerinception). So only a merge will tell if it works 😬

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -54,6 +54,13 @@ jobs:
           docker run --rm -idt -p 5000:5000 rigetti/qvm -S
           docker run --rm -idt -p 5555:5555 rigetti/quilc -R
 
+      - name: Fetch Jupyter cache
+        uses: actions/cache@v3
+        with:
+          path: ./docs/build/.jupyter_cache
+          key: jupyter-cache-${{ github.run_id }}
+          restore-keys: jupyter-cache
+
       - name: Build and test Sphinx docs
         run: |
           export BQSKIT_DOC_CHECK_OVERRIDE=1
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -55,11 +55,10 @@ jobs:
           docker run --rm -idt -p 5555:5555 rigetti/quilc -R
 
       - name: Fetch Jupyter cache
-        uses: actions/cache@v3
+        uses: actions/cache@v4
         with:
           path: ./docs/build/.jupyter_cache
           key: jupyter-cache-${{ github.run_id }}
-          restore-keys: jupyter-cache
 
       - name: Build and test Sphinx docs
         run: |


.github/workflows/build.yml
```diff
@@ -54,6 +54,13 @@ jobs:
           docker run --rm -idt -p 5000:5000 rigetti/qvm -S
           docker run --rm -idt -p 5555:5555 rigetti/quilc -R
 
+      - name: Fetch Jupyter cache
+        uses: actions/cache@v3
+        with:
+          path: ./docs/build/.jupyter_cache
+          key: jupyter-cache-${{ github.run_id }}
+          restore-keys: jupyter-cache
```
natestemen, line 62: 1. I'm not sure, is `restore-keys` required? If so, I think it might be missing a `-` at the end to match `key`?
2. My understanding of `key` is it is the thing that will control the use of a cache, so if the key changes, a new cache will be requested. Should we have the key be some sort of hash of the mitiq python files? It's done this way [here](https://github.com/lesteve/scikit-learn-mooc/blob/7243c2a8dfd5e1704986e7b9564c1d0111e9a334/.github/workflows/deploy-gh-pages.yml#L39).

natestemen (commented):
Great idea. Save engineering/CI time 💪🏻.

Had a few questions, and generally since we can't test it ourselves I figured it would be good to do some [searching across GitHub](https://github.com/search?q=actions%2Fcache+language%3AYAML+jupyter&type=code) to see what other people are doing.

.github/workflows/build.yml
```diff
@@ -54,6 +54,13 @@ jobs:
           docker run --rm -idt -p 5000:5000 rigetti/qvm -S
           docker run --rm -idt -p 5555:5555 rigetti/quilc -R
 
+      - name: Fetch Jupyter cache
+        uses: actions/cache@v3
```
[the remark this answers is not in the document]
cosenal: definitely! fixed

.github/workflows/build.yml
cosenal: 1. It makes sense only when you want to make a priority list of keys you want to restore, which is not our case. I don't think it was missing the `-` anyway, since it's just supposed to be a matching prefix, and `jupyter-cache` would match `jupyter-cache-${{ github.run_id }}`. Anyway, it's not required, I will remove it.

2. The key I picked is static as `github.run_id` effectively is the same for each workflow run. Here is my rationale for picking a static key: we are not using this action to create a cache mechanism, which is instead already provided by jupyter cache. All we want to do here is load the jupyter cache, and then sphinx/myst_nb will know how to handle that cache, the same way it happens locally when you run two times `make docs`. If we have a key that is a hash of the mitiq python files, or a hash of all mitiq files, then the cache is invalidated on each PR, which is not what we want. I know it's a bit counterintuitive, and that's because this is a bit of an unorthdox use of this action. We are using it to store and load an existing caching mechanism, rather than for implementing a caching mechanism.

cosenal:
@natestemen comments addressed, PTAL

.github/workflows/build.yml
```diff
@@ -54,6 +54,12 @@ jobs:
           docker run --rm -idt -p 5000:5000 rigetti/qvm -S
           docker run --rm -idt -p 5555:5555 rigetti/quilc -R
 
+      - name: Fetch Jupyter cache
```
natestemen, line 57: Just checking my understanding: this is both fetching and setting the cache, right?

natestemen (approved):
Let's try it out! I have had some instances locally where I needed to remove something from the cache in order to get the build to work correctly, but I haven't recreated that situation.

Just left one question, but not blocking. Thanks for answering my other questions :)

.github/workflows/build.yml
cosenal: Yes. My understanding is that it simply populate that path in the container with the content that's in cache.

cosenal:
Unfortunately, because of that linkcheck issue, the build doesn't succeed and the cache is not populated. I will merge anyway, and hopefully we will have a chance to try it out once the build is green.

purva-thakre:
@cosenal We could temporarily remove the `make linkcheck` line from the job

https://github.com/unitaryfund/mitiq/blob/12c7126ece12f89cdb9198c08ff19fa45cd15e01/.github/workflows/build.yml#L63

And add it back in when the other issue is resolved.

cosenal:
@purva-thakre Raise a PR and I will approve it right away :)

PR #2302, opened 2024-04-17
Remove unused devcontainer

cosenal:
## Description

Removing the development container, which is neither used nor maintained. 
It was hanging around as an example of how to containerize Mitiq, but for that we need more specific requirements.


natestemen (approved):
🚢

Issue #1201, opened 2022-04-01
Update `about()` and document latest package support


Aaron-Robertson:
Issue Description
-----------------

Latest supported versions of optional dependencies are not well documented. It is unclear which frontend version(s) to install after a user installs through `pip install mitiq`, but wants to use Qiskit, Pennylane, Braket, and/or PyQuil.

Proposed Solution
-----------------

- [x] Update `about()` to include Pennylane
- [x] Create a function `latest_supported()` to parse Qiskit, Pennylane (and Pennylane-Qiskit), Braket, and PyQuil versions from `dev_requirements.txt`
- [x] Add `Latest Supported` as a column in `about()` under `Optional Dependencies`
- [x] Create a user warning when package versions are ahead of latest (from Andrea):
> e.g. something like: `WARNING: Mitiq is tested with version xxx of Qiskit. Your version is higher and may be incompatible...`
- [x] Add warning to `about()`
- [x] Add warning to `QPROGRAM` type builder
- [ ] Add `latest_supported()` with output to docs under [Supported Frontends](https://mitiq.readthedocs.io/en/stable/#supported-frontends)

Additional References
---------------------

Discussed in Mitiq meeting 04/01/22. See this [thread on discord](https://discord.com/channels/764231928676089909/958388424472166400).

Aaron-Robertson:
@andreamari label/change this issue as you see fit. As written I think it would be a good first issue.

andreamari:
Thanks @Aaron-Robertson!

jordandsullivan:
So just getting my bearings here:
1. Was some version of the `latest_supported()` function ever actually merged into main? It's hard to tell between the discussion in #1220 and #1478.


The Mitiq Read the Docs is generated by sphinx, and the [supported backends section](https://mitiq.readthedocs.io/en/stable/index.html#supported-backends) comes specifically from the [README](https://github.com/unitaryfund/mitiq/blob/c5afbd4df6850cacc331a1aa1a477265d4bb584a/README.md?plain=1#L107).
2. We want to get this section to display the latest ne idea would be to use the workflow from the examples (i.e. use [myst-nb](https://myst-nb.readthedocs.io/en/latest/) to convert the README.md to a jupyter notebook, run that notebook to get the output of the `latest_supported()` function (if indeed it was implemented). Does this seem workable?

jordandsullivan:
@andreamari  Could you help answer the above?

natestemen:
[...]

Issue #2208, opened 2024-02-25
Any qiskit circuit that requires conversion to qasm breaks as of  qiskit 1.0.1


hrushikesh890:
## Issue Description
 
Example: Take the cdr case. We want to generate clifford training data with the input as qiskit quantum circuit. This fails with the error as
`'QuantumCircuit' object has no attribute 'qasm'`

## How to Reproduce
Minimum steps
```
from qiskit import QuantumCircuit
from mitiq.cdr import generate_training_circuits
qc = QuantumCircuit(3)
qc.h(0)
qc.cx(0,1)
qc.cx(1,2)

cd = generate_training_circuits(qc, 10, 0.1)
```

This generates following error
``` return from_qasm(circuit.qasm())

AttributeError: 'QuantumCircuit' object has no attribute 'qasm'
```

## Root Cause
QuantumCircuit object from qiskit 1.0.1 no longer has qasm as an attribute

## Possible solutions and workarounds
Fix - Replace circuit.qasm() in from_qiskit function with qiskit.qasm2.dumps(circuit)
This fix also works with qiskit 0.45 
---

<details>
<summary>Environment Context</summary>

Use the `about()` function to summarize information on operating system, python version and dependencies.
```python
from mitiq import about
about()
```

Additional Python Environment Details (`pip freeze` or `conda list`):

Mitiq version 0.34
Qiskit 1.0.1

```
Copy and paste the output of `pip freeze` or `conda list` here.
```

</details>

purva-thakre:
@hrushikesh890 Thank you for creating this issue. Yes, we are aware of the issues in `qiskit 1.0.1`  https://github.com/unitaryfund/mitiq/pull/2207. Our next release should hopefully fix these.

You will have to switch to an earlier version as specified in the dependencies of the project. 
https://github.com/unitaryfund/mitiq/blob/c5afbd4df6850cacc331a1aa1a477265d4bb584a/dev_requirements.txt#L1-L4

PR #2303, opened 2024-04-17
Introduce requirement setup for integrations

cosenal:
## Description

At the moment, we don't pin the version of integration packages (Braket, Qiskit, etc.)  in the Mitiq package that we ship to users. This causes incompatibility issues, e.g, https://github.com/unitaryfund/mitiq/issues/2208.

This PR introduces integration packages as extra dependencies.

Example: if a user wants to use Mitiq with Qiskit, they can run 
```
pip install mitiq[qiskit]
```
in their enviroment, and similarly for all other integrations we support. This will guarantee that the version of the integration packages we support are compatible with the ones in the user's environment. 

🎪 One important bonus is that now the main Mitiq package only depends on `cirq-core` as opposed to the full `cirq` package 🎪

-----

This fixes https://github.com/unitaryfund/mitiq/issues/1201 (see Issue description)

### Docs
Readme has been updated.

Note: There will be a lapse of time between merging this PR and the next release, when the README will not be synced with reality of the package. We may consider split the Readme change and ship that along with the release PR. (cc @FarLab as milestone manager.) 

### Testing
I tested that:
1. for Mitiq developers, everything works just like before with `pip install -e .[development]`
2. the code snippet from the README works with `pip install mitiq` (without any extra)
3. the different extra installations for the supported frontend also works (tested started from empty Conda environments)


jordandsullivan:
Good idea. I've already hit this issue several times.

natestemen (commented):
On a high level this looks great. I like this structure a lot! 👍🏻

cosenal:
@natestemen @jordandsullivan This is now ready to review. Please take a look.

setup.py
```diff
@@ -22,13 +25,21 @@
 with open("mitiq/_version.py", "w") as f:
     f.write(f"__version__ = '{__version__}'\n")
 
+third_party_integration_requirements = {
```
jordandsullivan, line 28: To confirm, when a user runs `pip install -e .[development]` or `make install`, this will install all the third party integrations as well, correct?

jordandsullivan: How are we enforcing the version compatibility of the various 3P requirements? Do pip or setuptools do this automatically?

README.md
```diff
@@ -96,12 +96,23 @@ We refer to any programming language you can write quantum circuits in as a _fro
...
 
-Note: Cirq is a core requirement of Mitiq and is installed when you `pip install mitiq`.
+You can install Mitiq support for these frontends by specifying them during installation, 
+as optional extras, along with the main package.
+To install Mitiq with one or more frontends, you can specify each frontend in square brackets as part of the installation command. 
+
+For example, 
+to install Mitiq with support for Qiskit and Qibo:
+```bash
+pip install mitiq[qiskit,qibo]
```
jordandsullivan, line 110: Is there a way to test this functionality before deploying this version on PyPI?

requirements/requirements-braket.txt
```diff
@@ -0,0 +1,2 @@
+amazon-braket-sdk~=1.69.0
+cirq-ionq>=1.0.0,<1.4.0
```
jordandsullivan, line 2: Why is `cirq-ionq` required for the `braket` integration?

requirements/requirements-cirq.txt
```diff
@@ -0,0 +1,3 @@
+# This file is redundant, since the internal representation of choice is Cirq.
```
jordandsullivan, line 1: I would rephrase just to make this a little clearer. Something like, "...since Cirq is the core internal representation for quantum circuits in Mitiq"

jordandsullivan (commented):
If we have updated `make install` to include all these requirements, can we also update [this line](https://github.com/unitaryfund/mitiq/blob/setup-integrations/CONTRIBUTING.md?plain=1#L25) in the main CONTRIBUTING guide to say `make install` rather than `pip install -e ".[development]"`?

requirements/requirements-qibo.txt
```diff
@@ -0,0 +1 @@
+qibo==0.2.7 # TODO: unpin this
```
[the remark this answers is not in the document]
cosenal: Notice I have just moved those dependencies around, namely I split them in multiple files, I haven't changed them.
The "unpin this" comes from https://github.com/unitaryfund/mitiq/pull/2221

requirements/requirements-braket.txt
cosenal: Because of this: https://github.com/unitaryfund/mitiq/blob/8c24a9bddd4d6213cdd39fae2d750e55247e82a9/mitiq/interface/mitiq_braket/conversions.py#L8

requirements/requirements-cirq.txt
cosenal: Yes, better

setup.py
cosenal: > this will install all the third party integrations as well, correct?
Yes, exactly the same as before.

> How are we enforcing the version compatibility of the various 3P requirements?
Same as we do now, that is, having all of them together in the development environment and running the test suite in such environment.

README.md
cosenal: Yes. To replicate the same example, you can run `pip install -e '.[qiskit,qibo]'` locally.

cosenal:
@jordandsullivan Thanks for the thorough review. I addressed / replied to all your comments. PTAL

jordandsullivan (approved):
Thanks, looks good to me!

natestemen (approved):
Thanks Alessandro!

Looking forward to this helping users use Mitiq with less dependency issues.

I was brainstorming potential issues that could still arise, and I came up with the following scenario.

1. User gets set up in their desired frontend (not cirq), writing ciruits and running them
2. They learn about Mitiq
3. They run `pip install mitiq` without looking through the docs too much
4. They hit compatibility issues

If this users then come to us, it should be easy to help solve their problem, as we can ask them to run `pip install mitiq[frontend]`.

This is not a blocker, but I just wanted to write this down both for my understanding[^1], and so we all know there are still _some_ potential issues this does not solve. I don't think there's any magic bullet for this problem, however.

[^1]: If I'm misunderstanding something, let me know!

cosenal:
@natestemen You're right, that's a valid scenario. One way to mitigate it would be to use the tools from `pkg_resources` to check at runtime (as early as possible) for the dependencies installed in the user environment. Example:
```python3
pkg_resources.working_set.require('qiskit~=2.0.0')

VersionConflict: (qiskit 1.0.2 (/opt/miniconda3/envs/mitiqenv/lib/python3.11/site-packages), Requirement.parse('qiskit~=2.0.2')) 
```
vs.
```python3
pkg_resources.working_set.require('qiskit~=1.0.2')

[qiskit 1.0.2 (/opt/miniconda3/envs/mitiqenv/lib/python3.11/site-packages)
...
]
```

Perhaps something for a follow-up PR, subject to prioritization, in case we get more dependency issues reported.

natestemen:
Yup agreed, lets only pursue further ideas if we get more complaints. Thanks for jotting down a potential next step :)

cosenal:
Merged this. After the two reviewers approvals, I have only merged main into the branch and solved the merge conflicts.

PR #2312, opened 2024-04-21
Remove duplicate import

cosenal:
## Description

1. Removes a duplicate import (and sort them while we are at it)
2. Rename a test folder to `tests` otherwise the tests themselves are picked up by codecov, because of this: https://github.com/unitaryfund/mitiq/blob/8c24a9bddd4d6213cdd39fae2d750e55247e82a9/.coveragerc#L3


cosenal:
The codecov/project failure looks like a hiccup on codecov. Upon inspection of the [coverage report](https://app.codecov.io/gh/unitaryfund/mitiq/pull/2312?dropdown=coverage&src=pr&el=continue&utm_medium=referral&utm_source=github&utm_content=checks&utm_campaign=pr+comments&utm_term=unitaryfund) there are no files affected.

cosenal:
> Hm, do you know why codecov is complaining that this line was an "indirect change?" https://app.codecov.io/gh/unitaryfund/mitiq/pull/2312/indirect-changes

It's unrelated to this PR. What is happening is that the order in which the tests get executed affects that line. Since the order test is random, that line ends up being covered or not. In fact, if you look at the [list of recent coverage reports](https://app.codecov.io/gh/unitaryfund/mitiq/pulls), you can see the score oscillates between 98.29% and 98.33%.
I will try to find the root cause of this "non-determinism"

jordandsullivan (approved):

PR #2319, opened 2024-04-24
Fix global tqdm module patch in test

### LINKED ISSUE #2318: Codecov report score fluctuates even when no changes
> Hm, do you know why codecov is complaining that this line was an "indirect change?" https://app.codecov.io/gh/unitaryfund/mitiq/pull/2312/indirect-changes

It's unrelated to this PR. What is happening is that the order in which the tests get executed affects that line. Since the order test is random, that line ends up being covered or not. In fact, if you look at the [list of recent coverage reports](https://app.codecov.io/gh/unitaryfund/mitiq/pulls), you can see the score oscillates between 98.29% and 98.33%.
I will try to find the root cause of this "non-determinism"

_Originally posted by @cosenal in https://github.com/unitaryfund/mitiq/issues/2312#issuecomment-2070925879_

cosenal:
Despite the patch of tqdm happens in a context manager, the fact that the module is reloaded inside the context makes it so that the patch affects all other tests. No test wasn't failing, because there was no assert on tqdm being there, but codecov reports were affected by this, i.e., this PR fixes #2318.

### DIFF
--- a/mitiq/shadows/test/test_quantum_processing.py
+++ b/mitiq/shadows/test/test_quantum_processing.py
@@ -45,6 +45,10 @@ def test_tqdm_import_not_available():
         )  # Reload the module to trigger the import
         assert mitiq.shadows.quantum_processing.tqdm is None
 
+    # Reload the module again to restore the original tqdm import.
+    # Otherwise, the rest of the tests are affected by the patch (issue #2318)
+    importlib.reload(mitiq.shadows.quantum_processing)
+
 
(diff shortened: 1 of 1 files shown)


cosenal:
The [test (macos-latest, 3.9)](https://github.com/unitaryfund/mitiq/actions/runs/8821438712/job/24217292103?pr=2319#logs) build failing is because of https://github.com/actions/setup-python/issues/696

FarLab (approved):

Issue #2150, opened 2024-01-05
Set up release workflow with PyPI Trusted Publishers


natestemen:
### Background

In April 2023 PyPI [released](https://blog.pypi.org/posts/2023-04-20-introducing-trusted-publishers/) a new publishing method using the [OpenID Connect](https://openid.net/developers/how-connect-works/) standard [[GitHub docs](https://docs.github.com/en/actions/deployment/security-hardening-your-deployments/about-security-hardening-with-openid-connect)]. This standard is described as

> an interoperable authentication protocol based on the OAuth 2.0 framework of specifications (IETF RFC 6749 and 6750). It simplifies the way to verify the identity of users based on the authentication performed by an Authorization Server [...] **Best of all, it removes the responsibility of setting, storing, and managing passwords which is frequently associated with credential-based data breaches.**

More details can be found in the PyPI docs [here](https://docs.pypi.org/trusted-publishers/#quick-background-publishing-with-openid-connect), but because we set and store long-term passwords for use in our github actions, this is a welcome addition.

### Fix

The changes required are laid out in the [PyPI docs](https://docs.pypi.org/trusted-publishers/adding-a-publisher/). One of the suggested changes (for increased security) is to create an environment in which the release originates. This allows for further restrictions and rules to be applied as to when the release happens. GitHub docs [here](https://docs.github.com/en/actions/deployment/targeting-different-environments/using-environments-for-deployment).

We will also need to change code in two places to use the [`pypa/gh-action-pypi-publish`](https://github.com/pypa/gh-action-pypi-publish) action:

- [`.github/workflows/publish-pypi.yml`](https://github.com/unitaryfund/mitiq/blob/master/.github/workflows/publish-pypi.yml)
- [`.github/workflows/publish-testpypi.yml`](https://github.com/unitaryfund/mitiq/blob/master/.github/workflows/publish-testpypi.yml)

This may potentially alter our [release docs](https://github.com/unitaryfund/mitiq/blob/master/docs/source/release.md) as well. If so, they should be updated accordingly.

Misty-W:
No PR in milestone 0.34.0.

PR #2320, opened 2024-04-25
Used trusted publishers for testpypi publishing

natestemen:
## Description

This PR fixes the failing testpypi publishing action which runs nightly to ensure Mitiq can be uploaded to PyPI.

In addition it adds a check to ensure the version uploaded to testpypi is what is then later downloaded.

resolves https://github.com/unitaryfund/mitiq/issues/2150

### ⚠️ before merging ⚠️:

- [x] create an environment (through GitHub) called `testpypi` [[docs](https://docs.github.com/en/actions/deployment/targeting-different-environments/using-environments-for-deployment#creating-an-environment)]
- [ ] ~create an environment (through GitHub) called `pypi`~ (this will be done in a follow on PR when this change takes effect for the main release workflow)
- [x] ensure testpypi is set up properly to accept trusted publishing [[docs](https://docs.pypi.org/trusted-publishers/creating-a-project-through-oidc/)]

### DIFF
--- a/.github/workflows/publish-testpypi.yml
+++ b/.github/workflows/publish-testpypi.yml
@@ -21,7 +21,7 @@ jobs:
       - name: Install Python dependencies
         run: |
           python -m pip install --upgrade pip
-          make install requirements
+          make install
           pip install setuptools wheel twine
       - name: Build and publish
         env:
--- a/.github/workflows/publish-testpypi.yml
+++ b/.github/workflows/publish-testpypi.yml
@@ -9,7 +9,7 @@ on:
 
 jobs:
   build:
-    if: github.repository_owner == 'unitaryfund' && github.event_name == 'push'
+    if: github.repository_owner == 'unitaryfund'
     runs-on: ubuntu-latest
     steps:
       - name: Check out mitiq
--- a/.github/workflows/publish-testpypi.yml
+++ b/.github/workflows/publish-testpypi.yml
@@ -9,6 +9,7 @@ on:
 
 jobs:
   build:
+    name: Build Mitiq
     if: github.repository_owner == 'unitaryfund'
     runs-on: ubuntu-latest
     steps:
@@ -31,6 +32,7 @@ jobs:
           path: dist/*
 
   test-publish:
+    name: Upload release to test PyPi
     runs-on: ubuntu-latest
     needs: build
     environment:
@@ -57,7 +59,7 @@ jobs:
       - name: Set up Python
         uses: actions/setup-python@v5
         with:
-          python-version: 3.11
+          python-version: "3.11"
       - name: Install package from TestPyPI
         run: pip install -i https://test.pypi.org/simple/ --extra-index-url https://pypi.python.org/simple/ mitiq
       - name: Set version variable
--- a/.github/workflows/publish-testpypi.yml
+++ b/.github/workflows/publish-testpypi.yml
@@ -8,8 +8,8 @@ on:
     - cron: "0 0 * * *"
 
 jobs:
-  deploy:
-    if: github.repository_owner == 'unitaryfund'
+  build:
+    if: github.repository_owner == 'unitaryfund' && github.event_name == 'push'
     runs-on: ubuntu-latest
     steps:
       - name: Check out mitiq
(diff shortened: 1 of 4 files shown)


.github/workflows/publish-testpypi.yml
```diff
@@ -8,8 +8,8 @@ on:
     - cron: "0 0 * * *"
 
 jobs:
-  deploy:
-    if: github.repository_owner == 'unitaryfund'
+  build:
+    if: github.repository_owner == 'unitaryfund' && github.event_name == 'push'
```
cosenal, line 12: Since this is on a schedule, where is event_name `push`?

.github/workflows/publish-testpypi.yml
```diff
@@ -21,29 +21,55 @@ jobs:
       - name: Install Python dependencies
         run: |
           python -m pip install --upgrade pip
-          make install requirements
+          make install
```
cosenal, line 24: This will install development requirements, why do we need them here?

.github/workflows/publish-testpypi.yml
```diff
@@ -21,29 +21,55 @@ jobs:
...
           pip install setuptools wheel twine
-      - name: Build and publish
-        env:
-          TWINE_USERNAME: ${{ secrets.TESTPYPI_USERNAME }}
-          TWINE_PASSWORD: ${{ secrets.TESTPYPI_PASSWORD }}
+      - name: Build mitiq
+        run: python setup.py sdist bdist_wheel
```
cosenal, line 27: This way of building the wheel is deprecated, see table at the bottom of https://blog.ganssle.io/articles/2021/10/setup-py-deprecated.html#summary. The new way is
```
pip install build
python -m build
```

.github/workflows/publish-testpypi.yml
natestemen: Ahh good catch. I was reading a document on securing github actions workflows and I blindly applied this recommendation without much thinking!

.github/workflows/publish-testpypi.yml
natestemen: Done in https://github.com/unitaryfund/mitiq/pull/2320/commits/97e309f284e59445af330082b5d33c78ebb1c3fb.

.github/workflows/publish-testpypi.yml
natestemen: Good question. I'm not convinced they were necessary either. I've reduced the install requirements in https://github.com/unitaryfund/mitiq/pull/2320/commits/97e309f284e59445af330082b5d33c78ebb1c3fb.

.github/workflows/publish-testpypi.yml
natestemen: Removed in https://github.com/unitaryfund/mitiq/pull/2320/commits/c8abc2190b1d9cad346249087345000cea572ade.

natestemen:
**Reviewer Note:** This change is still somewhat experimental and is hard to test thoroughly. We may want to merge these changes with the exception of https://github.com/unitaryfund/mitiq/pull/2320/commits/20aeef4d21287c53ee651c2bbb04787533b55683 which adds trusted publishing for (main) PyPI. This way we can ensure TestPyPI uploads work correctly before adding it to the main upload process. Also if there are issues, it gives us time to fix things before the 0.38 release goes out this Friday (June 28, 2024).

cosenal:
> **Reviewer Note:** This change is still somewhat experimental and is hard to test thoroughly. We may want to merge these changes with the exception of [20aeef4](https://github.com/unitaryfund/mitiq/commit/20aeef4d21287c53ee651c2bbb04787533b55683) which adds trusted publishing for (main) PyPI. This way we can ensure TestPyPI uploads work correctly before adding it to the main upload process. Also if there are issues, it gives us time to fix things before the 0.38 release goes out this Friday (June 28, 2024).

How do you plan to exclude that commit? Are you going to create a separate PR?

natestemen:
> How do you plan to exclude that commit? Are you going to create a separate PR?

Easiest thing to do (IMO) would be to add a new commit that is the anti-diff of https://github.com/unitaryfund/mitiq/pull/2320/commits/20aeef4d21287c53ee651c2bbb04787533b55683 (i.e. revert the commit), so that in a follow on PR the commit can be [cherry-picked](https://git-scm.com/docs/git-cherry-pick) and applied.

cosenal:
> > How do you plan to exclude that commit? Are you going to create a separate PR?
> 
> Easiest thing to do (IMO) would be to add a new commit that is the anti-diff of [20aeef4](https://github.com/unitaryfund/mitiq/commit/20aeef4d21287c53ee651c2bbb04787533b55683) (i.e. revert the commit), so that in a follow on PR the commit can be [cherry-picked](https://git-scm.com/docs/git-cherry-pick) and applied.

OK, sounds good to me 👍

cosenal:
@natestemen LGTM. Do you want to push the anti-diff commit before I approve?

cosenal (approved):

natestemen:
`testpypi` environment set up and testPyPI set up to know about this github action as a publisher. merging. will be tested tonight when this runs.

PR #2329, opened 2024-04-26
RFC: Layerwise Richardson Extrapolation

### LINKED ISSUE #2224: RFC: Layerwise Richardson Extrapolation (LRE)
Link to the RFC: https://docs.google.com/document/d/1yAQI8K8ShGI4jdn8OtpBv-H0X5SvkFVXA7DKtVOqm-8/edit?usp=sharing

purva-thakre:
[Layerwise Richardson Extrapolation RFC](https://docs.google.com/document/d/1oFRl4wMGMtn57V0c_1egaHh0WUUAbtgW-U_QxNL9_kY/edit?usp=sharing)

Fixes #2224

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -111,6 +111,7 @@ For any part of the template that you weren't able to complete please mention th
 ### List of accepted RFCs
 This is a list of accepted request-for-comments (RFC) documents by date of creation (reverse chronological order):
 
+- [Layerwise Richardson Extrapolation RFC](https://docs.google.com/document/d/1oFRl4wMGMtn57V0c_1egaHh0WUUAbtgW-U_QxNL9_kY/edit?usp=sharing) by Purva Thakre (@purva-thakre) and Vincent Russo (@vprusso) Apr 24, 2024
 - [Robust Shadow Estimation](https://docs.google.com/document/d/1B5FnqQDvoRYap5fGPqzcbp-RXIrUFjbBcLiWIUrLmuA) by Min Li (@Min-Li) Jun 16, 2023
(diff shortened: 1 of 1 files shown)


purva-thakre:
I'll be late to the Community call today due to another appointment right before. 

Apologies for the delay in adding this RFC! We can discuss the proposed changes async through the Google doc comments if there's not enough time during the community call.

purva-thakre:
Andrea left a comment on the RFC that we should keep the LRE and ZNE inference methods separate. Initially, I thought it would be better to move `mitiq.zne.inference` to `mitiq.inference` in order to avoid defining another module for the multivariate extrapolation methods. 

The refactor to the `Factory` class in mitiq would have comprised of :`polyfit` works for single variable extrapolation and there would be another option for the multivariate extrapolation. Based on Andrea's comments, this refactor would be a major breaking change. 

https://docs.google.com/document/d/1oFRl4wMGMtn57V0c_1egaHh0WUUAbtgW-U_QxNL9_kY/edit?disco=AAABMa8k_iI

FWIW I do agree with Andrea that this would be a major breaking change. I thought of proposing this breaking change to make it possible for us to extend multivariate extrapolation beyond Richardson extrapolation (in the near future). 

I would appreciate comments from others on the two comments in the `Backward Compatibility` section. I can work on making changes to the RFC afterward.

jordandsullivan:
> Andrea left a comment on the RFC that we should keep the LRE and ZNE inference methods separate. Initially, I thought it would be better to move `mitiq.zne.inference` to `mitiq.inference` in order to avoid defining another module for the multivariate extrapolation methods.

I agree that it would make sense to move all the inference methods into their own module as you describe @purva-thakre .

purva-thakre:
Current version of the RFC (without the linked Jupyter notebook) was accepted during this week's community call.

cosenal (approved):
Merging this, given that the RFC has been accepted.

PR #2287, opened 2024-04-11
Temporarily remove `linkcheck`

purva-thakre:
Related to #2279 

Will add this line back in as a separate scheduled workflow or a modified `build` workflow in #2283

PR #2332, opened 2024-04-28
Run linkcheck on 'release' PR workflow

### LINKED ISSUE #2283: Add a scheduled `linkcheck` workflow
Will add better details later. 

- have linkcheck ignore all doi/arxiv links

---
## Previous issue description

- In `.github/workflows`, add a workflow that will run on a [schedule](https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#schedule) (frequency could be daily at midnight or on a weekly basis, TBD). Currently, `make linkcheck` is run due to the triggers shown below in `.github/workflows/build.yml`

https://github.com/unitaryfund/mitiq/blob/35139fe291cc7c37e9d062a60c7a2079e0416e00/.github/workflows/build.yml#L3-L14

As the triggers for `make linkcheck` will be different, it might be better to define a separate workflow labeled `.github/workflows/linkcheck.yml`

- `make linkcheck` tends to fail due to a 403 related to IOPScience links in `docs/source/refs.bib`. Replace the IOPScience links with those allowed by arxiv. See https://github.com/unitaryfund/mitiq/pull/2277#issuecomment-2048129304 for more info. 

- `linkcheck` ignores some links in `docs/source/conf.py` because those would cause the job related to `make linkcheck` fail. Remove the [ignored links](https://github.com/unitaryfund/mitiq/blob/7bfa0fe9b8d292a0fe084047583c26c2374ea13b/docs/source/conf.py#L156) from `docs/source/conf.py` and replace the IOPScience links with arxiv links. Then, use a redirect pattern as discussed in https://github.com/unitaryfund/mitiq/pull/2277#issuecomment-2048165904

cosenal:
Fixes #2283 as agreed in Mitiq community call on April 26, 2024.

# Description
Linkcheck was disabled in #2287, because it was making every pull request (PR) build fail with false errors. Some websites (e.g., doi.org) have bot-preventive measures for which they return 4xx errors when a request is made. In this PR we ignore those websites, on the assumption that we trust their links are not going to break.

Some false errors may still occurr, so we decided to do a linkcheck pass only on 'release' PRs.

## How do you detect a 'release' PR?
We check whether the word 'release' is in the PR title (case-insensitive). This will have some _false positives_ (just like this PR itself 😝 ), but it should have no _false negatives_, see [this search](https://github.com/unitaryfund/mitiq/pulls?q=is%3Apr+release+in%3Atitle+is%3Aclosed) on past release PRs. On false positives we just run make linkcheck superfluously, not a big deal.

# Bonus
I fixed some linkcheck warning by explicitly replacing links that have been redirected since the time they were added to our docs. 

# Testing
The new step was tested locally with 
```bash
act pull_request -e act_event.json --container-architecture linux/amd64 -j docs
```
where `act_event.json` is a file with content
```
{
  "pull_request": {
    "title": "this is a relEAse PR"
  }
}
```


purva-thakre:
I think you should add the details from `How do you detect a 'release' PR?` in your PR description to [this section](https://mitiq.readthedocs.io/en/stable/release.html#prepare-the-master-branch) of the docs.

cosenal:
> I think you should add the details from `How do you detect a 'release' PR?` in your PR description to [this section](https://mitiq.readthedocs.io/en/stable/release.html#prepare-the-master-branch) of the docs.

@purva-thakre Yes, good idea

cosenal:
@purva-thakre comment addressed, PTAL

docs/source/conf.py
```diff
@@ -154,16 +154,10 @@
...
-    r"https://doi.org/10.1145/3466752.3480059",
-    r"https://doi.org/10.7566/jpsj.90.032001",
-    r"https://arxiv.org/abs/1612.02058",
-    r"https://arxiv.org/abs/1805.04492",
-    r"https://arxiv.org/abs/1807.05572",
+    r"https://arxiv.org/.*",
+    r"https://doi.org/.+",
+    r"https://link.aps.org/doi/.+",
+    r"https://www.sciencedirect.com/science/article/.+",
 ]
```
purva-thakre, line 161: What is the difference between `/` and `\` ?

Not requesting a suggested change, BTW. I am curious about `\d+/` used by [Sphinx docs](https://www.sphinx-doc.org/en/master/usage/configuration.html#confval-linkcheck_ignore).

cosenal: Oops, good catch! The `.` is a special character in regex and needs to be escaped with the `\`. Fixed.

docs/source/release.md
```diff
@@ -39,9 +39,18 @@ This task has two parts:
...
+[SemVer](https://semver.org/), so typically a release will involve changing the version from
 `vX.Y.Zdev` (development) to `vX.Y.Z` (released).
 
+### Create a release pull request
+
+A pull request with the changes mentioned above should be created against the _main_ branch as part of the release preparation. 
+The pull request must be reviewed by at least one Mitiq maintainer in addition to the milestone manager, 
+and its title should be "X.Y.Z Release", where `X.Y.Z` represents the version number to be released.
+
+Note: This pull request triggers a specialized build workflow due to the keyword 'release' in its title.
```
purva-thakre, line 51:
```suggestion
Note: This pull request triggers a specialized build workflow due to the keyword `release` in its title. 
```

Minor suggestion for the keyword.

purva-thakre (approved):
LGTM!

cosenal:
You can see this in action now! https://github.com/unitaryfund/mitiq/actions/runs/8899188557/job/24437953232?pr=2338
🎉  well.. linkcheck fails, but the change from this PR is working :)

PR #2345, opened 2024-05-03
Example stacking DDD and ZNE

jordandsullivan:
## Description

Tutorial to combine ZNE and DDD. 

---

### License


nathanshammah:
Thanks @jordandsullivan. To solve the issue that the notebook doesn't render in the documentation, you just need to add it to the TOC tree in the examples [file](https://github.com/unitaryfund/mitiq/blob/main/docs/source/examples/examples.md). Eventually we use md files, but as a temporary file, it should work fine also with .ipynb format. The [jupytext](https://mitiq.readthedocs.io/en/latest/contributing_docs.html#adding-code-examples) extension allows you to edit the md file as a notebook.

nathanshammah:
My suggestion is that the executor should be only called by ZNE. The DD part is a "pre-processing", with input the initial circuit (e.g., GHZ state) and output the DD compiled quantum circuit. This is fed to ZNE. This should remove incompatibility types. 
[`ddd.insert_ddd_sequences`](https://mitiq.readthedocs.io/en/latest/apidoc.html#mitiq.ddd.insertion.insert_ddd_sequences) should be used. See the documentation [here](https://mitiq.readthedocs.io/en/latest/guide/ddd-4-low-level.html#inserting-ddd-sequences) and [here](https://mitiq.readthedocs.io/en/latest/guide/ddd-3-options.html#nested-rules).

jordandsullivan:
> My suggestion is that the executor should be only called by ZNE. The DD part is a "pre-processing", with input the initial circuit (e.g., GHZ state) and output the DD compiled quantum circuit. This is fed to ZNE. This should remove incompatibility types. [`ddd.insert_ddd_sequences`](https://mitiq.readthedocs.io/en/latest/apidoc.html#mitiq.ddd.insertion.insert_ddd_sequences) should be used. See the documentation [here](https://mitiq.readthedocs.io/en/latest/guide/ddd-4-low-level.html#inserting-ddd-sequences) and [here](https://mitiq.readthedocs.io/en/latest/guide/ddd-3-options.html#nested-rules).

Thanks for the tip @nathanshammah , that did it. Now the question is mainly how to find a combination of error levels and circuit depth for which using DDD and ZNE together is in fact useful. Unfortunately, DDD seems to do best when you insert a lot of idle time in the circuit (which I modeled by inserting identity gates), while ZNE does significantly worse. ZNE itself shines in lower depth circuits. 

Perhaps it would be useful to do a characterization sweep of the range of noise levels or circuit depths at which each technique was useful?

docs/source/examples/combine_ddd_zne.md
```diff
@@ -0,0 +1,192 @@
...
+      jupytext_version: 1.16.1
+  kernelspec:
+    display_name: Python 3 (ipykernel)
+    language: python
+    name: python3
+---
+
+# Composing techniques: Digital Dynamical Decoupling and Zero Noise Extrapolation
+
+Noise in quantum computers can arise from a variety of sources, and sometimes applying multiple error mitigation techniques can be more beneficial than applying a single technique alone.
```
cosenal, line 17: It would be nice to give some intuition of when composing is beneficial and why in those cases it works. I see you touch on that later down in the notebook, but I think it's good to say it upfront. Ideally also citing a reference.

docs/source/examples/combine_ddd_zne.ipynb
cosenal, line 1: With the documentation setup we have currently, you can only upload either the ipynb or the md file, not both. In fact, how come gitignore didn't pick this up? 🤔

docs/source/examples/combine_ddd_zne.md
```diff
@@ -0,0 +1,192 @@
...
+    """
+    Execute a circuit with R_z dephasing noise of strength ``rz_noise`` and depolarizing noise ``depolar_noise``
+    """
+    # Simulate systematic dephasing (coherent RZ) on each qubit for each moment.
+    circuit = circuit.with_noise(cirq.rz(rz_noise))
+
+    # Simulate systematic depolarizing on each qubit for each moment.
+    circuit = circuit.with_noise(cirq.bit_flip(depolar_noise))
+
+    # Measure out all qubits
```
cosenal, line 89: Usually I am not a fan of a comment for each line of code like in this function, but I think for tutorials it works great 😊

docs/source/examples/combine_ddd_zne.md
```diff
@@ -0,0 +1,192 @@
...
+Since the default execute function includes noise in this case, we see that most of the results are all ones or all zeros, corresponding to perfect entanglement between all qubits.
+
+A few others however are one or more bitflips away from either of these states. It is these types of errors we want to mitigate. 
+
+
+## Observable
+
+In this example, we just want to check if we have achieved entanglement across all qubits. In this case, we will measure the observable $⨂_{i=1}^n​X_i$ or measuring `X` on all qubits. 
+
+This corresponds to projecting the state onto either the $∣+⟩^{⊗n}$ or $∣−⟩^{⊗n}$ basis. For a perfect GHZ state, this observable will have an expectation value of 1, which corresponds to all qubits being in the same state.
```
cosenal, line 122: LaTex syntax should be working here, is it not?

mitiq/benchmarks/mirror_circuits.py
```diff
@@ -149,7 +149,7 @@ def generate_mirror_circuit(
             ``cirq.Circuit``.
 
     Returns:
-        A randomized mirror circuit.
+        A randomized mirror circuit and correct bitstring.
```
cosenal, line 152: what does _correct_ mean here?

cosenal (commented):
Left some preliminary comments, but I haven't run the notebook yet.

docs/source/examples/combine_ddd_zne.md
jordandsullivan: It is, it was just quicker for me to copy/paste the unicode versions of these symbols. Happy to switch back to latex if there's an issue.

mitiq/benchmarks/mirror_circuits.py
jordandsullivan: Apologies, I accidentally committed this one to the wrong branch. It was intended as part of #2353, but I failed to notice I hadn't successfully switched branches. Mirror circuits are designed such that there is only one expected bitstring, and all others can be considered due to noise.

docs/source/examples/combine_ddd_zne.md
jordandsullivan: I am not aware of any references which specifically combine DDD and ZNE error mitigation. I think a more extensive study would be needed to determine the exact circumstances under which particular QEM techniques are useful to compose. 

Section IV C. of this [QEM survey paper](https://arxiv.org/pdf/2210.00921#section*.30) details some combinations which have been used in the existing literature, but does not mention DDD in combination with any other techniques, nor does it compare the relative efficacy of the various methods.

The paper overall leaves it as an open question to determine the optimal combination of techniques for a given use case in [section VI](https://arxiv.org/pdf/2210.00921#section*.36). 

I can certainly add this paper as a reference; it just won't necessarily be useful for the specific topic of this tutorial.

docs/source/examples/combine_ddd_zne.ipynb
jordandsullivan: Not sure, but I can delete it in my next commit.

docs/source/examples/combine_ddd_zne.md
cosenal: Thanks for those references! I agree that it will not be useful for the specific topic of DDD+ZNE. Time to write an article of our own on the topic :)

docs/source/examples/combine_ddd_zne.ipynb
cosenal: The docs build is failing because of the double files, so you don't have any other choice than deleting the ipynb 😬

docs/source/examples/combine_ddd_zne.md
jordandsullivan: I was thinking the same thing!

docs/source/examples/combine_ddd_zne.ipynb
jordandsullivan: Deleted it in most recent commit

jordandsullivan:
I'm not sure why the output does not seem to be getting included in the docs build :/ https://mitiq--2345.org.readthedocs.build/en/2345/examples/combine_ddd_zne.html

cosenal:
@jordandsullivan hm? I can see the output in the doc build.

jordandsullivan:
> @jordandsullivan hm? I can see the output in the doc build.

Ah I just needed to do a hard reload (CMD+SHIFT+R) in my browser. It's showing up for me now :)

docs/source/examples/combine_ddd_zne.md
```diff
@@ -0,0 +1,241 @@
...
+
+In the our noise model, we will apply errors after each moment. Adding the additional identity gates within each "moment" should amplify the effect of these time-correlated errors.
+
+**Note:** Due to cirq's definition of moments, the first two gates of the circuit (`cirq.H(qubits[0])` and `cirq.CNOT(qubits[0], qubits[i])`) are in the same moment. This means the idle steps cannot be added between these gates, and therefore the time correlated noise will have less of an effect for the start of the circuit. 
+
++++
+
+## Noise model and executor
+
+**Importantly**, since DDD is designed to mitigate time-correlated (non-Markovian) noise, we simulate systematic $R_z$ rotations and depolarising noise applied to each qubit after each time step. This corresponds to noise which is strongly time-correlated and, therefore, likely to be mitigated by DDD. We
```
cosenal, line 100: `We` hanging at the end of this paragraph.

docs/source/examples/combine_ddd_zne.md
```diff
@@ -0,0 +1,241 @@
...
+## Noise model and executor
+
+**Importantly**, since DDD is designed to mitigate time-correlated (non-Markovian) noise, we simulate systematic $R_z$ rotations and depolarising noise applied to each qubit after each time step. This corresponds to noise which is strongly time-correlated and, therefore, likely to be mitigated by DDD. We 
+
+We use an [executor function](../guide/executors.md) to run the quantum circuit with the noise model applied.
+
+```{code-cell}
+def execute(
+    circuit: cirq.Circuit, 
+    rz_noise: float = 0.01,
```
cosenal, line 107: How were those default values picked?

docs/source/examples/combine_ddd_zne.md
```diff
@@ -0,0 +1,241 @@
...
+```{code-cell}
+res = execute(circuit)
+res.to_dict() # Dictionary for more convenient visualization
+```
+
+The `'counts'` key represents the number of times a particular measurement bitstring occurred. 
+
+Since the default execute function includes noise in this case, we see that most of the results are all ones or all zeros, corresponding to perfect entanglement between all qubits.
+
+A few others, like `'110111'` however are one or more bitflips away from either of these states. It is these types of errors we want to mitigate.
```
cosenal, line 145: I like the level of details in this explanation!

docs/source/examples/combine_ddd_zne.md
```diff
@@ -0,0 +1,241 @@
...
+  display_name: Python 3 (ipykernel)
+  language: python
+  name: python3
+---
+
+# Composing techniques: Digital Dynamical Decoupling and Zero Noise Extrapolation
+
+Noise in quantum computers can arise from a variety of sources, and sometimes applying multiple error mitigation techniques can be more beneficial than applying a single technique alone.
+
+Here we apply a combination of Digital Dynamical Decoupling (DDD) and Zero Noise Extrapolation (ZNE) on a GHZ state.
```
cosenal, line 19: It would be nice to give some intuition on why GHZ state preparation has been picked for this example. It becomes more clear later in the notebook, but it'd be nice for the reader to have something upfront.

docs/source/examples/combine_ddd_zne.md
```diff
@@ -0,0 +1,241 @@
...
+print("Unmitigated expectation value:", "{:.5f}".format(noisy.real))
+
+print("Expectation value with DDD:", "{:.5f}".format(ddd_noisy.real))
+```
+
+### Zero Noise Extrapolation alone
+
++++
+
+For comparison, we then apply ZNE one our original circuit (without DDD sequences inserted).
```
cosenal, line 211: s/one/on

docs/source/examples/combine_ddd_zne.md
```diff
@@ -0,0 +1,241 @@
...
+
+### Digital Dynamical Decoupling + Zero noise extrapolation
+
++++
+
+Finally, we apply a combination of DDD and ZNE.
+DDD is applied first to apply the control pulses to each circuit which ZNE runs to do its extrapolation.
+
+```{code-cell}
+combined_executor = zne.mitigate_executor(execute, observable=obs, scale_noise=zne.scaling.folding.fold_global)
```
cosenal, line 233: hmm, shouldn't this use `noisy_exec` instead of `execute`? Otherwise you are comparing scenarios with different levels of noise. And in fact, shouldn't this just be `zne_executor` as defined above. And then you want to apply the ZNE executor to the DDD circuit to combine techniques, i.e.,
```
combined_result = zne_executor(ddd_circuit)
```

cosenal (changes requested):
The one at the bottom is the most important comment.

docs/source/examples/combine_ddd_zne.md
jordandsullivan: Thanks!

docs/source/examples/combine_ddd_zne.md
jordandsullivan: Done

docs/source/examples/combine_ddd_zne.md
jordandsullivan: Thanks, good catch on the `noisy_exec` call, those were comparing two different noise levels. I corrected that, and still find that the performance of DDD + ZNE is better than either individually.

docs/source/examples/combine_ddd_zne.md
jordandsullivan: ?

docs/source/examples/combine_ddd_zne.md
jordandsullivan: Added explanation

docs/source/examples/combine_ddd_zne.md
jordandsullivan: Deleted

docs/source/examples/combine_ddd_zne.md
jordandsullivan: True also that I can just call `zne_executor` on the `ddd_circuit` rather than instantiating a new executor called `combined_executor`. They are equivalent as written, but I agree it's unnecessary -- it was a holdover from an earlier way of combining techniques. I agree it's much clearer just to call `zne_executor(ddd_circuit)`. Updated in latest commit.

docs/source/examples/combine_ddd_zne.md
cosenal: I was just pointing out the typo. You wanted `on` instead of `one`. I was using the sed notation `s/match/replacement`, unix slang :D

docs/source/examples/combine_ddd_zne.md
cosenal: Good to hear that the combined value is still winning!

cosenal (approved):

cosenal:
LGTM!

PR #2347, opened 2024-05-04
Noise Scaling for LRE

### LINKED ISSUE #2307: LRE Functions: Noise Scaling
Use this issue to add the functionality for creating noise-scaled circuits using LRE. As discussed in the RFC, we are chooisng to create a separate module `mitiq/lre` instead of relying on `mitiq/zne`. 

 `mitiq/zne/scaling/layer_scaling.py` requires a list of which layers a user wants to fold. With the LRE method, each layer gets folded differently depending on the chosen `int` value known as the `fold_multiplier` in the Jupyter notebook and `minimum_noise_scaling_gap` in the LRE paper.  

If a user wants to fold only a specific list of layers, they won't be able to use multi-variate Richardson extrapolation.

1. A function to find the number of layers in the input circuit without terminal measurements. 
2. If a user wants to chunk a large input circuit into smaller chunks, a function to create the chunked layers. 
3. A function to generate the scale factor vectors required by the specific folding pattern
4. A function to generate multiple noise-scaled but logically equivalent circuits.

purva-thakre:
Fixes #2307 

## Description

---

### License

### DIFF
--- a/mitiq/interface/conversions.py
+++ b/mitiq/interface/conversions.py
@@ -6,7 +6,6 @@
 """Functions for converting to/from Mitiq's internal circuit representation."""
 
 from functools import wraps
-from itertools import chain
 from typing import Any, Callable, Dict, Iterable, Optional, Tuple, cast
 
 import cirq
@@ -318,12 +317,6 @@ def accept_qprogram_and_validate(
             )
             out_circuits = [out_circuit]
 
-        # for the case when the circuit modifier returns a list of circuits
-        # flatten the list of lists into a list
-        if isinstance(out_circuits[0], list):
-            new_out_circuits = list(chain(*out_circuits))
-            out_circuits = new_out_circuits
-
         circuits_to_return = []
         for out_circuit in out_circuits:
             # Post atomic conversion
@@ -386,13 +379,9 @@ def accept_qprogram_and_validate(
                     out_circuit.measure(q, c)
             circuits_to_return.append(out_circuit)
 
-        try:
-            if not one_to_many:
-                assert len(circuits_to_return) == 1
-                return circuits_to_return[0]
-        except AssertionError:
-            # for the case when the circuit modifier returns a list of circuits
-            assert len(out_circuits) == len(circuits_to_return)
+        if not one_to_many:
+            assert len(circuits_to_return) == 1
+            return circuits_to_return[0]
 
         return circuits_to_return
 
--- a/mitiq/interface/conversions.py
+++ b/mitiq/interface/conversions.py
@@ -6,6 +6,7 @@
 """Functions for converting to/from Mitiq's internal circuit representation."""
 
 from functools import wraps
+from itertools import chain
 from typing import Any, Callable, Dict, Iterable, Optional, Tuple, cast
 
 import cirq
@@ -317,6 +318,12 @@ def accept_qprogram_and_validate(
             )
             out_circuits = [out_circuit]
 
+        # for the case when the circuit modifier returns a list of circuits
+        # flatten the list of lists into a list
+        if isinstance(out_circuits[0], list):
+            new_out_circuits = list(chain(*out_circuits))
+            out_circuits = new_out_circuits
+
         circuits_to_return = []
         for out_circuit in out_circuits:
             # Post atomic conversion
@@ -379,9 +386,13 @@ def accept_qprogram_and_validate(
                     out_circuit.measure(q, c)
             circuits_to_return.append(out_circuit)
 
-        if not one_to_many:
-            assert len(circuits_to_return) == 1
-            return circuits_to_return[0]
+        try:
+            if not one_to_many:
+                assert len(circuits_to_return) == 1
+                return circuits_to_return[0]
+        except AssertionError:
+            # for the case when the circuit modifier returns a list of circuits
+            assert len(out_circuits) == len(circuits_to_return)
 
         return circuits_to_return
 
--- a/mitiq/lre/multivariate_scaling/layerwise_folding.py
+++ b/mitiq/lre/multivariate_scaling/layerwise_folding.py
@@ -136,6 +136,10 @@ def multivariate_layer_scaling(
 ) -> List[Circuit]:
     """Defines the noise scaling function required for Layerwise Richardson
     Extrapolation."""
+    if degree < 1:
+        raise ValueError("Multinomial degree not >= to 1.")
+    if fold_multiplier < 1:
+        raise ValueError("Fold multiplier not >= to 1.")
     circuit_copy = deepcopy(input_circuit)
     terminal_measurements = _pop_measurements(circuit_copy)
 
--- a/mitiq/lre/multivariate_scaling/layerwise_folding.py
+++ b/mitiq/lre/multivariate_scaling/layerwise_folding.py
@@ -27,12 +27,6 @@ def _get_num_layers_without_measurements(input_circuit: Circuit) -> int:
         Args:
             input_circuit: Circuit of interest
 
-        Raises:
-            UnfoldableCircuitError:
-                * If the circuit has intermediate measurements
-                * If the circuit has non-unitary channels which are not
-                    terminal measurements
-
         Returns:
             num_layers: the number of layers in the input circuit without the
                 terminal measurements
@@ -59,8 +53,6 @@ def _get_chunks(
                     returned
                 * when num_chunks == 1, the entire circuit is chunked into 1
                     layer
-
-
         Returns:
             split_circuit: Circuit of interest split into approximately equal
                 chunks
@@ -111,7 +103,7 @@ def _get_scale_factor_vectors(
     num_layers = len(circuit_chunks)
 
     # Find the exponents of all the monomial terms required for the folding
-    # pattern
+    # pattern.
     pattern_full = []
     for i in range(degree + 1):
         for j in itertools.combinations_with_replacement(range(num_layers), i):
@@ -170,15 +162,10 @@ def multivariate_layer_scaling(
     Raises:
         ValueError:
             When the degree for the multinomial is not greater than or
-                equal to 1.
-        ValueError:
-            When the fold multiplier to scale the circuit is greater than or
-                equal to 1.
-        ValueError:
-            When the number of chunks for a large circuit is 0.
-        ValueError:
-            When the number of chunks in a circuit is greater than the number
-                of layers in the input circuit.
+                equal to 1; when the fold multiplier to scale the circuit is
+                greater than/equal to 1; when the number of chunks for a
+                large circuit is 0 or when the number of chunks in a circuit is
+                greater than the number of layers in the input circuit.
 
     """
     if degree < 1:
--- a/mitiq/lre/multivariate_scaling/layerwise_folding.py
+++ b/mitiq/lre/multivariate_scaling/layerwise_folding.py
@@ -144,6 +144,16 @@ def multivariate_layer_scaling(
     Defines the noise scaling function required for Layerwise Richardson
     Extrapolation as defined in :cite:`Russo_2024_LRE`.
 
+    Note that this method only works for the multivariate extrapolation
+    methods. It does not allows a user to choose which layers in the input
+    circuit will be scaled.
+
+    .. seealso::
+
+        If you would prefer to choose the layers for unitary
+        folding, use :func:`mitiq.zne.scaling.layer_scaling.get_layer_folding`
+        instead.
+
     Args:
         input_circuit: Circuit to be scaled
         degree: Degree of the multivariate polynomial
--- a/mitiq/lre/multivariate_scaling/layerwise_folding.py
+++ b/mitiq/lre/multivariate_scaling/layerwise_folding.py
@@ -5,3 +5,34 @@
 
 """Functions for layerwise folding of input circuits to allow for multivariate
 extrapolation."""
+
+from copy import deepcopy
+
+import cirq
+
+from mitiq.utils import _pop_measurements
+from mitiq.zne.scaling.folding import _check_foldable
+
+
+def _get_num_layers_without_measurements(input_circuit: cirq.Circuit) -> int:
+    """Checks if the circuit has non-terminal measurements and returns the
+    number of layers in the input circuit without the terminal measurements.
+
+        Args:
+        circuit: Checks whether this circuit is able to be folded.
+
+        Raises:
+            UnfoldableCircuitError:
+                * If the circuit has intermediate measurements.
+                * If the circuit has non-unitary channels which are not
+                terminal measurements.
+
+        Returns:
+            num_layers: the number of layers in the input circuit without the
+            terminal measurements.
+    """
+    circuit = deepcopy(input_circuit)
+    _check_foldable(circuit)
+    _pop_measurements(circuit)
+    num_layers = len(circuit)
+    return num_layers
--- a/mitiq/lre/multivariate_scaling/layerwise_folding.py
+++ b/mitiq/lre/multivariate_scaling/layerwise_folding.py
@@ -15,6 +15,7 @@ import numpy as np
 from cirq import Circuit
 
 from mitiq import QPROGRAM
+from mitiq.interface import accept_qprogram_and_validate
 from mitiq.utils import _append_measurements, _pop_measurements
 from mitiq.zne.scaling import fold_gates_at_random
 from mitiq.zne.scaling.folding import _check_foldable
@@ -131,6 +132,7 @@ def _get_scale_factor_vectors(
     ]
 
 
+@accept_qprogram_and_validate
 def multivariate_layer_scaling(
     input_circuit: Circuit,
     degree: int,
--- a/mitiq/lre/multivariate_scaling/layerwise_folding.py
+++ b/mitiq/lre/multivariate_scaling/layerwise_folding.py
@@ -30,6 +30,7 @@ def _get_num_layers_without_measurements(input_circuit: Circuit) -> int:
         Returns:
             num_layers: the number of layers in the input circuit without the
                 terminal measurements
+
     """
 
     _check_foldable(input_circuit)
@@ -56,12 +57,21 @@ def _get_chunks(
         Returns:
             split_circuit: Circuit of interest split into approximately equal
                 chunks
+
+        Raises:
+            ValueError:
+                When the number of chunks for the input circuit is larger than
+                    the number of layers in the input circuit.
+
+            ValueError:
+                When the number of chunks is less than 1.
+
     """
     num_layers = _get_num_layers_without_measurements(input_circuit)
     if num_chunks is None:
         num_chunks = num_layers
 
-    if num_chunks <= 0:
+    if num_chunks < 1:
         raise ValueError(
             "Number of chunks should be greater than or equal to 1."
         )
--- a/mitiq/lre/multivariate_scaling/layerwise_folding.py
+++ b/mitiq/lre/multivariate_scaling/layerwise_folding.py
@@ -8,11 +8,14 @@ extrapolation."""
 
 import itertools
 from copy import deepcopy
+from typing import Callable
 
 import cirq
 import numpy as np
 
-from mitiq.utils import _pop_measurements
+from mitiq import QPROGRAM
+from mitiq.utils import _append_measurements, _pop_measurements
+from mitiq.zne.scaling import fold_gates_at_random
 from mitiq.zne.scaling.folding import _check_foldable
 
 
@@ -120,3 +123,39 @@ def _get_scale_factor_vectors(
         tuple(2 * num_folds + 1 for num_folds in pattern)
         for pattern in pattern_full
     ]
+
+
+def multivariate_layer_scaling(
+    input_circuit: cirq.Circuit,
+    degree: int,
+    fold_multiplier: int,
+    num_chunks: int = 1,
+    folding_method: Callable[
+        [QPROGRAM, float], QPROGRAM
+    ] = fold_gates_at_random,
+):
+    """Defines the noise scaling function required for Layerwise Richardson
+    Extrapolation."""
+    circuit_copy = deepcopy(input_circuit)
+    terminal_measurements = _pop_measurements(circuit_copy)
+
+    scaling_pattern = _get_scale_factor_vectors(
+        circuit_copy, degree, fold_multiplier, num_chunks
+    )
+
+    chunks = _get_chunks(circuit_copy, num_chunks)
+
+    multiple_folded_circuits = []
+    for scale_factor_vector in scaling_pattern:
+        folded_circuit = cirq.Circuit()
+        for chunk, scale_factor in zip(chunks, scale_factor_vector):
+            if scale_factor == 1:
+                folded_circuit += chunk
+            else:
+                chunks_circ = cirq.Circuit(chunk)
+                folded_chunk_circ = folding_method(chunks_circ, scale_factor)
+                folded_circuit += folded_chunk_circ
+            _append_measurements(folded_circuit, terminal_measurements)
+        multiple_folded_circuits.append((folded_circuit))
+
+    return multiple_folded_circuits
--- a/mitiq/lre/multivariate_scaling/layerwise_folding.py
+++ b/mitiq/lre/multivariate_scaling/layerwise_folding.py
@@ -61,7 +61,7 @@ def _get_chunks(
     if num_chunks is None:
         num_chunks = num_layers
 
-    if num_chunks == 0:
+    if num_chunks <= 0:
         raise ValueError(
             "Number of chunks should be greater than or equal to 1."
         )
--- a/mitiq/lre/multivariate_scaling/layerwise_folding.py
+++ b/mitiq/lre/multivariate_scaling/layerwise_folding.py
@@ -4,7 +4,8 @@
 # LICENSE file in the root directory of this source tree.
 
 """Functions for layerwise folding of input circuits to allow for multivariate
-extrapolation."""
+extrapolation as defined in :cite:`Russo_2024_LRE`.
+"""
 
 import itertools
 from copy import deepcopy
@@ -30,11 +31,11 @@ def _get_num_layers_without_measurements(input_circuit: Circuit) -> int:
             UnfoldableCircuitError:
                 * If the circuit has intermediate measurements.
                 * If the circuit has non-unitary channels which are not
-                terminal measurements.
+                    terminal measurements.
 
         Returns:
             num_layers: the number of layers in the input circuit without the
-            terminal measurements.
+                terminal measurements.
     """
     circuit = deepcopy(input_circuit)
     _check_foldable(circuit)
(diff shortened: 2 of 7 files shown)


mitiq/zne/scaling/layer_scaling.py
```diff
@@ -24,6 +24,15 @@ def layer_folding(
...
     """Applies a variable amount of folding to select layers of a circuit.
 
+    Note that this method only works for the univariate extrapolation methods. 
+    It allows a user to choose which layers in the input circuit will be scaled
+    The layerwise folding required for multivariate extrapolation is different
+    as the layers in the input circuit have to be scaled in a specific pattern.
+    The required specific pattern for multivariate extrapolation does not allow
+    a user to provide a choice of which layers to fold. If you would prefer to
+    use a multivariate extrapolation method for unitary
+    folding, use the functions available in `mitiq/lre` instead.
```
purva-thakre, line 34: When this PR is close to being merged, replace `mitiq/lre` with the `seealso` directive from Sphinx. 

https://www.sphinx-doc.org/en/master/usage/restructuredtext/directives.html#directive-seealso

mitiq/zne/scaling/layer_scaling.py
```diff
@@ -24,6 +24,15 @@ def layer_folding(
...
     """Applies a variable amount of folding to select layers of a circuit.
 
+    Note that this method only works for the univariate extrapolation methods.
+    It allows a user to choose which layers in the input circuit will be scaled
+    The layerwise folding required for multivariate extrapolation is different
+    as the layers in the input circuit have to be scaled in a specific pattern.
+    The required specific pattern for multivariate extrapolation does not allow
+    a user to provide a choice of which layers to fold. If you would prefer to
+    use a multivariate extrapolation method for unitary
+    folding, use the functions available in `mitiq/lre` instead.
```
purva-thakre, line 34: When this PR is close to being merged, replace `mitiq/lre` with the `seealso` directive from Sphinx. 

https://www.sphinx-doc.org/en/master/usage/restructuredtext/directives.html#directive-seealso

mitiq/lre/multivariate_scaling/layerwise_folding.py
```diff
@@ -0,0 +1,122 @@
...
+    )
+    return list(split_circuit)
+
+
+def _get_scale_factor_vectors(
+    input_circuit: cirq.Circuit,
+    degree: int,
+    fold_multiplier: int,
+    num_chunks: int = 1,
+) -> list[tuple[int]]:
```
purva-thakre, line 85: To do: Understand this `List comprehension has incompatible type List[Tuple[Any, ...]]; expected List[Tuple[int]]`

https://realpython.com/python-type-checking/
https://mypy.readthedocs.io/en/stable/kinds_of_types.html#tuple-types
https://mypy.readthedocs.io/en/stable/builtin_types.html

mitiq/lre/multivariate_scaling/layerwise_folding.py
```diff
@@ -0,0 +1,122 @@
...
+        for i in range(num_chunks)
+    )
+    return list(split_circuit)
+
+
+def _get_scale_factor_vectors(
+    input_circuit: cirq.Circuit,
+    degree: int,
+    fold_multiplier: int,
+    num_chunks: int = 1,
```
purva-thakre, line 84: It would be better to use the default value as `None` for the parameter `num_chunks`. Making changes to this currently leads to the circuit getting chunked for some reason. Figure this out; make changes.

mitiq/lre/multivariate_scaling/layerwise_folding.py
cosenal: 1. `tuple[int]` is the type of a tuple with only one element. If you want the tuple to have an arbitrary number of integer elements, then you want to type it as `tuple[int, ...]`
2. The `Any` is because mypy can't infer that the tuple elements that are being generated are integer. It should go away with fix 1.
3. Instead of built-in types `list` and `tuple`, it's better to use `List` and `Tuple` from the typing module, i.e., `from typing import List, Tuple` and then `List[Tuple[int, ...]]`

mitiq/lre/tests/test_layerwise_folding.py
```diff
@@ -0,0 +1,244 @@
...
+        test_input, degree, test_fold_multiplier
+    )
+
+    assert calculated_scale_factor_vectors == expected_scale_factor_vectors
+
+
+@pytest.mark.parametrize(
+    "test_input, degree, test_fold_multiplier, test_chunks, expected_size",
+    [
+        (test_circuit1, 1, 0, 1, 2),
```
purva-thakre, line 212: To do: Check if the expected size here is accurate here. 

On  this line, the entire circuit is chunked into 1 chunk, the degree is 1 and the fold_multiplier is 0 i.e. there should be no scaling. The issue here is with degree 1 and 1 layer in the input circuit and how the polynomial terms would look like.

mitiq/lre/multivariate_scaling/layerwise_folding.py
```diff
@@ -0,0 +1,161 @@
...
+def multivariate_layer_scaling(
+    input_circuit: Circuit,
+    degree: int,
+    fold_multiplier: int,
+    num_chunks: Optional[int] = None,
+    folding_method: Callable[
+        [QPROGRAM, float], QPROGRAM
+    ] = fold_gates_at_random,
+) -> List[Circuit]:
+    """Defines the noise scaling function required for Layerwise Richardson
```
purva-thakre, line 137: Work on this docstring

mitiq/lre/multivariate_scaling/layerwise_folding.py
```diff
@@ -0,0 +1,195 @@
...
+    Raises:
+        ValueError:
+            When the degree for the multinomial is not >= to 1.
+        ValueError:
+            When the fold multiplier to scale the circuit is not >= to 1.
+        ValueError:
+            When the number of chunks for a large circuit is 0.
+        ValueError:
+            When the number of chunks in a circuit is greater than the number
+                of layers in the input circuit.
```
purva-thakre, line 166: All of these `ValueError` messages could be listed as a bullet list. Attempts to do so resulted in build errors.

purva-thakre:
@cosenal @vprusso This is ready for a review up to using layerwise folding on cirq circuits. 

I am working on figuring out why layerwise folding on a pyquil circuit raises errors. Once this is fixed, laywerise folding for non-cirq circuits will be added.

mitiq/lre/multivariate_scaling/layerwise_folding.py
```diff
@@ -0,0 +1,195 @@
...
+def _get_chunks(
+    input_circuit: Circuit, num_chunks: Optional[int] = None
+) -> List[Circuit]:
+    """Splits a circuit into approximately equal chunks.
+
+    Adapted from:
+    https://stackoverflow.com/questions/2130016/splitting-a-list-into-n-parts-of-approximately-equal-length
+
+        Args:
+            input_circuit: Circuit of interest.
```
vprusso, line 56: [nit]: Sometimes comment lines are terminated with period--other times they are not. Either is fine, (but I prefer the former) but consistency would be good in either case.

mitiq/lre/multivariate_scaling/layerwise_folding.py
```diff
@@ -0,0 +1,195 @@
...
+def _get_scale_factor_vectors(
+    input_circuit: Circuit,
+    degree: int,
+    fold_multiplier: int,
+    num_chunks: Optional[int] = None,
+) -> List[Tuple[Any, ...]]:
+    """Returns the patterned scale factor vectors required for multivariate
+    extrapolation.
+
+        Args:
```
vprusso, line 97: The `num_chunks` variable is missing from this list.

mitiq/lre/multivariate_scaling/layerwise_folding.py
```diff
@@ -0,0 +1,195 @@
...
+
+    circuit_chunks = _get_chunks(input_circuit, num_chunks)
+    num_layers = len(circuit_chunks)
+
+    # find the exponents of all the monomial terms required for the folding
+    # pattern
+    pattern_full = []
+    for i in range(degree + 1):
+        for j in itertools.combinations_with_replacement(range(num_layers), i):
+            pattern = [0] * num_layers
```
vprusso, line 115: Instead of defining this as a list and then converting it to an `np.array`, you can just define it as a `np.array` of zeros

```py
pattern = np.zeros(num_layers)
```

then

```py
pattern_full.append(tuple(fold_multiplier * pattern))
```

mitiq/lre/multivariate_scaling/layerwise_folding.py
```diff
@@ -0,0 +1,195 @@
...
+            # get the monomial terms in graded lexicographic order
+            for index in j:
+                pattern[index] += 1
+            # use fold multiplier on the folding pattern to figure out which
```
vprusso, line 119: For comments, would be good to use proper grammatical syntax, i.e.:

```py
# Use the fold multiplier on the folding pattern to determine which layers will be scaled.
```

etc.

mitiq/lre/multivariate_scaling/layerwise_folding.py
```diff
@@ -0,0 +1,195 @@
...
+            the input circuit is unchanged.
+        folding_method: Unitary folding method. Default is
+            :func:`fold_gates_at_random`
+
+    Returns:
+        Multiple folded variations of the input circuit
+
+    Raises:
+        ValueError:
+            When the degree for the multinomial is not >= to 1.
```
vprusso, line 159: It's a bit confusing to read the `ValueError` comment and see the logic in the code as they are worded in a way that is stated as the negation of the thing being checked. Perhaps word these in the same way in which the logic is being invoked to check the condition?

So, for instance, this would be

        ValueError:
            When the degree for the multinomial is < 1

etc.

mitiq/lre/multivariate_scaling/layerwise_folding.py
```diff
@@ -0,0 +1,195 @@
...
+            When the fold multiplier to scale the circuit is not >= to 1.
+        ValueError:
+            When the number of chunks for a large circuit is 0.
+        ValueError:
+            When the number of chunks in a circuit is greater than the number
+                of layers in the input circuit.
+
+    """
+    if degree < 1:
+        raise ValueError("Multinomial degree not >= to 1.")
```
vprusso, line 170: I would word these conditions in a way that makes parsing the error a bit easier for the user. For instance:

```py
raise ValueError("Multinomial degree must be greater than 1.")
```

mitiq/zne/scaling/layer_scaling.py
```diff
@@ -24,6 +24,15 @@ def layer_folding(
 ) -> cirq.Circuit:
     """Applies a variable amount of folding to select layers of a circuit.
 
+    Note that this method only works for the univariate extrapolation methods.
+    It allows a user to choose which layers in the input circuit will be scaled
```
vprusso, line 28: Period at end of `scaled.`.

mitiq/lre/multivariate_scaling/layerwise_folding.py
```diff
@@ -0,0 +1,195 @@
...
+    num_layers = _get_num_layers_without_measurements(input_circuit)
+    if num_chunks is None:
+        num_chunks = num_layers
+
+    if num_chunks == 0:
+        raise ValueError("The number of chunks should be >= to 1.")
+
+    if num_chunks > num_layers:
+        raise ValueError(
+            "Number of chunks > the number of layers in the circuit."
```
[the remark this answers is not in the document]
purva-thakre: > maybe based on the mitiq line threshold it can't

Exactly this! line length = 79 for mitiq

mitiq/lre/multivariate_scaling/layerwise_folding.py
purva-thakre: missed this because its a private function and does not show up in the API doc.

mitiq/lre/multivariate_scaling/layerwise_folding.py
```diff
@@ -0,0 +1,195 @@
...
+def _get_num_layers_without_measurements(input_circuit: Circuit) -> int:
+    """Checks if the circuit has non-terminal measurements and returns the
+    number of layers in the input circuit without the terminal measurements.
+
+        Args:
+            input_circuit: Checks whether this circuit is able to be folded.
+
+        Raises:
+            UnfoldableCircuitError:
+                * If the circuit has intermediate measurements.
```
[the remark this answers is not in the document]
purva-thakre: I'll search around for what the usual convention for this is. 

Anyway, I need to figure out how to raise errors for a public function when private functions also raise other errors. https://github.com/unitaryfund/mitiq/pull/2347#discussion_r1633527850

mitiq/lre/multivariate_scaling/layerwise_folding.py
```diff
@@ -0,0 +1,195 @@
...
+    circuit = deepcopy(input_circuit)
+    _check_foldable(circuit)
+    _pop_measurements(circuit)
+    num_layers = len(circuit)
+    return num_layers
+
+
+def _get_chunks(
+    input_circuit: Circuit, num_chunks: Optional[int] = None
+) -> List[Circuit]:
```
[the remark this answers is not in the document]
purva-thakre: This is a good question. The validate workflow only runs for python 3.11 which is used for type checking by mypy. 

https://github.com/unitaryfund/mitiq/blob/837ea83546f16e39867e3ee74cd27996b1d7509d/.github/workflows/build.yml#L25-L32

mitiq/lre/multivariate_scaling/layerwise_folding.py
purva-thakre: We do want to allow `degree = 1` as well.

mitiq/interface/conversions.py
```diff
@@ -317,6 +318,12 @@ def new_function(circuit: QPROGRAM, *args: Any, **kwargs: Any) -> QPROGRAM:
             )
             out_circuits = [out_circuit]
 
+        # for the case when the circuit modifier returns a list of circuits
+        # flatten the list of lists into a list
+        if isinstance(out_circuits[0], list):
+            new_out_circuits = list(chain(*out_circuits))
+            out_circuits = new_out_circuits
+
```
purva-thakre, line 326: This works to the point of passing currently existing unit tests with the  `accept_qprogram_and_validate` decorator. 

If I use a qiskit circuit in ` multivariate_layer_scaling`, it fails with

```
CircuitConversionError: Circuit could not be converted from an internal Mitiq type to a circuit of type qiskit.
```

purva-thakre:
I think it's better to limit this PR to Cirq circuits only because I need to spend some more time understanding how the Mitiq interface decorators work. I have used the default args for the decorators in the past without any issues. This PR requires using the optional arguments.

Two different ways to implement the optional args led to errors. Either the `one_to_many` argument was ignored (first code block) or `_pop_measurement` was used on a Qiskit circuit instead of converting the circuit to a cirq circuit first (second code block). 

```python
@accept_qprogram_and_validate
def multivariate_layer_scaling( input_circuit, degree, ............, one_to_many = True)
```

Similar to how the optional arg is used in [PEC](https://github.com/unitaryfund/mitiq/blob/bf82e84410e1805437e15f1d26b0c5be8e8dcb8a/mitiq/pec/sampling.py#L199), I also tried changing `multivariate_layer_scaling` to `_cirq_multivariate_layer_scaling` such that the new `multivariate_layer_scaling` only accepted a QPROGRAM as input. 
```python
 noise_scaling_func = accept_qprogram_and_validate(
      _cirq_multivariate_layer_scaling,
      one_to_many=True,
  )
  scaled_circuits = noise_scaling_func(
      input_circuit, degree, fold_multiplier, num_chunks, folding_method
  )
 return scaled_circuits
```

jordandsullivan:
Sounds good Purva, we can limit this PR for Cirq for now.

mitiq/interface/conversions.py
purva-thakre: Resolving for now as there is a separate issue for this. 

https://github.com/unitaryfund/mitiq/issues/2416

purva-thakre:
Except for 1 unresolved comment https://github.com/unitaryfund/mitiq/pull/2347#discussion_r1633632306, this PR is ready for another review.

mitiq/lre/multivariate_scaling/layerwise_folding.py
```diff
@@ -0,0 +1,199 @@
...
+
+        Args:
+            input_circuit: Circuit of interest
+
+        Returns:
+            num_layers: the number of layers in the input circuit without the
+                terminal measurements
+    """
+
+    _check_foldable(input_circuit)
```
cosenal, line 35: This can raise an UnfoldableCircuitError, it should be documented in the docstring of this caller function as well.

mitiq/lre/multivariate_scaling/layerwise_folding.py
```diff
@@ -0,0 +1,199 @@
...
+    https://stackoverflow.com/questions/2130016/splitting-a-list-into-n-parts-of-approximately-equal-length
+
+        Args:
+            input_circuit: Circuit of interest
+            num_chunks: Number of desired approximately equal chunks
+                * when num_chunks == num_layers, the original circuit is
+                    returned
+                * when num_chunks == 1, the entire circuit is chunked into 1
+                    layer
+        Returns:
```
cosenal, line 56: Ditto as above; potential exceptions should be documented in docstrings

mitiq/lre/multivariate_scaling/layerwise_folding.py
```diff
@@ -0,0 +1,199 @@
...
+            split_circuit: Circuit of interest split into approximately equal
+                chunks
+    """
+    num_layers = _get_num_layers_without_measurements(input_circuit)
+    if num_chunks is None:
+        num_chunks = num_layers
```
cosenal, line 62: perhaps we can shortcut this case and `return input_circuit` directly

mitiq/lre/multivariate_scaling/layerwise_folding.py
```diff
@@ -0,0 +1,199 @@
...
+
+    if num_chunks == 0:
```
cosenal, line 64: may as well check for `< 1`, so it's protected from the case of negative values as well

mitiq/lre/multivariate_scaling/layerwise_folding.py
```diff
@@ -0,0 +1,199 @@
...
+                terminal measurements
+    """
+
+    _check_foldable(input_circuit)
+    circuit = deepcopy(input_circuit)
+    _pop_measurements(circuit)
+    return len(circuit)
+
+
+def _get_chunks(
```
cosenal, line 41: Even though it's a "private" function, it's better if we have unit tests for it.

mitiq/lre/tests/test_layerwise_folding.py
```diff
@@ -0,0 +1,269 @@
...
+        (1, 1, 5),
+        (9, 1, 1),
+        (5, 5, 1),
+        (5, 1, 5),
+        (1, 9, 1),
+        (1, 5, 5),
+        (1, 1, 9),
+    ]
+
+    i = 0
```
cosenal, line 53: you can use `enumerate` here.

```
for i, scale_factor_vector in enumerate(folding_pattern):
    ...
```

mitiq/lre/multivariate_scaling/layerwise_folding.py
purva-thakre: I had included it previously. https://github.com/unitaryfund/mitiq/pull/2347#discussion_r1633629577

We should agree on style guidelines for this project and update [the docs style guidelines section](https://mitiq.readthedocs.io/en/stable/contributing.html#style-guidelines) on docstrings. 

Couple of things that confused me:

- Since I am relying on private functions, a user won't be able to read the source of the error in the API. Which is why I felt it's better to document the raised errors in the public function. 
- Referring to Vincent's comment on raising errors from a function in a different module, the API-doc page only mentions [`UnfoldableCircuitError`](https://mitiq.readthedocs.io/en/stable/apidoc.html#mitiq.zne.scaling.folding.UnfoldableCircuitError) once as a defined exception. As `_check_foldable` is a private function, [the error raised by the function](https://github.com/unitaryfund/mitiq/blob/2f9c00a69739605ff86785e88757233af17af748/mitiq/zne/scaling/folding.py#L66) does not show up publicly.  
- Most of the raised errors in the public-facing function are specified in a paragraph. 
image
As we rely on google-style docstrings, there is no convention to allow lists in the description of the [different sections](https://google.github.io/styleguide/pyguide.html#doc-function-args) of a docstring. [Portions of our API-doc](https://mitiq.readthedocs.io/en/stable/apidoc.html#module-mitiq.cdr.cdr) have run-on sentences because someone used a list in the section description which appear as badly formatted sentences publicly.

| Intended      | in API-doc |
| ----------- | ----------- |
|image| image|

mitiq/lre/multivariate_scaling/layerwise_folding.py
purva-thakre: No, we want the input circuit to be split by each layer if `num_chunks = num_layers`. This makes it easier to define the steps in the main noise scaling function later. 

```
Original Circuit 

0: ───H───@───@───
          │   │
1: ───H───X───@───
              │
2: ───H───X───X───

When num_chunks is None, return original circuit

0: ───H───@───@───
          │   │
1: ───H───X───@───
              │
2: ───H───X───X───

When num_chunks is None, use num_chunks = num_layers

[0: ───H───

1: ───H───

2: ───H───,
 0: ───@───
      │
1: ───X───

2: ───X───,
 0: ───@───
      │
1: ───@───
      │
2: ───X───]
```

mitiq/lre/multivariate_scaling/layerwise_folding.py
purva-thakre: Ok. I figured it was covered by `test_get_scale_factor_vectors_with_chunking` as `_get_scale_factor_vectors` relies on `_get_chunks`.

purva-thakre:
This PR is ready to be merged once all checks pass. 

As of 1:45 pm CST, not all checks are being run when I pushed the latest commit to this PR. 

image

Looks like Github is aware of the issues and they are working on fixing it. https://www.githubstatus.com/

image

mitiq/lre/tests/test_layerwise_folding.py
```diff
@@ -0,0 +1,303 @@
...
+    "test_input, test_chunks, expected_chunks",
+    [
+        (test_circuit1 + test_circuit1 + test_circuit1, 3, 3),
+        (test_circuit1 + test_circuit1 + test_circuit1, 1, 1),
+        (test_circuit1 + test_circuit1 + test_circuit1, 5, 5),
+    ],
+)
+def test_get_num_chunks(test_input, test_chunks, expected_chunks):
+    """Verifies the chunking function works as expected."""
+    calculated_num_chunks = len(_get_chunks(test_input, test_chunks))
```
cosenal, line 86: Isn't the expectation here as simple as 
```
assert test_chunks = len(_get_chunks(test_input, test_chunks))
```
If so, we can get rid of the `expected_chunks` parameter and the test becomes easier to read it logically.

mitiq/lre/multivariate_scaling/layerwise_folding.py
```diff
@@ -0,0 +1,209 @@
...
+        folded_circuit = Circuit()
+        for chunk, scale_factor in zip(chunks, scale_factor_vector):
+            if scale_factor == 1:
+                folded_circuit += chunk
+            else:
+                chunks_circ = Circuit(chunk)
+                folded_chunk_circ = folding_method(chunks_circ, scale_factor)
+                folded_circuit += folded_chunk_circ
+        _append_measurements(folded_circuit, terminal_measurements)
+        multiple_folded_circuits.append((folded_circuit))
```
cosenal, line 207:
```suggestion
        multiple_folded_circuits.append(folded_circuit)
```

cosenal (approved):
LGTM

To re-run all checks, you will have to push another commit (even an empty one). I re-run it manually [here](https://github.com/unitaryfund/mitiq/actions/runs/9627892470) and everything looks fine.

mitiq/lre/tests/test_layerwise_folding.py
```diff
@@ -83,9 +83,7 @@ def test_get_num_layers(test_input, expected):
 )
 def test_get_num_chunks(test_input, test_chunks, expected_chunks):
     """Verifies the chunking function works as expected."""
-    calculated_num_chunks = len(_get_chunks(test_input, test_chunks))
-
-    assert calculated_num_chunks == expected_chunks
+    assert expected_chunks == len(_get_chunks(test_input, test_chunks))
```
cosenal, line 86: What I meant here is something further, that is, `test_chunks` is always equal to ` expected_chunks`, so you may as well get rid of one of the two parameters in the tests.

purva-thakre: ah, gotcha

cosenal (approved):
LGTM

mitiq/lre/multivariate_scaling/layerwise_folding.py
```diff
@@ -0,0 +1,209 @@
...
+    fold_multiplier: int,
+    num_chunks: Optional[int] = None,
+) -> List[Tuple[Any, ...]]:
+    """Returns the patterned scale factor vectors required for multivariate
+    extrapolation.
+
+        Args:
+            input_circuit: Circuit to be scaled
+            degree: Degree of the multivariate polynomial
+            num_chunks: Number of desired approximately equal chunks
```
vprusso, line 104: [nit]: Do we want to arrange the listing of arguments in the same order in which they are specified in the function prototype?

mitiq/lre/multivariate_scaling/layerwise_folding.py
```diff
@@ -0,0 +1,209 @@
...
+            fold_multiplier: Scaling gap required by unitary folding
+
+        Returns:
+            scale_factor_vectors: Multiple variations of scale factors for each
```
vprusso, line 108: [nit]: The term "multiple variations" might be misleading. There need not be any variations (i.e. they could all be the same across the board for each layer). Perhaps it might be better described as "A vector of scale factors where each component in the vector corresponds to the layer in the input circuit."

mitiq/lre/multivariate_scaling/layerwise_folding.py
```diff
@@ -0,0 +1,209 @@
...
+    circuit will be scaled.
+
+    .. seealso::
+
+        If you would prefer to choose the layers for unitary
+        folding, use :func:`mitiq.zne.scaling.layer_scaling.get_layer_folding`
+        instead.
+
+    Args:
+        input_circuit: Circuit to be scaled
```
vprusso, line 160: [nit]: Do we want to enforce that all comment lines end with a period (as is done for some of the items below?)

mitiq/lre/multivariate_scaling/layerwise_folding.py
```diff
@@ -0,0 +1,209 @@
...
+        degree: Degree of the multivariate polynomial
+        fold_multiplier: Scaling gap required by unitary folding
+        num_chunks: Number of desired approximately equal chunks. When the
+            number of chunks is the same as the layers in the input circuit,
+            the input circuit is unchanged.
+        folding_method: Unitary folding method. Default is
+            :func:`fold_gates_at_random`
+
+    Returns:
+        Multiple folded variations of the input circuit
```
vprusso, line 170: [nit]: Period at end of this sentence.

vprusso (approved):
(Minor comments only--but aside from those, LGTM!)

PR #2363, opened 2024-05-12
Add download notebook link in docs

### LINKED ISSUE #1619: Add download option button to tutorials in docs
## Issue Description

Add download option button to tutorials in docs, including `.ipynb` option. However, we would like to have the source code as it is, in terms of `.md` files, without adding .ipynb files in `mitiq/docs/source/examples/`, as Jupyter notebooks are hard to track in version control on Github. So, ideally, somehow the Jupyter notebooks should be created somehow (at documentation build or with another script), but still appear in the version of the docs that is hosted on readthedocs. It's a tricky issue :). 

## Proposed Solution

Find solution in other sphinx documentations.

## Additional References

See the top right button in this QuTiP notebook: https://nbviewer.org/urls/qutip.org/qutip-tutorials/tutorials-v4/time-evolution/003_qubit-dynamics.ipynb

cosenal:
Fixes #1619.

For a demo, see https://mitiq--2363.org.readthedocs.build/en/2363/examples/calibration-tutorial.html

### DIFF
--- a/docs/source/_templates/jupyter-download.html
+++ b/docs/source/_templates/jupyter-download.html
@@ -1,8 +1,0 @@
-{# Displays a link to download the Jupyter Notebook #}
-{% if notebook_link %}
-  <div class="tocsection sourcelink">
-    <a href="{{ pathto(notebook_link, true) }}">
-      <i class="fa-solid fa-book"></i> {{ _("Download Jupyter Notebook") }}
-    </a>
-  </div>
-{% endif %}
--- a/docs/source/_templates/jupyter-download.html
+++ b/docs/source/_templates/jupyter-download.html
@@ -1,0 +1,8 @@
+{# Displays a link to download the Jupyter Notebook #}
+{% if notebook_link %}
+  <div class="tocsection sourcelink">
+    <a href="{{ pathto('.' + notebook_link, true) }}">
+      <i class="fa-solid fa-book"></i> {{ _("Download Jupyter Notebook") }}
+    </a>
+  </div>
+{% endif %}
--- a/docs/source/_templates/jupyter-download.html
+++ b/docs/source/_templates/jupyter-download.html
@@ -1,7 +1,7 @@
 {# Displays a link to download the Jupyter Notebook #}
 {% if notebook_link %}
   <div class="tocsection sourcelink">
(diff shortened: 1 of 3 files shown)


jordandsullivan:
Nice! Thanks for implementing this. One thing I noticed is that it doesn't include the images referenced in the notebook, so for the example you link:

> For a demo, see https://mitiq--2363.org.readthedocs.build/en/2363/examples/calibration-tutorial.html

The image at the top doesn't load. In this case it's no big deal since the image is basically just a thumbnail
Screen Shot 2024-05-13 at 10 42 44 AM
Screen Shot 2024-05-13 at 10 40 30 AM

However, in cases where there's a diagram or figure included, it might be confusing to not have it missing. 

Looking through a handful of examples though, e.g. ([layerwise folding](https://mitiq--2363.org.readthedocs.build/en/2363/examples/layerwise-folding.html), [Pennylane](https://mitiq--2363.org.readthedocs.build/en/2363/examples/simple-landscape-pennylane.html),  it seems like the vast majority of images that are included in these docs are generated by the code itself, e.g. 
Screen Shot 2024-05-13 at 10 49 24 AM from the PennyLane example, so I think this isn't a big 

So I don't see this as a blocking issue.

cosenal:
@jordandsullivan I don't see any great solution for that. And in fact it's going to be the same in any implementation of the «Download notebook» if we want to export .ipynb, unless we overcomplicate it and package it as a zip with images included. Same problem with all implementations I saw around, for example see the notebook that you can download from Jupyter Book chapter: https://inferentialthinking.com/chapters/08/Functions_and_Tables.html

A solution could be to use absolute url for images, which are already in our repo. E.g. `https://github.com/unitaryfund/mitiq/blob/main/docs/source/_thumbnails/calibration.png` for your example.

jordandsullivan:
@cosenal Yeah like I said, I think it's fine given that it seems relatively rare the images contain anything crucial in our notebooks.

andre-a-alves:
Not to complicate things, but if we want to keep the images, we can always embed them as data directly in an `` tag.

cosenal:
@andre-a-alves that's an option too, but it will make the .md files more bulky and worse to review/edit.

andre-a-alves (approved):
LGTM. Nice job!

Issue #2365, opened 2024-05-13
Missing Unit Tests


purva-thakre:
Noticed the some lines in the following files show up as uncovered by unit tests in the pytest coverage report. This issue is to figure out if these lines require additional unit tests or if they erroneously show up as uncovered in the pytest report. 

- [ ] mitiq/_about.py
- [ ] mitiq/benchmarks/mirror_circuits.py
- [ ] mitiq/benchmarks/randomized_benchmarking.py
- [ ] mitiq/benchmarks/randomized_clifford_t_circuit.py
- [ ] mitiq/calibration/calibrator.py    
- [ ] mitiq/cdr/_testing.py             
- [ ] mitiq/cdr/clifford_training_data.py    
- [ ] mitiq/cdr/clifford_utils.py 
- [ ] mitiq/executor/executor.py 
- [ ] mitiq/interface/mitiq_braket/conversions.py 
- [ ] mitiq/interface/mitiq_qiskit/conversions.py    
- [ ] mitiq/observable/observable.py    
- [ ] mitiq/observable/pauli.py  
- [ ] mitiq/pec/representations/optimal.py     
- [ ] mitiq/pec/types/types.py 
- [ ] mitiq/shadows/quantum_processing.py     
- [ ] mitiq/shadows/shadows.py        
- [ ] mitiq/shadows/shadows_utils.py          
- [ ] mitiq/zne/inference.py       
- [ ] mitiq/zne/scaling/folding.py        
- [ ] mitiq/zne/zne.py 

Will update the issue description later with more details about the specific lines.

cosenal:
Happy to go through this exercise, with the caveat that one could make the [The Case Against 100% Code Coverage](https://about.codecov.io/blog/the-case-against-100-code-coverage/).

purva-thakre:
The caveat is understandable. Personally, I don't go by the coverage percentage. That value fluctuates inconsistently and it's still not clear to me how it is calculated. 

Typically, I look at the uncovered lines in the pytest coverage report to make pytest ignore some lines from the coverage report. 

For example, I don't think we need unit tests for `mitiq/_about.py` but `# pragma: no cover` is ignored by the pytest coverage report. So, we will have to create a PR to make pyest ignore these. 

https://github.com/unitaryfund/mitiq/blob/aefc5cec42d1913ae1760259130f2a939a6fc787/mitiq/_about.py#L26-L29

cosenal:
> That value fluctuates inconsistently

[...]

PR #2379, opened 2024-05-18
Ignore _about.py in pytest coverage

purva-thakre:
Related to two of the items in #2365 

1. mitiq/_about.py
2. mitiq/pec/representations/optimal.py

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -61,8 +61,6 @@ requirements: requirements/requirements.txt
 .PHONY: test
 test:
 	pytest -n auto -v --cov=mitiq --cov-report=term --cov-report=xml --ignore=mitiq/interface/mitiq_pyquil
-	pytest -rP mitiq/tests/test_about.py
-
 .PHONY: test-%
 test-%:
 	pytest -n auto -v --cov=mitiq --cov-report=term --cov-report=xml */$(*)/*
@@ -74,4 +72,3 @@ test-pyquil:
 .PHONY: test-all
 test-all:
 	pytest -n auto -v --cov=mitiq --cov-report=term --cov-report=xml
-	pytest -rP mitiq/tests/test_about.py
--- a/Makefile
+++ b/Makefile
@@ -61,6 +61,7 @@ requirements: requirements/requirements.txt
 .PHONY: test
 test:
 	pytest -n auto -v --cov=mitiq --cov-report=term --cov-report=xml --ignore=mitiq/interface/mitiq_pyquil
+
 .PHONY: test-%
 test-%:
 	pytest -n auto -v --cov=mitiq --cov-report=term --cov-report=xml */$(*)/*
--- a/Makefile
+++ b/Makefile
@@ -61,6 +61,7 @@ requirements: requirements/requirements.txt
 .PHONY: test
 test:
 	pytest -n auto -v --cov=mitiq --cov-report=term --cov-report=xml --ignore=mitiq/interface/mitiq_pyquil
+	pytest -rP mitiq/tests/test_about.py
 
 .PHONY: test-%
 test-%:
@@ -73,3 +74,4 @@ test-pyquil:
 .PHONY: test-all
 test-all:
 	pytest -n auto -v --cov=mitiq --cov-report=term --cov-report=xml
+	pytest -rP mitiq/tests/test_about.py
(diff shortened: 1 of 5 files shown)


purva-thakre:
@cosenal This one's ready to be merged as well. Unless we don't want to ignore this file from the pytest coverage report.

natestemen (commented):
As far as I understand, there are no existing tests for the `mitiq.about()` function[^1], so ignoring `mitiq/_about.py` will make coverage go up. Should there be tests for this function, however?

[^1]: `mitiq.about` is called in some of the notebooks, so that's an implicit test that it works, but no unit tests.

natestemen:
> I was confused if this file needs to have unit tests.

A test that ensures `mitiq.about()` returns a string seems reasonable. I wouldn't want to test much more than that, however. Definitely low priority.

purva-thakre:
> A test that ensures mitiq.about() returns a string seems reasonable.

Output of `mitiq.about()` is `NoneType` not `str`.

natestemen:
> Output of mitiq.about() is `NoneType` not `str`.

You're right! I shouldn't have said "_... returns a string_".

Exactly as you say, the type annotations denote **return** types, without any indication if something is printed to standard out, standard error, or any other external action is taken. What I'm lightly suggesting (again, low priority) is we ensure that `mitiq.about()` prints to standard out.

mitiq/tests/test_about.py
```diff
@@ -0,0 +1,13 @@
...
+# LICENSE file in the root directory of this source tree.
+
+"""Tests for mitiq.about()."""
+
+import mitiq
+
+
+def test_stdout():
+    """Tests function prints a str."""
+    mitiq.about()
```
natestemen, line 13: I don't think this tests that `mitiq.about()` prints to standard out (at least when run using `pytest mitiq/tests/test_about.py` which, IMO, we should support).

There is an example of capturing standard output in the pytest docs: https://docs.pytest.org/en/6.2.x/capture.html#accessing-captured-output-from-a-test-function

Makefile
```diff
@@ -61,6 +61,7 @@ requirements: requirements/requirements.txt
 .PHONY: test
 test:
 	pytest -n auto -v --cov=mitiq --cov-report=term --cov-report=xml --ignore=mitiq/interface/mitiq_pyquil
+	pytest -rP mitiq/tests/test_about.py
```
natestemen, line 64: Why is this needed? If I run the following command to list all the tests that will be run, `test_about.py::test_stdout` is included.
```bash
pytest --collect-only -n auto -v --cov=mitiq --cov-report=term --cov-report=xml --ignore=mitiq/interface/mitiq_pyquil

**Review by natestemen — CHANGES_REQUESTED** [review]
Thanks for cleaning this up! Looks like it's not quite working yet, but will be nice to have.

File: Makefile (hunk 2 above)
Comment (purva-thakre) [reply Makefile:64]: It's printing the output to verify it works. 

image

File: mitiq/tests/test_about.py (hunk 1 above)
Comment (purva-thakre) [reply mitiq/tests/test_about.py:13]: Yes, I did see that example before choosing to use `pytest -rP mitiq/tests/test_about.py`

I'll read more on the differences between capturing an stdout vs. testing an stdout.

File: Makefile (hunk 2 above)
Comment (natestemen) [reply Makefile:64]: I see that `-rP` means that extra info will be printed when tests pass **and** have output, but

1. I don't think the test is verifying there _is_ output,[^1] and
2. I don't think we need this to be printed every time `make test` is run. As long as pytest does in fact that the printing works, we don't need to see it :)

In fact won't `test_stdout` be run twice with this PR in the current state?

[^1]: I changed the `print` statement in `about()` to `return 0` and the test still passed.

File: Makefile [hunk 3]
```diff
@@ -61,7 +61,6 @@ requirements: requirements/requirements.txt
 .PHONY: test
 test:
 	pytest -n auto -v --cov=mitiq --cov-report=term --cov-report=xml --ignore=mitiq/interface/mitiq_pyquil
-
```
natestemen, line 64: Let's leave the spacing between the makefile commands. It makes the file a touch easier to read.

natestemen (approved):
LGTM! Just one small change requested before merging. Thanks!

Makefile
purva-thakre: Thanks for catching this! I shouldn't have made the change.

Issue #2346, opened 2024-05-04
In Pauli Twirling: allow user to pass a `noise_gate` or similar to simulate noise in between the twirled gates and CNOT / CZ gates.


Misty-W:
## Problem

The Pauli Twirling module `mitiq.pt` is currently difficult to use on a simulated backend. This is because incoherent/coherent noise is usually modeled using additional layers of noise that twirling does not account for.

```py
>>> a, b = cirq.LineQubit.range(2)
>>> c = cirq.Circuit(cirq.H.on(a), cirq.CNOT.on(a, b))
>>> cn = c.with_noise(cirq.NoiseModel.from_noise_model_like(cirq.depolarize(p=0.01)))
>>> print(cn)
0: ───H───D(0.01)[cirq.VirtualTag()]───@───D(0.01)[cirq.VirtualTag()]───
                                       │
1: ───────D(0.01)[cirq.VirtualTag()]───X───D(0.01)[cirq.VirtualTag()]───

>>> print(mitiq.pt.pauli_twirl_circuit(cn, num_circuits=1)[0])
0: ───H───D(0.01)[cirq.VirtualTag()]───Y───@───X───D(0.01)[cirq.VirtualTag()]───
                                           │
1: ───────D(0.01)[cirq.VirtualTag()]───Y───X───Z───D(0.01)[cirq.VirtualTag()]───
```

Because the added twirling gates ($Y\otimes Y$ and $X\otimes Z$ in the case above) do not surround the depolarizing noise, this version of twirling does not actually tailor **coherent** noise into **incoherent** noise. In essence, this means `mitiq.pt` is useless on simulators, unless the user manually inserts noise _after_ twirling.

Because $\mathrm{CNOT}$ / $\mathrm{CZ}$ gates themselves contain noise on hardware, this does not effect using `mitiq.pt` on a QPU.

## Requested solution

1. Allow the user to pass something like a `noise_gate` to apply in between the twirled gates and the $\mathrm{CNOT}$ / $\mathrm{CZ}$ gates.

2. Clarify in the documentation that if this is used in a simulator, the user(-defined executor) has to insert a noisy version of the CZ and CNOT gates.

More background is contained in #2201

ronitd2002:
Hello @cosenal I am Ronit Dutta, and I have working experience with google's cirq. I have been in programming in the same for my own research for 1 year. May I please be assigned this project?

cosenal:
[...]

PR #2397, opened 2024-06-03
Simulate noise for CNOT and CZ gates in Pauli Twirling

EmilianoG-byte:
## Description

Currently, the `pauli_twirl_circuit` function does not wrap correctly noise applied on CNOT and CZ gates when used in a simulator. Instead, it adds the twirled operators in between the noise and the two-qubit gates.

This PR implements an argument in `pauli_twirl_circuit` to pass a `noise_op` acting between CNOT/CZ gates and the twirled layers. Thus, it allows to properly Taylor the noise using Pauli Twirling on a simulated backend. 

Example:

Circuit without noise:
```python
0: ───H───@───@───
          │   │
1: ───────X───@───
```

Pauli Twirling before this PR:

```python
0: ───H────I───@───I───D(0.01)───I───@───Z───D(0.01)───
               │                     │
1: ────────X───X───X───D(0.01)───Y───@───Y───D(0.01)───
```

Pauli Twirling after this PR:

```python
0: ───H───I───@───D(0.01)───I───I───@───D(0.01)───Z───
              │                     │
1: ───────X───X───D(0.01)───X───Y───@───D(0.01)───Y────
```

Contribution as part of the Unitary Hack 2024

---

### License

Closes [#2346]

### DIFF
--- a/mitiq/pt/pt.py
+++ b/mitiq/pt/pt.py
@@ -4,9 +4,13 @@
 # LICENSE file in the root directory of this source tree.
 
 import random
+from functools import singledispatch
 from typing import List, Optional
 
 import cirq
+import pennylane as qml
+from cirq import Circuit as _Circuit
+from pennylane.tape import QuantumTape
 
 from mitiq import QPROGRAM
 from mitiq.interface import accept_qprogram_and_validate
@@ -49,11 +53,22 @@ CZ_twirling_gates = [
     (cirq.Z, cirq.Z, cirq.Z, cirq.Z),
 ]
 
+CIRQ_NOISE_OP = {
+    "bit-flip": cirq.bit_flip,
+    "depolarize": cirq.depolarize,
+}
+
+PENNYLANE_NOISE_OP = {
+    "bit-flip": qml.BitFlip,
+    "depolarize": qml.DepolarizingChannel,
+}
+
 
 def pauli_twirl_circuit(
     circuit: QPROGRAM,
     num_circuits: int = 10,
-    noise_op: Optional[cirq.Gate] = None,
+    noise_name: Optional[str] = None,
+    **kwargs,
 ) -> List[QPROGRAM]:
     r"""Return the Pauli twirled versions of the input circuit.
 
@@ -64,12 +79,13 @@ def pauli_twirl_circuit(
     Args:
         circuit: The input circuit to execute with twirling.
         num_circuits: Number of circuits to be twirled, and averaged.
-        noise_op: Noisy operator acting on CNOT and CZ gates. See warning.
+        noise_name: Name of the noisy operator acting on CNOT and CZ gates.
+        See warning.
 
     Warning:
         If this function is executed with a simulator backend, it is
         necessary to use the `noise_op` argument to apply noise on CNOT
-        and CZ gates. Otherwise, the twirled circuits will not include the
+        and CZ gates. Otherwise, the twirled circuits will not wrap the
         noise on these gates.
 
     Returns:
@@ -80,25 +96,39 @@ def pauli_twirl_circuit(
         twirl_CZ_gates(c, num_circuits=1)[0] for c in CNOT_twirled_circuits
     ]
 
-    if noise_op is not None:
+    if noise_name is not None:
         twirled_circuits = [
-            add_noise_to_two_qubit_gates(circuit, noise_op)
+            add_noise_to_two_qubit_gates(circuit, noise_name, **kwargs)
             for circuit in twirled_circuits
         ]
 
     return twirled_circuits
 
 
-@accept_qprogram_and_validate
 def add_noise_to_two_qubit_gates(
-    circuit: cirq.Circuit, noise_op: cirq.Gate
-) -> cirq.Circuit:
+    circuit: QPROGRAM, noise_name: str, **kwargs
+) -> QPROGRAM:
     """Add noise to CNOT and CZ gates on pre-twirled circuits.
 
     Args:
         circuit: Pre-twirled circuit
-        noise_op: noise operator to apply after CNOT and CZ gates
+        noise_name: name of noise operator to apply after CNOT and CZ gates
     """
+    # here we will validate if noise_op and kwargs are valid
+    return _add_noise_to_two_qubit_gates(circuit, noise_name, **kwargs)
+
+
+@singledispatch
+def _add_noise_to_two_qubit_gates(circuit, noise_name, **kwargs):
+    raise NotImplementedError(
+        f"Cannot add noise to Circuit of type {type(circuit)}."
+    )
+
+
+@_add_noise_to_two_qubit_gates.register
+def _cirq(circuit: _Circuit, noise_name: str, **kwargs):
+    noise_function = CIRQ_NOISE_OP[noise_name]
+    noise_op = noise_function(**kwargs)
 
     noisy_gates = [cirq.ops.CNOT, cirq.ops.CZ]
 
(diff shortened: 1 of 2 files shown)


mitiq/pt/tests/test_pt.py
```diff
@@ -144,3 +144,20 @@ def test_no_CNOT_CZ_circuit(twirl_func):
 
     for i in range(5):
         assert _equal(circuit, twirled_output[i])
+
+
+@pytest.mark.parametrize(
+    "noise_op", [cirq.bit_flip(p=0.1), cirq.depolarize(p=0.01)]
+)
+def test_noisy_CNOT_CZ_circuit(noise_op):
```
Misty-W, line 152: Can you check if this works on other Mitiq-supported circuit types?

mitiq/pt/pt.py
```diff
@@ -72,9 +80,39 @@ def pauli_twirl_circuit(
...
+        twirled_circuits = [
+            add_noise_to_two_qubit_gates(circuit, noise_op)
+            for circuit in twirled_circuits
+        ]
+
     return twirled_circuits
 
 
+@accept_qprogram_and_validate
+def add_noise_to_two_qubit_gates(
```
Misty-W, line 93: Am I understanding this function correctly, that it adds noise after _all_ CNOT and CZ gates in the circuit?
To solve this issue, we only want noise added to the gates that `pauli_twirl_circuit` adds because we assume the noise operations are already present. 
Apologies if this point was unclear in the original issue.

Misty-W (commented):
Good start @EmilianoG-byte ! 

There are a couple points to address before the PR can be accepted, let me know if you have any questions.

mitiq/pt/pt.py
purva-thakre: The function defined in this PR could be used before the twirled circuits are returned by the pauli twirling function:

https://github.com/unitaryfund/mitiq/blob/837ea83546f16e39867e3ee74cd27996b1d7509d/mitiq/pt/pt.py#L75

`pauli_twirling_function` would get an optional arg where if a noisy gate is provided, `add_noise_to_two_qubit_gates` is used to return the noisy circuit.

EmilianoG-byte: Hi @Misty-W  and @purva-thakre , thank you for the comments!
I am also slightly confused with the comment from @Misty-W. As shown in 

https://github.com/EmilianoG-byte/mitiq/blob/efc079fb126353adfa852c196f304ed59794ddaa/mitiq/pt/pt.py#L146

the twirl is applied to all CNOT gates in the circuit, and similarly for the CZ gates. So, as mentioned by @purva-thakre , these are where the noise we want to twirl should be present.

Could you elaborate on:
> we assume the noise operations are already present.

mitiq/pt/tests/test_pt.py
EmilianoG-byte: Hi @Misty-W! Well, that was a very good question, because indeed it does not work. the biggest problem here is that the Cirq function `to_qasm ` does not support noise operators. 

And even if this was the case, adding noise on Qiskit is done on execution (AFAIK). 

A similar issue happens on `QuantumTapes` on Pennylane with the `to_openqasm ` method. So also converting from a Pennylane noise channel to Cirq is also not trivial.

I honestly do not know how to proceed there 🤔

EmilianoG-byte: Hi, @purva-thakre. So you mean have some sort of look up dictionary based on the string and the circuit module? I guess that means trying to include as many noise channels as possible as long as all the supported `QPROGRAM`s have an equivalent, right?

This would still excluded qiskit as noise is added when executing the circuit. Is this correct?

purva-thakre: > So you mean have some sort of look up dictionary based on the string and the circuit module? I guess that means trying to include as many noise channels as possible as long as all the supported QPROGRAMs have an equivalent

Yes, this is what I was thinking but I don't know if this is the best possible solution. I am sure this will get discussed during the community call tomorrow. 

> This would still excluded qiskit as noise is added when executing the circuit

Did this change with the latest version of Qiskit? As far as I know, Qiskit used to allow a user to selectively add noise to a circuit.

EmilianoG-byte: I have added some minimal functionalities that work now with Cirq and Pennylane using your suggestion. I still think that it is kind of cumbersome to approach it like this, but since noise is not supported in QASM conversions, I cannot think of a better way.

I added the examples as unittests.

If you think this is the way to go, I can go ahead and register more dispatches for the other supported QPROGRAMS and add more noise models (the idea would be to include the ones supported by Cirq as well and raise an error if it's something else)

Please let me know what you think! 😄

mitiq/pt/pt.py
Misty-W: I'd thought that line would (unintentionally) cause the noise to be applied twice, but now I see that isn't the case here. Thanks for clarifying, I'll go ahead and resolve this part.

mitiq/pt/tests/test_pt.py
EmilianoG-byte: Thank you for the update! Any feedback on my proposed solution would also be great. For the moment I will assume there is a better way, so I will wait to hear from you 😊.

Misty-W:
hi @EmilianoG-byte good news, we will accept your solution for unitaryHACK, only change needed is to fix the mypy failures (we will give hints if needed). If you're interested in solving the multi-platform problem for fun, that would be awesome but not required to win the bounty.

EmilianoG-byte:
> hi @EmilianoG-byte good news, we will accept your solution for unitaryHACK, only change needed is to fix the mypy failures (we will give hints if needed). If you're interested in solving the multi-platform problem for fun, that would be awesome but not required to win the bounty.

Thank you @Misty-W , sounds great 😄. I have fixed the mypy issues, "except" for one where I ignored the type hinting since I am passing **kwargs  to the chosen noise function.

I am actually intrigued on what you and the other developers at Mitiq think about it. Do you agree with my solution being the best we can do (not just for the unitary hack)?

Misty-W (approved):
LGTM, congrats @EmilianoG-byte! 
Please comment on the issue #2346 so GitHub will allow me to assign you (how we keep track of bounties).

Misty-W:
> I am actually intrigued on what you and the other developers at Mitiq think about it. Do you agree with my solution being the best we can do (not just for the unitary hack)?

The alternative we discussed is to have the user specify a noise model in the simulator backend such that twirling gates surround the noisy CNOT and noisy CZ gates instead of the ideal CNOT and CZ. It's not clear whether we can get that approach to work on all Mitiq-supported platforms either, though.

EmilianoG-byte:
> @EmilianoG-byte the alternative we discussed is to have the user specify a noise model in the simulator backend such that twirling gates surround the noisy CNOT and noisy CZ gates instead of the ideal CNOT and CZ. It's not clear whether we can get that approach to work on all Mitiq-supported platforms either, though.

@Misty-W so that still means mapping them "manually"  as I am doing right? Since there is no automatic QASM conversion for noise models (?).

Misty-W:
> @Misty-W so that still means mapping them "manually" as I am doing right? Since there is no automatic QASM conversion for noise models (?).

I don't believe there is QASM conversion for noise models, but if the noise is added after converting the _twirled_ circuit back to the native frontend, then the noise model wouldn't have to be converted. I haven't tested this idea, so I might be missing something that prevents this flow from working as intended.

Issue #2354, opened 2024-05-08
Qiskit Quantum Fourier Transform (QFT) circuits error out during conversion


cosenal:
If a Qiskit input circuit is composed of a QFT subcircuit, it errors out once it's executed with error mitigation techniques.

Minimum non-working snippet (adapted from #558 to comply with Qiskit 1.0):
```python3
from qiskit_aer import QasmSimulator
from qiskit import QuantumCircuit, compiler
from qiskit.circuit.library import QFT

from mitiq.interface.mitiq_qiskit import initialized_depolarizing_noise
from mitiq.zne import execute_with_zne


def qs_noisy_simulation(circuit: QuantumCircuit, shots: int = 1) -> float:

    noise_model = initialized_depolarizing_noise(noise_level=0.02)
    backend = QasmSimulator(noise_model=noise_model)
    job = backend.run(circuit.decompose(), shots=shots)
    return job.result().get_counts().get("0", 0.0) / shots


circuit = QuantumCircuit(1)
circuit &= QFT(1)
circuit.measure_all()

print(circuit)

unmitigated = qs_noisy_simulation(circuit)
print(unmitigated)

# circuit = compiler.transpile(circuit, basis_gates=["u1", "u2", "u3", "cx"])
# circuit = circuit.decompose()

mitigated = execute_with_zne(circuit, qs_noisy_simulation)

print(mitigated)
```

Error:
```
CircuitConversionError: Circuit could not be converted to an internal Mitiq circuit. This may be because the circuit contains custom gates or Pragmas (pyQuil). If you think this is a bug or that this circuit should be supported, you can open an issue at https://github.com/unitaryfund/mitiq. 
```

Traceback:
```
...
QasmException: Syntax error: 'q0'
..._QFT q0 { h q0; }
qreg q[1];
creg meas[1];
gate_QFT q[0];
measure q[0] -> meas[0]
        ^
at line 3, column 15
...
```
This is somehow different from #558 and #1073, which were raised on versions of Qiskit pre-1.0, and where the error was about unsupported gates. 

One workaround is to either decompose or transpile the circuit before it's passed as an argument to quantum error mitigation routines (see commented lines in the snippet above)[^1]. At the moment, this workaround is not explicit to the user, so we could implement a fallback mechanism in Mitiq interface, where a circuit gets decomposed or transpiled whenever the conversion to the internal representation fails.
 
[...]

Issue #558, opened 2021-02-18
QFT circuit containing CU1 gates cannot be converted to internal mitiq circuit


mcoggins96:
Pre-Report Checklist
--------------------

- [X] I am running the latest version of mitiq
- [X] I checked to make sure that this bug has not already been reported

Issue Description
-----------------
Currently I'm trying to test the execute_with_zne() function with a QFT circuit in Qiskit. However when I run the code it gives back an error stating that the circuit cannot be converted to an internal mitiq circuit. It seems to be because it does not recognise cu1 gates which are used in the QFT circuit. 

### Code Snippet

```python
import qiskit
import mitiq

from qiskit import QuantumRegister, ClassicalRegister
from qiskit import QuantumCircuit, execute,IBMQ
from qiskit.tools.monitor import job_monitor
from qiskit.circuit.library import QFT
from mitiq import execute_with_zne

# Noise simulation packages
from qiskit.providers.aer.noise import NoiseModel
from qiskit.providers.aer.noise.errors.standard_errors import depolarizing_error

IBMQ.enable_account('ENTER API KEY HERE')
provider = IBMQ.get_provider(hub='ibm-q')

def qs_noisy_simulation(circuit: QuantumCircuit, shots: int = 1) -> float:

    #backend = provider.get_backend('ibmq_quito')
    backend = provider.get_backend('ibmq_qasm_simulator')

    job = execute(circuit, backend, shots=1000)

    job_monitor(job)

    counts = job.result().get_counts()

    print(counts)

    expval = counts["10101"] / shots
    return expval

q = QuantumRegister(5,'q')
c = ClassicalRegister(5,'c')

circuit = QuantumCircuit(q,c)
circuit.x(q[0])
circuit.x(q[2])
circuit.x(q[4])
circuit += QFT(num_qubits=5, approximation_degree=0, do_swaps=True, inverse=False, insert_barriers=True, name='qft')
circuit += QFT(num_qubits=5, approximation_degree=0, do_swaps=True, inverse=True, insert_barriers=True, name='qft')
circuit.measure(q,c)

print(circuit)

unmitigated = qs_noisy_simulation(circuit)
mitigated = execute_with_zne(circuit, qs_noisy_simulation)

print(unmitigated)
print(mitigated)
```

### Error Output

```
Traceback (most recent call last):
[...]

PR #2404, opened 2024-06-06
Issue #2354 Fix - qiskit QFT gates error during conversion

NnguyenHTommy:
fixes issue #2354 where a QFT qiskit circuit would error out during conversion. Implemented a fallback mechanism to decompose and transpile the qiskit circuit into native gates. Added a unit test in the ZNE module to test this with qiskit's QFT gate. 

This is my first open source contribution via unitary hack so any feedback would be appreciated! 

## Description

---

### License

- [x ] I added unit tests for new code.
- [ x] I used [type hints](https://www.python.org/dev/peps/pep-0484/) in function signatures.
- [ x] I used [Google-style](https://google.github.io/styleguide/pyguide.html#383-functions-and-methods) docstrings for functions.
- [x ] I [updated the documentation](../blob/main/docs/CONTRIBUTING_DOCS.md) where relevant.
- [ x] Added myself / the copyright holder to the AUTHORS file

### DIFF
--- a/mitiq/interface/mitiq_qiskit/conversions.py
+++ b/mitiq/interface/mitiq_qiskit/conversions.py
@@ -14,7 +14,7 @@ import cirq
 import numpy as np
 import qiskit
 from cirq.contrib.qasm_import import circuit_from_qasm
-from qiskit import qasm2
+from qiskit import compiler, qasm2
 from qiskit.transpiler import PassManager
 from qiskit.transpiler.layout import Layout
 from qiskit.transpiler.passes import SetLayout
@@ -248,7 +248,17 @@ def from_qiskit(circuit: qiskit.QuantumCircuit) -> cirq.Circuit:
     Returns:
         Mitiq circuit representation equivalent to the input Qiskit circuit.
     """
-    return from_qasm(qasm2.dumps(circuit))
+    try:
+        mitiq_circuit = from_qasm(qasm2.dumps(circuit))
+    except Exception:
+        # If the conversion fails, try to decompose and transpile the
+        # circuit to native gates
+        circuit = compiler.transpile(
+            circuit, basis_gates=["u1", "u2", "u3", "cx"]
+        )
+        circuit = circuit.decompose()
+        mitiq_circuit = from_qasm(qasm2.dumps(circuit))
+    return mitiq_circuit
 
 
 def from_qasm(qasm: QASMType) -> cirq.Circuit:
--- a/mitiq/interface/mitiq_qiskit/conversions.py
+++ b/mitiq/interface/mitiq_qiskit/conversions.py
@@ -15,7 +15,7 @@ import numpy as np
 import qiskit
 from cirq.contrib.qasm_import import circuit_from_qasm
 from cirq.contrib.qasm_import.exception import QasmException
-from qiskit import compiler, qasm2
+from qiskit import qasm2
 from qiskit.transpiler import PassManager
 from qiskit.transpiler.layout import Layout
 from qiskit.transpiler.passes import SetLayout
@@ -249,7 +249,13 @@ def from_qiskit(circuit: qiskit.QuantumCircuit) -> cirq.Circuit:
     Returns:
         Mitiq circuit representation equivalent to the input Qiskit circuit.
     """
-    return from_qasm(qasm2.dumps(circuit))
+    try:
+        mitiq_circuit = from_qasm(qasm2.dumps(circuit))
+    except QasmException:
+        # Try to decompose circuit before running
+        circuit = circuit.decompose()
+        mitiq_circuit = from_qasm(qasm2.dumps(circuit))
+    return mitiq_circuit
 
 
 def from_qasm(qasm: QASMType) -> cirq.Circuit:
@@ -261,15 +267,5 @@ def from_qasm(qasm: QASMType) -> cirq.Circuit:
     Returns:
         Mitiq circuit representation equivalent to the input QASM string.
     """
-    try:
-        qasm = _remove_qasm_barriers(qasm)
-        mitiq_circuit = circuit_from_qasm(qasm)
-    except QasmException:
-        # Try to decompose and transpile the qiskit
-        # circuit before running
-        circuit = qiskit.QuantumCircuit.from_qasm_str(qasm)
-        circuit = compiler.transpile(circuit)
-        circuit = circuit.decompose()
-        qasm = _remove_qasm_barriers(qasm2.dumps(circuit))
-        mitiq_circuit = circuit_from_qasm(qasm)
-    return mitiq_circuit
+    qasm = _remove_qasm_barriers(qasm)
+    return circuit_from_qasm(qasm)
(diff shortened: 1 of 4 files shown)


mitiq/interface/mitiq_qiskit/conversions.py
```diff
@@ -248,7 +248,17 @@ def from_qiskit(circuit: qiskit.QuantumCircuit) -> cirq.Circuit:
...
         Mitiq circuit representation equivalent to the input Qiskit circuit.
     """
-    return from_qasm(qasm2.dumps(circuit))
+    try:
+        mitiq_circuit = from_qasm(qasm2.dumps(circuit))
+    except Exception:
+        # If the conversion fails, try to decompose and transpile the
+        # circuit to native gates
+        circuit = compiler.transpile(
+            circuit, basis_gates=["u1", "u2", "u3", "cx"]
```
cosenal, line 257: This basis of gates was in the issue description as an excample that came from https://github.com/unitaryfund/mitiq/issues/558#issuecomment-781743206. However it requires adding some motivation on why it's the basis of choice here.

mitiq/zne/tests/test_zne.py
```diff
@@ -540,3 +541,22 @@ def execute(circuit: qiskit.QuantumCircuit, shots: int = 8192) -> float:
     # Note: Unmitigated value is also (usually) within 10% of the true value.
     # This is more to test usage than effectiveness.
     assert abs(zne_value - true_value) < 0.1
+
+
+def test_execute_zne_transpiled_qiskit_circuit_with_noise_and_QFT():
```
cosenal, line 546: This is useful, but it's more like an "integration" test where we actually execute and test many components. Additional to this, it'd be helpful to have a unit test with a smaller scope, with the unit being the `from_qasm` function.

cosenal (commented):
Thanks for your contribution, @NnguyenHTommy!
Good work so far, I left some comments.

mitiq/interface/mitiq_qiskit/conversions.py
```diff
@@ -248,7 +248,17 @@ def from_qiskit(circuit: qiskit.QuantumCircuit) -> cirq.Circuit:
     Returns:
         Mitiq circuit representation equivalent to the input Qiskit circuit.
     """
-    return from_qasm(qasm2.dumps(circuit))
+    try:
+        mitiq_circuit = from_qasm(qasm2.dumps(circuit))
+    except Exception:
```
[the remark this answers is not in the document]
NnguyenHTommy: I added the QasmException handler instead of a general exception in my most recent pull request. Is this what you meant by being more granular?

mitiq/interface/mitiq_qiskit/conversions.py
NnguyenHTommy: I removed specifying the basis and let qiskit handle this to make it more general

mitiq/zne/tests/test_zne.py
NnguyenHTommy: I added another unit test to test the from_qasm() function! In doing so, I realized that it would be better to make the exception catch in the from_qasm() function as that function is more general and is called by the from_qiskit() function. Thank you for the comment!

mitiq/interface/mitiq_qiskit/conversions.py
cosenal: Based on [Qiskit docs](https://docs.quantum.ibm.com/api/qiskit/compiler#qiskit.compiler.transpile), when basis_gates is `None` (default value), the circuit is not unrolled, so I would expect that `decompose()` is what is effective here. Have you tried to print the two circuits (before and after transpile&decompose) that come out off your example?

mitiq/zne/tests/test_zne.py
cosenal: Hmmm this is tricky. Despite the function `from_qasm` is in the mitiq_qiskit module, its signature `def from_qasm(qasm: QASMType) -> cirq.Circuit` (as well as its current definition) makes it very generic and not necessarely related to qiskit. Admittedly the function could be moved elsewhere, perhaps in a mitiq_qasm module. In view of such a possible refactoring in the future, I would move your new code back to the `from_qiskit()` function, since it's specific to qiskit circuits. What do you think?

cosenal:
@NnguyenHTommy Thanks for your second revision of this PR. I left some replies on the comments.. we are very close to merge this!

mitiq/interface/mitiq_qiskit/conversions.py
NnguyenHTommy: You're right, thank you for the feedback! Decompose is the only thing that is needed here. Transpile doesn't do anything and outputs the same circuit in this case. Decompose will decompose custom gates or packaged gates into standard qiskit gates.

mitiq/zne/tests/test_zne.py
NnguyenHTommy: With the most recent pull request, I moved the exception catch back to the from_qiskit() function. In doing so, I also had to change the test case to test the from_qiskit() function instead of the from_qasm() function. I agree that this makes more sense especially if some refactoring will be done. If anyone ever calls the from_qasm() function with a qasm string not from qiskit and **does not** want it to follow qiskit circuit formatting rules (for instance via mitiq!), then the way I had it would be very problematic. Unfortunately, this change means that if someone calls the from _qasm() function with a non-standard qiskit gate, it will throw an exception and won't go through the exception fallback which I believe is fine for the purpose of this issue since it wouldn't be from qiskit anymore?

cosenal, line 546: nit: I would rename this test (and the docstring on the line below). 
First of all, starting from the latest iteration, we realised transpiling is not necessary, but only decomposing is. More in general, this test is at a high level and should be unaware of how the circuit is converted under the hood. 
Perhaps just `test_execute_zne_on_qiskit_circuit_with_QFT()` ?

cosenal (commented):
Looks good. Left some small nits.

mitiq/interface/mitiq_qiskit/conversions.py
```diff
@@ -248,7 +249,13 @@ def from_qiskit(circuit: qiskit.QuantumCircuit) -> cirq.Circuit:
     Returns:
         Mitiq circuit representation equivalent to the input Qiskit circuit.
     """
-    return from_qasm(qasm2.dumps(circuit))
+    try:
+        mitiq_circuit = from_qasm(qasm2.dumps(circuit))
+    except QasmException:
+        # Try to decompose circuit before running
```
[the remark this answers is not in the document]
NnguyenHTommy: Added the comment in the most recent pull request!

mitiq/zne/tests/test_zne.py
NnguyenHTommy: Thanks for catching this. I updated the name and doc string as per suggested!

cosenal (approved):
LGTM.

🎉 Thanks for your contribution @NnguyenHTommy, and congratulations on the UnitaryHack bounty!

If you had fun and want to learn more about Mitiq, come hang out on our [Discord server](https://discord.gg/38wwdHUH).

Issue #2405, opened 2024-06-10
Warning raised during docs build


purva-thakre:
While running `make docs` locally, the following warning was raised:

The default value for `navigation_with_keys` will change to `False` in the next release. If you wish to preserve the old behavior for your site, set `navigation_with_keys=True` in the `html_theme_options` dict in your `conf.py` file. Be aware that `navigation_with_keys = True` has negative accessibility implications: https://github.com/pydata/pydata-sphinx-theme/issues/1492

purva-thakre:
I am not too sure if we need to worry about this warning. 

From my device, I cannot navigate to the next page of the documentation using &larr; and &rarr; keys on the keyboard. 
https://pydata-sphinx-theme.readthedocs.io/en/v0.12.0/user_guide/keyboard-shortcuts.html#change-pages

purva-thakre:
Reopening this because I misunderstood the warning. Indeed we can see this in the docs. If I go directly to the API age, and hit the right or left key on the keyboard it won't do anything. 

If I keyword search for `represent_operation_with_global_depolarizing_noise` on the API page, the page can suddenly move behind the left sidebar if I keep hitting the left arrow on the keyboard. 


https://github.com/unitaryfund/mitiq/assets/66048318/3bbc41b8-d181-4da7-8c4f-e8a19e81d0ba


This is partially related to https://github.com/unitaryfund/mitiq/issues/2331#issuecomment-2176567908 where we should get rid of the blank sidebar.

purva-thakre:
Closing as discussed in the `Description` section of #2424

Issue #2331, opened 2024-04-27
Fix navigation for the docs


purva-thakre:
Previously discussed here: https://github.com/unitaryfund/mitiq/issues/2278#issuecomment-2077751053

After moving to `PyData` theme for the docs, some of the left and right navigation sidebars appear blank. 

| Left Sidebar: [2nd Readme](https://mitiq.readthedocs.io/en/latest/readme.html)      | Right Sidebar: [User's Guide](https://mitiq.readthedocs.io/en/latest/guide/guide.html) |
| ----------- | ----------- |
| ![image](https://github.com/unitaryfund/mitiq/assets/66048318/3441ad0c-bade-45d6-abec-638f7f12cb47)| ![image](https://github.com/unitaryfund/mitiq/assets/66048318/4d9e2383-99d5-43f4-9178-2551e1bc0a39)|

To fix this issue, we will have to change the default options allowed by the theme in `conf.py`
https://pydata-sphinx-theme.readthedocs.io/en/stable/user_guide/layout.html#overview-of-theme-layout
https://pydata-sphinx-theme.readthedocs.io/en/stable/user_guide/layout.html#primary-sidebar-left
https://pydata-sphinx-theme.readthedocs.io/en/stable/user_guide/layout.html#secondary-sidebar-right

I vote for keeping either the left or right sidebars; not both.

jordandsullivan:
@purva-thakre Checking my understanding, it looks like on a normal page of the docs (i.e. not a top-level directory), the primary and secondary sidebars have different information: the high level section navigation and the page-level navigation, respectively:
<img width="1404" alt="Screen Shot 2024-05-09 at 3 58 37 PM" src="https://github.com/unitaryfund/mitiq/assets/15827191/9a9ea930-7fe1-489b-9c88-6c741e8bb5a4">

While I agree this may be a bit busy for some, I think it's still useful to have both in this circumstance. 

> I vote for keeping either the left or right sidebars; not both.

Do you mean to keep only one sidebar _across all of the docs_ or just on the specific pages you referenced above?

(Left Sidebar: [2nd Readme](https://mitiq.readthedocs.io/en/latest/readme.html) 	Right Sidebar: [User's Guide](https://mitiq.readthedocs.io/en/latest/guide/guide.html))

jordandsullivan:
@purva-thakre I don't see the original issue popping up on the linked pages. Do you?

purva-thakre:
[...]

PR #2424, opened 2024-06-27
Hide primary sidebar from certain pages of the documentation

purva-thakre:
## Description
Remove the left sidebar from `API-doc`, `Changelog`, and `References` because the primary sidebar is blank. 

Related to #2331 where we want to hide the blank sidebar on the left for some pages of the documentation. 

Note that the behavior from https://github.com/unitaryfund/mitiq/issues/2405#issuecomment-2183583772 is still present even if we hide the primary sidebar. 

This is not fixable by `navigation_with_keys=False` which makes me think the behavior from the screencast is unrelated to #2405 

https://github.com/unitaryfund/mitiq/assets/66048318/7cef51f8-f2c7-4276-80fd-a12c23700685

---

### License

### DIFF
--- a/docs/source/conf.py
+++ b/docs/source/conf.py
@@ -100,8 +100,8 @@ extensions = [
     "sphinx_gallery.load_style",
 ]
 
-# hide primary sidebar from apidoc page
-html_sidebars = {"apidoc": []}
+# hide primary sidebar from the following pages
+html_sidebars = {"apidoc": [], "changelog": [], "bibliography": []}
 
 intersphinx_mapping = {
     "python": ("https://docs.python.org/3.9", None),
--- a/docs/source/conf.py
+++ b/docs/source/conf.py
@@ -100,6 +100,11 @@ extensions = [
     "sphinx_gallery.load_style",
 ]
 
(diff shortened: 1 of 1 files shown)


cosenal:
My only concern is the one raised by @natestemen, that is, the content view moving around may be distracting for users. 
Perhaps @francespoblete has an opinion on this. 

Frances, the views to compare are:
(1) https://mitiq--2424.org.readthedocs.build/en/2424/
(2) https://mitiq.readthedocs.io/en/latest

Downside of (2): the API-doc page has an empty Section navigation bar on the left
Downside of (1): the content view of the API-doc page doesn't align with the content view of, say, the Examples page, or any other page that has a non-empty navigation bar on the left-hand side.

natestemen (approved):
IMO it's not ideal how content jumps around when switching to these pages, but it's nice not to have the sidebars when there is nothing in them.

It's wild how many options there are for [configuration](https://www.sphinx-doc.org/en/master/usage/configuration.html) in sphinx!

purva-thakre:
> IMO it's not ideal how content jumps around when switching to these pages

Should we create a separate issue for this?

PR #2425, opened 2024-06-27
bump min python version for intersphinx map

natestemen:
missed this in https://github.com/unitaryfund/mitiq/pull/2066. does what it says on the 🥫.


bdg221 (approved):
looks good

PR #2428, opened 2024-06-28
set version variable and pass it on

natestemen:
## Description

The upload to testpypi action is failing at the last stage where it compares the downloaded version against what's in `VERSION.txt`. This is because Mitiq is (rightfully) not checked out when downloading Mitiq from testpypi. Details [here](https://github.com/unitaryfund/mitiq/actions/runs/9705189820/job/26786831916#step:4:13). To fix this I've set a `VERSION` variable when Mitiq is checked out and passed the data to the verify version portion of the action.

I have a concern that even though the first two stages passed, the uploaded version seems to be [`0.38.0.dev0`](https://test.pypi.org/project/mitiq/0.38.0.dev0/) (with an extra 0 on the end). Hence even if this works, we might still have a problem.

### DIFF
--- a/.github/workflows/publish-testpypi.yml
+++ b/.github/workflows/publish-testpypi.yml
@@ -12,6 +12,8 @@ jobs:
     name: Build Mitiq
     if: github.repository_owner == 'unitaryfund'
     runs-on: ubuntu-latest
+    outputs:
+      VERSION: ${{ steps.export-version.outputs.VERSION }}
     steps:
       - name: Check out mitiq
         uses: actions/checkout@v4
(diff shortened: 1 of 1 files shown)


natestemen:
Ah so it looks like when pypi sees a version number that ends in `dev`, it automatically assumes it is a pre-release. Hence when we download from testpypi after the successful upload, we will always get the latest stable version, which as of 06/27/2024 is 0.9.0.

This is the standard for versioning numbers/info: https://peps.python.org/pep-0440/

cosenal (approved):
LGTM

I don't mind at all you iterating on this task through PRs, and happy to review all of them. However, I am curious whether those are testable locally through the `act` tool.

natestemen:
superceded by work in https://github.com/unitaryfund/mitiq/pull/2436

Issue #2431, opened 2024-07-01
Flaky test in `rem/tests/test_inverse_confusion_matrix.py`


purva-thakre:
Noticed this failure twice in a CI run for Windows. If I rerun the test, it passes. 

https://github.com/unitaryfund/mitiq/actions/runs/9748119578/job/26902405324?pr=2347#step:6:4326

![image](https://github.com/unitaryfund/mitiq/assets/66048318/e125c290-5a82-4ad1-b0ff-53599500a998)

purva-thakre:
Noticed the same failure for windows-python 3.11

![image](https://github.com/unitaryfund/mitiq/assets/66048318/7cd7f81e-6b5b-4655-a2b0-79581df8203e)

https://github.com/unitaryfund/mitiq/actions/runs/9766045422/job/26958119314?pr=2432#step:6:4325

purva-thakre:
The same failure on windows-python3.12

https://github.com/unitaryfund/mitiq/actions/runs/9765921429/job/26957722256#step:6:4326

purva-thakre:
I am unassigning this for myself because I cannot replicate this locally. My laptop does not have enough memory for local docker containers.

natestemen:
Looks like this is also happening on ubuntu: https://github.com/unitaryfund/mitiq/actions/runs/10267476608/job/28408109210?pr=2452#step:6:4486

natestemen:
Took a look at this in the mitiq coding call today and made some progress.

### Problem

https://github.com/unitaryfund/mitiq/blob/56ee1731514aa19ea60e86d41e8d7609bdc963c3/mitiq/rem/tests/test_inverse_confusion_matrix.py#L32-L33

This test (in essence)
1. generates a **random** sequence of 1000 `0`s and `1`s
2. sums that sequence (as ints)
3. Asserts that the resulting sum is between 450 and 550

Some stats tell us the sum should fall in this range 99.94% of the time, meaning that on average we should see a value outside this range every 1,667 runs.

### Solutions

We considered the following solutions
1. Removing the test. Since this is non-deterministic behavior and the test is mostly testing to [`np.random.choice`](https://github.com/unitaryfund/mitiq/blob/56ee1731514aa19ea60e86d41e8d7609bdc963c3/mitiq/rem/inverse_confusion_matrix.py#L30), it is not totally needed.
2. Increasing the valid range of values (say it must fall between 400 and 600).
3. Setting a seed to make the sum deterministic.

[...]

PR #2436, opened 2024-07-02
make upload versions unique for TestPyPI

natestemen:
## Description

TestPyPI uploads are failing (again) because once a version is uploaded with a specific version specifier (say 0.38.0dev), another with the same version **cannot** be uploaded. To work around this, we added a date the end of the `VERSION.txt` file so each upload is unique. We also removed the `verify-version` portion of the action since as long as the upload passes, this gives us enough confidence that our workflow is effective.

### DIFF
--- a/.github/workflows/publish-testpypi.yml
+++ b/.github/workflows/publish-testpypi.yml
@@ -23,6 +23,10 @@ jobs:
         run: |
           python -m pip install --upgrade pip
           pip install build
+      - name: change mitiq version
+        run: |
+          truncate -s -1 VERSION.txt
+          echo "$(date +"%Y%m%d")" >> VERSION.txt
       - name: Build mitiq
         run: python -m build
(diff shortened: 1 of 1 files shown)


purva-thakre:
@natestemen I noticed you reran the failing windows workflow. 

Do you remember if it was the same as #2431 ?

natestemen:
Action logs are stored for 30 days. You can find logs from the first run [here](https://github.com/unitaryfund/mitiq/actions/runs/9765921429/job/26957722256#step:6:4326). Indeed it is the same failure. Thanks for opening the issue, and bringing it to my attention here!

purva-thakre:
> Action logs are stored for 30 days. You can find logs from the first run [here](https://github.com/unitaryfund/mitiq/actions/runs/9765921429/job/26957722256#step:6:4326).

I missed checking this. Thanks!

PR #2437, opened 2024-07-03
Flaky test on Windows for REM

purva-thakre:
## Description

Fixes #2431 by increasing the sample size. Increasing the sample size does not increase the run time by a lot. 

| Sample Size      | Time (in Seconds)|
| ----------- | ----------- |
| 1000      | 0.02572114800204872       |
| 10000   | 0.2219564580009319      |

~~This PR also splits up the docs build portion of the github workflows from everything else. This lets us rerun the pytest workflows without requiring us to rerun the docs job.~~ 

---

### License

- [ ] I license this contribution under the terms of the GNU GPL, version 3 and grant Unitary Fund the right to provide additional permissions as described in section 7 of the GNU GPL, version 3.

Before opening the PR, please ensure you have completed the following where appropriate.

- [ ] I added unit tests for new code.
- [ ] I used [type hints](https://www.python.org/dev/peps/pep-0484/) in function signatures.
- [ ] I used [Google-style](https://google.github.io/styleguide/pyguide.html#383-functions-and-methods) docstrings for functions.
- [ ] I [updated the documentation](../blob/main/docs/CONTRIBUTING_DOCS.md) where relevant.
- [ ] Added myself / the copyright holder to the AUTHORS file

PR #2441, opened 2024-07-05
Separate docs build workflow

purva-thakre:
## Description

As discussed during the community call today, this PR splits the docs build job into a separate workflow. This makes it easier for us to rerun pytest related jobs without waiting for the docs build to be completed. 

After this PR is merged, #2437 will be rebased. 

---

### License

### DIFF
--- a/.github/workflows/docs-build.yml
+++ b/.github/workflows/docs-build.yml
@@ -28,7 +28,7 @@ jobs:
       - name: Install Python dependencies
         run: |
           python -m pip install --upgrade pip
-          make install requirements
+          make install
 
       - name: Run the quilc & qvm Docker images
         run: |
--- a/.github/workflows/docs-build.yml
+++ b/.github/workflows/docs-build.yml
@@ -1,0 +1,62 @@
+name: docs-build
+
+on:
+  pull_request:
+    types:
+      - opened
+      - reopened
+      - synchronize
+      - ready_for_review
+    branches:
+      - main
+  push:
+    branches:
+      - main
+
+jobs:
+  docs:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Check out mitiq
+        uses: actions/checkout@v4
+
+      - name: Set up Python
+        uses: actions/setup-python@v5
+        with:
+          python-version: "3.12"
+
+      - name: Install Python dependencies
+        run: |
+          python -m pip install --upgrade pip
+          make install requirements
+
+      - name: Run the quilc & qvm Docker images
+        run: |
+          docker run --rm -idt -p 5000:5000 rigetti/qvm -S
+          docker run --rm -idt -p 5555:5555 rigetti/quilc -R
+
+      - name: Generate hash from all files excluding example notebooks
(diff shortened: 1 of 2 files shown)


purva-thakre:
Pinging @natestemen because I want to rerun the jobs in #2573 without waiting for the docs build.

.github/workflows/docs-build.yml
```diff
@@ -0,0 +1,62 @@
...
+
+      - name: Set up Python
+        uses: actions/setup-python@v5
+        with:
+          python-version: "3.12"
+
+      - name: Install Python dependencies
+        run: |
+          python -m pip install --upgrade pip
+          make install requirements
```
natestemen, line 31: IIRC this should do the same thing. Can you double check? Maybe in the past it was a different command, but now I think `make install xyz` performs the same action as `make install`.

```suggestion
          make install
```

natestemen (approved):
I don't see any problems with this. We'll see how it goes, and can revert if any problems arise.

.github/workflows/docs-build.yml
purva-thakre: I do not know how a makefile works that well. It's on my list of todos.  

> Maybe in the past it was a different command,

Looking through the targets, were we using `make install install` or `make install requirements` in the past? According to [this](https://superuser.com/a/360180), `make install` will install from the first installation target which is `install`. 

https://github.com/unitaryfund/mitiq/blob/231319df9051dfbc61db79d02bee201979528a36/Makefile#L49-L51

https://github.com/unitaryfund/mitiq/blob/231319df9051dfbc61db79d02bee201979528a36/Makefile#L59-L61

I see `development` is utilized in `setup.py` where we install things listed in `requirements/`. 

Why do we have two separate installation targets in the makefile? IMO both appear to behave the same, unless I am missing something.

purva-thakre:
Going to go ahead and merge this as one failure is what's in https://github.com/unitaryfund/mitiq/issues/2431

.github/workflows/docs-build.yml
natestemen: > `pip install -e .[development]`

This will install
1. an editable installation of Mitiq (`-e` is short for `--editable`)
2. **all** the dependencies (the `[development]` "tag" is defined in `setup.py` [here](https://github.com/unitaryfund/mitiq/blob/02ebc4bf5291b1d8abfec7516dd3bc5a9911cc37/setup.py#L37-L40) to mean all of the dependencies found in `requirements/`.)

> `pip install -r requirements/requirements.txt`

This will _only_ install the requirements listed in `requirements/requirements.txt` which is currently `numpy`, `scipy`, `cirq-core`, and `tabulate`. I can't find any uses of this command in Mitiq and it should be safe to remove. I've added a note for the community call this week to discuss.

cosenal: Discussed in the community call. Green light from everyone on removing it. 
@purva-thakre is going to create a PR to remove it (no need for a ticket, we can just refer to this discussion) and there we will make sure all the builds pass.

PR #2443, opened 2024-07-07
git ignore all coverage reports

cosenal:
E.g. coverage reports with filename `.coverage.Alessandros-MacBook.local.*` are not git ignored at the moment.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -2,7 +2,6 @@
 .vscode
 .python-version
 .idea/
-.coverage
 *.dat
 *.ipynb
 *.hdf5
@@ -14,5 +13,9 @@ docs/build/
 mitiq.egg-info/
 dist/
 build/
-coverage.xml
 jupyter_execute/
+
+# Coverage reports
+coverage.xml
+.coverage
+.coverage.*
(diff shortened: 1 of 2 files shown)


natestemen (approved):
We should probably also delete these files when `make clean` is run. Worth adding here if you agree

https://github.com/unitaryfund/mitiq/blob/e61c058777b11cc34b81be05e78b1c4bc91d7775/Makefile#L29

bdg221 (approved):
Looks good. I checked to see if there were any other files we wanted to ignore as well, but I think we have them all covered now.

I do like Nate's idea of cleaning up the files when `make clean` is run.

cosenal:
> We should probably also delete these files when `make clean` is run. Worth adding here if you agree
> 
> https://github.com/unitaryfund/mitiq/blob/e61c058777b11cc34b81be05e78b1c4bc91d7775/Makefile#L29

good catch, addressed in [7aa8d6d](https://github.com/unitaryfund/mitiq/pull/2443/commits/7aa8d6d6b4a178cef3ece4a92abaa2aa86968ca0)

natestemen (approved):
🚢 

CI failure is an instance of https://github.com/unitaryfund/mitiq/issues/2431. okay to merge.

PR #2447, opened 2024-07-10
LRE Inference Functions

### LINKED ISSUE #2372: Multivariate inference for LRE
Due to the specific scale factors list LRE requires, there will be a separate inference module for LRE.

Here, we want to define the monomials required for multivariate extrapolation which will be used to create the sample matrix. From the sample matrix, the coefficients of linear combination allow us to extrapolate to the zero-noise value.

purva-thakre:
## Description

Fixes #2372

To Dos

- [x] Add details to docstrings
- [x] Add unit tests with chunking
- [x] Make mypy happy
- [x] Unit tests for sample matrix
- [x] Function and unit test for linear combination coefficients function

---

### License


mitiq/lre/inference/multivariate_richardson.py
```diff
@@ -0,0 +1,128 @@
...
+
+    num_layers = _get_num_layers_without_measurements(input_circuit)
+
+    scale_factor_vectors = _get_scale_factor_vectors(
+        input_circuit, degree, fold_multiplier, num_chunks
+    )
+
+    monomial_terms = full_monomial_basis(num_layers, degree)
+    if len(monomial_terms) != len(scale_factor_vectors):
+        raise ValueError("Sample matrix will not be a square matrix.")
```
purva-thakre, line 113: Check if we need to raise this error.

mitiq/lre/inference/multivariate_richardson.py
```diff
@@ -0,0 +1,128 @@
...
+    fold_multiplier: int,
+    num_chunks: Optional[int] = None,
+):
+    """Calculates the sample matrix required for extrapolation."""
+    if degree < 1:
+        raise ValueError(
+            "Multinomial degree must be greater than or equal to 1."
+        )
+    if fold_multiplier < 1:
+        raise ValueError("Fold multiplier must be greater than or equal to 1.")
```
purva-thakre, line 103: These errors are already raised in the main noise scaling function. Do we need to raise these again?

mitiq/lre/inference/multivariate_richardson.py
```diff
@@ -0,0 +1,156 @@
...
+        _get_scale_factor_vectors(
+            input_circuit, degree, fold_multiplier, num_chunks
+        )
+    )
+    input_sample_matrix = sample_matrix(
+        input_circuit, degree, fold_multiplier, num_chunks
+    )
+
+    if np.linalg.det(input_sample_matrix) == 0:
+        raise ValueError("Sample matrix is singular.")
```
purva-thakre, line 145: It might be better to remove this check. 

If the determinant is zero, we should get the division by zero error.  

image

mitiq/lre/tests/test_multivariate_richardson.py
```diff
@@ -0,0 +1,239 @@
...
+                0.027777777777777804,
+                0.09722222222222232,
+            ],
+        ),
+        (
+            test_circuit2,
+            2,
+            2,
+            [
+                2.4062499999999956,
```
purva-thakre, line 163: Need to approximate before comparing the expected and calculated values. 

https://github.com/unitaryfund/mitiq/actions/runs/9984803425/job/27594634059?pr=2447#step:6:4462

mitiq/lre/inference/multivariate_richardson.py
```diff
@@ -0,0 +1,199 @@
...
+    Args:
+        num_layers: Number of layers in the input circuit.
+
+    Returns:
+        Variables required to create the monomial basis.
+    """
+    return [f"λ_{i}" for i in range(1, num_layers + 1)]
+
+
+def _create_variable_combinations(num_layers: int, degree: int) -> List[Any]:
```
vprusso, line 36: Should `List[Any]` be `List[List[Tuple[str, ...]]]` instead?

mitiq/lre/inference/multivariate_richardson.py
```diff
@@ -0,0 +1,199 @@
...
+        degree: Degree of the multivariate polynomial.
+
+    Returns:
+        Variable combinations required for the monomial basis.
+    """
+    variables = _get_variables(num_layers)
+    variable_combinations = []
+    for i in range(degree, -1, -1):
+        # Generate combinations for the current degree
+        # Ranges from max degree, max degree -1, ...0.
```
vprusso, line 50: `-1, ...0.` -> `-1, ..., 0.`

mitiq/lre/inference/multivariate_richardson.py
```diff
@@ -0,0 +1,199 @@
...
+        num_layers: Number of layers in the input circuit.
+        degree: Degree of the multivariate polynomial.
+
+    Returns:
+        Variable combinations required for the monomial basis.
+    """
+    variables = _get_variables(num_layers)
+    variable_combinations = []
+    for i in range(degree, -1, -1):
+        # Generate combinations for the current degree
```
vprusso, line 49: (Period at end of comment)

mitiq/lre/inference/multivariate_richardson.py
```diff
@@ -0,0 +1,199 @@
...
+        # Ranges from max degree, max degree -1, ...0.
+        combos = list(combinations_with_replacement(variables, i))
+        variable_combinations.append(combos)
+
+    # return a flattened list
```
vprusso, line 54: [nit]: Consistent comment casing and punctuation `Return a flattened list.`.

mitiq/lre/inference/multivariate_richardson.py
```diff
@@ -0,0 +1,199 @@
...
+    Degree of the polynomial dictates the max and min degree of the monomial
+    terms.
+
+    Args:
+        num_layers: Number of layers in the input circuit.
+        degree: Degree of the multivariate polynomial.
+
+    Returns:
+        Monomial basis terms required for multivariate
+            extrapolation upto max degree
```
vprusso, line 74: `upto` -> `up to` and period at end of sentence.

mitiq/lre/inference/multivariate_richardson.py
```diff
@@ -0,0 +1,199 @@
...
+        # Generate combinations for the current degree
+        # Ranges from max degree, max degree -1, ...0.
+        combos = list(combinations_with_replacement(variables, i))
+        variable_combinations.append(combos)
+
+    # return a flattened list
+    return list(chain(*variable_combinations))
+
+
+def full_monomial_basis(num_layers: int, degree: int) -> List[str]:
```
vprusso, line 58: I'm not sure if this pattern is any better, but one could simplify this function (and reduce the need for some of the private methods above) as:

```py
from itertools import combinations_with_replacement

def generate_monomials(num_layers: int, degree: int) -> Iterator[str]:
    variables = [f"λ_{i}" for i in range(1, num_layers + 1)]
    
    for d in range(degree, -1, -1):
        for var_tuple in combinations_with_replacement(variables, d):
            yield create_monomial(var_tuple)

def create_monomial(var_tuple: tuple[str, ...]) -> str:
    if not var_tuple:
        return "1"
    
    counts = Counter(var_tuple)
    monomial_parts = [
        f"{var}**{count}" if count > 1 else var
        for var, count in sorted(counts.items())
    ]
    return "*".join(monomial_parts)

def full_monomial_basis(num_layers: int, degree: int) -> list[str]:
    """
    Generate the monomial basis terms for a given number of layers and max degree.
    
    Args:
        num_layers: Number of layers in the input circuit.
        degree: Maximum degree of the multivariate polynomial.
    
    Returns:
        List of monomial basis terms, from highest to lowest degree.
    """
    return list(generate_monomials(num_layers, degree))
```

mitiq/lre/inference/multivariate_richardson.py
```diff
@@ -0,0 +1,199 @@
...
+        raise ValueError("Sample matrix will not be a square matrix.")
+    sample_matrix = np.zeros((len(monomial_terms), len(monomial_terms)))
+
+    # Evaluate the monomial terms using the values in the scale factor vectors
+    # and insert in the sample matrix
+    # each row is specific to each scale factor vector
+    # each column is a term in the monomial basis
+    for i, point in enumerate(scale_factor_vectors):
+        for j, monomial in enumerate(monomial_terms):
+            var_mapping = {f"λ_{k+1}": point[k] for k in range(num_layers)}
```
vprusso, line 154: Creating the `var_mapping` dictionary in each iteration of the nested loop is inefficient. We can move this outside the inner loop.

mitiq/lre/inference/multivariate_richardson.py
```diff
@@ -0,0 +1,199 @@
...
+            sample_matrix[i, j] = eval(monomial, {}, var_mapping)
```
vprusso, line 155: Using `eval()` can be [dangerous](https://stackoverflow.com/questions/1832940/why-is-using-eval-a-bad-practice). 

Consider writing and using something like this as an alternative: 

```py
def evaluate_monomial(monomial: str, var_mapping: dict[str, float]) -> float:
    """Safely evaluate a monomial string."""
    if monomial == "1":
        return 1.0
    terms = monomial.split("*")
    result = 1.0
    for term in terms:
        if "**" in term:
            var, power = term.split("**")
            result *= var_mapping[var] ** int(power)
        else:
            result *= var_mapping[term]
    return result
    ```

File: mitiq/lre/inference/multivariate_richardson.py [hunk 13]
```diff
@@ -0,0 +1,199 @@
...
+    Args:
+        input_circuit: Circuit to be scaled.
+        degree: Degree of the multivariate polynomial.
+        fold_multiplier: Scaling gap required by unitary folding.
+        num_chunks: Number of desired approximately equal chunks. When the
+            number of chunks is the same as the layers in the input circuit,
+            the input circuit is unchanged.
+
+    Returns:
+        Matrix of the evaluated monomial basis terms using the scale factor
```
vprusso, line 178: Should the return statement here instead be 

```
        Array of coefficients for the linear combination of scaled circuits.
```

mitiq/lre/inference/multivariate_richardson.py
```diff
@@ -0,0 +1,199 @@
...
+            sample_matrix[i, j] = eval(monomial, {}, var_mapping)
+    return sample_matrix
+
+
+def linear_combination_coefficients(
+    input_circuit: Circuit,
+    degree: int,
+    fold_multiplier: int,
+    num_chunks: Optional[int] = None,
+) -> List[int]:
```
vprusso, line 164: Is the type here `NDArray[np.float64]` instead of `List[int]`?

mitiq/lre/inference/multivariate_richardson.py
```diff
@@ -0,0 +1,199 @@
...
+            input_circuit, degree, fold_multiplier, num_chunks
+        )
+    )
+    input_sample_matrix = sample_matrix(
+        input_circuit, degree, fold_multiplier, num_chunks
+    )
+
+    coeff_list = []
+    for i in range(num_layers):
+        sample_matrix_copy = input_sample_matrix.copy()
```
vprusso, line 192: Copying the entire matrix for each iteration. Can these be moved out of the loop?

mitiq/lre/inference/multivariate_richardson.py
```diff
@@ -0,0 +1,199 @@
...
+        sample_matrix_copy[i] = np.array([[1] + [0] * (num_layers - 1)])
+        coeff_list.append(
+            np.linalg.det(sample_matrix_copy)
```
vprusso, line 195: Using determinants for coefficient calculation can be numerically unstable for larger matrices. We may want to have a comment or raise a warning about this if the matrix is particularly large.

mitiq/lre/tests/test_multivariate_richardson.py
```diff
@@ -0,0 +1,239 @@
...
+from mitiq.lre.multivariate_scaling.layerwise_folding import (
+    _get_num_layers_without_measurements,
+    _get_scale_factor_vectors,
+)
+
+
+@pytest.mark.parametrize(
+    "test_num_layers, expected_list",
+    [
+        (2, ["λ_1", "λ_2"]),
```
vprusso, line 30: We may also want to include a test for when the number of layers is 1 (and also circuits with lots of layers)

```py
@pytest.mark.parametrize("num_layers", [1, 2, 10, 100])
def test_get_variables_various_sizes(num_layers):
    variables = _get_variables(num_layers)
    assert len(variables) == num_layers
    assert all(f"λ_{i}" in variables for i in range(1, num_layers + 1))
```

mitiq/lre/tests/test_multivariate_richardson.py
```diff
@@ -0,0 +1,239 @@
...
+)
+def test_get_variables(test_num_layers, expected_list):
+    calculated_variables = _get_variables(test_num_layers)
+    assert len(calculated_variables) == test_num_layers
+    assert calculated_variables == expected_list
+
+
+@pytest.mark.parametrize(
+    "test_num_layers, test_degree",
+    [(2, 2), (3, 2), (4, 2), (2, 3), (3, 3)],
```
vprusso, line 43: If the test timing is not too unreasonable, I think we should include tests for higher layers and degrees.

mitiq/lre/tests/test_multivariate_richardson.py
```diff
@@ -0,0 +1,239 @@
...
+                    [1.0, 3.0, 1.0, 3.0, 9.0, 3.0, 1.0, 9.0, 3.0, 9.0],
+                    [1.0, 1.0, 5.0, 1.0, 1.0, 5.0, 25.0, 1.0, 5.0, 1.0],
+                    [1.0, 3.0, 3.0, 1.0, 9.0, 9.0, 9.0, 3.0, 3.0, 1.0],
+                    [1.0, 5.0, 1.0, 1.0, 25.0, 5.0, 1.0, 5.0, 1.0, 1.0],
+                ]
+            ),
+        ),
+    ],
+)
+def test_sample_matrix(test_circ, test_degree, expected_matrix):
```
vprusso, line 138: We may want to include tests for the stability of the numerical stability of the sample matrix calculation as well. Something like:

```py
def test_sample_matrix_numerical_stability():
    large_circuit = Circuit([ops.H.on(LineQubit(i)) for i in range(100)])
    matrix = sample_matrix(large_circuit, 5, 1000)
    assert np.isfinite(matrix).all()
    assert not np.isnan(matrix).any()
```

mitiq/lre/tests/test_multivariate_richardson.py
```diff
@@ -0,0 +1,239 @@
...
+from math import comb
+
+import numpy as np
+import pytest
+from cirq import Circuit, LineQubit, ops
+
+from mitiq.lre.inference.multivariate_richardson import (
+    _create_variable_combinations,
+    _get_variables,
+    full_monomial_basis,
```
vprusso, line 17: We could provide more thorough coverage for `full_monomial_basis` and `linear_combination_coefficients` as these could use more complete coverage. 

Similarly, with the helper functions `_get_variables` and `create_variable_combinations` (although these are used and tested implicitly via your other tests, so perhaps that's not a high priority).

mitiq/lre/tests/test_multivariate_richardson.py
```diff
@@ -0,0 +1,239 @@
...
+                0.06249999999999997,
+                0.06249999999999997,
+                0.15624999999999956,
+                0.06249999999999997,
+                0.15624999999999956,
+            ],
+        ),
+    ],
+)
+def test_coeffs(test_circ, test_degree, test_fold_multiplier, expected_matrix):
```
vprusso, line 177: Instead of checking equality here, should we consider using something like`np.isclose`?

mitiq/lre/tests/test_multivariate_richardson.py
```diff
@@ -0,0 +1,239 @@
+# Copyright (C) Unitary Fund
+#
+# This source code is licensed under the GPL license (v3) found in the
+# LICENSE file in the root directory of this source tree.
+
+"""Unit tests for multivariate extrapolation inference functions."""
```
vprusso, line 6: It might be nice to include at least a few end-to-end integration tests as well that test the entire workflow from circuit creation to coefficient calculation.

mitiq/lre/tests/test_multivariate_richardson.py
```diff
@@ -0,0 +1,239 @@
...
+    """
+    num_layers = _get_num_layers_without_measurements(test_input)
+    calculated_basis = full_monomial_basis(num_layers, degree)
+    calculated_scale_factor_vectors = _get_scale_factor_vectors(
+        test_input, degree, test_fold_multiplier
+    )
+    assert len(calculated_basis) == len(calculated_scale_factor_vectors)
+
+
+def test_lre_inference_with_chunking():
```
vprusso, line 235: We could probably also include tests that make use of the `num_chunks` parameter:

```py
@pytest.mark.parametrize("num_chunks", [None, 2, 4])
def test_linear_combination_coefficients_with_chunking(num_chunks):
    coeffs = linear_combination_coefficients(test_circuit2, 2, 2, num_chunks)
    assert len(coeffs) == _get_num_layers_without_measurements(test_circuit2)
    assert np.isclose(sum(coeffs), 1.0)  # Coefficients should sum to 1
```

vprusso (commented):
Overall looks great. Nice work, @purva-thakre ! 

Added some comments, but many of them are quite minor.

mitiq/lre/tests/test_multivariate_richardson.py
purva-thakre: I'll try a different method of implementing this. 

`_get_num_layers_without_measurements` does not use `num_chunks`.

https://github.com/unitaryfund/mitiq/blob/231319df9051dfbc61db79d02bee201979528a36/mitiq/lre/multivariate_scaling/layerwise_folding.py#L23

mitiq/lre/tests/test_multivariate_richardson.py
purva-thakre: Yes, this is a better option than using `np.round`

mitiq/lre/inference/multivariate_richardson.py
purva-thakre: At the next step, we are replacing 1 row by `np.array([[1] + [0] * (num_layers - 1)])`.

If we don't replace `sample_matrix_copy` at each iteration, we are going to use the matrix created in the next row.

mitiq/lre/inference/multivariate_richardson.py
purva-thakre: No, it's a list. See `test_coeffs` in the unit tests.

mitiq/lre/inference/multivariate_richardson.py
purva-thakre: `mypy` expected it to be `*List[List[Tuple[str, ...]]]` but using `*` led to syntax issues. So, I chose the easy way out.

mitiq/lre/tests/test_multivariate_richardson.py
purva-thakre: I was thinking of adding these unit tests when I have defined the `execute_with_lre` function.

mitiq/lre/inference/multivariate_richardson.py
natestemen: I'm joining the code golf:
```py
import itertools

def full_monomial_basis(num_layers: int, degree: int) -> list[str]:
    variables = [f"λ_{i+1}" for i in range(num_layers)]
    terms = []

    for exponents in itertools.product(range(degree + 1), repeat=num_layers):
        if sum(exponents) <= degree:
            term = " * ".join(
                f"{variables[i]}^{exponents[i]}"
                for i in range(num_layers)
                if exponents[i] > 0
            )
            if term == "":
                term = "1"
            terms.append(term)
    return terms


print(full_monomial_basis(3, 2))
>>> ['1', 'λ_3^1', 'λ_3^2', 'λ_2^1', 'λ_2^1 * λ_3^1', 'λ_2^2', 'λ_1^1', 'λ_1^1 * λ_3^1', 'λ_1^1 * λ_2^1', 'λ_1^2']
```

mitiq/lre/inference/multivariate_richardson.py
natestemen: I think we should generally try to avoid storing the monomials as strings as this requires us to do lots of string manipulation. Why use strings when we can live in number land the whole time?

To this end, what if we store the monomials as tuples of exponents instead of strings? E.g. $x_0^0x_1^3x_2^1x_3^2$ gets stored as $(0, 3, 1, 2)$ whereupon we can then use an evaluate function that looks a little simpler:[^1]
```py
def evaluate_monomial(exponents: tuple[int, ...], values: list[float]) -> float:
    result = 1
    for exp, value in zip(exponents, values):
        result *= value ** exp
    return result
```

WDYT? If we went this route, the `full_monomial_basis` function can be simplified quite a bit to something like
```py
def full_monomial_basis(n: int, d: int) -> List[Tuple[int, ...]]:
    return [exps for exps in itertools.product(range(d + 1), repeat=n) if sum(exps) <= d]
```


[^1]: code not guaranteed to work :)

natestemen (changes requested):
Very cool stuff here.

I left some comments related to things Vincent raised. I think we can make some big simplifications with how we do monomials, but otherwise looking good.

mitiq/lre/inference/multivariate_richardson.py
purva-thakre: > To this end, what if we store the monomials as tuples of exponents instead of strings?

I like this idea!

mitiq/lre/inference/multivariate_richardson.py
purva-thakre: I had thought of adding this warning. I decided to skip it because I figured an overflow error or a runtime warning would get raised anyway for large matrices.

mitiq/lre/inference/multivariate_richardson.py
```diff
@@ -0,0 +1,201 @@
...
+    num_layers = len(
+        _get_scale_factor_vectors(
+            input_circuit, degree, fold_multiplier, num_chunks
+        )
+    )
+    input_sample_matrix = sample_matrix(
+        input_circuit, degree, fold_multiplier, num_chunks
+    )
+
+    coeff_list = []
```
purva-thakre, line 192: Haven't been able to replicate this warning in numpy yet. Numpy raises this error for large martices is taken from https://stackoverflow.com/q/39676437

mitiq/lre/inference/multivariate_richardson.py
```diff
@@ -0,0 +1,211 @@
...
+        )
+    )
+    input_sample_matrix = sample_matrix(
+        input_circuit, degree, fold_multiplier, num_chunks
+    )
+
+    try:
+        det = np.linalg.det(input_sample_matrix)
+
+    except (RuntimeWarning, det == 0, np.isnan(det)):
```
purva-thakre, line 196: Haven't been able to replicate this warning in numpy yet. Numpy will raise this error for large matrices according to https://stackoverflow.com/q/39676437

mitiq/lre/inference/multivariate_richardson.py
purva-thakre: @vprusso How do you decide when to use a generator and when to use a list?

```python
    for d in range(degree, -1, -1):
        for var_tuple in combinations_with_replacement(variables, d):
            yield create_monomial(var_tuple)
```
My first instinct here is to use list comprehension `list(combinations_with_replacement(variables, i))`

vprusso: Using a generator is better because it handles large combinations efficiently by yielding items one at a time. This approach saves memory and processes each monomial as needed.

purva-thakre: > Using a generator is better because it handles large combinations efficiently by yielding items one at a time.

Ok. For this PR I will rely on list comprehension. I will create a separate issue to look at the cases when we might need this. So far, based on the unit tests (not pushed yet), I don't see a need for generators.

natestemen: My general rule for when to use a generator is "am I _just_ going to iterate over this?" If the answer is yes, and I don't need indices, or to modify the order, then generator it is!

> list comprehension `list(combinations_with_replacement(variables, d))`

This is casting as a list, but not exactly a list comprehension. A list comprehension in python is something like `[x.something for x in another_iterable]`. Just FYI!

mitiq/lre/inference/multivariate_richardson.py
```diff
@@ -0,0 +1,228 @@
...
+    for i in var_exp_sorted:
+        var_prod_i = []
+        for j in range(1, num_var + 1):
+            if i[j] > 0:
+                var_prod_i.append((str_var[j - 1]) ** (i[j]))
+            else:
+                var_prod_i.append(1)
+        var_prod.append(var_prod_i)
+
+    return [np.prod(item) for item in var_prod]
```
purva-thakre, line 85: @natestemen @vprusso I changed the pattern as discussed for the full basis function. I'd appreciate it if you could take a look at 
`full_monomial_basis_terms`and `_full_monomial_basis_term_exponents`.

I still haven't pushed the remaining unit tests as I need to make changes to the spaghetti code in `sample_matrix`.

mitiq/lre/inference/multivariate_richardson.py
```diff
@@ -0,0 +1,228 @@
...
+    terms.
+
+    Args:
+        num_layers: Number of layers in the input circuit.
+        degree: Degree of the multivariate polynomial.
+
+    Returns:
+        Monomial basis terms required for multivariate
+            extrapolation up to max degree
+    """
```
purva-thakre, line 64: Note: This function is not used to evaluate the elements of the sample matrix. This is now a helper function for someone who wants to know which monomial terms are used to define the rows of the sample matrix.

mitiq/lre/inference/multivariate_richardson.py
```diff
@@ -0,0 +1,234 @@
...
+from sympy import Symbol
+
+from mitiq.lre.multivariate_scaling.layerwise_folding import (
+    _get_scale_factor_vectors,
+)
+
+
+def _full_monomial_basis_term_exponents(
+    num_layers: int, degree: int
+) -> List[Dict[int, int]]:
```
natestemen, line 27: Instead of using the dictionary to track the degree of each monomial, what if we make the return type `list[tuple[int, ...]]` and ensure they are ordered in the manner that is required later? I was able to achieve this using
```py
exponents = [
    exps
    for exps in itertools.product(range(degree + 1), repeat=num_layers)
    if sum(exps) <= degree
]

return sorted(exponents, key=lambda term: (sum(term), term[::-1]))
```

This does seem to be a bit slower, but `exponents` could be generated differently, and then still make user of the `sorted` call/key.

I think this might help clean up some of the code in `sample_matrix` as well.

mitiq/lre/inference/multivariate_richardson.py
```diff
@@ -0,0 +1,234 @@
...
+            + "the matrix is too large. Consider chunking your"
+            + " input circuit. "
+        )
+    assert det != 0.0
+
+    coeff_list = []
+    for i in range(num_layers):
+        sample_matrix_copy = input_sample_matrix.copy()
+        sample_matrix_copy[i] = np.array([[1] + [0] * (num_layers - 1)])
+        coeff_list.append(np.linalg.det(sample_matrix_copy) / det)
```
natestemen, line 232: This looks like we are computing the determinant of the minor. If so, can we just delete the values and compute the determinant? Then we can avoid the copying.
```suggestion
        sample_matrix_minor = np.delete(np.delete(sample_matrix, (0), axis=1), (i), axis=0)
        coeff_list.append(np.linalg.det(sample_matrix_minor) / det)
```

natestemen (commented):
Nice changes! I'm hopeful we can simplify things a bit more, but I might not have the full picture in my head to understand why certain things need to be that way.

mitiq/lre/inference/multivariate_richardson.py
```diff
@@ -0,0 +1,234 @@
...
+
+import warnings
+from collections import Counter
+from itertools import combinations_with_replacement
+from typing import Any, Dict, List, Optional
+
+import numpy as np
+from cirq import Circuit
+from numpy.typing import NDArray
+from sympy import Symbol
```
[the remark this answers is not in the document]
purva-thakre: @natestemen Not use `sympy`?

mitiq/lre/inference/multivariate_richardson.py
```diff
@@ -0,0 +1,234 @@
...
+    for combo_key in variable_exp_counter:
+        for j in variables:
+            if j not in combo_key:
+                combo_key[j] = 0
+    # return a dictionary with variable number as the key with exponent values
+    # first term is the 0 degree term's exponent
+    return variable_exp_counter[::-1]
+
+
+def full_monomial_basis_terms(num_layers: int, degree: int) -> List[str]:
```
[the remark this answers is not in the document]
purva-thakre: FIne by me.

mitiq/lre/inference/multivariate_richardson.py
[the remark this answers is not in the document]
natestemen: Correct. But if we're deleting `full_monomial_basis_terms` as agreed in https://github.com/unitaryfund/mitiq/pull/2447#discussion_r1706284381, this is a moot point.

purva-thakre:
@natestemen This is ready with your suggestions. 

I removed 2 unit tests in [42fd2b1](https://github.com/unitaryfund/mitiq/pull/2447/commits/42fd2b162c3a4c4adcaa82f8b8db435e3252217f) because `np.sum` is not behaving properly. I am looking into this. 

When I add the unit test to check if all the coeffs sum up to 1, the tests fail. But if I manually check this in a jupyter notebook, they sum up close to 1. 🤷🏾‍♀️

mitiq/lre/inference/multivariate_richardson.py
```diff
@@ -0,0 +1,159 @@
...
+# This source code is licensed under the GPL license (v3) found in the
+# LICENSE file in the root directory of this source tree.
+
+"""Functions for multivariate richardson extrapolation as defined in
+:cite:`Russo_2024_LRE`.
+"""
+
+import warnings
+from itertools import product
+from typing import Any, List, Optional, Tuple
```
purva-thakre, line 12: it might be better to use the built-in default types here

purva-thakre:
> because `np.sum` is not behaving properly. I am looking into this.

I think there is an issue when we are using `assert` to verify `np.sum`.

```python
import numpy as np
import pytest

smple_array = np.array([0.28649533, 0.39635102, 0.53429875, 0.66818352, 0.44364975,
       0.88817929, 0.49318006, -0.9223426 , 0.33117286, -0.86472772,
       0.57138075, 0.91268509, 0.14781393, -0.16693806, 0.36417079])

# prints out to approximately 4.084....
print(np.sum(smple_array))
# results in AssertionError
assert np.isclose(np.sum(smple_array), 4)
# prints to approximately -0.08355.....
print(4-np.sum(smple_array))
# results in AssertionError
assert np.isclose(4-np.sum(smple_array), -0.08)

```

mitiq/lre/inference/multivariate_richardson.py
vprusso: I agree with this comment in general, but most of `mitiq` uses this pattern. I think it might be good to have a separate issue that just covers using native types for typing throughout the whole codebase and keeping this as is (for now) for consistency.

mitiq/lre/inference/multivariate_richardson.py
```diff
@@ -0,0 +1,159 @@
...
+
+    return sample_matrix
+
+
+def linear_combination_coefficients(
+    input_circuit: Circuit,
+    degree: int,
+    fold_multiplier: int,
+    num_chunks: Optional[int] = None,
+) -> List[np.float64]:
```
natestemen, line 104:
```suggestion
) -> List[float]:
```

mitiq/lre/inference/multivariate_richardson.py
```diff
@@ -0,0 +1,159 @@
...
+                # raise scale factor value by the exponent dict value
+                evaluated_terms.append(base**exp)
+            # multiply both elements in the list to create an evaluated
+            # monomial term
+            sample_matrix[rows, cols] = np.prod(evaluated_terms)
+
+    return sample_matrix
+
+
+def linear_combination_coefficients(
```
natestemen, line 99: This could use a more descriptive name and docstring. It's hard to get a grasp of what it's doing from these two data points.

mitiq/lre/inference/multivariate_richardson.py
```diff
@@ -0,0 +1,159 @@
...
+    input_circuit: Circuit,
+    degree: int,
+    fold_multiplier: int,
+    num_chunks: Optional[int] = None,
+) -> List[np.float64]:
+    r"""
+    Defines the sample matrix required for multivariate extrapolation as
+    defined in :cite:`Russo_2024_LRE`.
```
natestemen, line 107: This looks like it was copied from the `sample_matrix` docstring.

mitiq/lre/inference/multivariate_richardson.py
```diff
@@ -0,0 +1,159 @@
...
+        # taken from https://stackoverflow.com/a/19317237
+        warnings.warn(  # pragma: no cover
+            "To account for overflow error, required determinant of "
+            + "large sample matrix is calculated through "
+            + "`np.linalg.slogdet`."
+        )
+        sign, logdet = np.linalg.slogdet(  # pragma: no cover
+            input_sample_matrix
+        )
+        det = np.exp(logdet)  # pragma: no cover
```
natestemen, line 141: Need to keep the sign, right?
```suggestion
        det = sign * np.exp(logdet)  # pragma: no cover
```

mitiq/lre/tests/test_multivariate_richardson.py
```diff
@@ -0,0 +1,221 @@
...
+    assert (
+        _full_monomial_basis_term_exponents(test_num_layers, test_degree)
+        == expected
+    )
+
+
+@pytest.mark.parametrize(
+    "test_num_layers, test_degree",
+    [(1, 1), (2, 2), (3, 2), (10, 4)],
+    # Note need to add (100, 2) here. This makes the unit test very slow.
```
natestemen, line 71: This should either be removed, or we'll need to refactor `_full_monomial_basis_term_exponents` to speed it up. If we need to speed it up, what are the largest values it should be performant at?

mitiq/lre/tests/test_multivariate_richardson.py
```diff
@@ -0,0 +1,221 @@
...
+):
+    """Ensures that the args for the sample matrix
+    an error for an invalid value."""
+    with pytest.raises(ValueError, match=error_msg):
+        sample_matrix(test_input, test_degree, test_fold_multiplier)
+
+
+def test_lre_inference_with_chunking():
+    circ = test_circuit1 * 7
+    chunked_sample_matrix_dim = sample_matrix(circ, 2, 2, 4).shape
```
natestemen, line 201: Just to make it explicit that `num_chunks` is being used here.
```suggestion
    chunked_sample_matrix_dim = sample_matrix(circ, 2, 2, num_chunks=4).shape
```

mitiq/lre/tests/test_multivariate_richardson.py
```diff
@@ -0,0 +1,221 @@
...
+
+def test_sample_matrix_numerical_stability():
+    large_circuit = Circuit([ops.H.on(LineQubit(i)) for i in range(10000)])
+    matrix = sample_matrix(large_circuit, 5, 10000)
+    assert np.isfinite(matrix).all()
+    assert not np.isnan(matrix).any()
+
+
+@pytest.mark.parametrize("num_chunks", [2, 3])
+def test_eval(num_chunks):
```
natestemen, line 214: Need a more descriptive test name here.

natestemen (commented):
Nice clean up! Getting closer!

I'm lacking some understanding as to how the functions added here fit together, but maybe that will come in the final PR? This is usually where I would rely on the PR description to add context for the reviewer.

mitiq/lre/tests/test_multivariate_richardson.py
purva-thakre: > we'll need to refactor _full_monomial_basis_term_exponents to speed it up

I was thinking of leaving this for the immediate future.

natestemen: Sounds good. Let's finish the full implementation and then come back to it. Please add `TODO` to the comment so it's a bit more discoverable.

mitiq/lre/tests/test_multivariate_richardson.py
```diff
@@ -0,0 +1,221 @@
...
+                    [1.0, 1.0, 1.0, 5.0, 1.0, 1.0, 1.0, 5.0, 5.0, 25.0],
+                ]
+            ),
+        ),
+    ],
+)
+def test_sample_matrix(test_circ, test_degree, expected_matrix):
+    assert (
+        expected_matrix - sample_matrix(test_circ, test_degree, 1)
+    ).all() <= 1e-3
```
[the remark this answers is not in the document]
purva-thakre: Thanks for catching this! 

Apparently, the two failures I noticed in https://github.com/unitaryfund/mitiq/pull/2447#issuecomment-2293634688 were related to this.  

Because the tests passed, I thought there was a float addition error. I will work on finding a way around the `for` block where I was copying the array.

mitiq/lre/inference/multivariate_richardson.py
purva-thakre: @vprusso You missed an earlier discussion on this: https://github.com/unitaryfund/mitiq/pull/2464#discussion_r1718343500

mitiq/lre/inference/multivariate_richardson.py
purva-thakre: > a more descriptive name

@natestemen I could think of two names. Let me know which one appears descriptive:

1. `multivariate_richardson_coefficients`
2. `lre_coefficients`
3. `linear_combination_coefficients_from_sample_matrix`

mitiq/lre/inference/multivariate_richardson.py
```diff
@@ -0,0 +1,170 @@
...
+        elif i == mat_row - 1:  # last row
+            new_mat = np.concatenate(
+                (input_sample_matrix[:i], repl_row), axis=0
+            )
+        else:
+            frst_sl = np.concatenate(
+                (input_sample_matrix[:i], repl_row), axis=0
+            )
+            sec_sl = input_sample_matrix[i + 1 :]
+            new_mat = np.concatenate((frst_sl, sec_sl), axis=0)
```
purva-thakre, line 166: @natestemen @vprusso Does this sound like a better alternative to the [previous version](https://github.com/unitaryfund/mitiq/blob/7d8868b94fd7404e552f6703b335893e13dc17fc/mitiq/lre/inference/multivariate_richardson.py#L192) where I was copying the sample matrix? 

Both of you pointed out memory issues with using `.copy()` on a numpy array.

mitiq/lre/inference/multivariate_richardson.py
```diff
@@ -0,0 +1,170 @@
...
+        )
+        det = sign * np.exp(logdet)  # pragma: no cover
+
+    if np.isinf(det):
+        raise ValueError(  # pragma: no cover
+            "Determinant of sample matrix cannot be calculated as "
+            + "the matrix is too large. Consider chunking your"
+            + " input circuit. "
+        )
+    assert det != 0.0
```
vprusso, line 145: Do we want to check whether it's close to 0.0 and add an appropriate debug message in case it is not?

mitiq/lre/inference/multivariate_richardson.py
```diff
@@ -0,0 +1,170 @@
...
+
+    coeff_list = []
+    mat_row, mat_cols = input_sample_matrix.shape
+    assert mat_row == mat_cols
```
vprusso, line 149: Similar question here as to whether we want to provide more details to the end-user as to what the issue would have been here if this assert fails.

mitiq/lre/inference/multivariate_richardson.py
```diff
@@ -0,0 +1,170 @@
...
+        degree: Degree of the multivariate polynomial.
+        fold_multiplier: Scaling gap required by unitary folding.
+        num_chunks: Number of desired approximately equal chunks. When the
+            number of chunks is the same as the layers in the input circuit,
+            the input circuit is unchanged.
+
+    Returns:
+        List of the evaluated monomial basis terms using the scale factor
+            vectors.
+    """
```
vprusso, line 121: Should we add error handling? 

```py
if degree < 1:
    raise ValueError("Degree must be a positive integer")
if fold_multiplier < 1:
    raise ValueError("Fold multiplier must be a positive integer")
if num_chunks is not None and num_chunks < 1:
    raise ValueError("Number of chunks must be a positive integer")
```

mitiq/lre/inference/multivariate_richardson.py
vprusso: This seems okay from my perspective!

mitiq/lre/inference/multivariate_richardson.py
purva-thakre: Discussed with Nate to use Option 1.

mitiq/lre/inference/multivariate_richardson.py
purva-thakre: Discussed this with @natestemen yesterday. We are coming back to think about edge cases like this once the general LRE implementation is complete. 

I'll go ahead and create an issue to log all the todos I need to tackle in the immediate future.

mitiq/lre/inference/multivariate_richardson.py
purva-thakre: Discussed with @natestemen to revert back to using `.copy()` on a numpy array because `np.concatenate` might have the same memory issues as `.copy()`. 

The code is more readable with `.copy()`. Will add this as a todo as discussed in https://github.com/unitaryfund/mitiq/pull/2447#discussion_r1729114577

mitiq/lre/inference/multivariate_richardson.py
purva-thakre: See https://github.com/unitaryfund/mitiq/pull/2447#discussion_r1729114577

mitiq/lre/inference/multivariate_richardson.py
```diff
@@ -0,0 +1,195 @@
...
+    _get_scale_factor_vectors,
+)
+
+
+def _full_monomial_basis_term_exponents(
+    num_layers: int, degree: int
+) -> list[tuple[int, ...]]:
+    """Finds exponents of monomial terms required to create the sample matrix.
+
+    $Mj(λ_i, d)$ is the basis of monomial terms for $l$-layers in the input
```
natestemen, line 28: If you're going to reference a piece of notation make sure to provide a reference for where it comes from. There is a reference in the top of this file, but maybe an equation number would be helpful? WDYT?

mitiq/lre/inference/multivariate_richardson.py
```diff
@@ -0,0 +1,195 @@
...
+    sample_matrix = np.empty((len(variable_exp), len(variable_exp)))
+
+    for i, scale_factors in enumerate(scale_factor_vectors):
+        for j, exponent in enumerate(variable_exp):
+            evaluated_terms = []
+            for base, exp in zip(scale_factors, exponent):
+                # raise scale factor value by the exponent dict value
+                evaluated_terms.append(base**exp)
+            # multiply both elements in the list to create an evaluated
+            # monomial term
```
natestemen, line 114: The code is satisfactorily self-explanatory :)

```suggestion
```

Probably for the comment above this one as well, but up to you.

mitiq/lre/inference/multivariate_richardson.py
```diff
@@ -0,0 +1,195 @@
...
+            sample_matrix[i, j] = np.prod(evaluated_terms)
+
+    # verify the matrix is square
+    mat_row, mat_cols = sample_matrix.shape
+    assert mat_row == mat_cols
```
natestemen, line 119: We typically use tests for this type of thing. Is there a reason we should have this `assert` in the code here?

natestemen (approved):
Approved, but a few small cleanup things to do before merging!

Well done with this Purva!

mitiq/lre/inference/multivariate_richardson.py
purva-thakre: I am making my way through Dan Bader's 'Python Tricks: The Book'. He recommends using assertions like this.

mitiq/lre/inference/multivariate_richardson.py
```diff
@@ -0,0 +1,195 @@
...
+    $Mj(λ_i, d)$ is the basis of monomial terms for $l$-layers in the input
+    circuit up to a specific degree $d$. The linear combination defines our
+    polynomial of interest. In general, the number of monomial terms for a
+    variable $l$ up to degree $d$  can be determined through the Stars and
+    Bars method.
+
+    We assume the terms in the monomial basis are arranged in a graded
+    lexicographic order such that the terms with the highest total degree are
+    considered to be the largest and the remaining terms are arranged in
+    lexicographic order.
```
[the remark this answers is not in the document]
natestemen: Okay, so I think all the confusion boils down to a potential confusion in convention.

A tuple of exponents $(a_1, a_2, \ldots, a_n)$ can be taken to represent **either** $x_1^{a_1}x_2^{a_2}\cdots x_n^{a_n}$ **or** $x_1^{a_n}x_2^{a_{n-1}}\cdots x_n^{a_1}$. The choice here is unimportant in it's own right, but important that we know which one we are working in and stay consistent.

**If** we are working in the second convention (which seems a bit strange to me[^1]) then I think everything is correct. In my head, I was previously working with the former convention which means the graded lexicographic ordering on monomials with two terms and maximum degree two ($1 < x_2 < x_1 < x_2^2 < x_1x_2 < x_1^2$) would read
```math
(0, 0) < (0, 1) < (1, 0) < (0, 2) < (1, 1) < (2, 0)
```
whereas the latter convention reads
```math
(0, 0) < (1, 0) < (0, 1) < (2, 0) < (1, 1) < (0, 2).
```
The latter convention indeed yields what is currently implemented.

So the question is **does $(0, 1)$ represent $x_1$ or $x_2$**?

[^1]: It's convention, though, not gospel, so 🤷🏻.

PR #2451, opened 2024-07-25
CDR Tutorial with Qrack

### LINKED ISSUE #2368: Tutorial using CDR with Qrack as the near-Clifford circuit simulator
When updating the documentation recently on CDR, the need for efficient near-Clifford circuit simulators became apparent. One of the known efficient near-Clifford circuit simulators is Qrack. It only makes sense to have a tutorial showing CDR at scale using Qrack.

I mentioned this to @WrathfulSpatula and he was very encouraging. He pointed me to https://github.com/vm6502q/pyqrack-jupyter where there are examples of using the Clifford RZ simulation.

This issue does seem similar to #2003 though that did not involve CDR.

bdg221:
## Description
Fixes #2368 

This PR is to add a tutorial to the documentation using CDR with Qrack as the Near-Clifford simulator.

I started with the CDR section of the User Guide and replaced `compute_density_matrix` with Qrack for the Near-Clifford simulation and a Qiskit Fake Backend that uses a snapshot of the configurations and noise of the Yorktown device.

---

### License


bdg221:
@WrathfulSpatula and @natestemen I would appreciate if you could check this PR.

Dan, does everything look good from the Qrack side?

For the doc build in the PR, pyqrack is not included in the environment. I'm not sure how to do the install (since it requires a python kernel reboot.)

purva-thakre:
@bdg221 I haven't dug into what the error message says exactly. 

I looked through the diff and I think you should add `pyqrack ` as a dependency in `mitiq/requirements/`.

bdg221:
@purva-thakre exactly. The error is coming from the `os._exit(00)` call after the `pip install pyqrack --quiet`.

I don't think it makes sense to include pyqrack as a dependency, but is there a way to include it for building the docs?

purva-thakre:
@bdg221 Yes, I get what you are saying but there's an easy way to install it.

List your dependency below:

https://github.com/unitaryfund/mitiq/blob/02ebc4bf5291b1d8abfec7516dd3bc5a9911cc37/requirements/requirements-dev.txt#L9

You can see some dependencies are only for some docs examples.

bdg221:
@WrathfulSpatula when the docs are getting built and pyqrack is imported, the following error is seeing https://mitiq--2451.org.readthedocs.build/en/2451/examples/cdr_qrack.html:
```
/usr/lib/qrack/libqrack_pinvoke.so: cannot open shared object file: No such file or directory
```

You can see info about the environment (linux, python 3.12, pyqrack 1.30.0) here: https://github.com/unitaryfund/mitiq/actions/runs/10116764224/job/27980295325?pr=2451

WrathfulSpatula:
The docs-building runner is probably Linux, if it's looking for
libqrack_pinvoke.so. This shouldn't happen so long as you install PyQrack
either with `pip3 install pyqrack` or
```
sudo add-apt-repository ppa:wrathfulspatula/vm6502q
sudo apt install pyqrack
```

Are you installing PyQrack from source, from the repository? That might
explain the issue, since you're not building and installing the Qrack
shared library, and it's not packaged on the main branch, which will resort
to looking for the system-wide installation in `/usr` or `/usr/local`.

WrathfulSpatula:
Come to think of it, that explanation might have been close but not quite
on-the-money: if there's not a specific PyQrack wheel for the platform on
which you install, you get the source build, which is effectively the same
as head of main branch on the PyQrack repository. If you install Qrack
system-wide on the runner, it should fix it.

Alternatively, if you know a valid wheel platform label I can build that
PyQrack doesn't already provide, I can probably package it for you
immediately and add it as a new wheel to 1.30.0.

docs/source/examples/cdr_qrack.md
```diff
@@ -0,0 +1,177 @@
...
+
+unmitigated_expval = qiskit_noisy(circuit)
+print(f"Unmitigated expectation value from Qiskit Fake backend: {unmitigated_expval:.3f}")
+
+mitigated_expval = cdr.execute_with_cdr(
+    circuit,
+    qiskit_noisy,
+    observable=None,
+    simulator=qrack_simulate,
+    seed=0,
```
natestemen, line 165: I think this needs to be called `random_state` in order for it to have an impact.

natestemen (changes requested):
Great start Brian!

Biggest thing is a little more explanation is needed before the definition of the three executors as to what they're needed for. I would generally expect a reader to have a basic idea of the CDR workflow, but always good to quickly remind people.

.github/workflows/docs-build.yml
```diff
@@ -49,6 +49,9 @@ jobs:
 
       - name: Build and test Sphinx docs
         run: |
+          sudo add-apt-repository ppa:wrathfulspatula/vm6502q
+          sudo apt update
+          sudo apt install libqrack-dev
```
cosenal, line 54:
```suggestion
          sudo apt update
          sudo install ocl-icd-opencl-dev
```

Once you have `pyqrack` in the Python requirements, which you do, it's actually enough to install libOpenCL, you don't need libqrack from PPA. I figured this out by
1. running the workflow locally with the `act` tool (`act push --container-architecture linux/amd64 -j docs`)
2. entering the docker container spawned by 1. (`docker exec -it <container_id> bash`)
3. open a python console
4. run `from ctypes import *; CDLL("/usr/local/lib/python3.10/dist-packages/pyqrack/qrack_system/qrack_lib/libqrack_pinvoke.so")`, and observe the error
```
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/usr/lib/python3.10/ctypes/__init__.py", line 374, in __init__
    self._handle = _dlopen(self._name, mode)
OSError: libOpenCL.so.1: cannot open shared object file: No such file or directory
```
due to a missing libOpenCL library in the container. (cc @natestemen)

bdg221: Was able to use `pyqrack` inside the docs build by doing the following:

1. mitiq/requirements/requirements-dev.txt include pyqrack to the dependencies. (pyqrack==1.30.0)
2. mitiq/.github/workflows/docs-build.yml use apt to install opencl.
```
sudo apt update
sudo apt install ocl-icd-opencl-dev
```
3. mitiq/.readthedocs.yml add the apt_package to the build.
```
apt_packages:
  - ocl-icd-opencl-dev
  ```

Comment (cosenal) [reply .github/workflows/docs-build.yml]: @bdg221 Nice! 🎉 
Special shoutout to @natestemen for bringing the topic to the Mitiq coding session.

File: docs/source/examples/cdr_qrack.md [hunk 3]
```diff
@@ -0,0 +1,177 @@
...
+    jupytext_version: 1.16.3
+kernelspec:
+  display_name: Python 3
+  language: python
+  name: python3
+---
+
+# CDR with Qrack as Near-Clifford Simulator
+
+In this tutorial, Clifford Data Regression (CDR) is used with [Qrack](https://qrack.readthedocs.io/en/latest/) and a Qiskit fake backend. For more detailed information about CDR, check the [Users Guide](https://mitiq.readthedocs.io/en/stable/guide/cdr.html)
```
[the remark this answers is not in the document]
bdg221: Instead of your commit, I used a commit to change all possible links to be relative. Also, I removed the second sentence about the User Guide.

docs/source/examples/cdr_qrack.md
bdg221: @natestemen can you clarify what you mean.

natestemen: I don't think `execute_with_cdr` accepts a `seed` argument. According to the docstring it looks like it should be called `random_state`:
https://github.com/unitaryfund/mitiq/blob/357f2c0342ee68b551843f319281586422a5d1be/mitiq/cdr/cdr.py#L93

bdg221: Thanks @natestemen I've made a note that the User Guide needs to be updated with this as well.

docs/source/examples/cdr_qrack.md
```diff
@@ -0,0 +1,178 @@
...
+    cirq.rx(np.pi / 2).on(a),  # Clifford
+    cirq.rx(np.pi / 2).on(b),  # Clifford
+)
+
+# CDR works better if the circuit is not too short. So we increase its depth.
+circuit = 5 * circuit
+print(circuit)
+```
+
+## Qrack Near-Clifford Simulator
```
natestemen, line 73: Right before we jump into defining the three executors, can we, in a few sentences, let readers know why we need three different executors. Otherwise it may be confusing when after defining each executor, another is defined.

docs/source/examples/cdr_qrack.md
```diff
@@ -0,0 +1,178 @@
...
+```{code-cell}
+unmitigated_error = abs(unmitigated_expval - ideal_expval)
+mitigated_error = abs(mitigated_expval - ideal_expval)
+
+print(f"Unmitigated Error:           {unmitigated_error:.3f}")
+print(f"Mitigated Error:             {mitigated_error:.3f}")
+
+improvement_factor = unmitigated_error / mitigated_error
+print(f"Improvement factor with CDR: {improvement_factor:.2f}")
+```
```
natestemen, line 178: Let's just add one more thing to the end here. Something like "If you want to learn more about CDR check out the original reference and our user guide."

natestemen (commented):
Nice updates Brian! Just two smaller requests and then should be good to go.

WrathfulSpatula (approved):

docs/source/examples/cdr_qrack.md
bdg221: I added a header section for the executors along with an explanation. Let me know if it needs more.

docs/source/examples/cdr_qrack.md
bdg221: I added a section at the very end including links to the other CDR info and also how to contact us with questions.

natestemen (approved):
Great work @bdg221!!!

Issue #2439, opened 2024-07-04
Create two-stage API for ZNE


natestemen:
As part of our goals to make QEM techniques more modular and composable, we'd like to use the fact that most techniques can be seen as first generating circuits, and then second combining the results from said circuits.

A first step in this direction will be allowing users to more easily access the intermediary circuits for each technique. This issue will track the implementation of this idea for ZNE.

Previous discussion: https://github.com/unitaryfund/mitiq/discussions/2394

cosenal:
@natestemen Before jumping into a PR, please comment here with a small design of what it would like for ZNE specifically, so we can continue discussion #2394, but only in the scope of ZNE.

natestemen:
### Suggested design

Add two new functions (up to naming) in the ZNE module (at the top level; i.e. `zne.XYZ`):

1. `scaled_circuits`
2. `combine_and_extrapolate`

I propose they have the following signatures.

```py
def scaled_circuits(
    circuit: QPROGRAM,
    scale_factors: list[float],
    scale_method: Callable[[QPROGRAM, float], QPROGRAM],
) -> list[QPROGRAM]:
    ...

def combine_results(
    scale_factors: list[float],
    results: list[float],
    extrapolation_method: Callable[[list[float]], float],
) -> float:
    ...
```

### Imagined workflow

```py
# user generates the circuits and inspects them (can even change them if desired;
# although this would mess with the results (but maybe that's what they want!))
zne_scaled_circuits = scaled_circuits(...)

# user executes circuits as desired, storing the results in a list
results = []
for circuit in zne_scaled_circuits:
    results.append(circuit.execute(...))

# user combines results from runs using desired extrapolation
zne_value = combine_results(..., results, ...)
```

### Considerations

I wish we had a better way to store results together with the QEM configuration that was used with them. In "normal" mitiq workflows we have full control over the protocol and hence can keep everything nice and clean. This suggestion moves a lot more control to the user (which IMO is needed as this stage of NISQ), but means that we have to rely on users building potentially complicated objects. If they use an executor this problem can be mitigated since the executor stores results and the circuits that were run (but not QEM config as far as I'm aware).

cosenal:
Thanks for the suggested design overview. Some observations: 

1. How is this going to interplay with [Factories](https://mitiq.readthedocs.io/en/stable/guide/zne-3-options.html#extrapolation-methods-factory-objects)?
[...]

PR #2452, opened 2024-07-26
Two stage ZNE

FarLab:
This PR modularizes ZNE using the design described in https://github.com/unitaryfund/mitiq/issues/2439.

### DIFF
--- a/mitiq/zne/tests/test_zne.py
+++ b/mitiq/zne/tests/test_zne.py
@@ -609,5 +609,4 @@ def test_two_stage_zne(noise_scaling_method, extrapolation_factory):
         factory=extrapolation_factory(scale_factors),
         scale_noise=noise_scaling_method,
     )
-    # print(zne_res, two_stage_zne_res)
     assert np.isclose(zne_res, two_stage_zne_res)
--- a/mitiq/zne/tests/test_zne.py
+++ b/mitiq/zne/tests/test_zne.py
@@ -23,13 +23,18 @@ from qiskit_aer import AerSimulator
 from mitiq import QPROGRAM, SUPPORTED_PROGRAM_TYPES
 from mitiq.benchmarks.randomized_benchmarking import generate_rb_circuits
 from mitiq.interface import accept_any_qprogram_as_input, convert_from_mitiq
+from mitiq.interface.mitiq_braket import to_braket
 from mitiq.interface.mitiq_cirq import (
     compute_density_matrix,
     sample_bitstrings,
 )
+from mitiq.interface.mitiq_pennylane import to_pennylane
+from mitiq.interface.mitiq_pyquil import to_pyquil
+from mitiq.interface.mitiq_qibo import to_qibo
 from mitiq.interface.mitiq_qiskit import (
     execute_with_shots_and_noise,
     initialized_depolarizing_noise,
+    to_qiskit,
 )
 from mitiq.observable import Observable, PauliString
 from mitiq.zne import (
@@ -53,14 +58,6 @@ from mitiq.zne.scaling import (
     insert_id_layers,
 )
 from mitiq.zne.zne import combine_results, scaled_circuits
-
-from mitiq.interface.mitiq_qiskit import to_qiskit
-from mitiq.interface.mitiq_braket import to_braket
-from mitiq.interface.mitiq_pennylane import to_pennylane
-from mitiq.interface.mitiq_pyquil import to_pyquil
-from mitiq.interface.mitiq_qibo import to_qibo
-
-
 
 BASE_NOISE = 0.007
 TEST_DEPTH = 30
@@ -577,6 +574,7 @@ def test_execute_zne_on_qiskit_circuit_with_QFT():
     mitigated = execute_with_zne(circuit, qs_noisy_simulation)
     assert abs(mitigated) < 1000
 
+
 @pytest.mark.parametrize(
     "noise_scaling_method",
     [fold_gates_at_random, insert_id_layers, fold_global, fold_all],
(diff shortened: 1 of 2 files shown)


mitiq/zne/tests/test_zne.py
```diff
@@ -560,3 +563,51 @@ def qs_noisy_simulation(
 
     mitigated = execute_with_zne(circuit, qs_noisy_simulation)
     assert abs(mitigated) < 1000
+
+
+@pytest.mark.parametrize(
+    "noise_scaling_method",
+    [fold_gates_at_random, insert_id_layers, fold_global, fold_all],
```
natestemen, line 570: Great that it works with all of these!

natestemen (commented):
Great work Farrokh! Needs a little bit of flushing out, but also there's not a whole lot **to** flush out! Nice that it's so simple.

I think a few small updates should get this merged, and then we can keep iterating once it's merged.

mitiq/zne/tests/test_zne.py
```diff
@@ -560,3 +563,51 @@ def qs_noisy_simulation(
...
+)
+@pytest.mark.parametrize(
+    "extrapolation_factory", [RichardsonFactory, LinearFactory]
+)
+def test_two_stage_zne(noise_scaling_method, extrapolation_factory):
```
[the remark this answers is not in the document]
FarLab: I think for now the combined test is good enough. As we iterate in future PRs and the functions get more complicated, it might be better to separate them then.

mitiq/zne/tests/test_zne.py
```diff
@@ -560,3 +563,51 @@ def qs_noisy_simulation(
...
+    qreg = cirq.LineQubit.range(2)
+    cirq_circuit = cirq.Circuit(
+        cirq.H.on_each(qreg),
+        cirq.CNOT(*qreg),
+        cirq.CNOT(*qreg),
+        cirq.H.on_each(qreg),
+    )
```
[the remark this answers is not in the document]
FarLab: good point, I will add different frontends.

mitiq/zne/tests/test_zne.py
```diff
@@ -560,3 +573,74 @@ def qs_noisy_simulation(
...
+    elif to_frontend == to_pennylane:
+        assert all(
+            isinstance(circ, pennylane.tape.QuantumTape) for circ in circs
+        )
+    elif to_frontend == to_pyquil:
+        assert all(isinstance(circ, pyquil.quil.Program) for circ in circs)
+    elif to_frontend == to_qibo:
+        assert all(isinstance(circ, qibo.Circuit) for circ in circs)
+    else:
+        assert all(isinstance(circ, cirq.Circuit) for circ in circs)
```
natestemen, line 624: These tests are redundant since we already know our scaling method functions preserve input type. E.g. `fold_global` never converts a `qibo.Circuit` to anything but a `qibo.Circuit`. Since those [tests](https://github.com/unitaryfund/mitiq/blob/357f2c0342ee68b551843f319281586422a5d1be/mitiq/zne/scaling/tests/test_folding.py) cover this, let's leave it out from here for readability.

mitiq/zne/tests/test_zne.py
```diff
@@ -560,3 +573,74 @@ def qs_noisy_simulation(
...
+@pytest.mark.parametrize(
+    "noise_scaling_method",
+    [fold_gates_at_random, insert_id_layers, fold_global, fold_all],
+)
+@pytest.mark.parametrize(
+    "extrapolation_factory", [RichardsonFactory, LinearFactory]
+)
+@pytest.mark.parametrize(
+    "to_frontend",
+    [None, to_qiskit, to_braket, to_pennylane, to_pyquil, to_qibo],
```
natestemen, line 587:
```suggestion
    [convert_to_mitiq, to_qiskit, to_braket, to_pennylane, to_pyquil, to_qibo],
```

I think we can use this instead to leave the circuit as a `cirq.Circuit` object instead of doing the check below.

natestemen (approved):
LGTM! Just two minor asks.

mitiq/zne/zne.py
```diff
@@ -6,13 +6,62 @@
...
+    for scale_factor in scale_factors:
+        circuits.append(scale_method(circuit, scale_factor))
+
+    return circuits
+
+
+def combine_results(
+    scale_factors: Sequence[float],
+    results: list[float],
+    extrapolation_method: Callable[[list[float], list[float]], float],
```
natestemen, line 43: Might want to generalize these to `Sequence[float]` to make mypy happy.

mitiq/zne/tests/test_zne.py
FarLab: I see, that makes sense. I'll remove these tests!

mitiq/zne/tests/test_zne.py
FarLab: That makes it a bit cleaner indeed!

FarLab: Actually, convert_to_mitiq() returns a tuple (the circuit and a string). So you still need one check right (if to_frontend = convert_to_mitiq)?

nathanshammah:
Well done on having this in @FarLab. Later on it could be nice to have very short docs on the added functionality for modularization, maybe once we have more in other techniques, but we can also start with ZNE only.

mitiq/zne/zne.py
```diff
@@ -6,13 +6,62 @@
...
+            A list of built-in functions can be found in ``mitiq.zne.scaling``.
+
+    Returns:
+        The scaled circuits using the scale_method.
+    """
+    circuits = []
+    for scale_factor in scale_factors:
+        circuits.append(scale_method(circuit, scale_factor))
+
+    return circuits
```
cosenal, line 37: this should be a list comprehension fo sho
```
return [scale_method(circuit, s) for s in scale_factors]
```

mitiq/zne/zne.py
```diff
@@ -6,13 +6,62 @@
...
+        extrapolation_method: The function for scaling the noise of a
+            quantum circuit. A list of built-in functions can be found
+            in ``mitiq.zne.scaling``.
+
+    Returns:
+        The expectation value estimated with ZNE.
+    """
+    res = extrapolation_method(scale_factors, results)
+
+    return res
```
cosenal, line 62: unnecessary variable

```
return extrapolation_method(scale_factors, results)
```

mitiq/zne/tests/test_zne.py
```diff
@@ -560,3 +568,59 @@ def qs_noisy_simulation(
...
+)
+def test_two_stage_zne(
```
cosenal, line 584: There should be two separete unit tests for the two new functions you implemented. This is an integration test, and one bad symptom is that this test is violating the [AAA pattern](https://jamescooke.info/arrange-act-assert-pattern-for-python-developers.html).

cosenal (commented):
Sorry for the delayed (and at this point posthumous) review. I added some comments to improve on it.
It would have been also nice to ship an example (perahps in a tutorial) where we use the new workflow, along with this code changes.

PR #2464, opened 2024-08-09
Fix flaky REM test / refactor

natestemen:
## Description

fixes https://github.com/unitaryfund/mitiq/issues/2431 by setting a random seed to ensure deterministic test.

I also took the liberty to refactor the function at hand which looked to be more complicated than needed.

### DIFF
--- a/mitiq/rem/inverse_confusion_matrix.py
+++ b/mitiq/rem/inverse_confusion_matrix.py
@@ -14,37 +14,30 @@ from mitiq import Bitstring, MeasurementResult
 
 
 def sample_probability_vector(
-    probability_vector: npt.NDArray[np.float64], samples: int
-) -> List[Bitstring]:
+    probability_vector: Sequence[float], samples: int
+) -> list[str]:
     """Generate a number of samples from a probability distribution as
     bitstrings.
 
     Args:
         probability_vector: A probability vector.
+        samples: The number of samples to generate.
 
     Returns:
         A list of sampled bitstrings.
     """
-    # sample using the probability distribution given
     num_values = len(probability_vector)
-    choices = np.random.choice(num_values, size=samples, p=probability_vector)
-
-    # convert samples to binary strings
-    bit_width = int(np.log2(num_values))
-    binary_repr_vec = np.vectorize(np.binary_repr)
-    binary_strings = binary_repr_vec(choices, width=bit_width)
-
-    # split the binary strings into an array of ints
-    bitstrings = (
-        np.apply_along_axis(  # type: ignore
-            func1d=np.fromstring,  # type: ignore
-            axis=1,
-            arr=binary_strings[:, None],
-            dtype="U1",  # type: ignore
+    if not np.log2(num_values).is_integer():
+        raise ValueError(
+            "The length of the probability vector must be a power of 2."
         )
-        .astype(np.uint8)
-        .tolist()
+
+    sampled_indices = np.random.choice(
+        num_values, size=samples, p=probability_vector
     )
+
+    bit_width = int(np.log2(num_values))
+    bitstrings = [format(index, f"0{bit_width}b") for index in sampled_indices]
 
     return bitstrings
 
@@ -132,7 +125,7 @@ def generate_tensored_inverse_confusion_matrix(
 
 def closest_positive_distribution(
     quasi_probabilities: npt.NDArray[np.float64],
-) -> npt.NDArray[np.float64]:
+) -> List[float]:
     """Given the input quasi-probability distribution returns the closest
     positive probability distribution (with respect to the total variation
     distance).
@@ -163,7 +156,7 @@ def closest_positive_distribution(
         raise ValueError(
             "REM failed to determine the closest positive distribution."
         )
-    return result.x
+    return result.x.tolist()
 
 
 def mitigate_measurements(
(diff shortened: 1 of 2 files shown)


cosenal (approved):
Nice, first PR in a while on which we don't have to override build failures 😄

mitiq/rem/inverse_confusion_matrix.py
```diff
@@ -14,38 +14,31 @@
 
 
 def sample_probability_vector(
-    probability_vector: npt.NDArray[np.float64], samples: int
-) -> List[Bitstring]:
+    probability_vector: Sequence[float], samples: int
+) -> list[str]:
     """Generate a number of samples from a probability distribution as
     bitstrings.
```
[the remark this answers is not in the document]
natestemen: Considering this left us a bit confused when first encountering it, I think that's a great idea. Done in https://github.com/unitaryfund/mitiq/pull/2464/commits/f8725ebffde52a493b91876057d914f0773fb539.

mitiq/rem/inverse_confusion_matrix.py
```diff
@@ -14,38 +14,35 @@
 
 
 def sample_probability_vector(
-    probability_vector: npt.NDArray[np.float64], samples: int
-) -> List[Bitstring]:
+    probability_vector: Sequence[float], samples: int
+) -> list[str]:
```
purva-thakre, line 18: I am curious as to what's usually preferred when we specify the return type? 

I see you have used `list[str]` here but you could also have used `from typing import List` before specifying the return type as `List[str]`. 

Searching around for this shows that we [do not need to use the typing module](https://stackoverflow.com/a/62775680) for `>= Python 3.9`

cosenal: Didn't we have this exact same discussion with @vprusso in [another PR](https://github.com/unitaryfund/mitiq/pull/2347)? iirc there we learned that `list[str]` is the way to go.

mitiq/rem/tests/test_inverse_confusion_matrix.py
```diff
@@ -64,20 +70,20 @@ def test_bitstrings_to_probability_vector():
...
-                pv
-                - bitstrings_to_probability_vector(
-                    sample_probability_vector(pv, 1000)
-                )
-            ),
-            0,
-            abs_tol=0.1,
-        )
+@pytest.mark.parametrize("_", range(10))
+def test_probability_vector_roundtrip(_):
```
cosenal, line 74: micro-nit: what I meant in my previous comment is actually
 ```suggestion
@pytest.mark.parametrize("repeats", range(10))
def test_probability_vector_roundtrip(_):
```

File: mitiq/rem/inverse_confusion_matrix.py (hunk 2 above)
Comment (purva-thakre) [reply mitiq/rem/inverse_confusion_matrix.py:18]: Yes, we did. Which is why I was asking if we need to put this on our todos. 

We still use things like `from typing import List` elsewhere.

Comment (cosenal) [reply mitiq/rem/inverse_confusion_matrix.py:18]: We don't need to be so proactive. The aliases are deprecated, but they are unlikely to be removed in the foreseeable future.

File: mitiq/rem/tests/test_inverse_confusion_matrix.py (hunk 3 above)
Comment (natestemen) [reply mitiq/rem/tests/test_inverse_confusion_matrix.py:74]: Ahh yeah that would be a little more clear, I think. Unfortunately, that is not possible. The string argument to [`pytest.mark.parametrize`](https://docs.pytest.org/en/6.2.x/parametrize.html) has to match the argument names (up to some formatting), so having a completely different name like `repeats` and `_` do not work.
```
============================================================== short test summary info ==============================================================
ERROR mitiq/rem/tests/test_inverse_confusion_matrix.py - Failed: In test_probability_vector_roundtrip: function uses no argument 'repeats'
================================================================= 1 error in 3.29s ==================================================================
ERROR: found no collectors for /Users/nate/code/mitiq/mitiq/rem/tests/test_inverse_confusion_matrix.py::test_probability_vector_roundtrip
```

File: mitiq/rem/inverse_confusion_matrix.py (hunk 2 above)
Comment (natestemen) [reply mitiq/rem/inverse_confusion_matrix.py:18]: Agreed, we should not take action until we need to on this. In the meantime lets try to use the builtins instead of `typing.List`, but it's not a big concern since they both achieve the same goal.

If we do migrate eventually, we will have to import a large majority of the types from `collections.abc` since many of the types we rely on (e.g. [`typing.Sequence`](https://docs.python.org/3/library/typing.html#typing.Sequence)) are deprecated in favor of coming from `collections.abc`.

File: mitiq/rem/inverse_confusion_matrix.py [hunk 4]
```diff
@@ -132,7 +129,7 @@ def generate_tensored_inverse_confusion_matrix(
 
 def closest_positive_distribution(
     quasi_probabilities: npt.NDArray[np.float64],
-) -> npt.NDArray[np.float64]:
+) -> List[float]:
```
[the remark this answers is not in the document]
natestemen: Agree on the consistency part when modifying files, so have changed the other instances of `List` in this file to `list` in https://github.com/unitaryfund/mitiq/pull/2464/commits/65bef0a63df97a2b1d6b6247df766c5a67a43e97.

mitiq/rem/tests/test_inverse_confusion_matrix.py
cosenal: oh, well, thanks for trying

PR #2480, opened 2024-08-23
execute_with_cdr does not use seed but random_state

bdg221:
## Description

In the CDR section of the User Guide, the `execute_with_cdr` example includes the parameter `seed` which should be `random_state` [per the cdr.py file](https://github.com/unitaryfund/mitiq/blob/af7b992b3be15d9560a8adda698e729e7e17c933/mitiq/cdr/cdr.py#L93)

### License


cosenal (approved):
⭐

PR #2493, opened 2024-09-04
simplify batch creation in executor

natestemen:
## Description

Was just digging around in this code and found a place to simplify some code using [`itertools.batched`](https://docs.python.org/3/library/itertools.html#itertools.batched).

### DIFF
--- a/mitiq/executor/executor.py
+++ b/mitiq/executor/executor.py
@@ -7,6 +7,7 @@
 by error mitigation techniques to compute expectation values."""
 
 import inspect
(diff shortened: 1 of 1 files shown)


cosenal:
> Added in version 3.12.
😬

natestemen:
Ahhhh yeah, I saw that which is why I didn't immediately assign a reviewer. I was hoping that `itertools==3.12` could be used on earlier versions of python, but I guess that's not how it works. Closing.

PR #2499, opened 2024-09-11
LRE Executors

### LINKED ISSUE #2370: Executors for LRE
Similar to the other existing error mitigation methods, 

1. `execute_with_lre` compatible with all interfaces
2. `lre_decorator`
3. `mitigate_lre_executor`

purva-thakre:
## Description
Fixes #2370 

List of todos:

- [x] Docstrings
- [x] Additional unit tests

---

### License

### DIFF
--- a/mitiq/lre/lre.py
+++ b/mitiq/lre/lre.py
@@ -11,7 +11,7 @@ from typing import Any, Callable, Optional, Union
 import numpy as np
 from cirq import Circuit
 
-from mitiq import QuantumResult
+from mitiq import QPROGRAM, QuantumResult
 from mitiq.lre import (
     multivariate_layer_scaling,
     multivariate_richardson_coefficients,
@@ -25,8 +25,8 @@ def execute_with_lre(
     degree: int,
     fold_multiplier: int,
     folding_method: Callable[
-        [Union[Any], float], Union[Any]
-    ] = fold_gates_at_random,
+        [QPROGRAM, float], QPROGRAM
+    ] = fold_gates_at_random,  # type: ignore [has-type]
     num_chunks: Optional[int] = None,
 ) -> float:
     r"""
@@ -47,7 +47,8 @@ def execute_with_lre(
         input_circuit: Circuit to be scaled.
         executor: Executes a circuit and returns a `QuantumResult`
         degree: Degree of the multivariate polynomial.
-        fold_multiplier: Scaling gap required by unitary folding.
+        fold_multiplier: Scaling gap value required for unitary folding which
+            is used to generate the scale factor vectors.
         folding_method: Unitary folding method. Default is
             :func:`fold_gates_at_random`.
         num_chunks: Number of desired approximately equal chunks. When the
@@ -73,7 +74,14 @@ def execute_with_lre(
 
     # verify the linear combination coefficients and the calculated expectation
     # values have the same length
-    assert len(lre_exp_values) == len(linear_combination_coeffs)
+    if not len(lre_exp_values) == len(  # pragma: no cover
+        linear_combination_coeffs
+    ):
+        raise AssertionError(
+            "The number of expectation values are not equal "
+            + "to the number of coefficients required for "
+            + "multivariate extrapolation."
+        )
 
     return np.dot(lre_exp_values, linear_combination_coeffs)
 
--- a/mitiq/lre/lre.py
+++ b/mitiq/lre/lre.py
@@ -91,15 +91,15 @@ def mitigate_executor(
     error-mitigated with layerwise richardson extrapolation (LRE).
 
     Args:
-    input_circuit: Circuit to be scaled.
-    executor: Executes a circuit and returns a `QuantumResult`
-    degree: Degree of the multivariate polynomial.
-    fold_multiplier: Scaling gap required by unitary folding.
-    folding_method: Unitary folding method. Default is
-        :func:`fold_gates_at_random`.
-    num_chunks: Number of desired approximately equal chunks. When the
-        number of chunks is the same as the layers in the input circuit,
-        the input circuit is unchanged.
+        input_circuit: Circuit to be scaled.
+        executor: Executes a circuit and returns a `QuantumResult`
+        degree: Degree of the multivariate polynomial.
+        fold_multiplier: Scaling gap required by unitary folding.
+        folding_method: Unitary folding method. Default is
+            :func:`fold_gates_at_random`.
+        num_chunks: Number of desired approximately equal chunks. When the
+            number of chunks is the same as the layers in the input circuit,
+            the input circuit is unchanged.
 
 
     Returns:
@@ -132,15 +132,15 @@ def lre_decorator(
     layerwise richardson extrapolation (LRE).
 
     Args:
-    input_circuit: Circuit to be scaled.
-    executor: Executes a circuit and returns a `QuantumResult`
-    degree: Degree of the multivariate polynomial.
-    fold_multiplier: Scaling gap required by unitary folding.
-    folding_method: Unitary folding method. Default is
-        :func:`fold_gates_at_random`.
-    num_chunks: Number of desired approximately equal chunks. When the
-        number of chunks is the same as the layers in the input circuit,
-        the input circuit is unchanged.
+        input_circuit: Circuit to be scaled.
+        executor: Executes a circuit and returns a `QuantumResult`
+        degree: Degree of the multivariate polynomial.
+        fold_multiplier: Scaling gap required by unitary folding.
+        folding_method: Unitary folding method. Default is
+            :func:`fold_gates_at_random`.
+        num_chunks: Number of desired approximately equal chunks. When the
+            number of chunks is the same as the layers in the input circuit,
+            the input circuit is unchanged.
 
 
     Returns:
--- a/mitiq/lre/lre.py
+++ b/mitiq/lre/lre.py
@@ -11,7 +11,7 @@ from typing import Any, Callable, Optional, Union
 import numpy as np
 from cirq import Circuit
 
-from mitiq import QPROGRAM, QuantumResult
+from mitiq import QPROGRAM
 from mitiq.lre import (
     multivariate_layer_scaling,
     multivariate_richardson_coefficients,
@@ -21,7 +21,7 @@ from mitiq.zne.scaling import fold_gates_at_random
 
 def execute_with_lre(
     input_circuit: Circuit,
-    executor: Callable[[Circuit], QuantumResult],
+    executor: Callable[[Circuit], float],
     degree: int,
     fold_multiplier: int,
     folding_method: Callable[
@@ -45,7 +45,7 @@ def execute_with_lre(
 
     Args:
         input_circuit: Circuit to be scaled.
-        executor: Executes a circuit and returns a `QuantumResult`
+        executor: Executes a circuit and returns a `float`
         degree: Degree of the multivariate polynomial.
         fold_multiplier: Scaling gap value required for unitary folding which
             is used to generate the scale factor vectors.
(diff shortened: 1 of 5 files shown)


mitiq/lre/tests/test_lre.py
```diff
@@ -0,0 +1,107 @@
...
+
+noisy_val = execute(test_cirq)
+
+
+def test_lre_exp_value():
+    """Verify LRE executors work as expected."""
+    ideal_val = execute(test_cirq, noise_level=0)
+    assert abs(ideal_val - noisy_val) > 0
+    lre_exp_val = execute_with_lre(
+        test_cirq, execute, degree=2, fold_multiplier=2
```
purva-thakre, line 35: To do: add a parameterized test where other values for `degree` and `fold_mmultiplier` are used.

purva-thakre:
This test does not fail locally: https://github.com/unitaryfund/mitiq/actions/runs/10847148259/job/30101612145?pr=2499#step:6:4556

mitiq/lre/lre.py
```diff
@@ -0,0 +1,161 @@
...
+from mitiq.lre import (
+    multivariate_layer_scaling,
+    multivariate_richardson_coefficients,
+)
+from mitiq.zne.scaling import fold_gates_at_random
+
+
+def execute_with_lre(
+    input_circuit: Circuit,
+    executor: Callable[[Circuit], QuantumResult],
```
natestemen, line 24:
```suggestion
    executor: Callable[[Circuit], float],
```
I think this has to be a float, right? And right now, this just works with `cirq.Circuit` objects, right?

mitiq/lre/lre.py
```diff
@@ -0,0 +1,161 @@
...
+    degree: int,
+    fold_multiplier: int,
+    folding_method: Callable[
+        [Union[Any], float], Union[Any]
+    ] = fold_gates_at_random,
```
natestemen, line 29: Let's follow what happens in ZNE here for consistency.

https://github.com/unitaryfund/mitiq/blob/6f912f63f087ac1213290819d85cf16695da5b34/mitiq/zne/zne.py#L70

mitiq/lre/lre.py
```diff
@@ -0,0 +1,161 @@
...
+
+        If you would prefer to choose the layers for unitary
+        folding, use :func:`mitiq.zne.scaling.layer_scaling.get_layer_folding`
+        instead.
+
+    Args:
+        input_circuit: Circuit to be scaled.
+        executor: Executes a circuit and returns a `QuantumResult`
+        degree: Degree of the multivariate polynomial.
+        fold_multiplier: Scaling gap required by unitary folding.
```
natestemen, line 50: Some more explanation for the `fold_multiplier` param would be helpful.

mitiq/lre/tests/test_lre.py
```diff
@@ -0,0 +1,112 @@
...
+    """Verify LRE executors work as expected."""
+    ideal_val = execute(test_cirq, noise_level=0)
+    assert abs(ideal_val - noisy_val) > 0
+    lre_exp_val = execute_with_lre(
+        test_cirq,
+        execute,
+        degree=input_degree,
+        fold_multiplier=input_fold_multiplier,
+    )
+    assert lre_exp_val > noisy_val
```
natestemen, line 42: Why does the LRE value always have to be larger than the noisy value?

mitiq/lre/tests/test_lre.py
```diff
@@ -0,0 +1,112 @@
...
+            .simulate(noisy_circuit)
+            .final_density_matrix
+        )
+        return rho[0, 0].real
+
+    assert noisy_val < execute(test_cirq)
+
+
+def test_lre_decorator_raised_error():
+    """Verify error is raised when the defualt parameters for the decorator are
```
natestemen, line 70:
```suggestion
    """Verify an error is raised when the required parameters for the decorator are
```

natestemen (commented):
Nice work Purva! All your hard work makes this a nice and simple PR :)

Maybe we can do some debugging as to the `num_chunks` issue in tomorrows coding session.

mitiq/lre/lre.py
```diff
@@ -0,0 +1,161 @@
...
+    )
+
+    lre_exp_values = []
+    for scaled_circuit in noise_scaled_circuits:
+        circ_exp_val = executor(scaled_circuit)
+        lre_exp_values.append(circ_exp_val)
+
+    # verify the linear combination coefficients and the calculated expectation
+    # values have the same length
+    assert len(lre_exp_values) == len(linear_combination_coeffs)
```
cosenal, line 76: nit: I am not a fan of assert in production code (as opposed to in tests), because they are removed if one were to compile Mitiq or run it in an optimized mode. See https://docs.python.org/3/reference/simple_stmts.html#the-assert-statement

mitiq/lre/tests/test_lre.py
```diff
@@ -0,0 +1,112 @@
...
+    )
+    exp_val_from_mitigate_executor = mitigated_executor(test_cirq)
+    assert exp_val_from_mitigate_executor > noisy_val
+
+
+def test_lre_decorator():
+    """Verify LRE decorators work as expected."""
+
+    @lre_decorator(degree=2, fold_multiplier=2)
+    def execute(circuit, noise_level=0.025):
```
cosenal, line 56: Something more in the spirit of unit tests (and more future-proof) would be a test that only checks whether the correct functions are called, and doesn't actually run the circuit through the simulator. Here you are testing lots of things at the same time (including a Google's simulator), all of which could go wrong without giving much insights on the _unit_ you are testing here, which is the behaviour of the new decorator.

cosenal (approved):
I left two comments, not blockers.

mitiq/lre/lre.py
purva-thakre: Yes, this only works for cirq circuits right now. So, your suggestion does make sense. I decided to make it `QuantumResult` as `make check-types` failed with the following:

```
Argument 2 to "execute_with_lre" has incompatible type "Callable[[Circuit], Union[float, MeasurementResult, ndarray[Any, Any]]]"; expected "Callable[[Circuit], float]"  [arg-type]
```

mitiq/lre/lre.py
natestemen: Sorry, I'm missing how `num_to_average` is related here. I'm just suggesting we typehint the folding method differently.

mitiq/lre/tests/test_lre.py
```diff
@@ -0,0 +1,112 @@
...
+        assert noisy_val < execute(test_cirq)
+
+
+def test_lre_executor_with_chunking():
+    """Verify the executor works as expected for chunking a large circuit into
+    a smaller circuit."""
+    ideal_val = execute(test_cirq * 200, noise_level=0)
+    assert abs(ideal_val - noisy_val) > 0
+    lre_exp_val = execute_with_lre(
+        test_cirq * 200, execute, degree=2, fold_multiplier=2, num_chunks=5
```
cosenal, line 94: I misled you by not creating a variable for `test_cirq * 200` in y'day live coding*, but we do want one in the pull request.

*As @natestemen was bugging me about while I tried to hack things around as quick as possible 😅

mitiq/lre/lre.py
purva-thakre: That's weird. I was viewing this suggestion locally in VS code. I interpreted 'follow what happens in ZNE for consistency' as use the typehint as well as the additional param. 

image

mitiq/lre/tests/test_lre.py
purva-thakre: I am going to use the mock testing module for this. As Nate noticed yesterday, `test_cirq * 200` will increase the time taken by unit tests.

mitiq/lre/lre.py
purva-thakre: I hadn't modified anything in ZNE which is why it's weird that when you linked L70 in `zne.py`, L71 also showed up. 

The location of the suggestion was not changed.

mitiq/lre/tests/test_lre.py
purva-thakre: Quick path to run the tests.

mitiq/lre/tests/test_lre.py
```diff
@@ -0,0 +1,139 @@
...
+    ]
+    ideal_val = execute(test_cirq, noise_level=0)
+    assert abs(ideal_val - noisy_val) > 0
+    lre_exp_val = execute_with_lre(
+        test_cirq, execute, degree=2, fold_multiplier=2, num_chunks=10
+    )
+    assert abs(lre_exp_val - ideal_val) <= abs(noisy_val - ideal_val)
+
+
+@pytest.mark.parametrize(
```
purva-thakre, line 107: @natestemen Continuing our conversation about the parallels between LRE chunking/no_chunking and local/global folding, would it be considered good practice to include several tests that are expected to fail? A quick google search did not give me a definitive answer on this.

natestemen: If the failure modes are different, then yes.

mitiq/lre/tests/test_lre.py
purva-thakre: @cosenal Are you available for a quick call on Friday? I have been unsuccessful in trying to write a unit test that utilizes mock objects. 

It's unclear to me what needs to be a mock object and what must have a pre-defined value.

mitiq/lre/tests/test_lre.py
purva-thakre: Discussed with Nate and Alessandro to skip this unit for when we get to #2512 in the future. 

The code block that misbehaves when we have really large circuits is in one of the inference functions. 

https://github.com/unitaryfund/mitiq/blob/6f912f63f087ac1213290819d85cf16695da5b34/mitiq/lre/inference/multivariate_richardson.py#L165-L182

purva-thakre:
I don't know why the tests are failing right now. Will rerun them in a couple of hours.

mitiq/lre/tests/test_lre.py
```diff
@@ -0,0 +1,136 @@
...
+        test_cirq, execute, degree=2, fold_multiplier=2, num_chunks=14
+    )
+    assert abs(lre_exp_val - ideal_val) <= abs(noisy_val - ideal_val)
+
+
+@pytest.mark.parametrize(
+    "test_input", [(1), (2), (3), (4), (5), (6), (7), (8), (9)]
+)
+@pytest.mark.xfail
+def test_lre_executor_with_chunking_failures(test_input):
```
natestemen, line 112: 1. Better variable naming for `test_input` please! Maybe `num_chunks`?
2. The word "failure" is overloaded here. It's not that the code is failing, but that the technique does not perform well in this parameter-space. Maybe we can rename the test something like `test_large_circuit_with_small_chunks_poor_performance`, and then instead of `xfail`, `assert` the inverse of what we do now?
3. Why do we need to test so many values here?

mitiq/lre/tests/test_lre.py
```diff
@@ -0,0 +1,136 @@
...
+    rho = DensityMatrixSimulator().simulate(noisy_circuit).final_density_matrix
+    return rho[0, 0].real
+
+
+noisy_val = execute(test_cirq)
+ideal_val = execute(test_cirq, noise_level=0)
+
+
+@pytest.mark.parametrize(
+    "input_degree, input_fold_multiplier", [(2, 2), (2, 3), (3, 4)]
```
natestemen, line 32:
```suggestion
    "degree, fold_multiplier", [(2, 2), (2, 3), (3, 4)]
```

mitiq/lre/lre.py
natestemen: It should work if you change all the instances of `QuantumResult` in this file to `float`.

natestemen (approved):
A few minor changes requested, but looking good once those are addressed.

How long does it take to run the tests in `test_lre.py`?

purva-thakre:
>How long does it take to run the tests in test_lre.py?

I used `pytest --durations=0 test_lre.py` in `mitiq/mitiq/lre/tests` to get the following:

image

Chunking test where it is not expected to fail takes the longest. 

https://stackoverflow.com/a/55095253

mitiq/lre/tests/test_lre.py
purva-thakre: > Why do we need to test so many values here?

At the time, I was testing what values for chunking work in our favor. Which is why I felt it would be better to keep these in.

PR #2514, opened 2024-09-25
Address executor and observable incompatibility

### LINKED ISSUE #2449: REM functionality dependent on typehinting
The following snippet only works when the `execute` function has a typehint for the functions return type.

```py
import networkx as nx

from mitiq import MeasurementResult, benchmarks, rem
from mitiq.observable.observable import Observable
from mitiq.observable.pauli import PauliString

depth, num_qubits = 10, 4
circuit, bs = benchmarks.generate_mirror_circuit(
    nlayers=depth,
    two_qubit_gate_prob=1.0,
    connectivity_graph=nx.complete_graph(num_qubits),
    return_type="qiskit",
)


def execute(circuit, shots=100) -> MeasurementResult:
    fake_bitstrings = [[0, 0, 1, 1] for _ in range(shots)]

    return MeasurementResult(fake_bitstrings)


p_flip = 0.05
icm = rem.generate_inverse_confusion_matrix(len(bs), p_flip, p_flip)
obs = Observable(PauliString("ZZZZ"))

rem_value = rem.execute_with_rem(
    circuit,
    execute,
    obs,
    inverse_confusion_matrix=icm,
)

print(f"Expectation Value After REM: {rem_value}")
```

The error that is raised when removing the typehint is
```
ValueError: Executor and observable are not compatible. Executors
                returning expectation values as float must be used with
                observable=None
```
which comes from the `Executor` class. In particular a check to ensure the observable and return type of the executor are compatible.
https://github.com/unitaryfund/mitiq/blob/231319df9051dfbc61db79d02bee201979528a36/mitiq/executor/executor.py#L163-L172

---

Needless to say, this code should work regardless of the typehint. If investigation shows that the typehint truly is needed, then this requirement needs to be documented.

bdg221:
## Description
Fixes #2449 

This PR address code changes in the `Executor` and `Pauli` classes to account for when an `Observable` is passed to an executor along with a circuit that already has a measurement on a qubit effected by the `Observable`.

Also, I removed a check in the `Executor` class, with the assumption the user is passing a valid `executor` if an `observable` is used. 

Furthermore, the docs for the `Executor` and `Observable` have been updated with the annotation/typehinting requirement.

Finally, additional unit testing has been added to ensure that incompatible `Executor`s and `Observable`s are caught and an appropriate message is given to the user.

---

### License


bdg221:
I believe there are additional unit tests that need to be created to ensure that we cover as many scenarios as possible. This is especially true when we assume the user has defined a compatible executor when passing an observable. 

I plan to discuss this in more detail with @natestemen to ensure appropriate test coverage.

bdg221:
@natestemen @cosenal I ran into a bit of an issue and I was curious how you think I should proceed. I added logic into the executor that will check the scenario where the executor does not have a return type specified and an observable is not passed. In this scenario we assume that the return type is float, therefore I am checking the circuits to make sure that there is a measurement included, if not I check if the return type was passed in (error messaging telling the user to include measurements in their circuit since they meant for it to return a float) or not (error message saying to use type hinting and if meant to return a float then add measurements.) 

I believe this is what we want, however I found that we have over 200 tests that fail when the code is implemented. The executors in the tests that fail do not use type hinting and also do not include measurement gates (since typically a simple float is return like the length of the circuit or a random number.)

I can certainly go through and clean those up, however I am wondering if I should break that out to a new issue/PR because it is a bit involved.

mitiq/executor/executor.py
```diff
@@ -149,6 +150,59 @@ def evaluate(
...
+                                convert_to_mitiq(circuit)[
+                                    0
+                                ].findall_operations_with_gate_type(
+                                    MeasurementGate
+                                )
+                            )
+                        )
+                        == 0
+                    ):
+                        if self._executor_return_type is None:
```
bdg221, line 192: It is line 192 to 204 that is causing the tests to fail.

bdg221:
@natestemen @cosenal With this latest update, I am only checking now for obvious incompatibility. This means that if a user does use type hinting and does/does not include an observable it will throw errors in the following scenarios:

`FloatLike` should NOT have an observable
`DensityMatrixLike` MUST have an observable
`MeasurementResultLike` MUST have an observable

When the return type is `MeasurementResultLike`, we use the observable to add measurement gates to the circuit before call `self.run()`. Previously, if the return type was not specified this would result in an error from the backend because the circuit would not have measurements. However, now I added a test if `self.run()` fails to see if there is an observable and if so to add the measurement(s) to the circuit(s) and try `self.run()` again. 

After calling `self.run()` successfully, we normally parse the results. `FloatLike` grabs `np.real_if_close()`, `DensityMatrixLike` uses the `observable._expectation_from_density_matrix()`, and `MeasurementResultLike` uses the `observable._expectation_from_measurements()`. While previously the code was looking `self._executor_return_type`, I am also now using the return type of the first object in the sequence returned by `self.run()`. 

I added tests that check the obviously bad scenarios when type hinting is used. Also, I compare the results of typed versus non-typed to make that if we assume the user does things correctly without type hinting that they get the correct answer.

mitiq/observable/pauli.py
```diff
@@ -282,8 +282,27 @@ def _measure_in(
...
+            ]
+
         for pauli in paulis.elements:
             basis_rotations.update(pauli._basis_rotations())
+            for qubit in pauli._qubits_to_measure():
+                if qubit in existing_measurements:
+                    raise ValueError(
+                        f"""More than one measaurement found for qubit: """
+                        f"""{qubit}. Only a single measurement is allowed """
+                        f"""per qubit."""
```
natestemen, line 304:
```suggestion
                        f"More than one measaurement found for qubit: "
                        "{qubit}. Only a single measurement is allowed "
                        "per qubit."
```

AFAIK, the f-string only need to be stated once and then it will apply to the whole string. (suggestion is modulo formatting)

mitiq/observable/pauli.py
```diff
@@ -282,8 +282,27 @@ def _measure_in(
...
+        # Find any existing measurement gates in the circuit
+        existing_measurements = []
+        measurement_tuples = list(
+            circuit.findall_operations_with_gate_type(cirq.MeasurementGate)
+        )
+        if measurement_tuples:
+            existing_measurements = [
+                measurement_tuple[1].qubits[0]
+                for measurement_tuple in measurement_tuples
+            ]
```
natestemen, line 295: Two suggestions:

1. I'm not sure this code will work if the circuit has a "2d" measurement[^1]. We can instead get all the qubits used in measurement if we have a set that keeps track of the qubits themselves
    ```py
    qubits_with_measurements = set()
    for _, op, _ in c.findall_operations_with_gate_type(cirq.MeasurementGate):
        qubits_with_measurements.update(op.qubits)
    ```

2. Once we have `qubits_with_measurements`, we can make the check to raise the error a little simpler:
    ```py
    if support & qubits_with_measurements:
        raise ValueError(...)
    ```


[^1]: E.g. if the circuit has `cirq.measure(LineQubit(1), LineQubit(2))`.

mitiq/executor/tests/test_executor.py
```diff
@@ -199,53 +208,27 @@ def test_run_executor_preserves_order(s, b):
...
+    circuits = [
+        cirq.Circuit(cirq.X(q), cirq.M(q)),
+        cirq.Circuit(cirq.H(q), cirq.Z(q), cirq.M(q)),
+    ]
 
     executor = Executor(execute)
 
     results = executor.evaluate(circuits)
-    assert np.allclose(results, [1, 2])
+    assert np.allclose(results, [2, 3])
```
natestemen, line 219: Do we need to have this line, and the line below checking `quantum_results`?

mitiq/executor/tests/test_executor.py
```diff
@@ -199,53 +208,27 @@ def test_run_executor_preserves_order(s, b):
...
-    ],
-)
-def test_executor_observable_compatibility_check(execute, obs):
-    q = cirq.LineQubit(0)
-    circuits = [cirq.Circuit(cirq.X(q)), cirq.Circuit(cirq.H(q), cirq.Z(q))]
-
-    executor = Executor(execute)
-
-    with pytest.raises(ValueError, match="are not compatible"):
-        executor.evaluate(circuits, obs)
```
natestemen, line 244: Does this test need to be removed? I think I'm missing something because I don't quite see why these are not compatible.

mitiq/executor/executor.py
```diff
@@ -149,6 +151,29 @@ def evaluate(
...
+                raise ValueError(
+                    "When using a density matrix like result, an observable "
+                    "is required."
+                )
+            # Type hinted as MeasurementResulteLike but no observable is set
+            elif self._executor_return_type in MeasurementResultLike:
+                raise ValueError(
+                    "When using a measurement, or bitstring, like result, an "
+                    "observable is required."
+                )
```
natestemen, line 175: These checks make it impossible to use an `Executor` to run and return density matrices and counts. I don't think we want to limit users in that way, do we?

mitiq/executor/executor.py
```diff
@@ -160,38 +185,55 @@ def evaluate(
...
-            )
         else:
             all_circuits = circuits
             result_step = 1
 
         # Run all required circuits.
-        all_results = self.run(all_circuits, force_run_all, **kwargs)
+        try:
+            all_results = self.run(all_circuits, force_run_all, **kwargs)
+        except Exception:
```
natestemen, line 195: Catch specific exceptions when possible. What fails in the above?

mitiq/executor/executor.py
```diff
@@ -160,38 +185,55 @@ def evaluate(
...
+                all_circuits = [
+                    circuit_with_measurements
+                    for circuit in circuits
+                    for circuit_with_measurements in observable.measure_in(
+                        circuit
+                    )
+                ]
+                all_results = self.run(all_circuits, force_run_all, **kwargs)
+            else:
+                raise
```
natestemen, line 206: What happens here?

mitiq/executor/executor.py
```diff
@@ -160,38 +185,55 @@ def evaluate(
...
+
+        # check returned type
+        manual_return_type = None
+        if len(all_results) > 0:
+            manual_return_type = type(all_results[0])
```
natestemen, line 211: Should we even care about the `_executor_return_type` at this point?

natestemen (changes requested):
Thanks for getting started with this Brian!

Apologies on the long delay for the review. In general since the `Executor` is so core to a lot of things in Mitiq, I'm hesitant to make changes when we don't understand the side effects thoroughly. Hopefully we can mitigate some of those worries through good tests and some back and forth :)

mitiq/executor/executor.py
```diff
@@ -40,6 +40,8 @@
 FloatLike = [
     None,  # Untyped executors are assumed to return floats.
     float,
+    np.float32,
+    np.float64,
```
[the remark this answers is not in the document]
bdg221: I found the need to add these when running `make test`. Here is the result without those in FloatLike:
image
And with those 2 lines, I don't see the errors:
image

@natestemen do you see this behavior too?

mitiq/executor/executor.py
bdg221: It does make it impossible to use an `Executor` to run and return density matrices and counts if an observable is not provided. This limitation actually comes from parsing the results where `observable` is used by density matrices `_expectation_from_density_matrix()` and counts `_expectation_from_measurements()`.

mitiq/executor/executor.py
bdg221: This is checking if the observable is not none and trying to see if the result type was meant to be MeasurementResultLike. If the return type was specified then the measurements would have been added to the circuit(s).

mitiq/executor/tests/test_executor.py
bdg221: No. Deleting this line.

bdg221: Resolved with #1333ebf

mitiq/observable/pauli.py
bdg221: Turns out Ruff gets mad with just the single f-string stated since the first line does not have any placeholders.

mitiq/executor/tests/test_executor.py
bdg221: All of those executors return floats or list[floats]. We do not allow you to have an observables with an executor that is returning FloatLike. However, this was an old test for years ago. This is now covered in `def test_executor_float_with_observable_typed()`.

mitiq/executor/executor.py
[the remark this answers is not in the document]
bdg221: I remembered why these two are in `FloatLike`. Since we are trying to assume things if a return type is not specified, we check the type of what is returned from `self.run()`. If we do not include them, then later checks (especially if we no longer look at `self._executor_return_type` as mentioned https://github.com/unitaryfund/mitiq/pull/2514#discussion_r1819858219

mitiq/executor/executor.py
bdg221: If an executor returns something different than what is specified there will be other errors thrown. You are right, at this point we can stick with the `manual_return_type`.

bdg221: Addressed in https://github.com/unitaryfund/mitiq/pull/2514/commits/3c7b02737fad652628d740b23633fac6fcb78607

bdg221: Should I add more unit testing since I just introduced this `if` branching?

mitiq/observable/pauli.py
bdg221: As a note for historical purposes, https://github.com/unitaryfund/mitiq/pull/2514/commits/a0e4cda6a2d76e7dbff43e111cfa0371df123fbe changed the test to include a "2d" measurement to ensure we cover that use case.

mitiq/executor/executor.py
```diff
@@ -160,38 +185,51 @@ def evaluate(
...
+                all_circuits = [
+                    circuit_with_measurements
+                    for circuit in circuits
+                    for circuit_with_measurements in observable.measure_in(
+                        circuit
+                    )
+                ]
+                all_results = self.run(all_circuits, force_run_all, **kwargs)
+            else:
+                raise
```
FarLab, line 206: I like the idea of just executing one of the circuits and checking the return type. Does this mean we can move all the compatibility checks in lines 156-175 after this manual checking of return type? This would remove the need of _executor_return_type variable and no typehinting is needed anymore. Maybe I am missing something.

bdg221: A couple of things.
1. This is NOT checking the return type of a single circuit, but instead this is running all circuits with the `executor`. 

2. The checks in lines 156-175 are specifically checking the following cases based off limitations already in the `execute` function.

- FloatLike - expected the circuits to already include measurements and there should NOT be an observable. The `executor` does not handle observables for FloatLike results.
- DensityMatrixLike and MeasurementLike - MeasurementResultLike handles observables in lines 182-187 and both respectively use OBSERVABLE.`_expectation_from_density_matrix()` and OBSERVABLE.`_expectation_from_measurements()`

Since these are requirements, I would think it would be better to check and break out in bad scenarios, instead of trying to run the circuits and then error out.

FarLab: Ah yes, you are right that it is running all circuits! I guess I was looking at lines 209-216 where you use the manual return type (of the first circuit?) to parse results. I was thinking to just run the first circuit to get the manual return type and do all the checks on observable compatibility afterwards. This is a completely different approach than using _executor_return_type, so understandable if we don't want to that.

bdg221: The main use of `_executor_return_type` (outside of what I have done in this PR) is to check if it is `MeasurementResultLike` and there is an observable so that `observable.measure_in()` can be called to add measurements into the circuit. Without doing this, the backends complain. I can't think of a way to do this without the `_executor_return_type` since calling `run()` to try to get back a manual return type throws errors.

FarLab (approved):
It looks good to me!

mitiq/executor/executor.py
natestemen: We definitely should not be re-running on behalf of the user (imagine they're paying for hardware time). If something fails we need to let the user know, and let them fix the issue before continuing.

mitiq/executor/executor.py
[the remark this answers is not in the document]
natestemen: I'm a little worried about this being needed now, despite me inspiring the change that required this to be added. I didn't foresee the consequences of moving from `_executor_return_type` to `manual_return_type`.

Imagine a situation where a user writes an executor that uses some machine learning stuff in `torch` and returns a [`torch.Tensor.float`](https://pytorch.org/docs/stable/generated/torch.Tensor.float.html). Since this type is not in `FloatLike` we will raise a `ValueError`! As far as I understand, this would be a workflow that would be broken by this PR.

As far as addressing it, I think we should rely on `_executor_return_type` for parsing the results. WDYT?

natestemen (changes requested):
Thanks for the continued effort on this Brian! I know we all want to merge this, but as it stands there are a few blockers (like running more circuits and a user might have anticipated).

Independent of the specifics of this PR, your work has been incredibly valuable in helping me understand the difficulties surrounding the `Executor`.

mitiq/executor/executor.py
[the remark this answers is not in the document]
bdg221: I agree and think we are arriving back at executors requiring typehinting and that we cannot assume things are correct (though untyped executors will be treated as FloatLike.)

mitiq/executor/executor.py
bdg221: There does not seem to be a way to assume the user has the executor set correctly without running the executor. As this could result in wasted runs and money, it seems like typehinting is necessary.

natestemen: As a general theme for the PR I think we should take the typehint as a source of truth if it is available, and fall back on whatever the use has supplied via function arguments otherwise and assume compatibility.

bdg221: I believe then that the main update for this PR will be explicitly checking incompatible cases only (float with observable, density matrix or measurement without observable.) At least for the density matrix and measurement this will throw an error BEFORE running the user defined executor and then trying to parse the results (when those would have failed.)

natestemen: IMO that's still a nice improvement, and can potentially save users some money/time.

bdg221: Agreed. I corrected this in https://github.com/unitaryfund/mitiq/pull/2514/commits/e3d8e2a7c3a9a8ab4516194e8429d54071656738

mitiq/executor/executor.py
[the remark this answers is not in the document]
bdg221: Removed the user of `manual_return_type` and returned to `_executor_return_type` in https://github.com/unitaryfund/mitiq/pull/2514/commits/e3d8e2a7c3a9a8ab4516194e8429d54071656738

bdg221:
At this point, this PR is checking for explicit incompatibility between the executors and observables. It provides clear error messages and includes additional unit tests.

mitiq/executor/executor.py
```diff
@@ -147,6 +147,36 @@ def evaluate(
...
 
+        # Check executor and observable compatability with type hinting
+        # If FloatLike is specified as a return and observable is used
+        if self._executor_return_type in FloatLike and observable is not None:
+            # Type hinted as FloatLike and observable passed
+            if self._executor_return_type is not None:
+                raise ValueError(
+                    "When using a float like result, measurements should be "
+                    "included manually and an observable should not be "
+                    "used."
```
natestemen, line 158:
```suggestion
                    "When using an executor which returns a float-like result, measurements should be added before the circuit is executed instead of with an observable."
```

WDYT? Just want to provide as much info as possible to ensure a user can fix the issue.

mitiq/executor/executor.py
```diff
@@ -147,6 +147,36 @@ def evaluate(
...
+                )
+            else:
+                # Using an observable but no type hinting
+                raise ValueError(
+                    "When using an observable, the return type must be "
+                    "specified on the user defined executor."
```
natestemen, line 164:
```suggestion
                    "When using an observable, the return type of the executor must be "
                    "specified using typehinting."
```

mitiq/executor/executor.py
```diff
@@ -147,6 +147,36 @@ def evaluate(
...
+                )
+        elif observable is None:
+            # Type hinted as DensityMatrixLike but no observable is set
+            if self._executor_return_type in DensityMatrixLike:
+                raise ValueError(
+                    "When using a density matrix like result, an observable "
+                    "is required."
```
natestemen, line 171:
```suggestion
                    "When using a density matrix result, an observable "
                    "is required."
```

natestemen (approved):
Woohoo! LGTM!

I just made a few suggestions for the warnings that are raised (modulo formatting). Feel free to iterate on my suggestions if they need more tweaking, and then we can merge!

mitiq/executor/executor.py
bdg221: I used your wording, just corrected the formatting in https://github.com/unitaryfund/mitiq/pull/2514/commits/19934a9d22d4ea381f6b7de7b113f5052b0a6ef8

mitiq/executor/executor.py
bdg221: I used your wording, just corrected the formatting in https://github.com/unitaryfund/mitiq/pull/2514/commits/19934a9d22d4ea381f6b7de7b113f5052b0a6ef8

mitiq/executor/executor.py
bdg221: I used your wording, just corrected the formatting in https://github.com/unitaryfund/mitiq/pull/2514/commits/19934a9d22d4ea381f6b7de7b113f5052b0a6ef8

bdg221:
@natestemen I made the changes (with some formatting adjustments.) Go ahead and give it one more look and then merge.

natestemen (approved):
:shipit:

natestemen:
Thank you for all your hard work on this Brian! 🙏🏻

PR #2523, opened 2024-10-02
Raise TypeError when a dictionary is passed to MeasurementResult constructor

natestemen:
**Motivation**: Multiple students at the UMass Amherst summer school attempted to pass a dictionary to instantiate a `MeasurementResult` object while working through the [challenges](https://github.com/unitaryfund/qnumerics/blob/main/challenge.ipynb). This silently accepts, but provides erroneous results. E.g.:
```py
>>> from mitiq import MeasurementResult
>>> MeasurementResult({'001': 23, '000': 43})
MeasurementResult: {'nqubits': 3, 'qubit_indices': (0, 1, 2), 'shots': 2, 'counts': {'001': 1, '000': 1}}
```

**Considerations**: I could see the following two potential solutions to help with this issue.

1. Add a typecheck at runtime and raise a [`TypeError`](https://docs.python.org/3/library/exceptions.html#TypeError), letting the user know we have a method `MeasurementResult.from_counts` for this specific pattern.
2. Call `MeasurementResult.from_counts` in the `__post_init__` method to **re-instantiate** `self`.

I chose to raise solution 1 as it seems to be a simple approach to address the problem. Admittedly, I am less familiar with the technical details of re-instantiating an object within the `__post_init__` function, but happy to test that out if reviewers think it is more appropriate.

### DIFF
--- a/mitiq/typing.py
+++ b/mitiq/typing.py
@@ -112,6 +112,11 @@ class MeasurementResult:
             ``tuple(range(self.nqubits))``, where ``self.nqubits``
             is the bitstring length deduced from ``result``.
 
+    Example:
+        >>> mr = MeasurementResult(["001", "010", "001"])
+        >>> mr.get_counts()
+        {'001': 2, '010': 1}
+
     Note:
         Use caution when selecting the default option for ``qubit_indices``,
         especially when estimating an :class:`.Observable`
--- a/mitiq/typing.py
+++ b/mitiq/typing.py
@@ -125,6 +125,11 @@ class MeasurementResult:
 
     def __post_init__(self) -> None:
         # Validate arguments
+        if isinstance(self.result, dict):
+            raise TypeError(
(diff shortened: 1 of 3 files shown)


cosenal (approved):
Solution 1 sounds reasonable to me, and a strict improvement on the current behavior.

mitiq/typing.py
```diff
@@ -125,6 +125,11 @@ class MeasurementResult:
 
     def __post_init__(self) -> None:
         # Validate arguments
+        if isinstance(self.result, dict):
```
cosenal, line 128: We could have something stricter that checks for results not being an instance of Sequence[Bitstring], which is what we expect, but then both the check and the error message become more complicated, so I would leave it as is.

natestemen:
Thanks for the suggestion of adding an example in the docstring. I like that. Done in https://github.com/unitaryfund/mitiq/pull/2523/commits/71a0a63e11f67343bb66c743798acaa883709a66.

PR #2545, opened 2024-10-23
remove failing test; simplify layerwise ZNE tests

natestemen:
## Description

1. Remove a test that has been randomly failing on CI (https://github.com/unitaryfund/mitiq/commit/e6210e6add29c5aab8b2a8ac54438e93bc14f85c)
2. As part of a Mitiq coding call we found an instance of using an integration test instead of small unit tests. We moved the testing of layerwise folding with all `QPROGRAM` types into a smaller unit test so we do not need to E2E test the entire workflow with every `QPROGRAM` instance. (https://github.com/unitaryfund/mitiq/commit/60528db816f61a05ed7bf4bddd66987ebfc4bcff)

#### Reviewer 👀 

I kept a modified version of `test_layerwise_folding_with_zne`, but struggled with knowing what to `assert` at the end since we'd like to move away from making assertions about the accuracy of techniques unless there are proofs to back the claims. As far as I understand, the assertions in `test_layerwise_folding_with_zne` are redundant to the the unit tests in `test_layer_scaling.py`, but am not 100% sure. WDYT? Should we keep this test? Should it be altered?

### DIFF
--- a/mitiq/lre/tests/test_lre.py
+++ b/mitiq/lre/tests/test_lre.py
@@ -101,20 +101,6 @@ def test_lre_executor_with_chunking():
     assert abs(lre_exp_val - ideal_val) <= abs(noisy_val - ideal_val)
 
 
-@pytest.mark.parametrize("num_chunks", [(1), (2), (3), (4), (5), (6), (7)])
-def test_large_circuit_with_small_chunks_poor_performance(num_chunks):
-    """Verify chunking performs poorly when a large number of layers are
-    chunked into a smaller number of circuit chunks."""
-    # define a larger circuit
-    test_cirq = benchmarks.generate_rb_circuits(n_qubits=1, num_cliffords=15)[
-        0
-    ]
-    lre_exp_val = execute_with_lre(
-        test_cirq, execute, degree=2, fold_multiplier=2, num_chunks=num_chunks
-    )
-    assert abs(lre_exp_val - ideal_val) >= abs(noisy_val - ideal_val)
-
-
 @pytest.mark.parametrize("input_method", [(fold_global), (fold_all)])
 def test_lre_executor_with_different_folding_methods(input_method):
     """Verify the executor works as expected for using non-default unitary
--- a/mitiq/zne/scaling/tests/test_layer_scaling.py
+++ b/mitiq/zne/scaling/tests/test_layer_scaling.py
@@ -5,9 +5,16 @@
 
 """Unit tests for scaling by layer."""
 
+import random
+from itertools import product
+from unittest.mock import patch
+
+import pytest
 from cirq import Circuit, LineQubit, ops
 
-from mitiq.zne.scaling import layer_folding
+from mitiq import SUPPORTED_PROGRAM_TYPES
+from mitiq.interface import convert_from_mitiq, convert_to_mitiq
+from mitiq.zne.scaling import get_layer_folding, layer_folding
 
 
 def test_layer_folding_with_measurements():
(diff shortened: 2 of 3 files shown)


mitiq/lre/tests/test_lre.py
```diff
@@ -101,20 +101,6 @@ def test_lre_executor_with_chunking():
     assert abs(lre_exp_val - ideal_val) <= abs(noisy_val - ideal_val)
 
 
-@pytest.mark.parametrize("num_chunks", [(1), (2), (3), (4), (5), (6), (7)])
```
purva-thakre, line 104: the test that keeps failing is when `num_chunks=7`. 

I think it might be better to remove that particular one instead of the entire test.

mitiq/zne/scaling/tests/test_layer_scaling.py
```diff
@@ -53,24 +60,59 @@ def test_layer_folding():
...
+def test_get_layer_folding(mock_layer_folding):
+    a, b = LineQubit.range(2)
+    circuit = Circuit(ops.X(a), ops.CNOT(a, b), ops.Y(b))
+    layer_index = 1
+    scale_factor = 3
+
+    folding_func = get_layer_folding(layer_index)
+    folding_func(circuit, scale_factor)
+
+    mock_layer_folding.assert_called_once_with(
```
cosenal, line 116: 🎉  Mitiq Coding for the win!

mitiq/zne/scaling/tests/test_layer_scaling.py
```diff
@@ -53,24 +60,59 @@ def test_layer_folding():
...
-                    [ops.H.on_each(*qreg)] * (2 * (layers_to_fold[0]) + 1),
-                    # Layer-2
-                    [ops.CNOT.on(qreg[0], qreg[1])]
-                    * (2 * (layers_to_fold[1]) + 1),
-                    [ops.X.on(qreg[2])] * (2 * (layers_to_fold[1]) + 1),
-                    # Layer-3
-                    [ops.TOFFOLI.on(*qreg)] * (2 * (layers_to_fold[2]) + 1),
-                )
-                assert folded_circuit == correct
+    for i, j, k in product(range(total_folds), repeat=3):
```
cosenal, line 63: nit (and very much about personal taste): I like the use of product, but it's arguable that i,j,k are better indices than numbered ones, given we are talking about an ordered number of layers here

mitiq/zne/tests/test_zne.py
```diff
@@ -493,32 +495,34 @@ def test_execute_with_zne_with_supported_circuits(circuit_type):
...
 
-    zne_value = execute_with_zne(
-        circuit, generic_executor, factory=fac, scale_noise=fold_layer_func
+    execute_with_zne(
+        circuit, mock_executor, factory=factory, scale_noise=fold_layer_func
     )
-
-    # Test zero noise limit is better than unmitigated expectation value
-    assert abs(unmitigated - expected) > abs(zne_value - expected)
+    assert mock_executor.call_count == len(scale_factors)
```
cosenal, line 517: To answer your question from the description, I like what you are asserting on in here, and I agree with removing asserting on actual values. This will still test an e2e behavior, and it will not be flaky.

cosenal (approved):
nice! :shipit:

mitiq/lre/tests/test_lre.py
natestemen: If we want to keep the test, I think we should re-work the test so that it is not testing performance, but correctness. What would you suggest changing the assertion to?

mitiq/zne/scaling/tests/test_layer_scaling.py
natestemen: Fair point. Changed back in https://github.com/unitaryfund/mitiq/pull/2545/commits/c9fb3bcc936807e111af54e7cac801a10532720e.

mitiq/lre/tests/test_lre.py
natestemen: Purva and I spoke about this live and decided that it was okay to remove. Purva was partially using the unit test to track which values were causing issue, but this is now at least partially recorded here.

Going ahead with removal.

Issue #2416, opened 2024-06-21
Make LRE compatible with non-cirq circuits


purva-thakre:
As discussed in another PR: https://github.com/unitaryfund/mitiq/pull/2347#issuecomment-2182074934

Two different ways to implement the optional args led to errors. Either the `one_to_many` argument was ignored (first code block) or `_pop_measurement` was used on a Qiskit circuit instead of converting the circuit to a cirq circuit first (second code block). 

```python
@accept_qprogram_and_validate
def multivariate_layer_scaling( input_circuit, degree, ............, one_to_many = True)
```

Similar to how the optional arg is used in [PEC](https://github.com/unitaryfund/mitiq/blob/bf82e84410e1805437e15f1d26b0c5be8e8dcb8a/mitiq/pec/sampling.py#L199), I also tried changing `multivariate_layer_scaling` to `_cirq_multivariate_layer_scaling` such that the new `multivariate_layer_scaling` only accepted a QPROGRAM as input. 
```python
 noise_scaling_func = accept_qprogram_and_validate(
      _cirq_multivariate_layer_scaling,
      one_to_many=True,
  )
  scaled_circuits = noise_scaling_func(
      input_circuit, degree, fold_multiplier, num_chunks, folding_method
  )
 return scaled_circuits
```

PR #2547, opened 2024-10-24
Ensure LRE compatibility with all supported frontends

natestemen:
### Description

This PR ensures that the core functionality of LRE is compatible with all supported frontends.

fixes https://github.com/unitaryfund/mitiq/issues/2416

### DIFF
--- a/mitiq/interface/conversions.py
+++ b/mitiq/interface/conversions.py
@@ -161,6 +161,7 @@ def convert_from_mitiq(
         circuit: Mitiq circuit to convert.
         conversion_type: String specifier for the converted circuit type.
     """
+    conversion_type = conversion_type.lower()
     conversion_function: Callable[[cirq.Circuit], QPROGRAM]
     if conversion_type == "qiskit":
         from mitiq.interface.mitiq_qiskit.conversions import to_qiskit
--- a/mitiq/interface/conversions.py
+++ b/mitiq/interface/conversions.py
@@ -6,7 +6,18 @@
 """Functions for converting to/from Mitiq's internal circuit representation."""
 
 from functools import wraps
-from typing import Any, Callable, Dict, Iterable, Optional, Tuple, cast
+from typing import (
+    Any,
+    Callable,
+    Collection,
+    Concatenate,
+    Dict,
+    Optional,
+    ParamSpec,
+    Tuple,
+    TypeVar,
+    cast,
+)
 
 import cirq
 
@@ -199,13 +210,21 @@ def convert_from_mitiq(
     return converted_circuit
 
 
+P = ParamSpec("P")
+R = TypeVar("R")
+
+
 def accept_any_qprogram_as_input(
-    accept_cirq_circuit_function: Callable[[cirq.Circuit], Any],
-) -> Callable[[QPROGRAM], Any]:
+    accept_cirq_circuit_function: Callable[Concatenate[cirq.Circuit, P], R],
+) -> Callable[Concatenate[QPROGRAM, P], R]:
+    """Converts functions which take as input cirq.Circuit object (and return
+    anything), to function which can accept any QPROGRAM.
+    """
+
     @wraps(accept_cirq_circuit_function)
     def accept_any_qprogram_function(
-        circuit: QPROGRAM, *args: Any, **kwargs: Any
-    ) -> Any:
+        circuit: QPROGRAM, *args: P.args, **kwargs: P.kwargs
+    ) -> R:
         cirq_circuit, _ = convert_to_mitiq(circuit)
         return accept_cirq_circuit_function(cirq_circuit, *args, **kwargs)
 
@@ -245,15 +264,19 @@ def atomic_converter(
 
 
 def atomic_one_to_many_converter(
-    cirq_circuit_modifier: Callable[..., Iterable[cirq.Circuit]],
-) -> Callable[..., Iterable[QPROGRAM]]:
+    cirq_circuit_modifier: Callable[..., Collection[cirq.Circuit]],
+) -> Callable[..., Collection[QPROGRAM]]:
+    """Convert function which returns multiple cirq.Circuits into a function
+    which returns multiple QPROGRAM instances.
+    """
+
     @wraps(cirq_circuit_modifier)
     def qprogram_modifier(
         circuit: QPROGRAM, *args: Any, **kwargs: Any
-    ) -> Iterable[QPROGRAM]:
+    ) -> Collection[QPROGRAM]:
         mitiq_circuit, input_circuit_type = convert_to_mitiq(circuit)
 
-        modified_circuits: Iterable[cirq.Circuit] = cirq_circuit_modifier(
+        modified_circuits = cirq_circuit_modifier(
             mitiq_circuit, *args, **kwargs
         )
 
--- a/mitiq/lre/multivariate_scaling/layerwise_folding.py
+++ b/mitiq/lre/multivariate_scaling/layerwise_folding.py
@@ -15,6 +15,7 @@ import numpy as np
 from cirq import Circuit
 
 from mitiq import QPROGRAM
+from mitiq.interface import accept_qprogram_and_validate
 from mitiq.utils import _append_measurements, _pop_measurements
 from mitiq.zne.scaling import fold_gates_at_random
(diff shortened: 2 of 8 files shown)


mitiq/interface/conversions.py
```diff
@@ -199,13 +210,21 @@ def conversion_function(circ: cirq.Circuit) -> cirq.Circuit:
...
 
 
+P = ParamSpec("P")
+R = TypeVar("R")
+
+
 def accept_any_qprogram_as_input(
-    accept_cirq_circuit_function: Callable[[cirq.Circuit], Any],
-) -> Callable[[QPROGRAM], Any]:
+    accept_cirq_circuit_function: Callable[Concatenate[cirq.Circuit, P], R],
```
cosenal, line 218: This looks like the key change. Can you explain (ideally in the PR description) why we need this, and how LRE differs from other technique on this?

mitiq/typing.py
```diff
@@ -90,6 +80,33 @@ class SUPPORTED_PROGRAM_TYPES(EnhancedEnum):
...
+    BRAKET = "braket"
+    CIRQ = "cirq"
+    PENNYLANE = "pennylane"
+    PYQUIL = "pyquil"
+    QIBO = "qibo"
+    QISKIT = "qiskit"
+
+    @classmethod
+    def _python_type(cls, identifier: str) -> QPROGRAM:
+        if identifier == "braket":
```
cosenal, line 94: Better to use the enum names here, then you will get static checks and things like IDE autocomplete (e.g. if you go and change this to, say, "bracket", it will fail at static checks (or early compilation) instead of runtime.

mitiq/lre/tests/test_layerwise_folding.py
```diff
@@ -30,36 +32,41 @@
 test_circuit1_with_measurements.append(ops.measure_each(*qreg1))
 
 
-def test_multivariate_layerwise_scaling():
+@pytest.mark.parametrize("circuit_type", SUPPORTED_PROGRAM_TYPES.keys())
```
cosenal, line 35: Related to a comment I left below: better to use `SUPPORTED_PROGRAM_TYPES` as parameter.

mitiq/lre/multivariate_scaling/layerwise_folding.py
```diff
@@ -208,3 +209,8 @@ def multivariate_layer_scaling(
         multiple_folded_circuits.append(folded_circuit)
 
     return multiple_folded_circuits
+
+
+multivariate_layer_scaling = accept_qprogram_and_validate(
+    _multivariate_layer_scaling, one_to_many=True
+)
```
purva-thakre, line 216: @natestemen Can we spend next week's coding session discussing these circuit converter decorators? When I was struggling with this, I knew I had to use `one_to_many=True`. But I struggled to use it with ` accept_qprogram_and_validate`.

mitiq/lre/tests/test_layerwise_folding.py
```diff
@@ -30,36 +32,41 @@
...
-    for i, scale_factor_vector in enumerate(folding_pattern):
-        scale_layer1, scale_layer2, scale_layer3 = scale_factor_vector
-        expected_circuit = Circuit(
-            [ops.H.on_each(*qreg1)] * scale_layer1,
-            [ops.CNOT.on(qreg1[0], qreg1[1]), ops.X.on(qreg1[2])]
-            * scale_layer2,
-            [ops.TOFFOLI.on(*qreg1)] * scale_layer3,
-        )
-        assert expected_circuit == multiple_scaled_circuits[i]
+    if circuit_type == "cirq":
```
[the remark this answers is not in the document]
natestemen: > Say it fails, we will have to check which of the two concerns it is failing at.

Well, pytest does that for us! E.g. if we have `test_foo.py`
```py
def test_foo():
    assert 1 == 2
    assert "hi" in {"one", "two"}
```
and run `python -m pytest test_foo.py` you'll see

```
    def test_foo():
        assert 1 == 1
>       assert "hi" in {"one", "two"}
E       AssertionError: assert 'hi' in {'one', 'two'}

tt.py:3: AssertionError
================================================= short test summary info ==================================================
FAILED tt.py::test_foo - AssertionError: assert 'hi' in {'one', 'two'}
```
If both assertions are made to be false, it will alert you to the first one only.

---

As far as checking return type, I agree it's not ideal, but I'm not sure what else to do that's not **a lot** more complicated. We could make sure the function correctly folds each type of circuit, but that means defining the correct circuit in every `QPROGRAM` which is tedious.

natestemen:
I think I've resolved all the comments in c2eb37cb78f84c716cf26a895e35028b6e514800. Apologies for not splitting it up into different commits, but everything got intertwined as I started working on it.

mitiq/interface/conversions.py
natestemen: I can add it to the PR description, but it's actually not a "key change" for this PR.


There are two problems with the typehint prior to this PR

1. It's too strict. A function `(cirq.Circuit, int) -> anything` does not pass the type checking as the type hint only specifies a single, `cirq.Circuit` argument.
2. Type-wise, `accept_any_qprogram_as_input` was allowed to change the return type of the function it was applied to. This is not intended.

To fix point 1, the [`ParamSpec`](https://docs.python.org/3/library/typing.html#typing.ParamSpec) was added to allow for any args/kwargs,[^1] and to resolve point 2, I added a [`TypeVar`](https://docs.python.org/3/library/typing.html#typing.TypeVar) to keep the return value consistent.

[^1]: This change necessitated the addition of [`Concatenate`](https://docs.python.org/3/library/typing.html#typing.Concatenate) to make sure a circuit object is the first argument.

Here's a short description of what each thing does.
```
Callable[
    Concatenate[      # Allows us to prepend fixed arguments (like a circuit) to other arguments captured by P.
        cirq.Circuit, # The first argument to the callable, must be a cirq.Circuit instance.
        P             # ParamSpec representing the rest of the args/kwargs.
    ],
    R                 # TypeVar for the return type, ensuring the return type is preserved across transformations or decorations.
]
```

mitiq/typing.py
```diff
@@ -90,6 +81,16 @@ class SUPPORTED_PROGRAM_TYPES(EnhancedEnum):
 ]
 
 
+# Supported quantum programs.
+class SUPPORTED_PROGRAM_TYPES(EnhancedEnum):
+    BRAKET = _BKCircuit
```
cosenal, line 86: nice, I like this evolution! 🙌

mitiq/lre/tests/test_lre.py
```diff
@@ -40,8 +42,27 @@ def test_lre_exp_value(degree, fold_multiplier):
...
+
+    lre_exp_val = execute_with_lre(
+        circuit,
+        mock_executor,
+        degree=degree,
+        fold_multiplier=fold_multiplier,
+    )
+
+    assert isinstance(lre_exp_val, float)
+    assert mock_executor.call_count == 10
```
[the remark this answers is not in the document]
cosenal: This still pending? It's not a big deal, but it makes me feel that the test expectation was written based on what result we empirically got by running the code, instead of the other way around.

mitiq/interface/conversions.py
cosenal: Thanks for the explanation. At the last Mitiq Coding session I understood this wasn't the main point of the PR.

cosenal (approved):
LGTM. Left a reminder about a nit from the first review.

mitiq/lre/tests/test_lre.py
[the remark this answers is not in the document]
natestemen: Ah sorry, I changed this [here](https://github.com/unitaryfund/mitiq/pull/2547/files#diff-86f176b9c86340449dc49cbe0689ec2b47646820d5634eb524e3e38fc117ae31R46-R47), but not here! I'll update.

mitiq/lre/tests/test_layerwise_folding.py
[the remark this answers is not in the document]
natestemen: You're right! I've split them up in https://github.com/unitaryfund/mitiq/pull/2547/commits/97b02dc36c092523fa4bf4c1200c3a451f973544.

Issue #2110, opened 2023-12-06
Classical Shadows Documentation Cleanup


purva-thakre:
This issue is to keep track of all the things that need to be fixed, added etc. in the `Classical Shadows` section of the documentation. 

- [ ] Missing formula (in `CS` theory)
![image](https://github.com/unitaryfund/mitiq/assets/66048318/9e2635b7-616f-45d1-b527-c604282605a1)

- [ ] The theory portion of `Classical Shadows` only has a discussion about how the expectation value is calculated via this technique. There is nothing about state reconstruction. 
- [ ] Equations in user guide has hard-typed equation numbers. Some of the equation numbers are incorrect. Switch to equation refs. https://stackoverflow.com/questions/67856170/how-to-refer-to-an-equation-in-sphinx
- [ ] Users guide landing page cleanup
- [ ] Users guide does not follow the standard format. Missing a section for `when should I use this technique`
- [ ] Users guide does not follow the standard format. Missing a section for `what options are available for this technique`: [`Calibration Stage`](https://mitiq.readthedocs.io/en/stable/guide/shadows-1-intro.html#calibration-stage) in docs should be moved here. 
- [ ] Users guide does not follow the standard format. Missing a section for `what happens when I use this technique`
- [ ] Cleanup the existing portions of the user guide. Some of the info could be moved to the sections listed above. 
- [x]  Last line of first paragraph of the example is supposed to link to the robust shadows tutorial. ![image](https://github.com/unitaryfund/mitiq/assets/66048318/e5181dd0-f991-4da9-b4e1-fd49b126e2ac)
- [x] Typo at L76 of `How do I use CS` 
![image](https://github.com/unitaryfund/mitiq/assets/66048318/b0c09348-1f15-40f5-965d-7f07397355dd)
- [x] Typo at L83 of `Classical Shadows Example`
- [ ] Reference in docstring of `mitiq.shadows.shadows.pauli_twirling_calibrate` is not a link. 
- [x] ![image](https://github.com/unitaryfund/mitiq/assets/66048318/3a7036ed-d1a2-4090-b765-dab1b200e84f)

More items to be added later.

matti-sarjala:
I checked the current Classical Shadows guide against this cleanup issue and opened two intentionally narrow PRs:

- #3010 replaces hard-coded equation number references in `docs/source/guide/shadows-5-theory.md` with named MyST/Sphinx equation references.
[...]

PR #2574, opened 2024-11-14
Quick fix: Fix typo and headings in classical shadows tutorial

JMuff22:
## Description

I noticed some typos and an unrendered heading in the classical shadows documentation. These are fixed here. 

Additionally, if you (the maintainers) think it useful I think it would be good to highlight Remark 1 from the paper (https://arxiv.org/abs/2002.08953) 

> Remark 1 (Constants in Theorem 1). The numerical constants featuring in N and K result from a conservative
(worst case) argument that is designed to be simple, not tight. We expect that the actual constants are much
smaller in practice.

In the shadows_tutorial for equation (10). I think the sudden appearance of `34` is confusing in the guide. 

image

---

### License

### DIFF
--- a/docs/source/examples/shadows_tutorial.md
+++ b/docs/source/examples/shadows_tutorial.md
@@ -426,7 +426,7 @@ When we realize this code, it's important to consider that we record the equival
 
 Consequently, computing the mean estimator involves counting the number of exact matches between the observable and the classical shadow, and then multiplying the result by the appropriate sign. In the following, we present the function `expectation_estimation_shadow`, which allows for estimating any observable based on a classical shadow. This is realised by the main function `execute_with_shadows` when *state_reconstruction =* **False**.
 
-###4.3 Shadow Estimation Bound on Estimation of Expectation Values of Observables
+### 4.3 Shadow Estimation Bound on Estimation of Expectation Values of Observables
 
 The shadow estimation bound of operator expectation values is given by the following theorem:
 _________________________________________________________________________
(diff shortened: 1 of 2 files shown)


purva-thakre:
Hi @JMuff22, thank you for your correction! Cleaning up this tutorial is on our list of todos #2110 

> Additionally, if you (the maintainers) think it useful I think it would be good to highlight Remark 1 from the paper

I have a very high-level understanding of how this technique works. Is there a particular spot where this remark might be useful?

> In the shadows_tutorial for equation (10). I think the sudden appearance of 34 is confusing in the guide.

I think this equation is directly taken from the paper. See eq S13 in the supplementary info.

JMuff22:
> I think this equation is directly taken from the paper. See eq S13 in the supplementary info.
Yes S13 without remark 1 is confusing in my opinion.

natestemen (approved):
Good catch! Thanks for the patch.

I agree the remark would be helpful here. Is that something you'd like to add to this PR?

JMuff22:
By the way, is there a quicker way of building the documentation for quick fixes? I noticed that `make docs` builds and executes all the jupyter notebooks which is not ideal for testing how a small change like this would look when rendered.

JMuff22:
> Good catch! Thanks for the patch.
> 
> I agree the remark would be helpful here. Is that something you'd like to add to this PR?

Added my suggested fix.

natestemen (approved):
Thanks Jake! BTW are you using the `mitiq.shadows` module in practice? Would love to get your thoughts/opinions on how it's going if so.

natestemen:
> By the way, is there a quicker way of building the documentation for quick fixes? I noticed that make docs builds and executes all the jupyter notebooks which is not ideal for testing how a small change like this would look when rendered.

`make docs` will cache the results from notebooks, so if it doesn't change it won't rerun. There's no caching on CI, but locally it should speed up after the first build. I think @cosenal had a command to run a single file, but I cannot find it now.

JMuff22:
> Thanks Jake! BTW are you using the `mitiq.shadows` module in practice? Would love to get your thoughts/opinions on how it's going if so.

I am not really using it in practice. Maybe in the future. I was more using it to learn the protocol and apply it myself.

Issue #2578, opened 2024-11-16
# Decomposition problem with Qiskit circuits


andreamari:
If an error is obtained during the qiskit -> qasm -> cirq conversion, Mitiq tries to decompose the Qiskit circuit into
more elementary gates. 

This trick works in some simple cases but not in many other similar cases. It looks like the Qiskit circuit sometimes is not fully decomposed.

The consequences are quite relevant since one gets errors for e.g. the very common gate 'rzz' . 

**Code to reproduce the issue:**

```python
from mitiq.interface import convert_to_mitiq

test_qc = QuantumCircuit(2)
test_qc.rx(0.1, 0)  # If you remove this line it works!
test_qc.rzz(0.1, 0, 1)
print(test_qc)
convert_to_mitiq(test_qc)
```

gluonhiggs:
I have created a PR to resolve this issue [here](https://github.com/unitaryfund/mitiq/pull/2579).
P/s:
Initially I thought we could decompose the circuit before using `convert_to_mitiq`, however that's just a work around. Mitiq should automatically do that for us. Hence, I tried to resolve it.

cosenal:
Fixed by #2579

Issue #1073, opened 2022-01-06
Some Qiskit gates are not supported (during the internal conversion to Cirq)


andreamari:
Some Qiskit gates are not supported (during the internal conversion to Cirq).
It would be good to support more gates. 
Since we are relying on Cirq conversion functions, a possible way to do achieve can be via up-streaming these requests to Cirq.

Gates that are currently not supported (please add more if you know more):

- `sx` = \sqrt{X} : up-to-a-phase equivalent to RX(pi/2)
- `cu1`  (see #558)
-  `PhaseGate(theta)` = up-to-a-phase equivalent to RZ(theta)
- `u(theta)`
- `ECRGate`:  echoed cross-resonance (ECR) gate.

rmlarose:
Phase gate (https://discord.com/channels/764231928676089909/773957956659052605/897658569510502441). 

Logistically, I suggest closing #558 and using this issue to track to avoid duplicates.

natestemen:
Scope of this ticket is twofold:

1. Do some investigation to figure out if the gates mentioned above are still failing (and documenting the results)
2. Adding support for failing gates

MaldoAlberto:
I can work on this issue

MaldoAlberto:
Hi @andreamari  ! I'm working on this issue and I find Qiskit works now with OpenQASM3 and cirq with OpenQASM2, and using less gates than qiskit, so my question is convenient to do a function to converts OpenQASM3 to OpenQASM2 and include the equivalents gates, for example 

cu(pi/2, pi/2, pi/2, 0) q[1], q[0];

 will be 

p(0) q[1];
p((pi/2 + pi/2) / 2) q[1];
p((pi/2 - pi/2) / 2) q[0];
cx q[1], q[0];
u(-pi/2 / 2, 0, -(pi/2 + pi/2) / 2) q[0];
cx q[1], q[0];
u(pi/2 / 2, pi/2, 0) q[0];

andreamari:
Hi @MaldoAlberto,

Thanks for isolating the root of problem so well! 
I am not sure about writing from scratch a QASM3<->QASM2 within Mitiq. It would be a complex piece of code to maintain which could solve this bug, but could create 100 new bugs :-) 

Two less dangerous solutions are perhaps the following:

- Open some issues on Cirq or Qiskit about conversion problems. Having good Qiskit<-->Cirq conversions is in their interest and they may find good and stable solutions.
[...]

PR #2579, opened 2024-11-17
Fix converting Rxx and similar Qiskit gates 

gluonhiggs:
## Description
Resolves issue #2578

Basically, the error happens technically due to the fact that `rzz` is not recognized by QASM.
In general, this can also happen to other gates that are pre-built in any framework but not existing in QASM.
So, the idea is that we should `decompose()` some parts of the argument circuit such that these gate are more recognizable to QASM, and then `decompose()` one more time to obtain the decomposition by the basis gates.

My original thought was:

> Basically, the error happens technically due to the fact that rzz is not recognized by QASM.
> In general, this can also happen to other gates that are pre-built in any framework but not existing in QASM.
> So, the idea is that we should decompose() the argument circuit to semi-basic gate using QFT such that these gate are recognizable to QASM, and then decompose() one more time to obtain the decomposition by the basis gates. Turns out it was not working even passing all tests.

---

### License

- [x] ~~I [updated the documentation](../blob/main/docs/CONTRIBUTING_DOCS.md) where relevant.~~ no doc updated.

### DIFF
--- a/mitiq/interface/mitiq_qiskit/conversions.py
+++ b/mitiq/interface/mitiq_qiskit/conversions.py
@@ -249,17 +249,19 @@ def from_qiskit(circuit: qiskit.QuantumCircuit) -> cirq.Circuit:
     Returns:
         Mitiq circuit representation equivalent to the input Qiskit circuit.
     """
+
     try:
         mitiq_circuit = from_qasm(qasm2.dumps(circuit))
     except QasmException:
         # Try to decompose circuit before running
         # This is necessary for converting qiskit circuits with
         # custom packaged gates, e.g., QFT gates
+        gates_to_decompose = ["rxx", "rzz", "rzx", "ryy", "QFT"]
         circuit = circuit.decompose(
-            gates_to_decompose=["u3", "cx", "rx", "ry", "rz"]
+            gates_to_decompose=gates_to_decompose, reps=10
         )
-        circuit = circuit.decompose()
         mitiq_circuit = from_qasm(qasm2.dumps(circuit))
+
     return mitiq_circuit
 
 
--- a/mitiq/interface/mitiq_qiskit/conversions.py
+++ b/mitiq/interface/mitiq_qiskit/conversions.py
@@ -52,7 +52,9 @@ def _remove_qasm_barriers(qasm: QASMType) -> QASMType:
     return "".join(lines)
 
 
-def _map_bit_index(bit_index: int, new_register_sizes: List[int]) -> Tuple[int, int]:
+def _map_bit_index(
+    bit_index: int, new_register_sizes: List[int]
+) -> Tuple[int, int]:
     """Returns the register index and (qu)bit index in this register for the
     mapped bit_index.
 
--- a/mitiq/interface/mitiq_qiskit/conversions.py
+++ b/mitiq/interface/mitiq_qiskit/conversions.py
@@ -52,7 +52,9 @@ def _remove_qasm_barriers(qasm: QASMType) -> QASMType:
     return "".join(lines)
 
 
-def _map_bit_index(bit_index: int, new_register_sizes: List[int]) -> Tuple[int, int]:
+def _map_bit_index(
+    bit_index: int, new_register_sizes: List[int]
+) -> Tuple[int, int]:
     """Returns the register index and (qu)bit index in this register for the
     mapped bit_index.
 
@@ -253,7 +255,9 @@ def from_qiskit(circuit: qiskit.QuantumCircuit) -> cirq.Circuit:
         # Try to decompose circuit before running
         # This is necessary for converting qiskit circuits with
         # custom packaged gates, e.g., QFT gates
-        circuit = circuit.decompose(gates_to_decompose=["u3", "cx", "rx", "ry", "rz"])
+        circuit = circuit.decompose(
+            gates_to_decompose=["u3", "cx", "rx", "ry", "rz"]
+        )
         circuit = circuit.decompose()
         mitiq_circuit = from_qasm(qasm2.dumps(circuit))
     return mitiq_circuit
--- a/mitiq/interface/mitiq_qiskit/conversions.py
+++ b/mitiq/interface/mitiq_qiskit/conversions.py
@@ -16,6 +16,7 @@ import qiskit
 from cirq.contrib.qasm_import import circuit_from_qasm
 from cirq.contrib.qasm_import.exception import QasmException
 from qiskit import qasm2
+from qiskit import transpile
 from qiskit.transpiler import PassManager
 from qiskit.transpiler.layout import Layout
 from qiskit.transpiler.passes import SetLayout
@@ -52,9 +53,7 @@ def _remove_qasm_barriers(qasm: QASMType) -> QASMType:
     return "".join(lines)
 
 
-def _map_bit_index(
-    bit_index: int, new_register_sizes: List[int]
-) -> Tuple[int, int]:
+def _map_bit_index(bit_index: int, new_register_sizes: List[int]) -> Tuple[int, int]:
     """Returns the register index and (qu)bit index in this register for the
     mapped bit_index.
 
@@ -255,6 +254,7 @@ def from_qiskit(circuit: qiskit.QuantumCircuit) -> cirq.Circuit:
         # Try to decompose circuit before running
         # This is necessary for converting qiskit circuits with
         # custom packaged gates, e.g., QFT gates
+        circuit = circuit.decompose(gates_to_decompose=["qft"])
         circuit = circuit.decompose()
         mitiq_circuit = from_qasm(qasm2.dumps(circuit))
     return mitiq_circuit
--- a/mitiq/interface/mitiq_qiskit/conversions.py
+++ b/mitiq/interface/mitiq_qiskit/conversions.py
@@ -15,7 +15,7 @@ import numpy as np
 import qiskit
 from cirq.contrib.qasm_import import circuit_from_qasm
 from cirq.contrib.qasm_import.exception import QasmException
-from qiskit import qasm2
+from qiskit import qasm2, transpile
 from qiskit.transpiler import PassManager
 from qiskit.transpiler.layout import Layout
 from qiskit.transpiler.passes import SetLayout
@@ -52,9 +52,7 @@ def _remove_qasm_barriers(qasm: QASMType) -> QASMType:
     return "".join(lines)
 
 
-def _map_bit_index(
-    bit_index: int, new_register_sizes: List[int]
-) -> Tuple[int, int]:
+def _map_bit_index(bit_index: int, new_register_sizes: List[int]) -> Tuple[int, int]:
     """Returns the register index and (qu)bit index in this register for the
     mapped bit_index.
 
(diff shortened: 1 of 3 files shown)


mitiq/interface/mitiq_qiskit/conversions.py
```diff
@@ -255,6 +255,9 @@ def from_qiskit(circuit: qiskit.QuantumCircuit) -> cirq.Circuit:
         # Try to decompose circuit before running
         # This is necessary for converting qiskit circuits with
         # custom packaged gates, e.g., QFT gates
+        circuit = circuit.decompose(
+            gates_to_decompose=["u1", "u2", "u3", "cx", "rx", "ry"]
```
andreamari, line 259: First, thanks @gluonhiggs for the quick fix!
I am not sure about this line where you are decomposing simple one-qubit gates such as "cx", "rx", "ry". Those gates are supported by Mitiq/Cirq and we risk to over-decompose the circuit.

I understand that tests are passing, but this change may unnecessarily increase the depth of the output circuit and this is a problem for the efficiency of the quantum computation. For example, if you input a short qiskit circuit made by a single "rzz" gate and many "cx", "rx", "ry" , Mitiq would decompose all gates producing an extremely long output circuit.

I don't know if it works, but I would suggest,it  **after** the existing line `circuit=circuit.decompose()`,
to add something like this:

`circuit = circuit.decompose(gates_to_decompose=["rzz", "rzx", "rxx"], reps=10)`

In this way, Mitiq will first apply a general first decomposition step as usual.
Afterwards, if there are still "rzz", "rzx", "rxx" gates left in the circuit, it will simplify only those specific gates.

What do you think?

gluonhiggs: @cosenal I thought about that and tried to `transpile` not to the maximum set of the common gates, but to the set of the gates that can be used to construct the others, i.e., `["u1", "u2", "u3", "cx", "rx", "ry", "rz"]`, then `decompose()`. I also thought that this would be efficient because `transpile` come with `optimization_method`. However, the same error still happens. I even tried with a broader set, e.g., `["u3", "u2", "u1", "cx", "id", "x", "y", "z", "h", "s", "t", "rx", "ry", "rz"]`, but nothing has changed. Generally, I think it would be hard to determine the set because there will be more `n-qubit` gates, right?
@andreamari If we choose the set gate to decompose like `["rzz", "rzx", "rxx"]`, I'm wonder if we also have to consider other combinations of `x`, `y`, and `z` here like `ryy`, `ryz`, and so on?

cosenal: What I am proposing is something like
```python
circuit = qiskit.transpile(
    circuit,
    basis_gates=QASM2_GATE_SET,
    optimization_level=0,       
)
 mitiq_circuit = from_qasm(qasm2.dumps(circuit))
 ```
_without_ the `decompose()`, because that would bring the gate set to another set that may not be supported by QASM serializer/parser.

I tried with `QASM2_GATE_SET = ["u1", "u2", "u3", "cx", "rx", "ry", "rz"]` and it seems to work fine. I don't think that's the maximal set though. Having the maximal QASM-supported set there is the best we can do. And something smaller than the maximal will be sub-optimal when it comes to the size of the output circuit (as Andrea pointed out.)

Comment (gluonhiggs) [reply mitiq/interface/mitiq_qiskit/conversions.py:259]: Yeah. It works. Let me have some tests. @andreamari what do you think? I think we can increase the level of optimization or expose the option to the interface.

Comment (andreamari) [reply mitiq/interface/mitiq_qiskit/conversions.py]: Two approaches are possible:

1- Decomposing all the gates that **we don't like** (specified by `gates_to_decompose`) until all the gates we don't like are gone (usually it's enough to add `reps>5`. )

2- Transpiling, as suggested by @cosenal, into a get set that **we like** (e.g. `QASM2_GATE_SET` or a subset of `QASM2_GATE_SET` that is also compatible with Qasm-to-Cirq conversion).

Solution 1 is closer to the existing workflow and can be good compromise. However, Solution 2 looks better at least in theory. 

@gluonhiggs :  I think you are confusing `gates_to_decompose`  with `basis_gates` . The `gates_to_decompose` contains bad gates that we want to decompose. `basis_gates` contains good gates that we want to have in the circuit.  So, e.g., `gates_to_decompose` should not contain `"cx", "rx", "ry"` since these are good gates that we want to keep in the circuit.

@gluonhiggs Yes, I think `ryy, ryz` and any other gate that is not supported should also be in `gates_to_decompose` if you go for Solution 1.

Comment (cosenal) [reply mitiq/interface/mitiq_qiskit/conversions.py]: @andreamari Why would we need repeated decompose() if we add all gates we don't like in `gates_to_decompose`?

Comment (cosenal) [reply mitiq/interface/mitiq_qiskit/conversions.py]: @gluonhiggs Can you elaborate on the tests that you think don't pass with the approach I proposed?
Regarding Update 2, I am not sure about the difference between decompose and transpile in the case where you don't specify any target hardware architecture on the transpile. It's something we can ask Qiskit folks on their community call.

Comment (gluonhiggs) [reply mitiq/interface/mitiq_qiskit/conversions.py]: @andreamari I can distinguish `gates_to_decompose` and `basis_gates`. Perhaps, the way I expressed my thoughts was not clear at some point. The problem here is for some reasons, when I tried to decompose the "bad gates" (different lists of `rii` with this [reference](https://github.com/Qiskit/qiskit/tree/main/qiskit/circuit/library/standard_gates)), the error similar to #2578 (the original issue) still happens. 
@cosenal Here is the short test summary from `pytest`:
```bash
================================================================================= short test summary info =================================================================================
FAILED mitiq/interface/mitiq_qiskit/tests/test_conversions_qiskit.py::test_convert_with_qft - AssertionError: assert False
FAILED mitiq/zne/scaling/tests/test_folding.py::test_convert_qiskit_to_mitiq_circuit_conversion_error - Failed: DID NOT RAISE <class 'mitiq.interface.conversions.CircuitConversionError'>
FAILED mitiq/zne/scaling/tests/test_folding.py::test_folding_circuit_conversion_error_qiskit[fold_gates_at_random] - Failed: DID NOT RAISE <class 'mitiq.interface.conversions.CircuitConversionError'>
FAILED mitiq/zne/scaling/tests/test_folding.py::test_folding_circuit_conversion_error_qiskit[fold_global] - Failed: DID NOT RAISE <class 'mitiq.interface.conversions.CircuitConversionError'>
```
Should I paste the whole log here or move the discussion to discord?

gluonhiggs: Here are the tests failed:
Solution 1:
mitiq/interface/mitiq_qiskit/tests/test_conversions_qiskit.py::test_convert_with_qft
mitiq/zne/tests/test_zne.py::test_execute_zne_on_qiskit_circuit_with_QFT

Solution 2:
mitiq/interface/mitiq_qiskit/tests/test_conversions_qiskit.py::test_convert_with_qft
mitiq/zne/scaling/tests/test_folding.py::test_convert_qiskit_to_mitiq_circuit_conversion_error
mitiq/zne/scaling/tests/test_folding.py::test_folding_circuit_conversion_error_qiskit[fold_gates_at_random]
mitiq/zne/scaling/tests/test_folding.py::test_folding_circuit_conversion_error_qiskit[fold_global]
The error log is too lengthy that I can't even paste to discord.

andreamari: > @andreamari Why would we need repeated decompose() if we add all gates we don't like in `gates_to_decompose`?

I think you are right, if all bad gates are in gates_to_decompose you probably don't need to also keep the existing line circuit.decompose() (without options).

However I think you would still need the `reps=...` option, since it ensures that potential bad gates that may be created  during the decomposition process are further decomposed.

gluonhiggs: FYI: I have updated the code and it seems to work fine thanks to your suggestion @andreamari 
The story is:
I fell in the 2nd approach that used `transpile` to transform the circuits to some pre-declared `basis_gates`. I had tried dozens of combinations of the supported gates to form `basis_gates`. In the end, I realized that a very good choice of `basis_gates` passing almost all test cases but still can cause `transpile` to fail these tests
```bash
mitiq/zne/scaling/tests/test_folding.py::test_convert_qiskit_to_mitiq_circuit_conversion_error
mitiq/zne/scaling/tests/test_folding.py::test_folding_circuit_conversion_error_qiskit[fold_gates_at_random]
mitiq/zne/scaling/tests/test_folding.py::test_folding_circuit_conversion_error_qiskit[fold_global]
```
because `transpile` is too strong that can transform any circuit of both supported and unsupported gates to a circuit constructed from `basis_gates`. In the context of mitiq, to my understanding, we do not implement the noise mitigation for unsupported gates (or custom gates in the failed test) directly. Therefore, we expect that `convert_to_mitiq` will raise error in case there are custom gates, instead of transforming the circuit normally. So, this is too much. We should not use `transpile` here.

Back to the 1st approach, when these tests are failed
```bash
mitiq/interface/mitiq_qiskit/tests/test_conversions_qiskit.py::test_convert_with_qft
mitiq/zne/tests/test_zne.py::test_execute_zne_on_qiskit_circuit_with_QFT
```
so I tried to add `"QFT"` to the `gates_to_decompose` and it's worked since then. Previously, instead of using `"QFT"`, I used `"qft"`. That's why it didn't work.

cosenal: @andreamari 
>  since it ensures that potential bad gates that may be created during the decomposition process

I would be surprised if gates from the `gates_to_decompose` reappear during decomposition. Something we can ask to Qiskit folks.

cosenal: @gluonhiggs if we go with the approach I see now in the PR (good that it works), we could do the decomposition directly instead of doing it as a fallback when the conversion fails.

mitiq/interface/mitiq_qiskit/conversions.py
```diff
@@ -249,14 +249,19 @@ def from_qiskit(circuit: qiskit.QuantumCircuit) -> cirq.Circuit:
...
     """
+
     try:
         mitiq_circuit = from_qasm(qasm2.dumps(circuit))
     except QasmException:
         # Try to decompose circuit before running
         # This is necessary for converting qiskit circuits with
         # custom packaged gates, e.g., QFT gates
-        circuit = circuit.decompose()
+        gates_to_decompose = ["rxx", "rzz", "rzx", "ryy", "QFT"]
```
cosenal, line 259: nit: this should be capitalized as we want to treat it as a constant

mitiq/interface/mitiq_qiskit/conversions.py
```diff
@@ -249,14 +249,19 @@ def from_qiskit(circuit: qiskit.QuantumCircuit) -> cirq.Circuit:
...
+        circuit = circuit.decompose(
+            gates_to_decompose=gates_to_decompose, reps=10
```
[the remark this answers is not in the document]
gluonhiggs: yes. I agree!

mitiq/interface/mitiq_qiskit/conversions.py
gluonhiggs: I have updated the code. Thank you!

cosenal (approved):
LGTM. Thanks for iterating on it!

andreamari (approved):
One could consider adding to GATES_TO_DECOMPOSE also the gates mentioned in #1073.
But maybe this would go too much beyond the scope of this PR, that is fixing #2578.

So, LGTM! 
And thanks everyone for your help!

PR #2593, opened 2024-12-02
Throw Erorr on Multiple Measurements per Qubit with Observables

bdg221:
## Description

This PR addresses the issue of a circuit already including a measurement on a qubit and an observable being passed to the executor resulting in another measurement on the qubit.

To address this, before adding the measurement from the observable (PauliString), we check to if a measurement gate has already been applied to the qubit in the circuit.

Also, a test was added to confirm this is behaving properly.

### License

### DIFF
--- a/mitiq/observable/pauli.py
+++ b/mitiq/observable/pauli.py
@@ -279,10 +279,25 @@ class PauliStringCollection:
 
         basis_rotations = set()
         support = set()
+        qubits_with_measurements = set[cirq.Qid]()
+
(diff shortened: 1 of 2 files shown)


natestemen (approved):
Nice work pulling this out of the larger PR, Brian!

I thought for a while about whether this should be a warning or error, and I think you're right that an error is the right way to go. Broadly, someone might intentionally create a situation where there are two measurements on a single qubit, but that generally doesn't work inside Mitiq, so best to catch and throw and error.

Issue #2586, opened 2024-11-27
add QEM schema to repo


natestemen:
The development of a schema for quantum error mitigation techniques has been undergoing through an [RFC](https://docs.google.com/document/d/1yX1yzs2TuGfM62CPwFcFwoTNp0IjoWNkznlsIfYuZog/), as well as a private repository. They are now at a point where they should be added, so we can start integrating them into the `Calibrator` and other tools.

I suggest creating a new directory `mitiq/schema` to add the files for the schema and work with them from there. More issues to come, but lets start simple.

cc @ecarlander @ruheenirodi comment here if you have any questions about getting these added, and let me know who I should assign to the issue.

natestemen:
Oh also, please add a link to the repo if you're okay with making it public. If you'd like to keep it private that's okay, but I know others want to see!

ecarlander:
We copied the contents out our working repo into the mitiq/schema directory, but if we should instead reserve that for more developed work we can switch back to that and I'll share the link! I also mentioned this in the PR I opened

PR #2597, opened 2024-12-07
adding QEM schema directory

ecarlander:
## Description

Resolves issue #2586
Added directory mitiq/schema containing  @ruheenirodi and I's current work developing QEM schema. 

A lot of this work (all) is not finalized, so let us know if we should instead just start with an empty directory and only add work that is fully-developed. 

---

### License


natestemen:
Thanks for adding this here Ella! Good start. We'll need to pair this down quite a bit before merging it in. For the time being, lets just add the schemas, and the code to validate experiments for the techniques you've already completed. Then we can make sure to add tests, ensure the code is formatted correctly, etc. Baby steps help us build up confidence about new code!

cosenal:
Is this in progress, is it in review? @natestemen Has the pair programming session happened? I see the PR is now much more manageable.

natestemen:
We are going to hold off on this for now purely for the reason of having limited bandwidth (and not because this isn't valuable work).

Issue #2584, opened 2024-11-26
LRE error with qiskit circuit


purva-thakre:
When attempting to get the sample matrrix of a qiskit circuit, the process fails with `AttributeError: 'QuantumCircuit' object has no attribute 'are_all_measurements_terminal'`.  This error should not be raised as the circuit of interest does not contain any measurements.

The error points to the following line where the scale factor vectors are generated and used by the sample matrix function:

https://github.com/unitaryfund/mitiq/blob/be6a76ce940033ddd58538f3fe9e989b8e536850/mitiq/lre/inference/multivariate_richardson.py#L97-L99

The function to generate the scale factor vectors relies on 

https://github.com/unitaryfund/mitiq/blob/be6a76ce940033ddd58538f3fe9e989b8e536850/mitiq/lre/multivariate_scaling/layerwise_folding.py#L39

### Code

```py
ghz_circuit = generate_ghz_circuit(4, "qiskit")
circuit = ghz_circuit.compose(ghz_circuit.inverse())
print(circuit)

     ┌───┐                              ┌───┐
q_0: ┤ H ├──■────────────────────────■──┤ H ├
     └───┘┌─┴─┐                    ┌─┴─┐└───┘
q_1: ─────┤ X ├──■──────────────■──┤ X ├─────
          └───┘┌─┴─┐          ┌─┴─┐└───┘     
q_2: ──────────┤ X ├──■────■──┤ X ├──────────
               └───┘┌─┴─┐┌─┴─┐└───┘          
q_3: ───────────────┤ X ├┤ X ├───────────────
                    └───┘└───┘               


sample_matrix(circuit, 2, 2)
```

purva-thakre:
To resolve this issue, we should add the corresponding unit tests in `mitiq/lre/tests/test_multivariate_richardson.py`. 

Compatibility of LRE with non-Cirq frontends was ensured in #2547 

This is definitely related to https://github.com/unitaryfund/mitiq/pull/2542#discussion_r1821942480

natestemen:
> This is definitely related to https://github.com/unitaryfund/mitiq/pull/2542#discussion_r1821942480

Ahhh 🙈 sorry! Since I worked on this previously, would it be easiest if I take it?

PR #2599, opened 2024-12-10
Generalize extra LRE functions to work with all QPROGRAMs

natestemen:
## Description

Quick fix to ensure the LRE module is compatible with all `QPROGRAM` instances.

fixes https://github.com/unitaryfund/mitiq/issues/2584

### DIFF
--- a/mitiq/lre/inference/multivariate_richardson.py
+++ b/mitiq/lre/inference/multivariate_richardson.py
@@ -52,6 +52,7 @@ def _full_monomial_basis_term_exponents(
     return sorted(exponents, key=lambda term: (sum(term), term[::-1]))
 
 
+@accept_any_qprogram_as_input
 def sample_matrix(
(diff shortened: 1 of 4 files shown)


purva-thakre (approved):
LGTM! I wonder if we should make sure other public functions in LRE also have a similar compatibility?

I see `sample_matrix` only works for Cirq circuits. 

https://github.com/unitaryfund/mitiq/blob/76486485509fc20204e12cd27f3d52827844b93c/mitiq/lre/inference/multivariate_richardson.py#L55-L56

natestemen:
> I wonder if we should make sure other public functions in LRE also have a similar compatibility?

Good call, added https://github.com/unitaryfund/mitiq/pull/2599/commits/db1b3939ac20a0f8d4430ad7c296606388eec08a to address `sample_matrix`.

purva-thakre (approved):

PR #2601, opened 2024-12-13
PT + ZNE tutorial

### LINKED ISSUE #2374: Tutorial: Composing techniques: Pauli Twirling and Zero Noise Extrapolation
In the same way we have a tutorial for [REM+ZNE](https://mitiq.readthedocs.io/en/latest/examples/combine_rem_zne.html) and one for [DDD+ZNE](https://mitiq.readthedocs.io/en/latest/examples/combine_ddd_zne.html), we want to create one about composition of Pauli Twirling (PT) and ZNE.

This would be an exploratory jupyter notebook in the Tutorial secions of Mitiq docs. 

Ideally, the example would show that error(PT + ZNE) < error(PT) < error(ZNE)

purva-thakre:
## Description

Fixes #2374 

---

### License


purva-thakre:
@natestemen Do you mind taking a quick look at the general outline of this tutorial? I have to clean up the code a bit. So, you could ignore the code blocks for now. 

I am not too sure if I need to describe something in excruciating detail as the PT user guide is still WIP.

natestemen:
Took a quick look through things and this looks great! I think it's a nice split of theory and practice. I can imagine some of the material working its way in to the user guide, but lets not worry about that for now.

purva-thakre:
@natestemen The comparison plot is missing data related to only ZNE values. 

This is because the circuit I am using is very small for ZNE. If I increase the circuit size used for the comparison plot, the time taken by the tutorial during execution is a lot.

docs/source/examples/pt_zne.md
```diff
@@ -0,0 +1,378 @@
...
+
+This tutorial helps us understand the cases when noise tailoring is helpful for improving the performance of 
+quantum error mitigation techniques. 
+
+We will
+
+- compare coherent noise to incoherent noise
+- Pauli twirl coherent noise to be incoherent
+- compare the performance of [Zero-Noise Extrapolation](../guide/zne.md) (ZNE)
+on its own and in combination with [Pauli Twirling](../guide/pt.md) (PT) when the circuit is subjected to coherent noise
```
natestemen, line 26:
```suggestion
This tutorial explores how noise tailoring can improve the effectiveness of quantum error mitigation techniques.
Specifically, we analyze how converting coherent noise into incoherent noise through Pauli Twirling impacts the performance of [Zero-Noise Extrapolation](TODO) (ZNE).

In this tutorial, we will:

1. Define and compare coherent and incoherent noise
2. Apply Pauli Twirling to transform coherent noise into incoherent noise
3. Compare the performance of ZNE
    1. on its own, and
    2. in combination with [Pauli Twirling](TODO)

By the end of the example, you will understand when and why noise tailoring can enhance quantum error mitigation.
```

docs/source/examples/pt_zne.md
```diff
@@ -0,0 +1,378 @@
...
+    jupytext_version: 1.14.1
+kernelspec:
+  display_name: Python 3
+  language: python
+  name: python3
+---
+
+```{tags} cirq, zne, pt, basic
+```
+# ZNE with Noise Tailoring
```
natestemen, line 16:
```suggestion
# Zero-Noise Extrapolation with Twirling
```

Twirling or Noise Tailoring? I feel like more people have heard of twirling?

docs/source/examples/pt_zne.md
```diff
@@ -0,0 +1,378 @@
...
+
+**Incoherent noise** is a process that entangles the quantum system with its environment i.e. this type of noise is irreversible. The system and the environment are in a mixed state. It scales linearly in the small error limit. 
+The noise channel can be described using Pauli operators which makes it easy to analyze and simulate. Worst case error rate is directly proportional to the **average gate infidelity**. 
+
+```{note}
+If $\mathcal{F}$ is the fidelity defining the success of preparing an arbitrary pure state $\rho$, then
+$1-\mathcal{F}$ is the **average gate infidelity**.
+```
+
+For example, a depolarizing noise channel is a stochastic noise channel where a noiseless process is probabilistically mixed with orthogonal errors. If $\rho$ is a single qubit state, $p$ is the probabilistic error rate and $\epsilon(\rho)$ is the noise channel:
```
natestemen, line 40:
```suggestion
For example, a depolarizing noise channel is a stochastic noise channel where a noiseless process is probabilistically mixed with orthogonal errors. If $\rho$ is a single qubit state, $p$ is the probabilistic error rate and $\mathcal{E}(\rho)$ is the noise channel:
```

I've typically seen $\mathcal{E}$ used for the error channel. Just notation, so if you don't like it no worries, but nice to be in line with what else is out there.

docs/source/examples/pt_zne.md
```diff
@@ -0,0 +1,378 @@
...
+
+We will
+
+- compare coherent noise to incoherent noise
+- Pauli twirl coherent noise to be incoherent
+- compare the performance of [Zero-Noise Extrapolation](../guide/zne.md) (ZNE)
+on its own and in combination with [Pauli Twirling](../guide/pt.md) (PT) when the circuit is subjected to coherent noise
+
+## Coherent noise vs. Incoherent noise
+
```
natestemen, line 29: Add a quick intro before diving in to definitions here. Maybe something like "Noise on quantum devices can broadly be broken into two classes. Each has different properties that requires mitigation by different techniques ..."

docs/source/examples/pt_zne.md
```diff
@@ -0,0 +1,378 @@
...
+**Coherent noise** is a reversible process as long as the noisy unitary transformation is known beforehand which is not always the case. These types of noise maintain the purity of the state. But in a quantum circuit, they are easily carried across the circuit such that they scale up quadratically. Dealing with coherent noise requires a large resource overhead.
```
natestemen, line 30: > But in a quantum circuit, they are easily carried across the circuit such that they scale up quadratically.

How much work would this be to make a little more precise? "They scale up quadratically" in particular.

docs/source/examples/pt_zne.md
```diff
@@ -0,0 +1,378 @@
...
+Thus, the depolarizing channel can be redescribed using Paulis as shown below. 
+
+```{math}
+\epsilon(\rho) = (1-\frac{3p}{4}) \rho + \frac{p}{4} (X \rho X + Y \rho Y + Z \rho Z)
+
+```
+
+### Pauli Transfer Matrix
+
+Let $\Lambda(rho)$ be a $n-$qubit noise channel where $K_i$ are the Kraus operators.
```
natestemen, line 60:
```suggestion
Let $\Lambda(\rho)$ be a $n$-qubit noise channel with $K_i$ being the corresponding Kraus operators.
```

docs/source/examples/pt_zne.md
```diff
@@ -0,0 +1,378 @@
...
+
+```{math}
+
+\Lambda(\rho) = \sum_{i=1} K_i \rho {K_i}^\dagger
+```
+If $P_i$ and $P_j$ are $n-$qubit Paulis, the following expression defines the entries of a Pauli Transfer Matrix (PTM). Here, 
+$i$ defines the rows while $j$ defines the columns of the PTM. All entries of the PTM are real and in the interval $[-1, 1]$.
```
natestemen, line 67:
```suggestion
If $P_i$ and $P_j$ are $n-$qubit Paulis, the following expression defines the entries of a Pauli Transfer Matrix (PTM). All entries of the PTM are real and in the interval $[-1, 1]$.
```

docs/source/examples/pt_zne.md
```diff
@@ -0,0 +1,378 @@
...
+
+
+```{math}
+(R_{\Lambda})_{ij} = \frac{1}{2^n} \text{Tr} \{ P_i \Lambda(P_j)\}
+```
+A PTM allows us to distinguish between the two types of noise. The off-diagonal terms of the PTM are due to the effect of coherent noise while the diagonal terms are due to incoherent noise. PTM of a circuit is a product of the PTM of each layer in the circuit. Due to these, it is straightforward to see how coherent noise carries across different layers in the circuit and how incoherent errors are easier to deal with in the small error limit.
```
natestemen, line 73:
```suggestion
A PTM allows us to distinguish between the two types of noise. The off-diagonal terms of the PTM are due to the effect of coherent noise while the diagonal terms are due to incoherent noise. To find the PTM of an entire circuit, we only need to take the product of the PTM of each layer in the circuit. Due to this, it is straightforward to see how coherent noise carries across different layers in the circuit and how incoherent errors are easier to deal with in the small error limit. 
```

> how incoherent errors are easier to deal with in the small error limit

I don't actually see why this is the case. Can you add some more detail?

docs/source/examples/pt_zne.md
```diff
@@ -0,0 +1,378 @@
...
+
+
+```{math}
+(R_{\Lambda})_{ij} = \frac{1}{2^n} \text{Tr} \{ P_i \Lambda(P_j)\}
+```
+A PTM allows us to distinguish between the two types of noise. The off-diagonal terms of the PTM are due to the effect of coherent noise while the diagonal terms are due to incoherent noise. PTM of a circuit is a product of the PTM of each layer in the circuit. Due to these, it is straightforward to see how coherent noise carries across different layers in the circuit and how incoherent errors are easier to deal with in the small error limit. 
+
+The known fault tolerant thresholds for stochastic noise are higher than coherent noise which makes the former a 'preferable' type of noise compared to the latter. If we want to not deal with coherent noise, Pauli twirling can be used to tailor coherent noise to incoherent noise. Same as the expression above, when a coherent noise channel is Pauli twirled, the noise channel can be described using Paulis after averaging over multiple Pauli twirled circuits. Refer to the [Pauli Twirling user guide](../guide/pt.md) for additional information.
```
natestemen, line 75: The thing that I'm missing here is a discussion of how many twirled circuits one needs in order to effectively do the task of converting coherent errors to incoherent.

docs/source/examples/pt_zne.md
```diff
@@ -0,0 +1,378 @@
...
+```
+A PTM allows us to distinguish between the two types of noise. The off-diagonal terms of the PTM are due to the effect of coherent noise while the diagonal terms are due to incoherent noise. PTM of a circuit is a product of the PTM of each layer in the circuit. Due to these, it is straightforward to see how coherent noise carries across different layers in the circuit and how incoherent errors are easier to deal with in the small error limit. 
+
+The known fault tolerant thresholds for stochastic noise are higher than coherent noise which makes the former a 'preferable' type of noise compared to the latter. If we want to not deal with coherent noise, Pauli twirling can be used to tailor coherent noise to incoherent noise. Same as the expression above, when a coherent noise channel is Pauli twirled, the noise channel can be described using Paulis after averaging over multiple Pauli twirled circuits. Refer to the [Pauli Twirling user guide](../guide/pt.md) for additional information. 
+
+## Using Pauli Twirling in Mitiq
+
+```{code-cell} ipython3
+# Utility functions for Pauli Twirling.
+
```
natestemen, line 81:
```suggestion
```

docs/source/examples/pt_zne.md
```diff
@@ -0,0 +1,378 @@
...
+pauli_unitary_list = [
+    cirq.unitary((cirq.I)),
+    cirq.unitary((cirq.X)),
+    cirq.unitary((cirq.Y)),
+    cirq.unitary((cirq.Z)),
+]
+
+
+def _n_qubit_paulis(num_qubits: int) -> list[npt.NDArray[np.complex64]]:
```
natestemen, line 98: Lets not make these functions "private" since we are showing them to the reader.

It is possible to hide code cells, thought. Do you think we should make this code cell foldable so only the interested reader takes a look?

natestemen (commented):
Haven't finished reading through everything just yet, but wanted to get feedback soon so we can start iterating.

Overall this is a really great mix of theory and practice. Well done setting the bones up.

docs/source/_thumbnails/pt_zne.png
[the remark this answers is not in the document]
purva-thakre: Yeah, I have been debating if I should plot the difference or the expectation values themselves. 

Right now, I am comparing every other expectation value with the ideal value.

docs/source/examples/pt_zne.md
natestemen: Makes sense. Some comments on that would be good.

docs/source/_thumbnails/pt_zne.png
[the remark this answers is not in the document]
natestemen: I think just adding the ideal values to this would be great!

docs/source/examples/pt_zne.md
purva-thakre: Requires some math to show this.

natestemen: Okay lets leave that out, but it might be good to specify what "they" are.

docs/source/examples/pt_zne.md
```diff
@@ -0,0 +1,378 @@
...
+    CNOT.on(q2, q3),
+)
+
+print(circuit)
+
+```
+
+We are going to add coherent noise to this circuit and then get the error-mitigated expectation value. For a detailed discussion on this, refer to the [ZNE user guide](../guide/zne-1-intro.md). 
+
+As we are using a simulator, we have to make sure the twirling gates are sandwiching the coherent noise channel in the circuit.
```
cosenal, line 255: This sentence is confusing, and in fact was already object of confusion with that infamous UnitaryHack task. The way to describe this is simply that we define a simulator where the noise model introduces coherent noise. Look at this as a black box, it being an actual hardware or a simulator. You ask the black box to execute the circuit, and whenever the box is supposed to execute a controlled gate, it corrupts the result with noise. If we mix it with twirling, or use words like sandwiching, etc.. we confuse the reader unnecessarily.

docs/source/examples/pt_zne.md
```diff
@@ -0,0 +1,378 @@
...
+    return (
+        DensityMatrixSimulator(noise=get_noise_model(noise_level=noise_level))
+        .simulate(circuit)
+        .final_density_matrix[0, 0]
+        .real
+    )
+
+
+# Set the intensity of the noise
+NOISE_LEVEL = 7
```
cosenal, line 290: 7 pi/2 rotations? 🤔

docs/source/examples/pt_zne.md
```diff
@@ -0,0 +1,378 @@
...
+    """Substitute each CZ and CNOT gate in the circuit
+    with the gate itself followed by an Rx rotation on the output qubits.
+    """
+    rads = pi / 2 * noise_level
+    def noisy_c_gate(op):
+        if isinstance(op.gate, (CZPowGate, CXPowGate)):
+            return CircuitOperation(
+                Circuit(
+                    op.gate.on(*op.qubits), 
+                    Ry(rads=rads).on_each(op.qubits),
```
cosenal, line 273: you say Rx in the docstring, but apply Ry here

cosenal:
I restarted the RTD build; having a rendered version will make it easier to review.

docs/source/_thumbnails/pt_zne.png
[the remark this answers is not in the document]
purva-thakre: I decided to plot the expectation values only rather than a difference of the expectation values. The former makes a better plot at showing when combining PT and ZNE is better.

docs/source/examples/pt_zne.md
```diff
@@ -0,0 +1,392 @@
...
+
+q0 = LineQubit(0)
+q1 = LineQubit(1)
+
+circuit = Circuit(CNOT(q0, q1))
+print(circuit)
+```
+```{code-cell} ipython3
+ptmcnot = ptm_matrix(circuit, 2)
+ax = sns.heatmap(ptmcnot.real, linewidth=0.5)
```
natestemen, line 190: Currently the heatmap bounds are changing for each plot, making it difficult to compare plots against each other. Can you set the bounds for all heatmaps?

docs/source/examples/pt_zne.md
```diff
@@ -0,0 +1,392 @@
...
+    error_with_twirling_and_zne.append(abs(ideal_value - mitigated_twirled_result))
+
+
+plt.plot(noise_strength, error_no_zne_no_twirling,"", label=r"|Ideal - Noisy|", color="#1f77b4")
+plt.plot(noise_strength, zne_vals, "", label=r"|Ideal - ZNE|", color="#bcbd22")
+plt.plot(noise_strength, error_with_twirling,"", label=r"|Ideal - Twirling|", color="#ff7f0e")
+plt.plot(noise_strength, error_with_twirling_and_zne, "", label=r"|Ideal - (ZNE + Twirling)|", color="#2ca02c")
+
+plt.xlabel(r"Noise strength, Coherent noise:$R_y(\frac{\pi}{2} \times \text{noise_strength})$")
+plt.ylabel("Expectation Values Difference")
```
natestemen, line 374: Can we relabel this to "Absolute Error"?

docs/source/examples/pt_zne.md
```diff
@@ -0,0 +1,392 @@
...
+plt.title("Comparison of expectation values with ideal as a function of noise strength")
+plt.legend()
+plt.show()
```
natestemen, line 377: Do we understand why ZNE is so good when the noise strength = 0.5?

natestemen (changes requested):
Really solid tutorial! Can't wait to get it up!

docs/source/examples/pt_zne.md
purva-thakre: I am not too sure for the circuit used in this example. 

If I use a smaller GHZ circuit of 2 qubits, the noise does not do anything as the final state vector before and after adding coherent noise is the same. Interestingly, this behavior is not preserved as the number of qubits in the initial ghz circuit is increased. Worth looking into!

docs/source/examples/pt_zne.md
```diff
@@ -0,0 +1,392 @@
...
+### Pauli Transfer Matrix
+
+Let $\Lambda(\rho)$ be a $n$-qubit noise channel with $K_i$ being the corresponding Kraus operators.
+
+```{math}
+
+\Lambda(\rho) = \sum_{i=1} K_i \rho {K_i}^\dagger
+```
+If $P_i$ and $P_j$ are $n$-qubit Paulis, the following expression defines the entries of a Pauli Transfer Matrix (PTM). Here, 
+$i$ defines the rows while $j$ defines the columns of the PTM. All entries of the PTM are real and in the interval $[-1, 1]$.
```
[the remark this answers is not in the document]
natestemen: Not totally IMO. Part of the reason I mention it though, is because when you say "If $P_i$ and $P_j$ are Paulis,..." it seems like they can be any Paulis, but in reality they are specific ones, according to an ordering.

As long as that is cleared up somehow, I'm fine.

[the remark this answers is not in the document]
purva-thakre: I am still a bit confused about what you wanted me to add. Here are the newer changes:

image

docs/source/examples/pt_zne.md
```diff
@@ -0,0 +1,386 @@
...
+## Coherent noise vs. Incoherent noise
+
+Noise on quantum devices can broadly categorized into two types: Coherent, Incoherent. Each has different properties that can be unfavorable toward a quantum circuit in different ways.
+
+**Coherent noise** is a reversible process as long as the noisy unitary transformation is known beforehand which is not always the case. These types of noise maintain the purity of the state. But in a quantum circuit subjected to coherent noise, the errors are easily carried across the circuit. This can be discerned through the **average gate infidelity** $r(\mathcal{E})$. When coherent errors contribute to a portion of the total error-rate, the worst case infidelity can scale as $\sqrt{r(\mathcal{E})}$ {cite}`Wallman_2014` which in turn reduces the performance of a quantum device by orders of magnitude. Thus, dealing with coherent noise requires a large resource overhead to acquire inferred knowledge of the noise unitaries which can then be used to reverse the effects.
+
+```{note}
+If $\mathcal{F}$ is the fidelity defining the success of preparing an arbitrary pure state $\rho$, then
+$1-\mathcal{F}$ is the **average gate infidelity**.
+```
```
natestemen, line 40: I like that we're bringing in some mathematical results here, but I do wonder if there is some ambiguity around which fidelity these statements are in reference to. Is this a process fidelity or state fidelity? Is $\mathcal{F}$ equal to what's defined here https://docs.quantum.ibm.com/api/qiskit/0.19/qiskit.quantum_info.average_gate_fidelity?

docs/source/examples/pt_zne.md
```diff
@@ -0,0 +1,386 @@
...
+
+**Incoherent noise** is a process that results in the quantum system entangling with its environment i.e. this type of noise is irreversible. The system and the environment end up in a mixed state. It scales linearly in the small error limit. 
+The noise channel can be described using Pauli operators which makes it easy to analyze and simulate. Worst case error rate is directly proportional to the **average gate infidelity**. 
+
+For example, a depolarizing noise channel is a stochastic noise channel where a noiseless process is probabilistically mixed with orthogonal errors. If $\rho$ is a single qubit state, $p$ is the probabilistic error rate and $\mathcal{E}(\rho)$ is the noise channel:
```
natestemen, line 45: This is a great example, but it needs a connection back to (in)coherent noise. Can you add something, maybe at the end?

Do you think we should have an example for coherent noise too?

docs/source/examples/pt_zne.md
```diff
@@ -0,0 +1,386 @@
...
+
+The known fault tolerant thresholds for stochastic noise are higher than coherent noise which makes the former a 'preferable' type of noise compared to the latter. If we want to not deal with coherent noise, Pauli twirling can be used to tailor coherent noise to incoherent noise. Same as the expression above, when a coherent noise channel is Pauli twirled, the noise channel can be described using Paulis after averaging over multiple Pauli twirled circuits. Refer to the [Pauli Twirling user guide](../guide/pt.md) for additional information.
```
natestemen, line 80: > If we want to not deal with coherent noise, ...

To avoid dealing with coherent noise, Pauli twirling can be applied to convert coherent noise into incoherent noise.

> Same as the expression above, ...

Which expression above are you referring to?

docs/source/examples/pt_zne.md
```diff
@@ -0,0 +1,386 @@
...
+q0 = LineQubit(0)
+q1 = LineQubit(1)
+
+circuit = Circuit(CNOT(q0, q1))
+print(circuit)
+```
+```{code-cell} ipython3
+ptmcnot = ptm_matrix(circuit, 2)
+ax = sns.heatmap(ptmcnot.real, linewidth=0.5, vmin=-1, vmax=1)
+print("Ideal CNOT PTM")
```
natestemen, line 185: Can you add this as a title instead of printing it?

docs/source/examples/pt_zne.md
```diff
@@ -0,0 +1,386 @@
...
+# Average results executed over twirled circuits
+from functools import partial
+from mitiq import Executor
+pt_vals = Executor(partial(execute, noise_level=NOISE_LEVEL)).evaluate(twirled_circuits)
+twirled_result = np.average(pt_vals)
+
+
+print(f"Error without twirling: {abs(ideal_value - noisy_value) :.3}")
+print(f"Error with twirling: {abs(ideal_value - twirled_result) :.3}")
+```
```
natestemen, line 304: Sometimes the error with twirling is worse than the error without. This isn't unexpected, but I think deserves a comment. I worry someone might see that and think it's a problem.

docs/source/examples/pt_zne.md
```diff
@@ -0,0 +1,386 @@
...
+zne_pt_vals = []
+
+for i in twirled_circuits:
+    zne_pt_vals.append(execute_with_zne(i, executor))
+
+mitigated_result = np.average(zne_pt_vals)
+
+print(f"Error without twirling: {abs(ideal_value - noisy_value) :.3}")
+print(f"Error with ideal and twirling: {abs(ideal_value - twirled_result) :.3}")
+print(f"Error with ideal and ZNE + PT: {abs(ideal_value - mitigated_result) :.3}")
```
natestemen, line 321:
```suggestion
print(f"Error with twirling: {abs(ideal_value - twirled_result) :.3}")
print(f"Error with ZNE + PT: {abs(ideal_value - mitigated_result) :.3}")
```

natestemen (commented):
Left a few comments, but hopefully nothing major!

docs/source/examples/pt_zne.md
```diff
@@ -0,0 +1,386 @@
...
+noisy_twirled_circuits = []
+
+for circ in twirled_circuits:
+    split_circuit = Circuit(circ[0], circ[1], Ry(rads=np.pi/12)(q0), Ry(rads=np.pi/12)(q1), circ[-1])
+    noisy_twirled_circuits.append(split_circuit)
+
+print("Example noisy twirled circuit", noisy_twirled_circuits[-1], sep="\n")
+```
+
+The twirled PTM is averaged over each noisy twirled circuit such that the new PTM is close to that of the PTM of incoherent noise. We skip the step in this section as we require a very large number of twirled circuits to demonstrate the desired effect of averaging over multiple numpy arrays.
```
cosenal, line 232: A bit anti-climatic, I was looking forward to seeing the effect of this on those PTM heatmaps :)

docs/source/examples/pt_zne.md
purva-thakre: It's the qiskit one.

docs/source/examples/pt_zne.md
```diff
@@ -0,0 +1,386 @@
...
+
+## Noisy ZNE
+
+Lets define a larger circuit of CNOT, CZ and H gates.
```
cosenal, line 236: typo: Let's

also, the ghz circuit below doesn't have CZ.

docs/source/examples/pt_zne.md
```diff
@@ -0,0 +1,386 @@
...
+
+circuit = generate_ghz_circuit(n_qubits=7)
+
+print(circuit)
+
+```
+
+We are going to add coherent noise to this circuit and then get the error-mitigated expectation value. For a detailed discussion on this, refer to the [ZNE user guide](../guide/zne-1-intro.md). 
+
+As we are using a simulator, we have to make sure the noise model adds coherent noise to CZ/CNOT gates in our circuit. For this, `get_noise_model` is used to add noise to CZ/CNOT gates. See [PT user guide](../guide/pt-1-intro.md) for more.
```
cosenal, line 250: solid function that `get_noise_model` 📦 😃

docs/source/examples/pt_zne.md
purva-thakre: The depolarizing noise example is to show how Pauli twirling is equivalent to tailoring the noise into a depolarizing noise channel.

docs/source/examples/pt_zne.md
cosenal: 👍 Motivating this was my main blocker when I was writing this tutorial.

docs/source/examples/pt_zne.md
natestemen: Then I think it should read

> If $\mathcal{F}$ is the average gate fidelity ...

WDYT? Should we add references here?

docs/source/examples/pt_zne.md
natestemen: I don't quite see how this relates to pauli twirling. Can you help spell it out?

From my perspective the main idea in pauli twirling is to add gates that compile to the identity, but change the impact of noise. So when you say the depolarizing channel can be written in terms of paulis, I don't see how that is pauli twirling.

docs/source/examples/pt_zne.md
purva-thakre: Yes, well, this increases the notebook execution time by a lot. The coherent noise that I used is obviously not included in the noisy options allowed by `generate_pauli_twirl_variants`. We should create an issue to clean this portion of pauli twirling. 

So, if I want to generate 100 pauli twirled circuits, I have to splice each of the circuits to make sure the noise is inserted in the correct spot and the pauli twirling gates are also inserted accordingly. Then, the next step is to get their PTM's and sum them up.

natestemen: What if we included an image saying "this is what would happen if you generated the twirled PTM"?

docs/source/examples/pt_zne.md
purva-thakre: > From my perspective the main idea in pauli twirling is to add gates that compile to the identity,

Uh, no. Pauli twirling's goal is to transform some noise channel into a depolarizing noise channel. The circuits average over to a mixed noise channel which is what I was trying to show.

docs/source/examples/pt_zne.md
purva-thakre: I have added 3 examples. 

Ideally, I would prefer this to be a gif to show how the heatmap changes when as the number of pauli twirled circuits is increased. We could keep this as a unitary hackathon issue, if there's not a lot of non-quantum issues.

purva-thakre:
@natestemen @cosenal This is ready. Whenever you guys are free... 

Good luck with FOSDEM!

docs/source/examples/pt_zne.md
```diff
@@ -0,0 +1,400 @@
...
+noisy_twirled_circuits = []
+
+for circ in twirled_circuits:
+    split_circuit = Circuit(circ[0], circ[1], Ry(rads=np.pi/12)(q0), Ry(rads=np.pi/12)(q1), circ[-1])
+    noisy_twirled_circuits.append(split_circuit)
+
+print("Example noisy twirled circuit", noisy_twirled_circuits[-1], sep="\n")
+```
+
+The twirled PTM is averaged over each noisy twirled circuit such that the new PTM is close to that of the PTM of incoherent noise. We skip the step in this section as we require a very large number of twirled circuits to demonstrate the desired effect of averaging over multiple numpy arrays. The variations in pauli twirled PTMs are shown below when averaged over a different number of pauli twirled circuits: [3](../img/pt_zne_3_circuits.png), [5](../img/pt_zne_5_circuits.png), [30](../img/pt_zne_30_circuits.png) and [300](../img/pt_zne_100_circuits.png).
```
purva-thakre, line 238: IDK if there's a better way to provide a link to a couple of png images. These appear as images a user will have to download. 

image

docs/source/examples/pt_zne.md
natestemen: > Pauli twirling's goal is to transform some noise channel into a local depolarizing noise channel.

Right, yes, good point. That's the main goal. When adding gates surrounding each CNOT, you do it in such a way so that the CNOT is still the logical operation (up to global phase). In that sense are we not adding gates that end up compiling **out** (should you compile, which, you shouldn't)?

docs/source/examples/pt_zne.md
```diff
@@ -0,0 +1,400 @@
...
+
+The known fault tolerant thresholds for stochastic noise are higher than coherent noise which makes the former a 'preferable' type of noise compared to the latter. To avoid dealing with coherent noise, Pauli twirling can be used to tailor coherent noise to incoherent noise. Same as Eq {math:numref}`depolarizing_noise_paulis`, when a coherent noise channel is Pauli twirled, the noise channel can be described using Paulis after averaging over multiple Pauli twirled circuits. Refer to the [Pauli Twirling user guide](../guide/pt.md) for additional information. 
+
+It is worth noting that the number of Pauli twirled circuits required to transform coherent noise to incoherent noise differs
+across the circuit used, noise stength, etc. The higher the number of generated twirled circuits, the better the result. Similarly, we get better results from Pauli twirling when the coherent noise strength is closer to the small error limit.
```
natestemen, line 89:
```suggestion
It is worth noting that the number of Pauli twirled circuits required to transform coherent noise to incoherent noise depends on the circuit used, noise stength, etc.
Generally, the higher the number of generated twirled circuits, the better the result.
Similarly, better results are obtained more quickly when the coherent noise strength is low.
```

docs/source/examples/pt_zne.md
```diff
@@ -0,0 +1,400 @@
...
+there are additional sources of errors to deal with when coherent noise is acting on the CNOT gate. These can be reduced or tailored to be close to how the incoherent noise PTM appears through Pauli Twirling.
+
+```{code-cell} ipython3
+from mitiq.pt import generate_pauli_twirl_variants
+
+# Generate twirled circuits
+NUM_TWIRLED_VARIANTS = 3
+twirled_circuits = generate_pauli_twirl_variants(
+    circuit, num_circuits=NUM_TWIRLED_VARIANTS)
+print("Example ideal twirled circuit", twirled_circuits[-1], sep="\n")
```
natestemen, line 224:
```suggestion

NUM_TWIRLED_VARIANTS = 3

twirled_circuits = generate_pauli_twirl_variants(
    circuit,
    num_circuits=NUM_TWIRLED_VARIANTS,
)
print("Example ideal twirled circuit", twirled_circuits[-1], sep="\n")
```

docs/source/examples/pt_zne.md
```diff
@@ -0,0 +1,400 @@
...
+noisy_twirled_circuits = []
+
+for circ in twirled_circuits:
+    split_circuit = Circuit(circ[0], circ[1], Ry(rads=np.pi/12)(q0), Ry(rads=np.pi/12)(q1), circ[-1])
+    noisy_twirled_circuits.append(split_circuit)
+
+print("Example noisy twirled circuit", noisy_twirled_circuits[-1], sep="\n")
+```
+
+The twirled PTM is averaged over each noisy twirled circuit such that the new PTM is close to that of the PTM of incoherent noise. We skip the step in this section as we require a very large number of twirled circuits to demonstrate the desired effect of averaging over multiple numpy arrays. The variations in pauli twirled PTMs are shown below when averaged over a different number of pauli twirled circuits: [3](../img/pt_zne_3_circuits.png), [5](../img/pt_zne_5_circuits.png), [30](../img/pt_zne_30_circuits.png) and [100](../img/pt_zne_100_circuits.png).
```
natestemen, line 238: IMO no need to have the image links given the beautiful GIF is there!

docs/source/examples/pt_zne.md
```diff
@@ -0,0 +1,400 @@
...
+pt_vals = Executor(partial(execute, noise_level=NOISE_LEVEL)).evaluate(twirled_circuits)
+twirled_result = np.average(pt_vals)
+
+
+print(f"Error without twirling: {abs(ideal_value - noisy_value) :.3}")
+print(f"Error with twirling: {abs(ideal_value - twirled_result) :.3}")
+```
+
+```{caution}
+It is worth noting that Pauli twirling's goal is to tailor the noise from coherent to incoherent. Depending on the noise strength, type of coherent noise etc. this transformation might not give better results after the Pauli twirled circuit is executed. See the plot in the next section for more information.
```
natestemen, line 317:
```suggestion
It is worth noting that Pauli twirling's goal is to tailor the noise from coherent to incoherent.
Depending on the noise strength, type of coherent noise etc., this transformation alone might not give better results after the Pauli twirled circuit is executed.
```

docs/source/examples/pt_zne.md
purva-thakre: I am confused about the 'compile to identity' phrase. 

If we start out with a CNOT subjected to a coherent noise channel, after Pauli twirling it's a CNOT gate with incoherent noise.

docs/source/examples/pt_zne.md
```diff
@@ -0,0 +1,400 @@
...
+## Combining Pauli Twirling with ZNE
+
+```{code-cell} ipython3
+from mitiq.zne import execute_with_zne
+
+executor=partial(execute, noise_level=NOISE_LEVEL)
+zne_pt_vals = []
+
+for i in twirled_circuits:
+    zne_pt_vals.append(execute_with_zne(i, executor))
```
natestemen, line 329:
```suggestion
for twirled_circuit in twirled_circuits:
    zne_pt_vals.append(execute_with_zne(twirled_circuit, executor))
```

docs/source/examples/pt_zne.md
```diff
@@ -0,0 +1,400 @@
...
+
+```{warning}
+You can get better results if you control the number of samples in `noise_strength` in addition to using a higher number for
+`NUM_TWIRLED_VARIANTS`. We have chosen to not do so to reduce execution time for this tutorial.
+```
+
+## Conclusion
+
+In this tutorial, we've shown how to use a noise tailoring method with Zero-Noise Extrapolation.
+If you're interested in finding out more about these techniques, check out their respective sections of the users guide: [ZNE](../guide/zne.md), [Pauli Twilring](../guide/pt.md).
```
natestemen, line 400:
```suggestion
If you're interested in finding out more about these techniques, check out their respective sections of the users guide: [ZNE](../guide/zne.md), [Pauli Twirling](../guide/pt.md).

```

natestemen (approved):
Ready to go once my comments are addressed (and Alessandro's incoming).

docs/source/examples/pt_zne.md
```diff
@@ -0,0 +1,400 @@
...
+$$(depolarizing_noise_paulis)
+
+### Pauli Transfer Matrix (PTM)
+
+Let $\Lambda(\rho)$ be a $n$-qubit noise channel with $K_i$ being the corresponding Kraus operators. For a Pauli channel,
+we can use $P_i$ and $P_j$ to be the Kraus operators in Eq.{math:numref}`CPTP_map`.
+
+$$
+\Lambda(\rho) = \sum_{i=1} K_i \rho {K_i}^\dagger 
+$$(CPTP_map)
```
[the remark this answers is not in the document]
purva-thakre: I am going to ignore this suggestion as Alessandro had pointed out the CPTP map expression was disjointed and not discussed anywhere. In the context of Pauli noise channel, the kraus operators are n-qubit Paulis.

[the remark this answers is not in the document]
natestemen: Does that mean you're removing this equation?

docs/source/examples/pt_zne.md
natestemen: > I am confused about the 'compile to identity' phrase.

Yeah that was sloppy phrasing on my part, sorry! I think we agree that the effect of pauli twirling on an ideal CNOT is the identity map (PT(ideal CNOT) = ideal CNOT).

> Pauli twirling is equivalent to a local depoalrizing channel.

Is that true in all cases? I think Pauli twirling (in the infinite samples limit) only guarantees a Pauli channel is spit out, right?

BTW thanks for going back and forth on this a few times. It's really helpful for my understanding!

docs/source/examples/pt_zne.md
purva-thakre: >  See the plot in the next section for more information.

Why remove this line?

docs/source/examples/pt_zne.md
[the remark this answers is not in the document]
purva-thakre: No, changed my mind. I kept your suggestion and edited the `Pauli operators` to n-qubit Paulis as a segue to the next line about their definition.

docs/source/examples/pt_zne.md
natestemen: I'm not sure I meant to, but can we change it to

> See the plot in the next section for an example.

docs/source/examples/pt_zne.md
purva-thakre: > I think Pauli twirling (in the infinite samples limit) only guarantees a Pauli channel is spit out, right?

Depends on the noise strength, the number of Pauli twirled circuits etc. If you look at the gif example, I ended up at a finite number of pauli twirled circuits that gave us the ideal CNOT PTM. I'll edit the lines to discuss this. 

> BTW thanks for going back and forth on this a few times. It's really helpful for my understanding!

No worries!

docs/source/examples/pt_zne.md
```diff
@@ -0,0 +1,406 @@
...
+    mitigated_twirled_result = np.average(zne_pt_vals)
+    error_with_twirling_and_zne.append(abs(ideal_value - mitigated_twirled_result))
+
+
+plt.plot(noise_strength, error_no_zne_no_twirling,"", label=r"|Ideal - Noisy|", color="#1f77b4")
+plt.plot(noise_strength, zne_vals, "", label=r"|Ideal - ZNE|", color="#bcbd22")
+plt.plot(noise_strength, error_with_twirling,"", label=r"|Ideal - Twirling|", color="#ff7f0e")
+plt.plot(noise_strength, error_with_twirling_and_zne, "", label=r"|Ideal - (ZNE + Twirling)|", color="#2ca02c")
+
+plt.xlabel(r"Noise strength, Coherent noise:$R_y(\frac{\pi}{2} \times \text{noise_strength})$")
```
cosenal, line 386: nit: «Noise strength» as label vs. noise_strength as variable. Use the same, so it's clear that it's the actual values in the axis that parametrize the noise operation.

docs/source/examples/pt_zne.md
```diff
@@ -0,0 +1,406 @@
...
+error_no_zne_no_twirling = []
+error_with_twirling = []
+error_with_twirling_and_zne = []
+zne_vals = []
+ideal_values = []
+NUM_TWIRLED_VARIANTS = 30
+
+for strength in noise_strength:
+    ideal_value = execute(circuit, noise_level=0.0)
+    ideal_values.append(ideal_value)
```
cosenal, line 360: These go outside the for loop

docs/source/examples/pt_zne.md
```diff
@@ -0,0 +1,406 @@
...
+    twirled_noisy_diff = abs(ideal_value-twirled_result)
+    error_with_twirling.append(twirled_noisy_diff)
+
+    executor=partial(execute, noise_level=strength)
+    zne_vals.append(abs(ideal_value - execute_with_zne(circuit, executor)))
+    
+    zne_pt_vals = []
+    for i in twirled_circuits:
+        zne_pt_vals.append(execute_with_zne(i, executor))
+    mitigated_twirled_result = np.average(zne_pt_vals)
```
cosenal, line 377: This is basically the same code as in the previous section. Since it's the core part of the tutorial, at least when it comes to code, it should be in a function on its own.

docs/source/examples/pt_zne.md
```diff
@@ -0,0 +1,406 @@
...
+from mitiq import Executor
+pt_vals = Executor(partial(execute, noise_level=NOISE_LEVEL)).evaluate(twirled_circuits)
+twirled_result = np.average(pt_vals)
+
+
+print(f"Error without twirling: {abs(ideal_value - noisy_value) :.3}")
+print(f"Error with twirling: {abs(ideal_value - twirled_result) :.3}")
+```
+
+```{caution}
```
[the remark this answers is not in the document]
purva-thakre: This code block is only demonstrating how to get an execute the circuit in addition to utilizing Pauli twirling. 

> then warning the reader that they shouldn't pay attention to them is odd.

Due to the values I used, we are making a point to state that there are cases when Pauli twirling on its own is not too helpful.

docs/source/examples/pt_zne.md
purva-thakre: Not really. These lines are making sure the size of `ideal_values` is the same as the size of the other values we are comparing it to.

purva-thakre: Never mind. I changed this.

docs/source/examples/pt_zne.md
purva-thakre: I don't really get what you are trying to suggest here. Do you want me to create a function that compares the expectation values?

docs/source/examples/pt_zne.md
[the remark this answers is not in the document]
cosenal: Maybe you can make the same point in a more gentle way for the reader, as opposed to using the «caution» box.

docs/source/examples/pt_zne.md
cosenal: What I am suggesting here is that you could define a routine to encapsulate the code for ZNE + PT. Something on the lines of:
```python
def zne_with_pt(circuit, executor, num_twirled_variants):
    zne_pt_vals = []
    for twirled_circuit in generate_pauli_twirl_variants(circuit, num_circuits=num_twirled_variants):
        zne_pt_vals.append(execute_with_zne(twirled_circuit, executor))
    return np.average(zne_pt_vals)
```
And you can use it both in L379 and L333. This would highlight the steps that are involved in composing the two techniques.

docs/source/examples/pt_zne.md
```diff
@@ -0,0 +1,406 @@
...
+$\frac{I}{2}$ is the maximally mixed state which can be described using Paulis.
+
+$$
+\frac{I}{2} = \frac{1}{4} (\rho + X \rho X + Y \rho Y + Z \rho Z)
+$$
+
+Thus, the depolarizing channel can be redescribed using Paulis as shown below. 
+
+$$
+\mathcal{E}(\rho) = (1-\frac{3p}{4}) \rho + \frac{p}{4} (X \rho X + Y \rho Y + Z \rho Z)
```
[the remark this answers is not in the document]
purva-thakre: I think there was an earlier conversation in this PR where either you or Nate suggested providing more details on why coherent noise scales quadratically. 

Which is why I decided to add some information about $r(\mathcal{E})$ and the relationship to average gate infidelity.

docs/source/examples/pt_zne.md
purva-thakre: Ah, ok. I get what you are trying to say now. 

Previously, I thought you wanted me to define a function to make it easy to get all the expectation values in the `for` loop before the plot.

purva-thakre:
@natestemen I know you already approved this. Pinging you again in case you want to take a look at the changes in https://github.com/unitaryfund/mitiq/pull/2601/commits/aeb9d8f89321842f7a07f4c8a9bf98295608b697

I added a new sub-section to add more information about the quadratic and linear error rates.

docs/source/examples/pt_zne.md
```diff
@@ -0,0 +1,449 @@
...
+
+## Combining Pauli Twirling with ZNE
+
+```{code-cell} ipython3
+from mitiq.zne import execute_with_zne
+from typing import List
+
+executor=partial(execute, noise_level=NOISE_LEVEL)
+
+def combine_zne_pt(list_circuits, executor) -> List[float]:
```
cosenal, line 370: 🎉

cosenal (approved):
LGTM

natestemen (approved):

Issue #1896, opened 2023-07-03
ZNE test failing non-deterministically


andreamari:
The `test_with_observable_batched_factory` test is quite old but, for some reason, it seems to fail more often in the last weeks.
See e.g.:
https://github.com/unitaryfund/mitiq/actions/runs/5423886272/jobs/9862463130#step:5:3881

I would suggest to have a look under the hood  to check that everything works as expected, e.g. what circuits are executed, what results are obtained, etc.. 

If everything works, we should probably modify this test to make it deterministic.

purva-thakre:
The linked logs have expired.

cosenal:
Copying over @purva-thakre's report from #2442, which has got non-stale logs.

---
Noticed the failure here: https://github.com/unitaryfund/mitiq/actions/runs/9814019705/job/27101033866#step:6:4363

If I rerun the workflow, the test passes. 

![image](https://github.com/unitaryfund/mitiq/assets/66048318/328732a8-e757-4f5c-bc3d-f6516735f688)

https://github.com/unitaryfund/mitiq/blob/e61c058777b11cc34b81be05e78b1c4bc91d7775/mitiq/zne/tests/test_zne.py#L81-L101

https://github.com/unitaryfund/mitiq/blob/e61c058777b11cc34b81be05e78b1c4bc91d7775/mitiq/interface/mitiq_cirq/cirq_utils.py#L17

Shivansh20128:
Hey! This error resurfaced again [here](https://github.com/unitaryfund/mitiq/actions/runs/11798888714/job/32866119137#step:6:4421).

PR #2602, opened 2024-12-13
Fix flaky ZNE factory + observable test

natestemen:
## Description

Replaces a test that would fail occasionally with a deterministic assertion based on mocking the `BatchedFactory` object. I did keep one assertion regarding the actual value, but I'm pretty sure it's also deterministic.

fixes https://github.com/unitaryfund/mitiq/issues/1896

cc @Shivansh20128 since you ran into this recently.

### DIFF
--- a/mitiq/zne/tests/test_zne.py
+++ b/mitiq/zne/tests/test_zne.py
@@ -94,21 +94,23 @@ def executor(circuit) -> float:
 def test_with_observable_batched_factory(executor):
     observable = Observable(PauliString(spec="Z"))
     circuit = cirq.Circuit(cirq.H.on(cirq.LineQubit(0))) * 20
+    executor = functools.partial(
+        sample_bitstrings, noise_model_function=cirq.depolarize
+    )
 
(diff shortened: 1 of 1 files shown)


mitiq/zne/tests/test_zne.py
```diff
@@ -94,21 +94,23 @@ def executor(circuit) -> float:
 def test_with_observable_batched_factory(executor):
```
cosenal, line 94: you are not using the `executor` argument anywhere in the test, which means you are not using all the parameter values 🤔

cosenal (changes requested):

mitiq/zne/tests/test_zne.py
natestemen: Good catch! That was left over from testing and never got changed back. Fixed in https://github.com/unitaryfund/mitiq/pull/2602/commits/52765b73bd60da44f267d6fa2c96197841a10aa2.

cosenal (approved):
👍

PR #2604, opened 2024-12-13
Modularize PEC functionality

natestemen:
## Description

This PR introduces two functions

- `intermediary_sampled_circuits`, and
- `combine_results`

for generating PEC sampled circuits, and combining results from having run those circuits. The first function is used to simplify the code in `execute_with_pec`, but the second function cannot be used due to the ability to output PEC data via the `full_output` flag.

It also "simplifies" some of the PEC tests to move testing to the new `intermediary_sampled_circuits` function.

fixes https://github.com/unitaryfund/mitiq/issues/2563

### DIFF
--- a/mitiq/pec/pec.py
+++ b/mitiq/pec/pec.py
@@ -13,13 +13,11 @@ from typing import (
     Dict,
     Iterable,
     List,
-    Literal,
     Optional,
     Sequence,
     Tuple,
     Union,
     cast,
-    overload,
 )
 
 import numpy as np
@@ -40,32 +38,13 @@ _LARGE_SAMPLE_WARN = (
 )
 
 
-@overload
-def intermediary_sampled_circuits(
+def generate_sampled_circuits(
     circuit: QPROGRAM,
     representations: Sequence[OperationRepresentation],
     precision: float = 0.03,
     num_samples: int | None = None,
     random_state: int | np.random.RandomState | None = None,
-    return_signs: Literal[False] = False,
-) -> list[QPROGRAM]: ...
-@overload
-def intermediary_sampled_circuits(
-    circuit: QPROGRAM,
-    representations: Sequence[OperationRepresentation],
-    precision: float = 0.03,
-    num_samples: int | None = None,
-    random_state: int | np.random.RandomState | None = None,
-    *,
-    return_signs: Literal[True],
-) -> tuple[list[QPROGRAM], list[int], float, int]: ...
-def intermediary_sampled_circuits(
-    circuit: QPROGRAM,
-    representations: Sequence[OperationRepresentation],
-    precision: float = 0.03,
-    num_samples: int | None = None,
-    random_state: int | np.random.RandomState | None = None,
-    return_signs: bool = False,
+    full_output: bool = False,
 ) -> list[QPROGRAM] | tuple[list[QPROGRAM], list[int], float, int]:
     """Generates a list of sampled circuits based on the given
     quasi-probability representations.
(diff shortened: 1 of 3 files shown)


mitiq/pec/pec.py
```diff
@@ -37,6 +40,113 @@ class LargeSampleWarning(Warning):
...
+@overload
+def intermediary_sampled_circuits(
+    circuit: QPROGRAM,
+    representations: Sequence[OperationRepresentation],
+    precision: float = 0.03,
+    num_samples: int | None = None,
+    random_state: int | np.random.RandomState | None = None,
+    *,
+    return_signs: Literal[True],
+) -> tuple[list[QPROGRAM], list[int], float, int]: ...
```
purva-thakre, line 61:
```suggestion
```

mitiq/pec/pec.py
```diff
@@ -37,6 +40,113 @@ class LargeSampleWarning(Warning):
...
+        random_state=random_state,
+        num_samples=num_samples,
+    )
+
+    if return_signs:
+        return sampled_circuits, signs, norm, num_samples
+    return sampled_circuits
+
+
+def combine_results(
```
purva-thakre, line 127: @natestemen Wouldn't you want to use `combine_results` in `execute_with_pec` as well?

mitiq/pec/pec.py
```diff
@@ -37,6 +40,113 @@ class LargeSampleWarning(Warning):
...
+    """Combine expectation values coming from probabilistically sampled
+    circuits.
+
+    Warning:
+        The ``results`` must be in the same order as the circuits were
+        generated.
+
+    Args:
+        results: Results as obtained from running circuits.
+        norm:
```
purva-thakre, line 139: Missing description for `norm`.

mitiq/pec/pec.py
natestemen: It would be nice, but because there is `full_output` option of `execute_with_pec` it would require making `combine_results` a lot more complicated so I opted not to use this function internally.

mitiq/pec/pec.py
natestemen: Good catch. Added details in https://github.com/unitaryfund/mitiq/pull/2604/commits/f4bf4d14221412ac190fd9ef8a89da69c63cdf7e.

mitiq/pec/pec.py
purva-thakre: Maybe you should add a note in this docstring that if they want all the PEC data, they cannot use this function. 
They will have to use `execute_with_PEC` with `full_output=True`.

natestemen: The user can easily compute all of the items in `pec_data` if they have the results from each circuit run.

https://github.com/unitaryfund/mitiq/blob/7e3070fd3bc1639a6125027ccaa791bee4ac608a/mitiq/pec/pec.py#L148-L156

purva-thakre: Yes, that's what I am trying to say. It is only available through `execute_with_pec` right now. 

Maybe direct a user to these lines for `pec_data` in the docstring of `combine_results` or when you update the user guide, show a user how to get this data by using the modularized functions. Up to you.

purva-thakre (approved):
Nicely done with the mock tests! I always struggle with these.

mitiq/pec/pec.py
natestemen: I think the situation you're describing[^1] is slightly too niche for me to feel like is worth documenting. If they are aware of `full_output` it is also described in some detail in the **Returns** section of the [`execute_with_pec`](https://mitiq.readthedocs.io/en/stable/apidoc.html#mitiq.pec.pec.execute_with_pec) docs.

[^1]: 1. User is aware of the `full_output=True` option to `execute_with_pec`
    2. User decides to use circuit generation/combination workflow
    3. User wants to compute items in `full_output`, but doesn't know how.

PR #2608, opened 2024-12-20
LRE failure for chunking

purva-thakre:
## Description

Noticed in #2601 

https://github.com/unitaryfund/mitiq/actions/runs/12433520415/job/34715234002?pr=2601#step:6:4591

Originally, the test was written to confirm chunking a circuit performs worse in certain scenarios. However, this performance varies non-deterministically. Due to this reason, now the test is verifying there is a difference between two expectation values. 

---

### License

### DIFF
--- a/mitiq/lre/tests/test_lre.py
+++ b/mitiq/lre/tests/test_lre.py
@@ -113,7 +113,7 @@ def test_lre_decorator_raised_error():
 
 def test_lre_executor_with_chunking():
     """Verify the executor works as expected for chunking a large circuit into
-    a smaller circuit. Note that this does not verify whether the chunked 
+    a smaller circuit. Note that this does not verify whether the chunked
     circuit gets better results compared to a non-chunked circuit."""
     # define a larger circuit
     test_cirq = benchmarks.generate_rb_circuits(n_qubits=1, num_cliffords=12)[
@@ -123,8 +123,7 @@ def test_lre_executor_with_chunking():
         test_cirq, execute, degree=2, fold_multiplier=2, num_chunks=14
     )
     # verify we get an expectation value
-    assert(lre_exp_val_chunking) > 0
-    
+    assert (lre_exp_val_chunking) > 0
 
 
 @pytest.mark.parametrize("input_method", [(fold_global), (fold_all)])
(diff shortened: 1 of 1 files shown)


purva-thakre:
Converting this back to draft because @natestemen suggested to write a better test.

mitiq/lre/tests/test_lre.py
```diff
@@ -113,15 +113,18 @@ def execute(circuit, noise_level=0.025):
...
         0
     ]
-    lre_exp_val = execute_with_lre(
+    lre_exp_val_chunking = execute_with_lre(
         test_cirq, execute, degree=2, fold_multiplier=2, num_chunks=14
     )
-    assert abs(lre_exp_val - ideal_val) <= abs(noisy_val - ideal_val)
+    # verify we get an expectation value
+    assert(lre_exp_val_chunking) > 0
+
```
purva-thakre, line 127: @natestemen I cannot think of a better test here. 

`test_layerwise_folding` already has a test comparing the number of noise-scaled layers for a chunked and a non-chunked circuit.

natestemen (commented):
I would argue that this test is unnecessary.

We already have tests that ensure that chunking works as expected ([here](https://github.com/unitaryfund/mitiq/blob/4705a43e8ffac59e4229942b5eeb8eb95a899ae2/mitiq/lre/tests/test_layerwise_folding.py#L107), and [here](https://github.com/unitaryfund/mitiq/blob/4705a43e8ffac59e4229942b5eeb8eb95a899ae2/mitiq/lre/tests/test_layerwise_folding.py#L260)). The only thing this test could be doing (aside from asserting accuracy) is testing the number of circuits run is as expected, but this would be redundant from the above linked tests.

Should we remove it. WDYT?

purva-thakre:
This test is verifying the executor works with chunking without any issues. I prefer to keep it.

purva-thakre:
that the lre executor takes `num_chunks` without any issues. I don't think this is tested elsewhere.

natestemen:
Ah okay that's right! In that case, I think it should be sufficient to mock the executor, and ensure that it is run the correct number of times.

purva-thakre:
> I think it should be sufficient to mock the executor, and ensure that it is run the correct number of times.

Done!

natestemen (approved):
🚢

PR #2610, opened 2024-12-20
Modularize DDD functionality

### LINKED ISSUE #2562: Modularize DDD
Similar to the work done in #2452, we want to create an API where a user has access to intermediary circuits and can control the inference process.

We also want to update the documentation such that the two-stage API functions are demonstrated in the user guide.

bdg221:
## Description
---
This PR introduces two functions

- `generate_circuits_with_ddd`
- `generate_ddd_value`

for generating circuits with DDD sequences and using the results from running those circuits to get an expectation value. The first function moves generating the circuits outside of `execute_with_ddd`, allowing users to generate and return DDD circuits.

The second function takes the results from all of the DDD circuits used in `execute_with_ddd` and averages the results to produce a single expectation value (float) that is returned.

fixes #2562

### DIFF
--- a/mitiq/ddd/tests/test_ddd.py
+++ b/mitiq/ddd/tests/test_ddd.py
@@ -233,7 +233,7 @@ def test_ddd_decorator_with_rule_args():
     assert not np.isclose(mitigated_large_spacing, mitigated_small_spacing)
 
 
-@mark.parametrize("num_trials", [1, 3, 5])
+@mark.parametrize("num_trials", [1, 10, 20, 30])
 def test_num_trials_generates_circuits(num_trials: int):
     """Test that the number of generated circuits follows num_trials."""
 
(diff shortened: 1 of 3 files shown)


bdg221:
@natestemen Do you think the second function `generate_ddd_value` is necessary since it is just calling `np.average`?

bdg221:
On the community call, it was agreed that the second function is important for consistency with ZNE and PEC. Also, I changed `generate_ddd_value` to `combine_results` for consistency.

mitiq/ddd/tests/test_ddd.py
```diff
@@ -226,3 +231,14 @@ def exec_xx_small_spacing(circuit):
     # What is important to test is getting different results.
     assert not np.isclose(unmitigated, mitigated_small_spacing)
     assert not np.isclose(mitigated_large_spacing, mitigated_small_spacing)
+
+
+@mark.parametrize("num_trials", [1, 3, 5])
```
natestemen, line 236: Since this test should be fast (correct me if i'm wrong) can we add a few more trials? Say up to 30?

natestemen (approved):
LGTM! Left one comment, and my only other hold-up was thinking `generate_circuits_with_ddd` is a bit of a mouthful, but I can't think of a better name!

natestemen:
Going to merge before CI passes. I ran the new test locally and it looks good!

PR #2649, opened 2025-01-23
First PR for VD, helper function that copies a circuit M times

chrispy-chicken:
### Description

- To implement virtual distillation in mitiq, first we need a couple helper functions. This function is the first one and copies a circuit M times in a way that they are stacked on top of eachother. For example, with M=2, a N qubit circuit will return a 2N qubit circuit. This PR also has a vd_tests.py file ready for the tests that are coming with the full virtual distillation implementation.
---

### License


AUTHORS
```diff
@@ -60,4 +60,5 @@ Francesc Sabater
 Emiliano Godinez
 Tommy Nguyen
 Duong H. D. Tran
-Mudassir Chapra
\ No newline at end of file
+Mudassir Chapra
+chrispy-chicken
```
natestemen, line 64: Do you want to add everyone's name/handle from the project?

mitiq/vd/vd.py
natestemen, line 1: 1. Lets move these functions into a file `mitiq/vd/utils.py` since they are not the "main" functions for this module.
    1. when you do this, you should rename the test file to `mitiq/vd/tests/test_vd_utils.py`
3. Any function which we don't intend users to use (e.g. this one), we indicate this by beginning the function with an underscore `_`. This is a python standard way to denote something is private.

mitiq/vd/vd.py
```diff
@@ -0,0 +1,60 @@
...
+    circuit = cirq.Circuit()
+    qubits = cirq.LineQubit.range(N*M)
+
+    for i in range(M):
+        circuit += rho.transform_qubits(lambda q: qubits[q.x + N*i])
+
+    return circuit
+
+# add unit tests
+def test_M_copies_of_rho():
```
natestemen, line 35: 1. This test should be moved to the test file.
2. We should add a separate test to ensure each circuit that comes out of this function has the intended number of qubits.

mitiq/vd/vd.py
```diff
@@ -0,0 +1,60 @@
...
+    #     print("warning: M_copies_of_rho is not needed for M <= 1")
+    #     return rho
+
+    N = len(rho.all_qubits())
+
+    circuit = cirq.Circuit()
+    qubits = cirq.LineQubit.range(N*M)
+
+    for i in range(M):
+        circuit += rho.transform_qubits(lambda q: qubits[q.x + N*i])
```
natestemen, line 30: I believe this line will only work when using `cirq.LineQubit`s. E.g. if someone defines a circuit with `cirq.GridQubit`s then there are no `x` attributes.

Do you know any other ways of writing this function? If there are no simple ways we can go forward and make a note to come back later.

mitiq/vd/vd.py
```diff
@@ -0,0 +1,60 @@
...
+def test_M_copies_of_rho():
+     
+    M = 2
+    qubits = cirq.LineQubit.range(2)
+    circuit = cirq.Circuit(cirq.H(qubits[0]), cirq.CNOT(qubits[0], qubits[1]))
+    new_circuit = M_copies_of_rho(circuit, M)
+
+    expected_qubits = cirq.LineQubit.range(2*M)
+    expected_circuit = cirq.Circuit(cirq.H(expected_qubits[0]), cirq.CNOT(expected_qubits[0], expected_qubits[1]))
+    expected_circuit += cirq.Circuit(cirq.H(expected_qubits[2]), cirq.CNOT(expected_qubits[2], expected_qubits[3]))
```
natestemen, line 44:
```suggestion
    expected_circuit = cirq.Circuit(
        cirq.H(expected_qubits[0]),
        cirq.CNOT(expected_qubits[0], expected_qubits[1]),
        cirq.H(expected_qubits[2]),
        cirq.CNOT(expected_qubits[2], expected_qubits[3]),
    )
```

Defining the circuit as a single call makes most sense. If you agree, this should be changed below as well.

mitiq/vd/vd.py
```diff
@@ -0,0 +1,60 @@
...
+    
+    assert len(new_circuit) == 4
```
natestemen, line 46:
```suggestion
    assert len(circuit) == len(new_circuit)
```

I believe the new and old circuits should have the same length, right?

natestemen (changes requested):
Great work! I've left some comments for you to review and make some changes, but this is a good start.

mitiq/vd/vd.py
chrispy-chicken: yeah and i guess then the assert below that fully checks equality already does that

mitiq/vd/vd.py
chrispy-chicken: it is then that the == operator fails to see that both the circuits are equal, due to some weird way the gates are applied. I now compare unitary matrices

mitiq/vd/vd.py
chrispy-chicken: fixed by working with unitary matrices. Output becomes a bit more unreadable, but the user should not work with this function. We can also revert back and look for another solution if necessary

mitiq/vd/vd_utils.py
```diff
@@ -0,0 +1,34 @@
...
+
+    N = len(rho.all_qubits())
+
+    circuit = cirq.Circuit()
+    qubits = cirq.LineQubit.range(N*M)
+
+    unitary_gate = cirq.MatrixGate(cirq.unitary(rho))
+    for i in range(M):
+        qubit_subset = qubits[i * N : (i+1) * N]
+        circuit.append(unitary_gate(*qubit_subset))
```
natestemen, line 32: Sorry to ask to change this again, but I think best to keep the same circuit structure as what the user passed. A user may have already compiled their circuit when sending it through Mitiq, so we want to keep as much structure as we can.

For now lets go back to the previous implementation.

natestemen (commented):
Nice progress!

Reminder to make sure the formatting and `mypy` are happy before you push to your branch. This will make the continuous integration (CI) go green to indicate the code is formatted correctly and the typehinting makes sense.

If you need help resolving any of the `mypy` issues just let me know and I can help sort them out. It's confusing at first.

chrispy-chicken:
Done!

mitiq/vd/vd_utils.py
```diff
@@ -0,0 +1,51 @@
...
+
+    Given a circuit rho that acts on N qubits,
+    this function returns a circuit that copies rho M times in parallel.
+    This means the resulting circuit has N * M qubits.
+
+    Args:
+        rho:
+            The circuit to be copied.
+        M:
+            The number of copies of rho to be made.
```
natestemen, line 19: These arguments need to renamed to reflect the current function arguments. Also if you can remove mentions of $\rho$ I think that's best.

natestemen (approved):
Nice solution accepting both line and grid qubits.

I left one comment on the docstring, but after that it should be good to merge.

mitiq/vd/vd_utils.py
chrispy-chicken: agreed, overlooked that one

chrispy-chicken:
everything should be ready!

natestemen (approved):
woohoo! Well done with the first PR!

PR #2655, opened 2025-01-31
Create `scale_amplifications.py` for Probabilistic Error Amplification (PEA)

### LINKED ISSUE #2636: Create `scale_representations.py` for Probabilistic Error Amplification
## Issue Description

The first of the two top-level PEA functions, scale_circuit_represntations, generates the representations of the circuit parameterized by the noise level and a set of scale factors. The scaled circuit representations are noise-amplified quasiprobabilisty representations of the circuit at different noise scale factors.

Misty-W:
## Description

Fixes #2636.

This PR contains functions for the first part of the probabilistic error amplification (PEA) technique. In PEA, error amplification (gate) sequences are applied probabilistically to scale the noise level of the computation. PEA was proposed and applied in the paper [Evidence for the utility of quantum computing before fault tolerance | Nature](https://www.nature.com/articles/s41586-023-06096-3).

In  the `construct_circuits()` function in `pea.py`, different implementable circuits from the noise-amplified representation of the input circuit, corresponding to each of the input noise `scale_factors`, are sampled.
The noise-amplified circuits are constructed according to a set of scale factors (inspired by ZNE), noise model (currently local and global depolarizing noise models are supported), and noise level.
 
The above circuit construction step is in preparation for the later steps in the PEA workflow (in draft PR #2785), i.e. evaluating the noisy expectation values associated to the sampled circuits through the `executor`, estimating the ideal expectation value from a suitable linear combination of the noisy ones at each noise scale factor, and extrapolating the expectation values obtained at each noise scale factor to the zero noise limit .

---

### License


Misty-W:
> @Misty-W what is the status here? Is docs the only missing part here?

@natestemen, since this is only half of the PEA technique, I'd say we can add docs in a separate PR.
I added the missing tests and updated the docstrings, so I've marked the PR for review.

mitiq/pea/sample_amplifications.py
```diff
@@ -0,0 +1,126 @@
...
+def sample_circuit_amplifications(
+    ideal_circuit: Circuit,
+    executor: Union[Executor, Callable[[QPROGRAM], QuantumResult]],
+    scale_factors: List[float],
+    noise_model: str,
+    epsilon: float,
+    observable: Observable | None = None,
+    num_samples: Optional[int] = None,
+    force_run_all: bool = True,
+) -> List[float]:
```
natestemen, line 61: I'm a little confused by this function. It seems like it's orchestrating the everything except for the extrapolation.

Do we want to have this function given most of our techniques are available through either

1. `execute_with_xyz`, or
2. the two-step application of the technique with `construct_circuits` and `combine_results`?

I would think we would want to follow the same pattern here, and I'm not sure how this function fits into that pattern.

mitiq/pea/sample_amplifications.py
```diff
@@ -0,0 +1,126 @@
...
+from mitiq.pec import OperationRepresentation
+from mitiq.pec.sampling import sample_circuit
+
+
+def scale_circuit_amplifications(
+    ideal_circuit: Circuit,
+    scale_factors: float,
+    noise_model: str,
+    epsilon: float,
+) -> Sequence[OperationRepresentation]:
```
natestemen, line 21: Making sure I understand here. This function should generate sampled circuits that implement noise-scaled circuits at the given `scale_factors`. From the [PEA RFC](https://docs.google.com/document/d/1l-74EFdMA0CSFUpHjqCyQYb3ZKCmY77seB1_mOZo5Co/edit?tab=t.0#heading=h.4wr9qx2m833b) this part of the diagram:
Screenshot 2025-05-20 at 8 49 10 PM

I'd expect in this function an argument to say how many circuits to sample at each scale factor. Is this where that would be added?

Also, the docstring indicates that this function returns circuits, but the typehint says otherwise.

Finally, I think this function should be the `construct_circuits` function of the `mitiq.pea` module. Or is there a different one you had in mind?

natestemen (commented):
Really great work Misty! This first review is mostly an architecture review, so I haven't had a look at all the code in detail, but more the structure.

I can't wait to test this out!

Misty-W:
Note: although `codecov/project` is failing, according to the latest Codecov report  https://github.com/unitaryfoundation/mitiq/pull/2655#issuecomment-2628540494: "all modified and coverable lines are covered by tests."

natestemen:
I don't think I can this week :/. Maybe one of @cosenal @vprusso or @FarLab can?

Otherwise, I'll plan to take a look early next week.

Misty-W:
> Might be good to add a PR description to add some context to the PR in case someone else wants to review.

Good catch, I'll do that.

mitiq/pea/amplifications/amplify_depolarizing.py
```diff
@@ -0,0 +1,295 @@
+# Copyright (C) Unitary Fund
```
natestemen:
```suggestion
# Copyright (C) Unitary Foundation
```

mitiq/pea/amplifications/amplify_depolarizing.py
```diff
@@ -0,0 +1,295 @@
...
+#
+# This source code is licensed under the GPL license (v3) found in the
+# LICENSE file in the root directory of this source tree.
+
+"""Functions related to amplifications with depolarizing noise."""
+
+import copy
+from itertools import product
+from typing import List
```
natestemen: We can use the built-int `list` constructor since we are on python 3.10+.

mitiq/pea/pea.py
```diff
@@ -0,0 +1,110 @@
...
+    LargeSampleWarning,
+    sample_circuit,
+)
+from mitiq.typing import QPROGRAM
+
+
+def construct_circuits(
+    circuit: Circuit,
+    scale_factors: List[float],
+    noise_model: str,
```
natestemen, line 25: Maybe this is a later addition, but should we allow someone to pass a list of representations here if they have them?

mitiq/pea/pea.py
```diff
@@ -0,0 +1,110 @@
...
+    epsilon: float,
+    random_state: Optional[np.random.RandomState] = None,
+    precision: float = 0.1,
+    num_samples: Optional[int] = None,
+    full_output: bool = True,
+) -> (
+    List[List[QPROGRAM]]
+    | tuple[List[List[QPROGRAM]], List[List[int]], List[float]]
```
natestemen: If we wanted to maintain as much consistence across the `mitiq.xyz.construct` circuit functions, then the circuits should come as `list[QPROGRAM]`. It makes sense what you've done here grouping the circuits for each scale factor into its own list, but I guess it would be possible to flatten them into a single list. Did you consider this at all? Maybe there's something I'm missing that would make this impossible or really cumbersome.

Regardless of what we do here, I think this only support `cirq.Circuit` objects, right? If so the typehint should be updated.

mitiq/pea/pea.py
```diff
@@ -0,0 +1,110 @@
...
+)
+from mitiq.typing import QPROGRAM
+
+
+def construct_circuits(
+    circuit: Circuit,
+    scale_factors: List[float],
+    noise_model: str,
+    epsilon: float,
+    random_state: Optional[np.random.RandomState] = None,
```
natestemen:
```suggestion
    random_state: np.random.RandomState | int | None = None,
```

Looks like `int` is accepted in the code.

mitiq/pea/scale_amplifications.py
```diff
@@ -0,0 +1,54 @@
+# Copyright (C) Unitary Foundation
+#
+# This source code is licensed under the GPL license (v3) found in the
+# LICENSE file in the root directory of this source tree.
+
+"""Tools for construting the noise-amplified representations of ideal
+operations.
+"""
+
+from typing import Sequence
```
natestemen: Since https://github.com/unitaryfoundation/mitiq/pull/2775 was merged, we are importing type-hints from `collections.abc` where possible since many `typing` objects are being deprecated.

mitiq/pea/tests/test_pea.py
```diff
@@ -0,0 +1,146 @@
...
+from mitiq.pec import (
+    OperationRepresentation,
+)
+from mitiq.pec.pec import LargeSampleWarning
+
+
+# Noisy representations of Pauli and CNOT operations for testing.
+def get_pauli_and_cnot_representations(
+    base_noise: float,
+    qubits: Optional[List[cirq.Qid]] = None,
```
natestemen: This second argument is never used. I guess we can remove.

mitiq/pea/amplifications/amplify_depolarizing.py
```diff
@@ -0,0 +1,295 @@
...
+        ideal_circuit: The ideal circuit, whose ideal operations should be
+            represented.
+        noise_level: The (gate-independent) depolarizing noise level.
+
+    Returns:
+        The list of quasi-probability amplifications associated to
+        the operations of the input ``ideal_circuit``.
+
+    .. note::
+        Measurement gates are ignored (not represented).
```
natestemen:
```suggestion
    .. note::
        Measurement gates are ignored.
```

mitiq/pea/amplifications/amplify_depolarizing.py
```diff
@@ -0,0 +1,295 @@
...
+
+    .. note::
```
natestemen:
```suggestion
    .. warning::
```

mitiq/pea/amplifications/amplify_depolarizing.py
```diff
@@ -0,0 +1,295 @@
...
+    for each of them, generates the corresponding quasi-probability
+    amplification (linear combination of implementable noisy operations).
+
+    This function assumes that the tensor product of ``k`` single-qubit
+    depolarizing channels affects each implemented operation, where
+    ``k`` is the number of qubits associated to the operation.
+
+    This function internally calls
+    :func:`amplify_operation_with_local_depolarizing_noise` (more details
+    about the quasi-probability amplification can be found in its docstring).
```
natestemen: Unnecessary IMO.

mitiq/pea/tests/test_pea.py
```diff
@@ -0,0 +1,146 @@
...
+    # we expect num_samples = 1/precision^2:
+    assert np.allclose(
+        [precision**2 * len(c) for c in scaled_circuits],
+        [1] * len(scaled_circuits),
+        atol=0.2,
+    )
+
+
+def test_precision_ignored_when_num_samples_present():
+    """Check precision is ignored when num_samples is given."""
```
natestemen:
```suggestion
```

mitiq/pea/tests/test_pea.py
```diff
@@ -0,0 +1,146 @@
...
+        construct_circuits(
+            oneq_circ,
+            scale_factors=[1],
+            noise_model="global_depolarizing",
+            epsilon=0.02,
+            num_samples=100_001,
+        )
+
+
+def test_full_output():
```
natestemen: In theory shouldn't this be caught by `mypy`?

mitiq/pea/tests/test_pea.py
natestemen, line 1: One extra test request: Ensure that `sample_circuit` and `scale_circuit_amplifications` are called the appropriate number of times when `construct_circuits` is called.

mitiq/pea/pea.py
```diff
@@ -0,0 +1,110 @@
...
+
+    Args:
+        circuit: The ideal circuit from which an implementable
+            sequence is sampled.
+        scale_factors: A list of (positive) numbers by which the baseline
+            noise level is to be amplified.
+        noise_model: A string describing the noise model to be used for the
+            noise-scaled representations, e.g. "local_depolarizing" or
+            "global_depolarizing".
+        epsilon: Baseline noise level.
```
natestemen, line 51: I'm kind of confused by this argument. Can you expand on the argument description here? Is it the (average) local/global depolarizing noise rate on a device? I might have a misunderstanding of PEA, but is this parameter required?

In the IBM paper they talk about "noise gain levels"[^1]. How do those relate to epsilon?

[^1]: page 502, second column near the top

mitiq/pea/tests/test_scale_amplifications.py
```diff
@@ -0,0 +1,65 @@
...
+            "local_depolarizing",
+            amplify_noisy_ops_in_circuit_with_local_depolarizing_noise,
+        ],
+        [
+            "global_depolarizing",
+            amplify_noisy_ops_in_circuit_with_global_depolarizing_noise,
+        ],
+    ],
+)
+def test_scale_circuit_amplifications(epsilon, noise_model):
```
natestemen: Two suggestions and a question:

1. More specific test name to indicate what the function is testing (that the norm for each sampled circuit)
2. Separate the `noise_model` parameter into two (noise model string and amplification function).
3. It seems this test is fundamentally testing the amplification functions as opposed to `scale_circuit_amplifications`. Do we need to call that function here?

mitiq/pea/amplifications/amplify_depolarizing.py
```diff
@@ -0,0 +1,295 @@
...
+
+from mitiq import QPROGRAM
+from mitiq.interface.conversions import (
+    append_cirq_circuit_to_qprogram,
+    convert_to_mitiq,
+)
+from mitiq.pec.types import NoisyOperation, OperationRepresentation
+
+
+def amplify_noisy_op_with_global_depolarizing_noise(
```
natestemen, line 23:
```suggestion
@accept_any_qprogram_as_input
def amplify_noisy_op_with_global_depolarizing_noise(
```

You might've already tried this and maybe it didn't work, but it keeps the function body a bit conceptually cleaner.

mitiq/pea/amplifications/amplify_depolarizing.py
```diff
@@ -0,0 +1,295 @@
...
+    # The single-qubit case: linear combination of 1Q Paulis
+    if len(qubits) == 1:
+        q = tuple(qubits)[0]
+
+        alpha_pos = 1.0 - noise_level
+        alpha_neg = noise_level / 3
+
+        alphas = [alpha_pos] + 3 * [alpha_neg]
+        post_ops = [[]]  # for alpha_pos, we do nothing, rather than I
+        post_ops += [[P(q)] for P in [X, Y, Z]]  # 1Q Paulis
```
natestemen:
```suggestion
        post_ops += [[P(q)] for P in [X, Y, Z]]  # 1Q Paulis
```
Does this work? Unsure why we would need the definition above.

mitiq/pea/amplifications/amplify_depolarizing.py
```diff
@@ -0,0 +1,295 @@
...
+        alpha_pos = 1.0 - noise_level
+        alpha_neg = noise_level / 15
+
+        alphas = [alpha_pos] + 15 * [alpha_neg]
+        post_ops = [[]]  # for alpha_pos, we do nothing, rather than I x I
+        post_ops += [[P(q0)] for P in [X, Y, Z]]  # 1Q Paulis for q0
+        post_ops += [[P(q1)] for P in [X, Y, Z]]  # 1Q Paulis for q1
+        post_ops += [
+            [Pi(q0), Pj(q1)] for Pi in [X, Y, Z] for Pj in [X, Y, Z]
+        ]  # 2Q Paulis
```
natestemen: These are still single qubit paulis. Should they be something else?

mitiq/pea/amplifications/amplify_depolarizing.py
```diff
@@ -0,0 +1,295 @@
...
+        alphas = [alpha_pos] + 3 * [alpha_neg]
+        post_ops = [[]]  # for alpha_pos, we do nothing, rather than I
+        post_ops += [[P(q)] for P in [X, Y, Z]]  # 1Q Paulis
+
+    # The two-qubit case: linear combination of 2Q Paulis
+    elif len(qubits) == 2:
+        q0, q1 = qubits
+
+        alpha_pos = 1.0 - noise_level
+        alpha_neg = noise_level / 15
```
natestemen, line 76: Where does this (noise/15) come from?

mitiq/pea/amplifications/tests/test_amplify_depolarizing.py
```diff
@@ -0,0 +1,156 @@
...
+def test_three_qubit_local_depolarizing_amplification_error():
+    q0, q1, q2 = LineQubit.range(3)
+    with pytest.raises(ValueError):
+        amplify_noisy_op_with_local_depolarizing_noise(
+            Circuit(CCNOT(q0, q1, q2)),
+            0.05,
+        )
+
+
+@pytest.mark.parametrize("circuit_type", ["cirq", "qiskit", "pyquil"])
```
natestemen: Does it work with all accepted qprogram types?

mitiq/pea/amplifications/amplify_depolarizing.py
Misty-W: I see the built-in type replacement was merged last week, so I've made PEA consistent with the new way.

mitiq/pea/pea.py
Misty-W: We could flatten the list, but the downside is we'd have to assume the same number of circuits per scale factor. That's fine for now, but it might not always hold if we tune the number of samples at each scale factor.

mitiq/pea/pea.py
Misty-W: Good idea. I'd prefer to make that a slightly later addition, possibly in #2785, and just focus on making a clean MVP in this PR.

mitiq/pea/tests/test_pea.py
Misty-W: Good catch, I missed that when I adapted some PEC tests for PEA.

mitiq/pea/tests/test_pea.py
Misty-W: Come to think of it, `mypy` does the same thing as this test. I'll go ahead and remove.

mitiq/pea/amplifications/amplify_depolarizing.py
Misty-W: Not exactly because we need the empty op as well, in place of identity. This is consistent with the pattern in `mitiq.pec`.

mitiq/pea/amplifications/amplify_depolarizing.py
Misty-W: No, they should be single qubit Paulis. I guess the `# 2Q Paulis` comment is confusing, so I've reworded it to indicate that it's Paulis on both qubits instead.

mitiq/pea/pea.py
Misty-W: Epsilon is an estimate of the average noise rate on a device as in PEC. The noise gain levels are the scale factors by which the noise rate is amplified .

mitiq/pea/amplifications/tests/test_amplify_depolarizing.py
Misty-W: It does now 🙂

mitiq/pea/tests/test_scale_amplifications.py
Misty-W: Since the intention is to test `scale_circuit_amplifications`, I've addressed this by:
1. Comparing the scaled reps directly rather than using the norm
2. Separating the noise_model parameter into two as suggested

mitiq/pea/tests/test_pea.py
Misty-W: Since `test_precision_option_used_in_num_samples` and `test_precision_ignored_when_num_samples_present` already check for the correct number of circuits at each scale factor, I think the remaining piece was testing that the length of the list of circuit lists is equal to the length of the list of scale factors. 
I've added a test for that.

mitiq/pea/amplifications/amplify_depolarizing.py
Misty-W: Removed.

mitiq/pea/amplifications/amplify_depolarizing.py
Misty-W: Removed.

mitiq/pea/tests/test_pea.py
Misty-W: Removed.

mitiq/pea/amplifications/amplify_depolarizing.py
Misty-W: Replaced as suggested.

mitiq/pea/amplifications/amplify_depolarizing.py
Misty-W: Added and test is passing on `SUPPORTED_QPROGRAM_TYPES`.

mitiq/pea/amplifications/amplify_depolarizing.py
Misty-W: Fixed.

mitiq/pea/pea.py
Misty-W: Suggestion applied.

Misty-W:
thanks for the review @natestemen! I think I resolved or replied to everything. Happy to chat about anything that needs more clarification before merging, but hopefully we can merge very soon!

mitiq/pea/amplifications/amplify_depolarizing.py
natestemen: @Misty-W is there a reference for this?

mitiq/pea/scale_amplifications.py
```diff
@@ -0,0 +1,54 @@
...
+) -> Sequence[OperationRepresentation]:
+    r"""Generates a list of implementable sequences from the noise-amplified
+    representation of the input ideal circuit based on the input noise model
+    and baseline noise level.
+
+    Args:
+        ideal_circuit: The ideal circuit from which an implementable
+            sequence is sampled.
+        scale_factor: A list of (positive) numbers by which the baseline
+            noise level is to be amplified.
```
natestemen:
```suggestion
        scale_factor: A (positive) number by which the baseline noise
            level is to be amplified.
```

natestemen:
Once CI passes I'll go ahead and merge this, but @Misty-W can you provide a reference regarding https://github.com/unitaryfoundation/mitiq/pull/2655#discussion_r2155075741?

PR #2666, opened 2025-02-11
Default scaling method in modularized ZNE function

### LINKED ISSUE #2640: `scaled_circuits` in ZNE should have a default noise scaling method
Similar to the other top-level functions in ZNE, we should specify a non-optional scaling method

```py
def scaled_circuits(
    circuit: QPROGRAM,
    scale_factors: list[float],
    scale_method: Callable[[QPROGRAM, float], QPROGRAM]  = fold_gates_at_random,
) -> list[QPROGRAM]:
```
https://github.com/unitaryfund/mitiq/blob/93e6ec36789c089fa846d56c20f521fe83c3916f/mitiq/zne/zne.py#L16

Unit test for the default scaling method should be added in https://github.com/unitaryfund/mitiq/blob/93e6ec36789c089fa846d56c20f521fe83c3916f/mitiq/zne/tests/test_zne.py#L590

purva-thakre:
## Description
Fixes #2640 

---

### License

### DIFF
--- a/mitiq/zne/tests/test_zne.py
+++ b/mitiq/zne/tests/test_zne.py
@@ -630,3 +630,50 @@ def test_two_stage_zne(
         scale_noise=noise_scaling_method,
     )
     assert np.isclose(zne_res, two_stage_zne_res)
+
+
+@pytest.mark.parametrize(
+    "extrapolation_factory", [RichardsonFactory, LinearFactory]
+)
+@pytest.mark.parametrize(
+    "to_frontend",
+    [None, to_qiskit, to_braket, to_pennylane, to_pyquil, to_qibo],
+)
+def test_two_stage_zne_default_scaling(extrapolation_factory, to_frontend):
+    qreg = cirq.LineQubit.range(2)
+    cirq_circuit = cirq.Circuit(
+        cirq.H.on_each(qreg),
+        cirq.CNOT(*qreg),
+        cirq.CNOT(*qreg),
+        cirq.H.on_each(qreg),
+    )
+    if to_frontend is not None:
+        frontend_circuit = to_frontend(cirq_circuit)
+    else:
+        frontend_circuit = cirq_circuit
+
+    scale_factors = [1, 3, 5]
+    circs = scaled_circuits(frontend_circuit, scale_factors)
+
+    assert len(circs) == len(scale_factors)
+
+    np.random.seed(42)
+
+    def executor(circuit):
+        return np.random.random()
+
+    results = [executor(cirq_circuit) for _ in range(3)]
+    extrapolation_method = extrapolation_factory.extrapolate
+    two_stage_zne_res = combine_results(
+        scale_factors, results, extrapolation_method
+    )
+
+    assert isinstance(two_stage_zne_res, float)
+
+    np.random.seed(42)
+    zne_res = execute_with_zne(
+        cirq_circuit,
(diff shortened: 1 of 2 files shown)


mitiq/zne/tests/test_zne.py
```diff
@@ -630,3 +630,50 @@ def executor(circuit):
...
+
+    assert isinstance(two_stage_zne_res, float)
+
+    np.random.seed(42)
+    zne_res = execute_with_zne(
+        cirq_circuit,
+        executor,
+        factory=extrapolation_factory(scale_factors),
+    )
+    assert np.isclose(zne_res, two_stage_zne_res)
```
natestemen, line 679: These last two tests (assertions) seem unrelated to the code change in `zne.py`.

1. The first tests that the circuits generated by `fold_gates_at_random` return floats. This should be covered by tests in `test_folding.py`.
2. The second tests the performance of a specific ZNE configuration, which I think we want to avoid as it creates non-deterministic tests.

mitiq/zne/tests/test_zne.py
```diff
@@ -630,3 +630,50 @@ def executor(circuit):
...
+
+
+@pytest.mark.parametrize(
+    "extrapolation_factory", [RichardsonFactory, LinearFactory]
+)
+@pytest.mark.parametrize(
+    "to_frontend",
+    [None, to_qiskit, to_braket, to_pennylane, to_pyquil, to_qibo],
+)
+def test_two_stage_zne_default_scaling(extrapolation_factory, to_frontend):
```
natestemen, line 642: `to_frontend` makes it seem like there should be an argument `frontend` that is going to be passed.

```suggestion
def test_two_stage_zne_default_scaling(extrapolation_factory, conversion_func):
```

natestemen (approved):
I'm not convinced that any tests are needed for this code change. What needs to be tested with this change? The only thing I can think of is that `fold_gates_at_random` is called when no `scale_method` is used, but even then that feels unnecessary.

mitiq/zne/tests/test_zne.py
purva-thakre: yeah, I copy-pasted the existing test to understand if I needed to have any newer unit tests. 

Sorry, this PR is not ready yet!

purva-thakre:
@natestemen This PR is now ready.

mitiq/zne/tests/test_zne.py
```diff
@@ -609,6 +609,12 @@ def test_two_stage_zne(
 
     assert len(circs) == len(scale_factors)
 
+    circs_default_scaling_method = scaled_circuits(
+        frontend_circuit, scale_factors
+    )
+
+    assert len(circs_default_scaling_method) == len(scale_factors)
```
natestemen, line 616: This feels unnecessary to me since it's actually a test of a for loop with an append call. If we are going to keep it, it should be in it's own test, not an existing one.

natestemen: @purva-thakre out of curiosity why do you think that the default argument needs a test? It's my first time seeing a test like that, and I'm not really convinced it's needed, but you seem to want to keep it!

purva-thakre: @natestemen It's to make code coverage happy. In the past, I noticed the default option line would show up as uncovered if there wasn't a unit test for it.

PR #2681, opened 2025-03-01
Adding support for observables in LRE executors

### LINKED ISSUE #2513: Make sure LRE executors work with observables
All the top-level functions in LRE should work with `observable`.  This issue is to make sure of this as well as add unit tests for the same.

Shivansh20128:
This PR aims to add support for observables in the LRE executors.
#Resolves #2513

## Description

1. Added required parameters and code changes in lre.py.
2. Added tests in the test file to check observable support.
---

### License


Shivansh20128:
Hi! I tried adding the test for a batched executor as well but could not prepare an `executor` that would support an `observable` as discussed [here](https://github.com/unitaryfund/mitiq/issues/2513#issuecomment-2691225754). 
If its desirable, can someone suggest how I can proceed.

I was trying something like this:

```
@pytest.mark.parametrize(
    "degree, fold_multiplier, observable",
    [(2, 2, Observable[PauliString(spec="X")])],
)
def test_lre_batched_executor_with_observable(degree, fold_multiplier, observable):
    """Verify LRE batch executor with observable works as expected."""
    test_batched_executor = Executor(batched_executor, max_batch_size=200)
    lre_exp_val_batched = execute_with_lre(
        test_cirq,
        test_batched_executor,
        degree=degree,
        fold_multiplier=fold_multiplier,
        observable=observable,
    )
    assert isinstance(lre_exp_val_batched, float)

    assert test_batched_executor.calls_to_executor == 1
    assert (
        test_batched_executor.executed_circuits
        == multivariate_layer_scaling(test_cirq, degree, fold_multiplier)
    )

```

You can find the `batched_executor` [here](https://github.com/unitaryfund/mitiq/blob/34eb6cedda1cc218f2f097704a9278bf15650d1f/mitiq/lre/tests/test_lre.py#L38).

The problem is that I was unable to formulate a `batched_executor` that supports `observables`.

bdg221:
If you check [test_executor.py](https://github.com/unitaryfund/mitiq/blob/ff15f44b8bdc57bc6f4e7feef14dc3d3f2b7f86d/mitiq/executor/tests/test_executor.py#L79C1-L84C76) you can see a lot of examples of executors. I would recommend using `executor_density_matrix_typed` and `executor_density_matrix_batched`.
You should then be able to do:
```
test_batched_executor = Executor(executor_density_matrix_batched, max_batch_size=200)
```

mitiq/lre/tests/test_lre.py
```diff
@@ -75,6 +98,27 @@ def test_lre_executor(degree, fold_multiplier):
     )
 
 
+@pytest.mark.parametrize(
+    "degree, fold_multiplier, observable",
+    [(2, 2, obs_z), (2, 3, obs_y), (3, 4, obs_x)],
+)
+def test_lre_executor_with_observables(degree, fold_multiplier, observable):
```
bdg221, line 105: This looks to be the exact same as `test_lre_exp_value_with_observable`. I would recommend moving the assert lines to that test and removing this one.

bdg221 (changes requested):
@Shivansh20128 Let me know if you have any more questions about the batched executors with observables. Hopefully, the test_exectuor.py examples are helpful.

mitiq/lre/tests/test_lre.py
Shivansh20128: Thank you for the review. I have made the proposed changes.

bdg221 (approved):
LGTM! Great job!

mitiq/lre/lre.py
```diff
@@ -46,10 +47,14 @@ def execute_with_lre(
...
         input_circuit: Circuit to be scaled.
-        executor: Executes a circuit and returns a `float`
+        executor: Executes a circuit and returns a `float`.
         degree: Degree of the multivariate polynomial.
         fold_multiplier: Scaling gap value required for unitary folding which
             is used to generate the scale factor vectors.
+        observable: Observable to compute the expectation value of. If
+            ``None``, the ``executor`` must return an expectation value.
+            Otherwise, the ``float`` returned by ``executor`` is used
+            to compute the expectation of the observable.
```
natestemen, line 57:
```suggestion
        observable: Observable to compute the expectation value of. If
            ``None``, the ``executor`` must return an expectation value.
            Otherwise, the ``MeasurementResult`` returned by ``executor`` is used
            to compute the expectation of the observable.
```

Should this be measurement result or density matrix?

mitiq/lre/tests/test_lre.py
```diff
@@ -55,6 +71,29 @@ def test_lre_exp_value(degree, fold_multiplier):
...
+    """Verify LRE executors with observables work as expected."""
+    test_executor = Executor(mitiq_cirq.compute_density_matrix)
+    lre_exp_val = execute_with_lre(
+        test_cirq,
+        test_executor,
+        degree=degree,
+        fold_multiplier=fold_multiplier,
+        observable=observable,
+    )
+    assert abs(lre_exp_val - ideal_val) <= abs(noisy_val - ideal_val)
```
natestemen, line 88:
```suggestion
```

I think the assertions you have below are great. This kind of assertion is great in theory, but in practice can lead to tests that fail randomly which is annoying.

mitiq/lre/tests/test_lre.py
```diff
@@ -55,6 +71,29 @@ def test_lre_exp_value(degree, fold_multiplier):
     assert abs(lre_exp_val - ideal_val) <= abs(noisy_val - ideal_val)
 
 
+@pytest.mark.parametrize(
+    "degree, fold_multiplier, observable",
+    [(2, 2, obs_z), (2, 3, obs_z), (3, 4, obs_z)],
+)
```
natestemen, line 77: Can we make one of the tests on a setup (circuit and observable) that's on two qubits? There could be something that breaks when more qubits are introduced.

mitiq/lre/lre.py
Shivansh20128: Density matrix would make more sense as discussed [here](https://github.com/unitaryfund/mitiq/issues/2513#issuecomment-2691225754).

natestemen: When an observable is passed, the executor should also be able to return bitstrings **or** a density matrix.

Shivansh20128: Then maybe I should write "the ``DensityMatrix`` or ``Bitstrings`` returned". Because writing ``MeasurementResult`` does not indicate the type of object being returned.

natestemen: Agreed!

mitiq/lre/lre.py
```diff
@@ -25,6 +25,7 @@ def execute_with_lre(
     executor: Union[Executor, Callable[[QPROGRAM], float]],
```
natestemen, line 25: Return type of `Callable` should be generalized to `QuantumResult` similar to to how it exists in ZNE: https://github.com/unitaryfund/mitiq/blob/ff15f44b8bdc57bc6f4e7feef14dc3d3f2b7f86d/mitiq/zne/zne.py#L66

mitiq/lre/tests/test_lre.py
natestemen: I removed two of the single qubit tests cases and simplified the 2 qubit case in https://github.com/unitaryfund/mitiq/pull/2681/commits/e7986ab29dc5f6b65bb9ab8bf678cc2ff8c461fe just so it could run faster. Crazy that it takes 15 minutes! Once CI goes green on this I'll merge it!

natestemen (approved):
Thanks @Shivansh20128!

PR #2698, opened 2025-03-18
Adding utility functions for Virtual Distillation.

FarLab:
Add apply_symmetric_observable and apply_cyclic_system_permutation functions for virtual distillation technique.

### DIFF
--- a/mitiq/vd/vd_utils.py
+++ b/mitiq/vd/vd_utils.py
@@ -1,6 +1,8 @@
-from typing import List, cast
+from typing import List, cast, Optional
+from mitiq.observable import Observable
 
 import cirq
+import numpy as np
 
 
 def _copy_circuit_parallel(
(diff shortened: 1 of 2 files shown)


mitiq/vd/vd_utils.py
```diff
@@ -51,3 +55,137 @@ def map_for_grid_qubits(qu: cirq.Qid) -> cirq.Qid:
...
+        for j in range(2**N_qubits)
+        for i in range(0, 2 ** (M * N_qubits), 2**N_qubits)
+    ]
+
+    # Some fancy index magic to permute the rows in O(n) time
+    # and space (n=2**(M*N_qubits))
+    idx = np.empty_like(permutation)
+    idx[permutation] = np.arange(len(permutation))
+
+    # allow to work with lists or arrays of np.ndarrays
```
natestemen, line 91: Is this comment a leftover? I don't think this should work with lists.

mitiq/vd/vd_utils.py
```diff
@@ -51,3 +55,137 @@ def map_for_grid_qubits(qu: cirq.Qid) -> cirq.Qid:
             new_circuit += circuit.transform_qubits(map_for_grid_qubits)
 
     return new_circuit
+
+
+def _apply_cyclic_system_permutation(
+    matrix: NDArray[np.complex128], N_qubits: int, M: int = 2
```
natestemen, line 61:
```suggestion
    matrix: NDArray[np.complex64], N_qubits: int, M: int = 2
```

Lets use 64-bit precision throughout since we use that in other parts of the code.

mitiq/vd/vd_utils.py
```diff
@@ -51,3 +55,137 @@ def map_for_grid_qubits(qu: cirq.Qid) -> cirq.Qid:
...
+                ]
+            )
+        else:
+            raise ValueError("matrix should be 2D or 3D ndarray")
+
+    else:
+        obs_array = (
+            observable
+            if isinstance(observable, np.ndarray)
+            else np.array(observable)
```
natestemen, line 175:
```suggestion
            else observable.matrix()
```

`np.array(observable)` creates some array based on ... something?

```py
>>> ob = Observable(PauliString("XY"))
>>> np.array(ob)
array(<mitiq.observable.observable.Observable object at 0x16a872930>,
      dtype=object)
```

mitiq/vd/vd_utils.py
```diff
@@ -51,3 +55,137 @@ def map_for_grid_qubits(qu: cirq.Qid) -> cirq.Qid:
...
+        sym_observable_diagonals: List[NDArray[np.complex128]] = []
+        for i in range(N_qubits):
+            observable_i_diagonal = np.array(
+                [
+                    j
+                    for k in range(2 ** (i))
+                    for j in [1.0, -1.0]
+                    for ll in range(2 ** (N_qubits - i - 1))
+                ]
+            )
```
natestemen, line 137: A quadruple `for` loop 😱! I have no immediate ideas how to simplify/optimize this (if it's possible), but definitely a code-smell.

We should be able to get rid of the `k` and `ll` variables though, just by replacing them as `_` since they're unused (right?).

mitiq/vd/vd_utils.py
```diff
@@ -51,3 +55,137 @@ def map_for_grid_qubits(qu: cirq.Qid) -> cirq.Qid:
...
+            # turn [a,b,c] into [a,b,c,a,b,c,a,b,c]. This is the same as
+            # tensoring the N_qubit identity on the left
+            observable_i_diagonal_system2 = np.array(
+                [observable_i_diagonal for _ in range(2**N_qubits)]
+            ).flatten("C")
+            # add the symmetric observable
+            # Create the combined diagonal and explicitly convert to complex128
+            combined_diagonal = 0.5 * (
+                observable_i_diagonal_system1 + observable_i_diagonal_system2
+            )
```
natestemen, line 153: nit:
```suggestion
            combined_diagonal = (
                observable_i_diagonal_system1 + observable_i_diagonal_system2
            ) / 2
```

mitiq/vd/vd_utils.py
```diff
@@ -51,3 +55,137 @@ def map_for_grid_qubits(qu: cirq.Qid) -> cirq.Qid:
...
+
+def _apply_symmetric_observable(
+    matrix: NDArray[np.complex128],
+    N_qubits: int,
+    observable: Optional[Union[Observable, NDArray[np.complex128]]] = None,
+    M: int = 2,
+) -> NDArray[np.complex128]:
+    """
+    Function that applies a symmetric observable to a matrix or vector.
+    The observable is assumed to be the Z observable by default.
```
natestemen, line 109: 1. Remove the sentence about $Z$ being default since it says so in the argument description.
```suggestion
    Function that applies a symmetric observable to a matrix or vector.
```
2. It's unclear what exactly the requirements are for `observable`. Does it need to be symmetric? If so, the code should check that and error if it's not. Is the observable on `N_qubits` or `N_qubits * M`?

mitiq/vd/vd_utils.py
```diff
@@ -51,3 +55,137 @@ def map_for_grid_qubits(qu: cirq.Qid) -> cirq.Qid:
...
+    else:
+        raise TypeError("matrix should be 2D or 3D ndarray")
+    return matrix
+
+
+def _apply_symmetric_observable(
+    matrix: NDArray[np.complex128],
+    N_qubits: int,
+    observable: Optional[Union[Observable, NDArray[np.complex128]]] = None,
+    M: int = 2,
```
natestemen, line 105: Should we rename `M` to `num_registers` to be more explicit?

mitiq/vd/vd_utils.py
```diff
@@ -51,3 +55,137 @@ def map_for_grid_qubits(qu: cirq.Qid) -> cirq.Qid:
...
+) -> NDArray[np.complex128]:
+    """
+    Function that applies a symmetric observable to a matrix or vector.
+    The observable is assumed to be the Z observable by default.
+
+    Args:
+        matrix: The matrix or vector that should be shifted.
+        N_qubits: The number of qubits in each register.
+        observable: The observable that should be applied.
+        If None, the Z observable is used.
```
natestemen, line 115:
```suggestion
        observable: The observable that should be applied.
            If None, the Z observable is used.
```

natestemen (commented):
Great work adding tests for this! Looks good, but a few comments before merging.

mitiq/vd/vd_utils.py
FarLab: I'll remove this comment.

mitiq/vd/vd_utils.py
FarLab: good question, I am assuming they expect a matrix in list form? which doesn't make sense because the type hint suggests either np array or Observable. so instead of np.array(ob) we should use the matrix form of the Observable object. what do you think?

mitiq/vd/vd_utils.py
FarLab: yes, we should remove the k and ll variable as they are unused.

mitiq/vd/vd_utils.py
natestemen: I do think it's good to align with source material in general, but when it comes to implementing mathematical algorithms I don't think it makes as much sense. Single letter variables make (some) sense in math and is basically a tradition. Single letter variables in code make sense, [sometimes](https://softwareengineering.stackexchange.com/questions/285288/is-the-use-of-one-letter-variables-encouraged), but only in particular scenarios.

Since $M$ has a simple interpretation (number of registers (as far as I understand)), I think it is best to spell it out.

mitiq/vd/vd_utils.py
natestemen: Yup that's what I think as well! Thanks for making the change.

mitiq/vd/vd_utils.py
FarLab: makes sense, I'll update it!

Issue #2669, opened 2025-02-11
Mitiq requires `Typing` typehint to create batched executor


natestemen:
## Issue Description

Mitiq's `Executor` class as a method `is_batched_executor` which indicates whether it has the capability of running multiple circuits with a single call. It knows whether it can do this by looking at the annotation of the python function passed to the `Executor` upon instantiation.

https://github.com/unitaryfund/mitiq/blob/f404e288aa15f90fd21e74b742e805ca1ad47431/mitiq/executor/executor.py#L355-L359

The way this is written requires the annotated return type to use a `typing` class. We should generalize this code to allow for users to create batched executors while using modern python type-hints using data structure constructors (`list`, `set`, `tuple`, etc).

## How to Reproduce

```py
def batched_ex(circuits: list[...]) -> list[float]: ...

ex = mitiq.Executor(batched_ex)

ex.is_batched_executor
>>> False  # this should be True
```

PR #2700, opened 2025-03-21
allow list/tuple constructor in executor typehints

natestemen:
## Description

This PR started really simple, but ended up taking me a bit longer because of all the cases we need to support. This PR does two main things:

1. DRYs up duplicated logic to see if an executor can batch (**removing the `is_batched_executor` static method**)
2. Ensures support for executors typehinted with `list` and `tuple` constructors instead of just `typing.List`/`typing.Tuple`.

fixes https://github.com/unitaryfund/mitiq/issues/2669

### Details

I originally tried to make a much more robust solution revolving around checking if the return type is a subclass of `Iterable` which works well, but resulted in cascade of issues.

1. **Numpy arrays are iterable.** Not a problem, add an exception for any numpy object. Doable (albeit somewhat messy).
2. There are **more undesirable subclasses of `Iterable`** that we don't want to allow. E.g. `str` and `dict`. Either add another exception, or ignore those as they are pathological since they are not typical executor return types.
3. Once `list[MeasurementResult]` is supported as a batche-able return type in `can_batch`, it causes a cirq simulator error (thankfully caught in out tests) since it is not in [**`MeasurementResultLike`**](https://github.com/unitaryfund/mitiq/blob/5bd29cebc1ebcd87d7a24348eae0f1e653703307/mitiq/executor/executor.py#L48-L54).

At this point it made sense to me to go the simple route and just manually add a few new types to the existing lists instead of putting in a bunch of time to reconstruct how we manage typehints across Mitiq.

### DIFF
--- a/mitiq/executor/executor.py
+++ b/mitiq/executor/executor.py
@@ -39,6 +39,7 @@ DensityMatrixLike = [
     Tuple[np.ndarray],
     npt.NDArray[np.complex64],
     list[npt.NDArray[np.complex64]],
+    list[np.ndarray],  # type: ignore
     tuple[npt.NDArray[np.complex64]],
 ]
 FloatLike = [
--- a/mitiq/executor/executor.py
+++ b/mitiq/executor/executor.py
@@ -6,6 +6,7 @@
 """Defines utilities for efficiently running collections of circuits generated
 by error mitigation techniques to compute expectation values."""
 
+import collections
 import inspect
 import warnings
 from collections import Counter
@@ -117,7 +118,15 @@ class Executor:
 
         return return_type in (
             BatchedType[T]  # type: ignore[index]
-            for BatchedType in [Iterable, List, Sequence, Tuple, list, tuple]
+            for BatchedType in [
+                Iterable,
+                List,
+                Sequence,
+                Tuple,
+                list,
+                tuple,
+                collections.abc.Sequence,
+            ]
             for T in get_args(QuantumResult)
         )
 
--- a/mitiq/executor/executor.py
+++ b/mitiq/executor/executor.py
@@ -105,7 +105,7 @@ class Executor:
 
         Otherwise, it is considered "serial".
 
-        Batched executors can _run several quantum programs in a single call_.
+        Batched executors can *run several quantum programs in a single call*.
 
         Returns:
             True if the executor is detected as batched, else False.
--- a/mitiq/executor/executor.py
+++ b/mitiq/executor/executor.py
@@ -19,6 +19,7 @@ from typing import (
     Tuple,
     Union,
     cast,
+    get_args,
 )
 
 import numpy as np
@@ -36,6 +37,8 @@ DensityMatrixLike = [
     Sequence[np.ndarray],  # type: ignore
     Tuple[np.ndarray],
     npt.NDArray[np.complex64],
+    list[npt.NDArray[np.complex64]],
+    tuple[npt.NDArray[np.complex64]],
 ]
 FloatLike = [
     None,  # Untyped executors are assumed to return floats.
@@ -44,6 +47,8 @@ FloatLike = [
     List[float],
     Sequence[float],
     Tuple[float],
+    list[float],
+    tuple[float],
 ]
 MeasurementResultLike = [
     MeasurementResult,
@@ -51,6 +56,8 @@ MeasurementResultLike = [
     List[MeasurementResult],
     Sequence[MeasurementResult],
     Tuple[MeasurementResult],
+    list[MeasurementResult],
+    tuple[MeasurementResult],
 ]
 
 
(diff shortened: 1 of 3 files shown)


mitiq/executor/executor.py
```diff
@@ -84,10 +91,34 @@ def __init__(
...
+        include:
+
+        * ``Iterable[QuantumResult]``
+        * ``List[QuantumResult]``/``list[QuantumResult]``
+        * ``Sequence[QuantumResult]``
+        * ``Tuple[QuantumResult]``/``tuple[QuantumResult]``
+
+        Otherwise, it is considered "serial".
+
+        Batched executors can _run several quantum programs in a single call_.
```
bdg221, line 108: Sphinx is hitting a problem with this since the underscore can represent italics and a reference/link. I would recommend changing to using an asterisk instead of underscore to avoid this confusion.

bdg221 (changes requested):
I saw the docs were failing to build. Should be a quick change to help with the docstring.

mitiq/executor/executor.py
```diff
@@ -36,6 +37,8 @@
     Sequence[np.ndarray],  # type: ignore
     Tuple[np.ndarray],
     npt.NDArray[np.complex64],
+    list[npt.NDArray[np.complex64]],
```
cosenal, line 40: We could align the rest of types to have the same more specific parameters `npt.NDArray[np.complex64]` instead of `np.ndarray`

mitiq/executor/executor.py
```diff
@@ -44,13 +47,17 @@
     List[float],
     Sequence[float],
     Tuple[float],
+    list[float],
+    tuple[float],
 ]
 MeasurementResultLike = [
     MeasurementResult,
     Iterable[MeasurementResult],
```
cosenal, line 55: Doesn't this line make all other iterables (list, tuple, etc.) redundant?

mitiq/executor/executor.py
```diff
@@ -84,10 +91,34 @@ def __init__(
...
+        return_type = self._executor_return_type
+
+        if return_type is None:
+            return False
+
+        return return_type in (
             BatchedType[T]  # type: ignore[index]
-            for BatchedType in [Iterable, List, Sequence, Tuple]
-            for T in QuantumResult.__args__  # type: ignore[attr-defined]
+            for BatchedType in [Iterable, List, Sequence, Tuple, list, tuple]
```
cosenal, line 120: `typing.Sequence` is also [deprecated](https://docs.python.org/3/library/typing.html#typing.Sequence) in favor of `collections.abc.Sequence`

cosenal (approved):
left some small comments, not blocking

mitiq/executor/executor.py
natestemen: No, because of the way we use `MeasurementResultLike`. E.g.

https://github.com/unitaryfund/mitiq/blob/7e9641014f5e27a15942775c99067a94e88a5969/mitiq/executor/executor.py#L174

If we did something like checking if the return type is a subclass then perhaps this would work, but then we would need to have different checks for batched/not batched executors.
```py
if any(issubclass(self._executor_return_type, Iterable[MeasurementResult])):
    ...  # I haven't tried this
```

mitiq/executor/executor.py
natestemen: Actually I think all the data types in `typing` are deprecated and have been moved to `collections.abc`. I've opened https://github.com/unitaryfund/mitiq/issues/2701 to ensure we upgrade, but I think that is best done in a separate PR.

cosenal: What I meant is that how we added `list`, and `tuple`, which are the non deprecated alternative of `List` and `Tuple`, then we should add `abc.Sequence` as a non-deprecated alternative for `typing.Sequence`.

mitiq/executor/executor.py
natestemen: This is tightly linked with my comment in https://github.com/unitaryfund/mitiq/pull/2700#discussion_r2008809787 about how we actually use these types (i.e. how we use member inclusion instead of subclassing). I've tried to replace the `np.ndarray` generics but then tests start failing because `np.ndarray` (a potential typehint for a density matrix executor) fails to be in `DensityMatrixLike`.

I'm not sure what the best solution here is because someone could use `np.ndarray` as a typehint, or `npt.NDArray[...]`. How far do we go to support every typehint?

mitiq/executor/executor.py
natestemen: I see. Added an additional type in https://github.com/unitaryfund/mitiq/pull/2700/commits/d2ba012f13b0bb43e361a9c0a251d071df1f5e4f.

This made me realize that if someone typehints an executor with `collections.abc.Iterable`, we will misclassify it as a serial executor since `typing.Iterable != collections.abc.Iterable`. Ugh.

cosenal: The bottom-line problem is this use of type checks at runtime. In the next Mitiq Coding session, let's brainstorm if/why we need those at all.

natestemen:
Merging is blocked, either because

1. there have been added commits since an approval
2. changes have been requested and an approval has not followed

@bdg221 PTAL :)

bdg221 (approved):
LGTM!

PR #2709, opened 2025-04-01
Modularization uniformity for ZNE, PEC, DDD, and LRE

### LINKED ISSUE #2692: Modularization Uniformity for ZNE, PEC, DDD, and LRE
### Discussed in https://github.com/unitaryfund/mitiq/discussions/2677

<div type='discussions-op-text'>

<sup>Originally posted by **bdg221** February 19, 2025</sup>
Looking at the different techniques that have been modularized, there are inconsistencies. 
ZNE, PEC, and DDD have one function that generates the circuits and another that provides the final expectation value.

LRE has a function that generates the circuits and one that generates the coefficients used in generating the final expectation value. 

ZNE: `scaled_circuits` and `combine_results`
PEC: `generate_sampled_circuits` and `combine_results`
DDD: `generate_circuits_with_ddd` and `combine_results`
LRE: `multivariate_layer_scaling` and `multivariate_richardson_coefficients`

I think it would be useful to have the function that generates the circuits to have the same naming convention. Also, I think adding a `combine_results` to LRE makes sense for consistency.</div>

The discussion concluded with the agreement of using `construct_circuits` and `combine_results`.

bdg221:
Fixes #2692 

## Description

The mitigation techniques in the title now have `construct_circuits` and `combine_results` to allow for two stage implementations.

ZNE, PEC, and DDD were straight forward renaming of existing functions. LRE required creating the two functions and a test. Also, during this PR I also made input parameter names conform to ZNE and the other technique's naming scheme.

For LRE's `construct_circuits`, I used ZNE as a template and included a parameter `scale_method` which is defaulted to 
`multivariate_layer_scaling`. My concern is that this is the only scaling method currently available to users. While it leaves open the flexibility for future, I do not know if the plan is to include other options for scaling.

For testing LRE's new functions, I used the ZNE two stage test as an example allowing for testing both the `construct_circuits` and `combine_results`.

> [!NOTE]
> LRE's documentation has NOT been updated yet. It requires a bit more of a re-write than just a few name changes. All other docs have been updated.
---

### License

### DIFF
--- a/docs/source/examples/lre-zne-comparison.md
+++ b/docs/source/examples/lre-zne-comparison.md
@@ -146,7 +146,7 @@ Next for ZNE.
 
 ```{code-cell} ipython3
 from mitiq.zne.scaling import fold_gates_at_random
-from mitiq.zne.zne import construct_circuits
+from mitiq.zne import construct_circuits
 
 scale_factors = [1.0, 2.0, 3.0]
 
--- a/docs/source/examples/lre-zne-comparison.md
+++ b/docs/source/examples/lre-zne-comparison.md
@@ -146,7 +146,7 @@ Next for ZNE.
 
 ```{code-cell} ipython3
 from mitiq.zne.scaling import fold_gates_at_random
-from mitiq.zne.zne import scaled_circuits
+from mitiq.zne.zne import construct_circuits
 
 scale_factors = [1.0, 2.0, 3.0]
 
@@ -154,7 +154,7 @@ avg_num_scaled_circuits_zne = 0.0
 avg_depth_scaled_circuits_zne = 0.0
 
 for circuit in circuits:
-    noise_scaled_circuits_zne = scaled_circuits(
+    noise_scaled_circuits_zne = construct_circuits(
         circuit=circuit,
         scale_factors=[1.0, 2.0, 3.0],
         scale_method=fold_gates_at_random,
--- a/mitiq/lre/lre.py
+++ b/mitiq/lre/lre.py
@@ -43,8 +43,6 @@ def construct_circuits(
             number of chunks is the same as the layers in the input circuit,
             the input circuit is unchanged.
 
-        circuit: The input circuit to execute with ZNE.
-        scale_factors: An array of noise scale factors.
 
     Returns:
         The scaled circuits using the multivariate_layer_scaling.
--- a/mitiq/lre/lre.py
+++ b/mitiq/lre/lre.py
@@ -20,8 +20,90 @@ from mitiq.lre.multivariate_scaling import (
 from mitiq.zne.scaling import fold_gates_at_random
 
 
+def construct_circuits(
+    circuit: QPROGRAM,
+    degree: int,
+    fold_multiplier: int,
+    folding_method: Callable[
+        [QPROGRAM, float], QPROGRAM
+    ] = fold_gates_at_random,  # type: ignore [has-type]
+    num_chunks: Optional[int] = None,
+    scale_method: Callable[
+        [
+            QPROGRAM,
+            int,
+            int,
+            Optional[int],
+            Callable[[QPROGRAM, float], QPROGRAM],
+        ],
+        list[QPROGRAM],
+    ] = multivariate_layer_scaling,
+) -> list[QPROGRAM]:
+    """Given a circuit, degree, fold_multiplier, folding_method, num_chunks,
+       and an optional scale_method with a default of
+       multivariate_layer_scaling, outputs a list of circuits that will be used
+       in LRE.
+
+    Args:
+        circuit: Circuit to be scaled.
+        degree: Degree of the multivariate polynomial.
+        fold_multiplier: Scaling gap value required for unitary folding which
+            is used to generate the scale factor vectors.
+        folding_method: Unitary folding method. Default is
+            :func:`fold_gates_at_random`.
+        num_chunks: Number of desired approximately equal chunks. When the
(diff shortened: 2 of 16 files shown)


mitiq/lre/lre.py
```diff
@@ -20,8 +20,90 @@
...
+    scale_method: Callable[
+        [
+            QPROGRAM,
+            int,
+            int,
+            Optional[int],
+            Callable[[QPROGRAM, float], QPROGRAM],
+        ],
+        list[QPROGRAM],
+    ] = multivariate_layer_scaling,
```
natestemen, line 40: as noted in your PR description we have no plans to add different scale methods here. Let's remove this as an option until we need to add it.

natestemen (changes requested):
Just one small blocker, but otherwise this looks good! Thanks for taking a deeper look than simply renaming everything :)

mitiq/lre/lre.py
bdg221: Removed in https://github.com/unitaryfoundation/mitiq/pull/2709/commits/bee9c8ab993765398fa347dad49be10bb869145a

docs/source/examples/lre-zne-comparison.md
```diff
@@ -146,15 +146,15 @@ Next for ZNE.
 
 ```{code-cell} ipython3
-from mitiq.zne.zne import scaled_circuits
+from mitiq.zne.zne import construct_circuits
```
natestemen, line 149:
```suggestion
from mitiq.zne import construct_circuits
```

mitiq/lre/lre.py
```diff
@@ -20,8 +20,74 @@
...
+        fold_multiplier: Scaling gap value required for unitary folding which
+            is used to generate the scale factor vectors.
+        folding_method: Unitary folding method. Default is
+            :func:`fold_gates_at_random`.
+        num_chunks: Number of desired approximately equal chunks. When the
+            number of chunks is the same as the layers in the input circuit,
+            the input circuit is unchanged.
+
+        circuit: The input circuit to execute with ZNE.
+        scale_factors: An array of noise scale factors.
```
natestemen, line 47:
```suggestion
```

mitiq/lre/lre.py
```diff
@@ -20,8 +20,74 @@
...
+
+    Returns:
+        The scaled circuits using the multivariate_layer_scaling.
```
natestemen, line 50:
```suggestion
        The scaled circuits using the multivariate layer scaling.
```

Not sure what the best wording is here, but if we want to reference a function, lets make sure it's formatted as code.

mitiq/lre/lre.py
```diff
@@ -20,8 +20,74 @@
...
+
+    Args:
+        circuit: Circuit to be scaled.
+        degree: Degree of the multivariate polynomial.
+        fold_multiplier: Scaling gap value required for unitary folding which
+            is used to generate the scale factor vectors.
+        num_chunks: Number of desired approximately equal chunks. When the
+            number of chunks is the same as the layers in the input circuit,
+            the input circuit is unchanged.
+        results: An array storing the results of running the scaled circuits.
```
natestemen, line 78: `results` should come first since it is the first argument.

mitiq/lre/lre.py
```diff
@@ -20,8 +20,74 @@
...
+    """
+    noise_scaled_circuits = multivariate_layer_scaling(
+        circuit, degree, fold_multiplier, num_chunks, folding_method
+    )
+    return noise_scaled_circuits
+
+
+def combine_results(
+    results: list[float],
+    circuit: QPROGRAM,
```
natestemen, line 60: Kind of strange that the user has to pass the circuit to the `combine_results` function, but I guess it makes sense as the `results` arg does not correspond uniquely to a set of LRE parameters.

[no action needed, just thinking aloud]

mitiq/lre/lre.py
```diff
@@ -107,7 +173,6 @@ def mitigate_executor(
     error-mitigated with layerwise richardson extrapolation (LRE).
 
     Args:
-        input_circuit: Circuit to be scaled.
```
natestemen, line 110: good find!

natestemen (approved):
Just a few minor comments, but looks good once those are resolved. Feel free to merge once they are :)

mitiq/lre/lre.py
bdg221: I agree it's a bit unusual, but the other option would be to expect the user to run `multivariate_richardson_coefficients` on their own and pass in those along with the results at which point `combine_results` would simply be doing a dot product of the two input parameters. Definitely, something to think about, but I'm not sure there is a perfect way to handle it.

natestemen: agreed, that would be even more complicated. I think the choice you made is reasonable.

mitiq/lre/lre.py
bdg221: solved in [32675e2](https://github.com/unitaryfoundation/mitiq/pull/2709/commits/32675e2cbe6b510bc4cc234f10b6143716400d55)

mitiq/lre/lre.py
bdg221: I updated this to use:
```
:func:`mitiq.lre.multivariate_scaling.layerwise_folding.multivariate_layer_scaling`
```

I plan to wait for the doc builds to complete to verify the links work properly in the API docs. Previously, the `:func:` was used for `fold_gates_at_random` without the path, but I also updated that path. Once the docs are built and the functionality is confirmed, then I will look to merge.

PR #2724, opened 2025-04-17
Use uv and pyproject.toml for dependencies and configs

### LINKED ISSUE #867: Migrate to pyproject.toml for project configuration
Pre-Request Checklist
---------------------

Issue Description
-----------------

By newly accepted [PEP guides](https://www.python.org/dev/peps/pep-0631/) for best practices, project configuration/parameters should be set in a `pyproject.toml` instead of requirements files, setup.py, and other conf files. Many other tools in the Python ecosystem already support getting configuration from these toml files (just another markdown sort of syntax). One really nice thing about this is that there is _one_ file (not script) that you can look at for everything you need to configure the project. It allows for so many specifications like optional and dev dependencies, easy version numbering based on git tags, and [so much more](https://www.python.org/dev/peps/pep-0518/).

Proposed Solution
-----------------
Move any tools/configuration to the `pyproject.toml` file we already have (black requires it for conf).

Tools confirmed to use:
- [black](https://ichard26-testblackdocs.readthedocs.io/en/refactor_docs/pyproject_toml.html)
- [pytest](https://docs.pytest.org/en/6.2.x/customize.html#pyproject-toml)
- [mypy](https://mypy.readthedocs.io/en/stable/config_file.html#using-a-pyproject-toml-file)

Tools that we use that don't default support `pyproject.toml`, but have shims to make it work:
- [flake8 ](https://github.com/csachs/pyproject-flake8)
- [sphinx](https://sphinx-pyproject.readthedocs.io/en/latest/)
  - MyST by extension

Additional References
---------------------

Blogs from Python core maintainers/community that are helpful in understanding the changes and how to migrate:
- https://snarky.ca/what-the-heck-is-pyproject-toml/
- http://ivory.idyll.org/blog/2021-transition-to-pyproject.toml-example.html
- https://discuss.python.org/t/where-to-get-started-with-pyproject-toml/4906

Examples of files in the wild:
- https://github.com/sphinx-toolbox/sphinx-pyproject/blob/master/pyproject.toml
-

bdg221:
fixes #867 

## Description

This PR is meant to move the project away from requirements and `setup.py` and instead use `pyproject.toml` and `uv`.

This is my first time using `uv`, so any help would be appreciated.

First, I built out `[project.dependencies]` and `[project.optional-dependencies]` to remove the requirements files. Next, I updated the `[project]` attributes from setup.py, including setting up dynamic versioning by looking at `VERSION.txt`. I then added `[tool.uv.dependencies]` to match the optional `dev` dependencies. After removing the requirements directory and files, along with the `setup.py`, I have tested a basic end-user install with `pip install mitiq` to confirm that dependency installation matches the older method of using `setup.py` and requirements. I have not tested the optional installation pieces yet.

`[development]` in `setup.py` was built by including `requirements-dev.txt` and also pulling in all other optional-dependencies (requirements/requirements-XXX.txt) listed in `./INTEGRATION.txt.` 

My understanding is if the optional-dependencies should be included with development then they must be explicitly and manually included. While this is not a problem, keeping track of the versions will be more complicated using the single `pyproject.toml` file. Is there a cleaner or easier way to do this since the updates to dependencies will have to be done in the specific `[project.optional-dependencies]` section, the “development” section under `[project.optional-dependencies]`, and `[tools.uv.dependencies]`.

This leads to the question, do we want to have both [project.optional-dependencies] with `development` and [tools.uv.dependencies], which controls a development environment through `uv`? We obviously want to update the documentation to tell contributors the new way to set up a development environment, but should we retain the old method?

Currently, the version of Mitiq is pulled from `./VERSION.txt`. I have set up the `pyproject.toml` to use this file. However, do we want to continue to do this? Also, `setup.py` changed `mitiq/_version.txt`, though I am not sure why this code exists. Is that script for changing `mitiq/_version.txt` still necessary?

After the questions above are answered, I will test the more specific installs with optional-dependencies, utilize `uv` for a `uv.lock` and test the development environment, and then look into CI/CD pipeline to see how it can be changed. If there are any suggestions for additional testing, please let me know.

---

### License


bachase:
Great work @bdg221. For your questions:

### On `development` dependencies

TL;DR -- I'd drop the one under `[project.optional-dependencies]`

I've not contributed to `mitiq` myself, but my mental model is that `[project.optional-dependencies]` is more for `mitiq` users, where these options become the "extras" in what features are installed with `mitiq`, e.g. `pip install mitiq[cirq]`. `[tool.uv.dependencies]` captures dependencies only needed for management, and are *not* included as dependencies in the published packages. Note that UV now [supports](https://docs.astral.sh/uv/concepts/projects/dependencies/#development-dependencies) `[dependency-groups]` that are from PEP 735.

As you note, it seems the `setup.py` mixed this a bit, where "development" was also exposed as an extra for users. I think this change to `uv` should break that, where we are now more prescriptive. If you want to develop/contribute to mitiq, you should use `uv` to manage development dependencies. In that case, would drop the `development` group under the `[project.optional-dependences]`. 

I haven't tested, but I think 🤞  with such a setup, `uv` would install and manage resolution of all of these together, and then when the project is published, only the `project.optional-dependencies` ones make it to PyPI.

### On version.txt
Would want @natestemen to weigh in, but I like it where the single version is just the one in explicitly set in the `pyproject.toml`, and then the code references this file to get the version. For example, in `ucc`, this sets the version at the package level here https://github.com/unitaryfoundation/ucc/blob/main/ucc/_version.py. I think the others uses of version.py in mitiq could work like done in `ucc`, e.g. the docs or ci would import the package and get the version there. But there may be other nuances in mitiq release process that prevent this. But this would follow your thinking to eliminate the extra scripting/layers here and keep it simple

bdg221:
A few comments/questions/concerns.

### VERSION.txt
Brad mentioned using the `pyproject.toml` file for this and while that seemed straight-forward, I did find that in [publish-testpypi.yml](https://github.com/unitaryfoundation/mitiq/blob/9d071cd9529f4802d59dec9a5b66230a6969faae/.github/workflows/publish-testpypi.yml#L29) the `VERSION.txt` file is used.

Nate, what are your thoughts on this? Do we want to keep this info in `VERSION.txt` and set the `pyproject.toml` to dynamically pull it from this file? I think that might be easiest and at least for now limit the number of changes in this PR.

### pyproject.toml, [development], and Makefile
As Brad pointed out, `uv` supports `[dependency-groups]` which includes `[dev]` by default. This means when running `uv sync` the `[dev]` group is included by default. One interesting thing is that `pip install .[development]` actually installs all of the 3rd party requirements as well. So the more accurate `uv` command would be `uv sync --all-extras` so that the other dependencies are installed too. Currently, the Mitiq [contributor documentation](https://mitiq.readthedocs.io/en/stable/contributing.html#development-environment) says to use `make install`. I have updated the `Makefile` to use `uv sync --all-extras`. Does this seem reasonable?

I have a concern with other `Makefile` use-cases. For example, if a developer run `make install` then `make test`, this will fail. After `make install`, users would need to run `uv run make test`. I could update all the calls inside of the `Makefile` to use `uv run`, but I'm not sure if that is the best approach. Any thoughts or ideas?

One more note about `pyproject.toml` and the current `requirements` is that the default dependencies include `cirq-core` and the `mitiq[cirq]` (or `requirements-cirq.txt`) also includes `cirq-core` and ONLY `cirq-core`. Is there any reason other than uniformity to keep `cirq` as a now `[project.optional-dependencies]`?


### publish-pypi.yml and publish-testpypi.yml 
I have not yet addressed these files. I have been looking at [ucc](https://github.com/unitaryfoundation/ucc/blob/main/.github/workflows/python-publish.yml) as an example, and it seems much cleaner. This may due to the project be younger and not going through as many changes/iterations. First, `uv` does include a `uv build` option that could replace the current `python -m build`. Also, `uv` includes `uv publish`. [Here](https://docs.astral.sh/uv/guides/package/) is the documentation. Again, I do not know if we want to include this in this PR or create a new issue to address is independently.

### build.yml and pyquil extra step
I am guessing at some point `cirq-core` may have had `pyquil` as a dependency, but it does not anymore. Therefore, the line in [build.yml](https://github.com/unitaryfoundation/mitiq/blob/9d071cd9529f4802d59dec9a5b66230a6969faae/.github/workflows/build.yml#L55) which removes pyquil after installing the base dependencies no longer makes sense. Currently, this line actually results in [WARNING: Skipping pyquil as it is not installed.](https://github.com/unitaryfoundation/mitiq/actions/runs/14586844755/job/40913813314?pr=2725#step:4:116)

pyproject.toml
```diff
@@ -34,3 +127,12 @@ disallow_untyped_defs = true
 no_implicit_optional = true
 warn_redundant_casts = true
 warn_unused_configs = true
+
+[tool.setuptools.packages.find]
```
bachase, line 131: I'm not familiar with `tool.setuptools.packages.find` and docs online aren't super clear to me. Can you explain how its used for mitiq?

pyproject.toml
```diff
@@ -34,3 +127,12 @@ disallow_untyped_defs = true
...
+where = ["."]
+
+[[tool.uv.index]]
```
bachase, line 134: Is this one needed for the current publish workflow?

bachase (commented):
Two small question but overall 👍 on the changes thus far. 

My view on your questions

1. I agree that `uv sync --all-extras` makes sense for the default install 
2. For the `make` targets, I'd lean towards putting `uv run` inside each way to simplify this change. I think it's probably worth revisiting the use of the Makefile at somepoint, but agree better to separate from this change.
3. On `cirq-core`, I'd lean towards eliminating it as an optional dependency group. It does make that different than all the other types, but suppose that just gets documented. Also +1 on removing the pyqill bit that is no longer needed.
4. For publishing, one difference is `mitiq` runs the test pypi flow nightly, where as `ucc` has that as a step in the release process (GH release -> test pypi -> real pypi). To minimize changes, I think better to leave the two flows and consider unifying as a separate PR/issue. For the same reason, maybe wait on `uv publish`, unless its an easy drop-in replacement.

pyproject.toml
bdg221: This section comes from migrating the following in [setup.py](https://github.com/unitaryfoundation/mitiq/blob/9d071cd9529f4802d59dec9a5b66230a6969faae/setup.py#L43C5-L43C30):
```
packages=find_packages(),
```

I was following [this Medium post](https://medium.com/the-python-project/migrating-from-setup-py-to-pyproject-toml-bd840125e07a) which also used `find_packages`. Checking the [setuptools documentation](https://setuptools.pypa.io/en/latest/userguide/package_discovery.html#finding-simple-packages), they provide a useful example (allowing you to switch from setup.py to pyproject.toml.) Most examples I have seen pass information to `find_packages` and then that information is used in the `[tools.setuptools.packages.find]`. We however are not limiting or qualifying the search for `__init__.py` files.

If you are aware of a better/different way to do this, I would appreciate any insight.

bachase: Got it. Good to know. It looks like `mitiq` only has one package, so maybe its not necessary? But IMHO, ok to leave in to keep the translation simple.

.github/workflows/build.yml
```diff
@@ -42,22 +45,24 @@ jobs:
...
         with:
           python-version: "3.12"
+
       - name: Install Mitiq
         # Since requirements.txt includes cirq, which in turn has pyquil as dependency,
         # we explicitly remove pyquil from the installed packages after installing mitiq
-        run: |
-          python -m pip install --upgrade pip
-          pip install .
-          pip uninstall -y pyquil
```
natestemen, line 55: The "uninstall pyquil" is likely vestigial from when we transitioned from relying on `cirq` to `cirq-core`. Good to clean this up!

.github/workflows/build.yml
```diff
@@ -19,20 +19,23 @@ jobs:
...
-        run: |
-          python -m pip install --upgrade pip
-          make install
+          
+      - name: Install the project
+        run: uv sync --all-extras
 
       - name: Check types with mypy
-        run: make check-types
+        run: uv run make check-types
```
natestemen, line 35: I like @bachase's suggestions of prefixing every command in the makefile with `uv run` so it doesn't have to change elsewhere. WDYT?

natestemen (commented):
This is looking great! Thanks a bunch Brian for modernizing this, and thanks Brad for the help.

I want to wait to give an approval until I've had a chance to pull it down and see what it's like doing development with it, but wanted to give some feedback/ask some questions quicker. Overall, I agree with Brad in his latest comment.

> Nate, what are your thoughts on this? Do we want to keep this info in VERSION.txt and set the pyproject.toml to dynamically pull it from this file? I think that might be easiest and at least for now limit the number of changes in this PR.

If there's an easy workaround to replace the `VERSION.txt`, great! I can't think of an easy one, and if you can't either I'd say lets worry about that in a later PR.

Does this change will mean users will lose the capability to run `pip install mitiq[qiskit]` (for example)? It's not the end of the world if that is the case, but I just want to be aware.

bdg221:
@natestemen in this last few commits I separated out the `docs` requirements into its own `dependency-groups`. While `make install` will still include this, running `uv sync --all-extras` installs all dev and third party dependencies except those for `docs`. This is useful for #2728 since these commits update the github workflow so that only the `doc-build` workflow includes the `docs` dependencies since it is pinned at Python 3.12. Any new dependencies for the Mitiq documentation would be added to `pyproject.toml` which is added to the [contributor documentation](https://mitiq--2724.org.readthedocs.build/en/2724/contributing_docs.html#adding-code-examples).

CONTRIBUTING.md
```diff
@@ -12,19 +12,25 @@ The rest of this document describes the technical details of getting set up to d
...
-4. Inside the Mitiq repository (`cd mitiq`), activate a virtual environment. With conda this is done using the following command.
-```
-conda create --name myenv python=3
-conda activate myenv
-```
-5. Install the dependencies. First, to get an updated version of [`pip`](https://pypi.org/project/pip/) inside the virtual environment run `conda install pip` followed by
-```
+Mitiq uses [uv](https://docs.astral.sh/uv/) for packaging and dependency management. It may also be used for managing the virtual environment. 
+
+1. Ensure you have [uv installed](https://docs.astral.sh/uv/getting-started/installation/), for example using `pip install uv`.
```
natestemen, line 17:
```suggestion
1. Ensure you have [`uv` installed](https://docs.astral.sh/uv/getting-started/installation/).
```

Probably best to leave it up to the `uv` docs to describe how to install.

CONTRIBUTING.md
```diff
@@ -12,19 +12,25 @@ The rest of this document describes the technical details of getting set up to d
...
+
+You should now have a development environment set up to work on Mitiq! 🎉 To go forward with making the desired changes, please consult the ["Making changes" section](https://www.asmeurer.com/git-workflow/#making-changes) of the `git` workflow article. If you've encountered any problems thus far, please let us know by opening an issue! More information about workflow can be found below in the [lifecycle](#lifecycle) section.
+
+> [!Note]
+> Since uv uses a virtual environment, any commands to be run inside the virtual environment will need to be prefaced with `uv run` or the uv [managed virtual environment](https://docs.astral.sh/uv/pip/environments/#using-a-virtual-environment) will need to be activated by running `source .venv/bin/activate` in your shell. 
+>
+> All commands using the Makefile do NOT require `uv run`.
```
natestemen, line 33:
```suggestion
> Running `make` commands do NOT require prepending `uv run`.
```

.github/workflows/build.yml
```diff
@@ -19,14 +19,17 @@ jobs:
     steps:
       - name: Check out mitiq
         uses: actions/checkout@v4
+
+      - name: Install uv
+        uses: astral-sh/setup-uv@v5
```
natestemen, line 24:
```suggestion
        uses: astral-sh/setup-uv@v6
```

[v6](https://github.com/astral-sh/setup-uv/releases/tag/v6.0.0) is out!

.github/workflows/build.yml
```diff
@@ -72,21 +77,27 @@ jobs:
...
         uses: actions/setup-python@v5
         with:
           python-version: ${{ matrix.python-version }}
-      - name: Install Python dependencies
-        run: |
-          python -m pip install --upgrade pip
-          make install
+
+      - name: Install Python dependencies except for docs
+        run: uv sync --all-extras
```
natestemen, line 90: I thought `--all-extras` includes the docs group?

natestemen (approved):
Thanks for updating the contributing docs! This is getting close.

.github/workflows/build.yml
bdg221: TLDR; `--all-extras` refers to optional dependencies `--all-groups` refers to optional **development** dependencies

With uv there are multiple ways to add dependencies in the `pyproject.toml` file.

1.) The [main basic dependencies](https://docs.astral.sh/uv/concepts/projects/dependencies/#project-dependencies) required when installing the package. Installed by default:
[project]
dependencies = [...]

2.) [Optional dependencies](https://docs.astral.sh/uv/concepts/projects/dependencies/#optional-dependencies) when installing the package. These are installed with `--all-extras` and for users can be installed with `pip install mitiq[qiskit]`:
[project.optional-dependencies]
braket = [...]
qiskit = [...]

3.) [Development dependency default `dev` group](https://docs.astral.sh/uv/concepts/projects/dependencies/#default-groups). This is installed by default with `uv sync`:
[dependency-groups]
dev = [..]

4.) Other [development dependency groups](https://docs.astral.sh/uv/concepts/projects/dependencies/#dependency-groups). These requires `--all-groups` for installation with `uv sync`.
[dependency-groups]
dev = [...]
**docs = [...]**


Since I separated out the docs group from the default dev group, `make install` now calls `uv sync --all-extras --all-groups` to include the docs group.

bdg221:
That is a great point! I believe that `publish-pypi.yml` does need to be updated. However, I am not sure how we can test this. @natestemen do you have any ideas? If not, I'll research this and figure it out.

natestemen:
The publish workflows (for pypi and test pypi) are indeed a bit cumbersome to test. Given we can always circumvent the github release by using the manual process documented [here](https://github.com/unitaryfoundation/mitiq/blob/main/docs/source/release.md#release-the-new-version-on-pypi) I say lets merge (after the conflict is resolved) and try it!

natestemen:
Failure is purely related to codecov, so going to merge! 🎉

Thanks a lot Brian for upgrading our stack 🙏🏻

PR #2728, opened 2025-05-02
Add tutorial demonstrating basic usage of mitiq & ucc

natestemen:
## Description

Does what it says on the 🥫.

Once the [research notebook](https://github.com/unitaryfoundation/research/pull/31) is more nailed down it would be good to add one of the plots demonstrating the impact of compilation in a more robust way than the one cherry-picked example in this tutorial.

fixes https://github.com/unitaryfoundation/mitiq/issues/2705


bdg221:
@natestemen I tested this locally and hit an issue. I started with a clean environment, installed python=3.12, then ran `make install`. I then tried to walk through the notebook version of `ucc.md`. When doing the initial imports, I get an error:
ModuleNotFoundError: No module named 'pytket.extensions'

natestemen:
Oh perfect! This is the same error that RTD is hitting, that I wasn't able to reproduce. Can you try installing [`pytket-cirq`](https://docs.quantinuum.com/tket/extensions/pytket-cirq/)?

bdg221:
That fixed it! All cells run properly. I did get the following through when running `pip install pytket-cirq`:

ERROR: pip's dependency resolver does not currently take into account all the packages that are installed. This behaviour is the source of the following dependency conflicts.
cirq-ionq 1.4.1 requires cirq-core==1.4.1, but you have cirq-core 1.5.0 which is incompatible.
mitiq 0.45.0.dev0 requires cirq-core<1.5.0,>=1.4.0, but you have cirq-core 1.5.0 which is incompatible.
cirq-rigetti 1.4.1 requires cirq-core==1.4.1, but you have cirq-core 1.5.0 which is incompatible.

docs/source/examples/ucc.md
```diff
@@ -0,0 +1,153 @@
...
+```{code-cell} ipython3
+header = "{:<15} {:<10} {:<10}"
+row = "{:<15} {:<10.4f} {:<10.4f}"
+
+print(header.format("", "Uncompiled", "Compiled"))
+print(row.format("Ideal", ideal_uncompiled, ideal_compiled))
+print(row.format("Noisy", noisy_uncompiled, noisy_compiled))
+print(row.format("Mitigated", mitigated_uncompiled, mitigated_compiled))
+print(
+    row.format(
```
jordandsullivan, line 133: I think this would be nice to see as a simple visualization, say a bar chart with some color/texture coding, e.g.
Screenshot 2025-05-06 at 3 42 35 PM

With just the text, it's a little hard to interpret at a glance.

jordandsullivan:
> I did get the following through when running `pip install pytket-cirq`:
> 
> ERROR: pip's dependency resolver does not currently take into account all the packages that are installed. This behaviour is the source of the following dependency conflicts. cirq-ionq 1.4.1 requires cirq-core==1.4.1, but you have cirq-core 1.5.0 which is incompatible. mitiq 0.45.0.dev0 requires cirq-core<1.5.0,>=1.4.0, but you have cirq-core 1.5.0 which is incompatible. cirq-rigetti 1.4.1 requires cirq-core==1.4.1, but you have cirq-core 1.5.0 which is incompatible.

I got the same error as @bdg221 trying to install this dev version of mitiq, and when separately running `pip install pytket-cirq`. So the install seems to fail, but the notebook itself runs fine 🤷

jordandsullivan (changes requested):
The installation errors are a bit of a concern for me. Visualization is secondary but I think important for the storytelling of UCC + Mitiq.

docs/source/examples/ucc.md
natestemen: I like the idea of making this more visual. Here was my attempt inspired by your drawing.

image

natestemen:
> I got the same error as @bdg221 trying to install this dev version of mitiq, and when separately running pip install `pytket-cirq`. So the install seems to fail, but the notebook itself runs fine 🤷

Do you need to install `pytket-cirq` separately? It is in the requirements files, so should be installed when running `make install`. I'm not seeing this locally, so I might need some help debugging if you are.

### Demo circuits

I fussed with this a lot to try to find "the right" circuit to demo this on, but everything I tried seemed to be limited by the fact that UCC simplifies 2-qubit circuits so well that sometimes EM doesn't actually do anything useful (i.e. the circuit is so short that EM basically isn't needed). Either that, or UCC finds the structure (e.g. mirror circuits, or RB, or even rotated RB) and then eliminates a huge majority of it. This is great for UCC, but makes it hard to create this demo. That's why I left this with a random clifford+$T$ circuit.

### non-determinism

I tried setting all the seeds I could, but there's still some non-determinism I haven't been able to pin down. It could be numerical instability in the extrapolation. This needs to be changed before merging.

### build failures

> The installation errors are a bit of a concern for me though.

Because UCC requires python 3.12 and all dependencies are required to be installed to run the tests, this breaks the tests for 3.10 and 3.11. Brian's PR to move our build/dev infra to `uv` https://github.com/unitaryfoundation/mitiq/pull/2724 should fix this, so we will wait until that is merged to merge this and ensure it does fix the problem. If not, we'll have to rethink how to merge this (probably as a pre-executed notebook).

bdg221:
> > I got the same error as @bdg221 trying to install this dev version of mitiq, and when separately running pip install `pytket-cirq`. So the install seems to fail, but the notebook itself runs fine 🤷
> 
> Do you need to install `pytket-cirq` separately? It is in the requirements files, so should be installed when running `make install`. I'm not seeing this locally, so I might need some help debugging if you are.

I can confirm with the latest commits and a new environment that `pytket-cirq` does get installed with `make install` and the docs are built without errors.

docs/source/examples/ucc.md
jordandsullivan: I like it. For me though it's confusing to have it horizontal rather than vertical. I would have the bar charts going up and down with the uncompiled result on the left. Color is up to you, but to me the gray shading seems to imply there's uncertainty in the results, Which I don't think is the intention.

jordandsullivan:
### Demo circuits 
@natestemen  did you try all of the 10 qubit circuits we have in UCC bench? Just curious if any of them showed interesting results. 

### installation
I get the pytket-cirq error about version incompatibility when I just run `make install`, So I figured I would try to install an older version of high ticket that everything was compatible with. So for some reason I'm still getting an error on install but the notebook here runs fine. 

### non-determinism
That is odd, have you tried using a debugger tool to step through each call and see where the numerics may be changing? You would just need to convert the notebook to a script to debug.

natestemen:
> did you try all of the 10 qubit circuits we have in UCC bench? 

That was a great idea! The tutorial now takes ~4 minutes to run on my machine, but it's a more robust example. Using the square heisenberg circuit from ucc-bench is much more stable numerically as well, so the lack of determinism is no longer an issue!

I'm stumped on your installation issue since I can't reproduce. Happy to debug if you'd like together.

docs/source/examples/ucc.md
natestemen: I ended up keeping horizontal bars, but livened up the colors. Let me know if you want to switch the bar order.

image

docs/source/examples/ucc.md
```diff
@@ -0,0 +1,206 @@
...
+
+```bash
+pip install mitiq
+pip install ucc
+pip install cirq
+```
+
+## 🔁 Workflow
+
+We’ll:
```
bdg221, line 33:
```suggestion
We'll be completing the following steps:
```

Not critical and could just be me being picky but "We'll:" looks a [little weird](https://mitiq--2728.org.readthedocs.build/en/2728/examples/ucc.html) on its own.

bdg221 (approved):
Good job using `uv`! Approved with a small suggestion.

bachase (approved):
👍 

Was able to run it locally. The plot is great and easy to understand!

nathanshammah (approved):
this reads great, thanks @natestemen.
Minor comments:
- On the topic of compiling a Heisenberg model simulation, I am just ignorant of what should and should not be done. Or maybe anything can be done, as long as the total unitary is the same. Pinging also @Changhao-Li for his take on this. Note that in any case the circuit can just be an example, devoid of its physical meaning. 
- I find confusing the plot with the histogram (I read the conversation above): my suggestion is to use standard plot, possibly with error bars, as that makes clearer what is the expected value of each modality (right now the unmitigated results are/seem covered by the mitigated style choice).
- About the warning on "The above plot is a single instance of a given circuit" at the end of the plot. I see that there are two options, `simulate` and `run` for Cirq [DensityMatrixSimulator](https://quantumai.google/reference/python/cirq/DensityMatrixSimulator): running the circuit could provide repetitions – is the problem there not having a seed? For high enough repetition the randomization should not be an issue for the purpose of the result.

Changhao-Li:
> this reads great, thanks @natestemen. Minor comments:
> 
> * On the topic of compiling a Heisenberg model simulation, I am just ignorant of what should and should not be done. Or maybe anything can be done, as long as the total unitary is the same. Pinging also @Changhao-Li for his take on this. Note that in any case the circuit can just be an example, devoid of its physical meaning.
> * I find confusing the plot with the histogram (I read the conversation above): my suggestion is to use standard plot, possibly with error bars, as that makes clearer what is the expected value of each modality (right now the unmitigated results are/seem covered by the mitigated style choice).
> * About the warning on "The above plot is a single instance of a given circuit" at the end of the plot. I see that there are two options, `simulate` and `run` for Cirq [DensityMatrixSimulator](https://quantumai.google/reference/python/cirq/DensityMatrixSimulator): running the circuit could provide repetitions – is the problem there not having a seed? For high enough repetition the randomization should not be an issue for the purpose of the result.

On the first point of which circuit to use, I think the Heisenberg model dynamics simulation on square lattice is a nice example for testing compilations etc. The model has rich dynamics (magnetic ordering, phase transition) and has been broadly studied using digital quantum computers in several experiments (see this [review paper](https://www.nature.com/articles/s41467-024-46402-9#Abs1)). Indeed it has also been used for device transpilation tests in [Benchpress](https://www.nature.com/articles/s43588-025-00792-y) and other compilation work ([example](https://dl.acm.org/doi/10.1145/3470496.3527394)).

Actually I have a naive question on combining ucc circuit compilation with error mitigation in `ucc.md`, would the result change if we switch the order of compilation and error mitigation? If the goal is to explore how compilation affects noise sensitivity and how mitigation can restore accuracy, the current order makes sense to me, but I'm just curious in general settings how to decide which to do first.

natestemen:
> I find confusing the plot with the histogram

The plot is a bar chart that shows 4 numbers.

1. raw
2. compiled
3. mitigated
4. compiled + mitigated

There is nothing to histogram here, and hence no errors bars to show.

> ... DensityMatrixSimulator ...

We use the `simulate` method which gives access to the final density matrix, which means there is no shot error included in these results.

The warning at the end currently reads:

> The above plot is a single instance of a given circuit. The behavior of compilation and mitigation acting beneficially is a common phenomenon, but not always guaranteed.

This is just a disclaimer that compilation + mitigation is not guaranteed to always be the most effective of the 4 compilation/mitigation combinations.

jordandsullivan (approved):

Changhao-Li:
> This is just a disclaimer that compilation + mitigation is not guaranteed to always be the most effective of the 4 compilation/mitigation combinations.

shouldn't there be 5 compilation/mitigation combinations, including the order of compilation/mitigation? (following my question above).

natestemen:
Ah yeah sorry I forgot the address that question.

Many error mitigation techniques introduce gates into the circuit to be run that compile to the identity, so compiling after EM has been applied will potentially void the EM protocol. E.g. in Zero-Noise Extrapolation a circuit $C$ will get turned into $\\{C, C C^\dagger C, C C^\dagger C C^\dagger C, \ldots\\}$. If compilation happens on this set then all that will be run is $$\\{ C, C, C, \ldots \\}$.

That said, there are other techniques such as PEC where compilation post mitigation might actually make sense.

PR #2743, opened 2025-05-08
Lighter docs build option

natestemen:
## Description

Adds a make target to build documentation without running any of the executable content. Just structure/markdown rendering.

fixes https://github.com/unitaryfoundation/mitiq/issues/2738 (maybe?)

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -44,7 +44,7 @@ docs-clean:
 
 .PHONY: docs-lite
 docs-clean:
-	DOCS_LITE=1 make -C docs html
+	uv run DOCS_LITE=1 make -C docs html
 
 .PHONY: linkcheck
 linkcheck:
--- a/Makefile
+++ b/Makefile
@@ -42,6 +42,10 @@ docs-clean:
 	uv run make -C docs clean
 	uv run make -C docs html
 
+.PHONY: docs-lite
+docs-clean:
+	DOCS_LITE=1 make -C docs html
+
 .PHONY: linkcheck
 linkcheck:
 	uv run make -C docs linkcheck
--- a/Makefile
+++ b/Makefile
@@ -43,7 +43,7 @@ docs-clean:
 	uv run make -C docs html
 
 .PHONY: docs-lite
-docs-clean:
+docs-lite:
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -42,6 +42,10 @@ docs-clean:
 	make -C docs clean
 	make -C docs html
 
+.PHONY: docs-lite
+docs-clean:
+	DOCS_LITE=1 make -C docs html
```
bdg221, line 47: There is a new Makefile using `uv`.
```suggestion
	uv run DOCS_LITE=1 make -C docs html
```

bdg221 (changes requested):
I would recommend bringing in the recent changes to main in order to utilize `uv` in the tests.

Makefile
```diff
@@ -42,6 +42,10 @@ docs-clean:
 	uv run make -C docs clean
 	uv run make -C docs html
 
+.PHONY: docs-lite
+docs-clean:
```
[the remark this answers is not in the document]
natestemen: Good eye. Must have copied this from `docs-clean`.

Makefile
```diff
@@ -42,6 +42,10 @@ docs-clean:
...
+	uv run DOCS_LITE=1 make -C docs html
```
bdg221, line 47:
```suggestion
	uv run env DOCS_LITE=1 make -C docs html
```

This tells `uv` that `DOCS_LITE=1` is setting an environment variable.

natestemen: Do you have a link to docs that show this? I'm not seeing anything here: https://docs.astral.sh/uv/reference/cli/#uv-run

natestemen: I'm dumb, I didn't realize this is the standalone [`env`](https://www.man7.org/linux/man-pages/man1/env.1.html) command.

bdg221: It is a [Linux command](https://www.man7.org/linux/man-pages/man1/env.1.html) and not part of `uv`. It appears that `uv` is expecting either a command, file, or directory and using `env` accomplishes this.

bdg221 (approved):
Looks good. Tested the latest version and it is working as expected!

PR #2756, opened 2025-05-20
Fix mutable default args

willzeng:
## Summary
- prevent shared state by using `None` defaults in DDD, PEC, and QSE modules
- add runtime initialization for optional dictionaries
- ensure imports and formatting comply with lint rules

## Testing
- `ruff check mitiq/ddd/ddd.py`
- `ruff check mitiq/pec/representations/learning.py`
- `ruff check mitiq/pec/sampling.py`
- `ruff check mitiq/qse/qse.py`
- `ruff check mitiq/qse/qse_utils.py`
- `pytest -k nothing` *(fails: command not found)*

### DIFF
--- a/mitiq/qse/qse.py
+++ b/mitiq/qse/qse.py
@@ -22,7 +22,8 @@ def execute_with_qse(
     check_operators: Sequence[PauliString],
     code_hamiltonian: Observable,
     observable: Observable,
-    pauli_string_to_expectation_cache: Dict[PauliString, complex] | None = None,
+    pauli_string_to_expectation_cache: dict[PauliString, complex]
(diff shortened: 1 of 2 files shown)


willzeng:
@natestemen This is an example of an otherwise unconfigured pull-request built made entirely by OpenAI Codex: https://openai.com/index/introducing-codex/

I think it would have caught the bug if it had been configured on the Codex side to be able to run the tests (for some reason it was trying pytest instead of uv).

If you want to polish this up and merge it then go for it. Alternatively, would be worth you taking a look into Codex to see if it can help you maintain and develop on mitiq.

PR #2769, opened 2025-05-29
Fix 0.45.0 download issues

natestemen:
## Description

Need to do a patch release to fix the missing `VERSION.txt` error. fixes https://github.com/unitaryfoundation/mitiq/issues/2767

I'll also plan to

1. release this as 0.45.1
2. yank the 0.45.0 release


bachase (approved):
Confirming this fixes the issue for me when installed locally.

```
$ uv init
$ uv pip install git+https://github.com/unitaryfoundation/mitiq@849b4d6e465a2fc98aa3e7e5f1c05ac88fc9f46c
$ uv run python -c "import mitiq"
```

natestemen:
And just to confirm, installing `main` currently fails, right?

bdg221 (approved):
I can also confirmed it is fixed (and the 0.45.0 had the issue.) I tested with `pip install`.

Issue #2624, opened 2025-01-08
Add functionality to plot ZNE results with multiple extrapolation techniques


natestemen:
## Motivation

When testing ZNE it can be hard to know which extrapolation technique to use given your data. Visualizing the results from the scaled circuit, as well as what linear/polynomial/exponential fits all look like on the same plot of your data could be very helpful in helping users pick the correct fitting function.

## Proposed Solution

```py
from mitiq import zne


folded_circuits = zne.scaled_circuits(circuit, scale_factors, scale_method)

results: list[float] = ...  # circuit execution happens here from the user

# ------------
# NEW FUNCTION
# ------------
zne.visualize_fits(scale_factors, results)  # plots data and various fits
```

Function signature might need to change, as well as naming.

q-inho:
Submitted PR #2777 implementing the requested fix. Please let me know if any changes are needed.

limitless-100:
@natestemen is this issue still open to new solutions too?

natestemen:
Technically (according to unitaryHACK rules), yes the bounty is still open. That said, #2777 is pretty close to shipping so I wouldn't encourage you to pick up this particular issue.

PR #2777, opened 2025-05-31
add `visualize_fits` to plot ZNE results with multiple extrapolation techniques

q-inho:
## Description

This PR adds a plotting helper, `visualize_fits`, to the ZNE package.  
It allows users to visualize ZNE data with linear, polynomial (order configuratble), and exponential extrapolations.
This addition resolves issue #2624 by enabling visualization of ZNE results with multiple extrapolation techniques.

---

### License

### DIFF
--- a/mitiq/zne/zne.py
+++ b/mitiq/zne/zne.py
@@ -5,12 +5,41 @@
 
 """High-level zero-noise extrapolation tools."""
 
+import warnings
 from functools import wraps
-from typing import Callable, List, Optional, Sequence, Union
+from typing import (
+    Any,
+    Callable,
+    List,
+    Optional,
+    Protocol,
+    Sequence,
+    Union,
+    cast,
+)
+
+import matplotlib.pyplot as plt
+import numpy as np
+from matplotlib.figure import Figure
 
 from mitiq import QPROGRAM, Executor, Observable, QuantumResult
-from mitiq.zne.inference import Factory, RichardsonFactory
+from mitiq.zne.inference import (
+    ExpFactory,
+    ExtrapolationError,
+    ExtrapolationResult,
+    Factory,
+    LinearFactory,
+    PolyFactory,
+    RichardsonFactory,
+)
 from mitiq.zne.scaling import fold_gates_at_random
+
+
+class SupportsExtrapolate(Protocol):
+    """Objects providing an ``extrapolate`` method."""
+
+    @staticmethod
+    def extrapolate(*args: Any, **kwargs: Any) -> ExtrapolationResult: ...
 
 
 def construct_circuits(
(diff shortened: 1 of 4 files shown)


q-inho:
Examples of the Figures by random circuits
```
from mitiq.zne import construct_circuits, combine_results
from mitiq import zne, benchmarks

import numpy as np
from cirq import DensityMatrixSimulator, depolarize
from mitiq.interface import convert_to_mitiq

circuit = benchmarks.generate_rb_circuits(n_qubits=2, num_cliffords=2)[0]

def execute(circuit, noise_level=0.01):
    mitiq_circuit, _ = convert_to_mitiq(circuit)
    noisy_circuit = mitiq_circuit.with_noise(depolarize(p=noise_level))
    rho = DensityMatrixSimulator().simulate(noisy_circuit).final_density_matrix
    return rho[0, 0].real

scale_factors = [1.0, 2.0, 3.0, 4.0]
scale_method = zne.scaling.fold_gates_at_random


folded_circuits = zne.construct_circuits(circuit, scale_factors, scale_method)
results = [execute(circ) for circ in folded_circuits]
print(results)

plot = zne.visualize_fits(scale_factors, results)
```

demonstration

Another example:
demonstration_3order

mitiq/zne/zne.py
```diff
@@ -218,3 +247,126 @@ def decorator(
...
+        scale_factors,
+        exp_values,
+        "o",
+        markersize=10,
+        markeredgecolor="black",
+        alpha=0.8,
+        label=None,
+    )
+
+    smooth_scale_factors = np.linspace(0, max(scale_factors), 20)
```
natestemen: Instead of 20, can we do something like 5x the number of scale factors?

natestemen (changes requested):
This is looking really good! I've got a few suggestions, but hopefully nothing major.

1. Lets add some horizontal padding on each side of the plot and extend the extrapolations so users can see a bit beyond $x = 0$ and the max scale factor.
2. Add an optional argument for the ideal value so a point can be plotted at $x = 0$.
3. Can richardson extrapolation (`RichardsonFactory`) be added here as a default? Do you think that is overcrowding the plot?
4. Merging this PR would mean that `matplotlib` is now a mandatory dependency. Can you move it into the main `dependencies` group (from `doc`)?
5. Lets move the `visualize_fits` function into a newly created file `mitiq/zne/viz.py` and then import them in `mitiq/zne/__init__.py`.
6. [question] Exponential fitting can be performed with less than 4 data points. Any reason you're requiring 4 here?

mitiq/zne/zne.py
q-inho: Yes. I think it would be much better. I will modify it.

q-inho:
After incorporating your suggestions, the example output is as follows.
Initially I implemented the condition for the exponential fitting requiring more than 3 data because it used to give Warning.
```
/mitiq/zne/inference.py:133: ExtrapolationWarning: The extrapolation fit may be ill-conditioned. Likely, more data points are necessary to fit the parameters of the model.
  opt_params, params_cov = curve_fit(
```
However, you are right that it can still draw the plot.

zne_fit

mitiq/zne/zne.py
```diff
@@ -6,13 +6,24 @@
...
+    RichardsonFactory,
+)
 from mitiq.zne.scaling import fold_gates_at_random
 
 
+class SupportsExtrapolate(Protocol):
+    """Objects providing an ``extrapolate`` method."""
+
+    @staticmethod
+    def extrapolate(*args: Any, **kwargs: Any) -> ExtrapolationResult: ...
```
natestemen: I think this can be moved to `viz.py` since it's only used there, right?

natestemen (approved):
One small suggestion, but will merge once that's addressed! Well done, and thank you!

mitiq/zne/zne.py
q-inho: Done!

PR #2799, opened 2025-06-26
Use qiskit throught layerwise folding tutorial

natestemen:
## Summary
- keep the layerwise folding tutorial consistently in Qiskit
- update code samples to construct, fold, and execute Qiskit circuits

fixes https://github.com/unitaryfoundation/mitiq/issues/2768

------
https://chatgpt.com/codex/tasks/task_e_685dc94c8c7c832c853b98e4ffcaaaf6

### DIFF
--- a/docs/source/examples/layerwise-folding.md
+++ b/docs/source/examples/layerwise-folding.md
@@ -16,41 +16,39 @@ kernelspec:
 
 # ZNE with Qiskit: Layerwise folding
 
-
 This tutorial shows an example of how to mitigate noise on IBMQ backends using
 layerwise folding in contrast with global folding.
 
 One may ask why folding by layer is potentially beneficial to consider. One
 reason is that applying global folding will increase the length of the entire
 circuit while layerwise folding on a subset of only the noisiest layers will
-increase the circuit by a smaller factor. 
+increase the circuit by a smaller factor.
 
 If running a circuit on hardware is bottle-necked by the cost of running a long
 circuit, this technique could potentially be used to arrive at a better result
 (although not as good as global folding) but with less monetary cost.
 
-More information on the layerwise folding technique can be found in 
-*Calderon et al. Quantum (2023)* {cite}`Calderon_2023_Quantum`.
-
+More information on the layerwise folding technique can be found in
+_Calderon et al. Quantum (2023)_ {cite}`Calderon_2023_Quantum`.
 
 ## Setup
 
 ```{code-cell} ipython3
+import matplotlib.pyplot as plt
 import numpy as np
 import qiskit
-import matplotlib.pyplot as plt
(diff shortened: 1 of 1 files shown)


docs/source/examples/layerwise-folding.md
```diff
@@ -16,55 +16,39 @@ kernelspec:
...
-from mitiq.interface.mitiq_qiskit.qiskit_utils import initialized_depolarizing_noise
-from mitiq.interface.mitiq_qiskit.conversions import to_qiskit
-from mitiq.interface.mitiq_cirq.cirq_utils import sample_bitstrings
+from mitiq.interface.mitiq_qiskit.qiskit_utils import sample_bitstrings
+from mitiq.zne.scaling.layer_scaling import layer_folding
 
-from cirq.contrib.svg import SVGCircuit
-from qiskit_aer import QasmSimulator
+noise_model = NoiseModel()
+noise_model.add_all_qubit_quantum_error(depolarizing_error(0.02, 1), ["x"])
```
bachase, line 50: This looks different than what the prior function was doing: 
https://github.com/unitaryfoundation/mitiq/blob/c051fd2fcbf304a200329030c42044c8a451f010/mitiq/interface/mitiq_qiskit/qiskit_utils.py#L24

Is it intended to add to `"x"` whereas the old version was adding to other gates?

natestemen: Previously, the tutorial had two executions of running circuits:

1. [`mitiq.interface.mitiq_cirq.sample_bitstrings`](https://github.com/unitaryfoundation/mitiq/blob/c051fd2fcbf304a200329030c42044c8a451f010/mitiq/interface/mitiq_cirq/cirq_utils.py#L17) which by default has amplitude damping noise built in (unless you override). This is what was used to generate the plot showing how much error each layer accrued.
2. A `qiskit_aer.QasmSimulator` with a noise model constructed from [`mitiq.interface.mitiq_qiskit.qiskit_utils.initialized_depolarizing_noise`](https://github.com/unitaryfoundation/mitiq/blob/c051fd2fcbf304a200329030c42044c8a451f010/mitiq/interface/mitiq_qiskit/qiskit_utils.py#L24). That function adds a depolarizing channel after any instances of `{U1, U2, U3, CX}`. This worked with Cirq circuits because they were being translated to Qiskit and effectively re-basis-ed to include `Ux` gates (and hence adding noise).

Now that we've removed cirq from the tutorial (and hence frontend translations), circuits like an all-$X$ circuit don't get rewritten into gates affected by `initialized_depolarizing_noise`, so they would appear noiseless under the Qiskit backend.

To simplify the flow, I replaced both execution methods with a single one using the custom noise model that explicitly applies noise to the dummy circuit.

PR #2809, opened 2025-07-17
Add `compare_cost` utility function to show overhead of QEM circuits

### LINKED ISSUE #2790: Add feature to show overhead of QEM circuits
## Motivation

We added `mitiq.xyz.construct_circuits` as a workflow to many techniques, but users are still missing tools to compare how much more this will cost them. The goal with this feature is to help users understand how much a QEM experiment will cost.

## Suggested workflow

```py
import mitiq

circuit = ...

qem_circuits = mitiq.xyz.construct_circuits(circuit, ...)

mitiq.compare_cost(circuit, qem_circuits) # <<< new function
```

This function should tell you how many extra circuits are required, how many extra gates (broken down by arity). Having an optional `shots` argument could be nice to show how many shots each QEM circuit will have

q-inho:
## Description
Add feature, `compare_cost` function in `utils.py` to show overhead of QEM circuits. 
`compare_cost` function begins by translating the original circuit and each mitigated variant into a consistent internal format, then counts how many one‑qubit, two‑qubit, and larger‑arity gates appear across all mitigated circuits. By subtracting the gate totals of the original circuit, it determines the extra gates introduced by the mitigation procedure.
Resolve #2790 .

For example
```python
import mitiq
import cirq
import numpy as np

q0, q1, q2 = cirq.LineQubit.range(3)
circuit = cirq.Circuit(
    cirq.H(q0),
    cirq.CNOT(q0, q1),
    cirq.measure(q0, q1),
)
qem_circuit = mitiq.zne.construct_circuits(
    circuit, [1.0, 3.0], mitiq.zne.scaling.fold_global
)
cost = mitiq.compare_cost(circuit, qem_circuit)
print(f"Cost of the original circuit 1: {cost}")

circuit2 = cirq.Circuit(
    cirq.H(q0),
    cirq.H(q1),
    cirq.CNOT(q0, q1),
    cirq.CCNOT(q0, q1, q2),
    cirq.measure(q0, q1, q2),
)
qem_circuit2 = mitiq.zne.construct_circuits(
    circuit2, [1.0, 3.0, 5.0], mitiq.zne.scaling.fold_global
)
cost2 = mitiq.compare_cost(circuit2, qem_circuit2, shots=100)
print(f"Cost of the original circuit 2: {cost2}")
```
now it gives
```
Cost of the original circuit 1: {'extra_circuits': 1, 'gate_overhead': {'1q': 3, '2q': 3, 'nq': 0}}
Cost of the original circuit 2: {'extra_circuits': 2, 'gate_overhead': {'1q': 16, '2q': 8, 'nq': 8}, 'total_shots': 300}
```

---

### License


mitiq/utils.py
```diff
@@ -7,7 +7,7 @@
 
 from copy import deepcopy
 from itertools import product
-from typing import Any
+from typing import Any, Sequence
```
natestemen: `Sequence` should be imported from `collections.abc` since `typing.Sequence` is deprecated.

mitiq/utils.py
```diff
@@ -385,3 +388,61 @@ def qem_methods() -> dict[str, str]:
...
+        key: total_counts[key] - base_counts.get(key, 0)
+        for key in total_counts
+    }
+
+    result: dict[str, int | dict[str, int]] = {
+        "extra_circuits": len(qem_circuits) - 1,
+        "gate_overhead": gate_overhead,
+    }
+    if shots is not None:
+        result["total_shots"] = shots * len(qem_circuits)
```
natestemen: Total shots should be kept constant when comparing experiments, so it makes more sense to divide total shots by the number of qem circuits.

mitiq/tests/test_utils.py
natestemen, line 1: Did you try any testing with non-cirq circuits?

natestemen (commented):
Thanks for getting this started! I think there are going to be some edge cases where the depth of circuits changes when converting from X to cirq to do the gate counts if we architect it this way. I'm not sure there's a better way around that, but maybe we should warn the user? WDYT?

mitiq/utils.py
q-inho: Thanks. I modify this `total_shots` to `shots_per_circuit`.

mitiq/tests/test_utils.py
q-inho: I only tested with Qiskit for non-cirq circuits by demo code. Do you think it would be better to add all available framework to this test file?

q-inho:
> Thanks for getting this started! I think there are going to be some edge cases where the depth of circuits changes when converting from X to cirq to do the gate counts if we architect it this way. I'm not sure there's a better way around that, but maybe we should warn the user? WDYT?

@natestemen Yes. I think it would be problem, since cirq will not provide and not one-to-one compatiable with each other. Since, most of framework provide function counting their operations like Qiskit's `QuantumCircuit.count_ops()`, maybe I can map each available framework for counting their operations. What do you think about this method? Do you have any suggestion?

mitiq/tests/test_utils.py
natestemen: If you wouldn't mind trying: testing this function using all supported frontend types can be done with a pytest parametrization:

https://github.com/unitaryfoundation/mitiq/blob/4fab6701627e041c79ac2cdc800c4387e9420b64/mitiq/zne/tests/test_zne.py#L481

Even if this PR doesn't end up supporting all of them because of a difficulty, it would be good to know why that is.

natestemen:
> most of framework provide function that couns their circuit's operations like Qiskit's `QuantumCircuit.count_ops()`

I'm happy for you to explore this if you want to see what's available in each frontend. If it doesn't end up working, lets stick with the simple thing of adding a warning saying there may be some discrepancies due to conversions.

q-inho:
@natestemen According to my exploration of the mapping for each program, it seems the current folding conversion seems quite not working correctly especially for Pyquil due to its conversion to cirq, when it do `fold_global`.
For Braket and Pennylane, I am still exploring.

natestemen (approved):
Apologies for the delay! This looks good. The one request I would make is that we move most of the utility functions from `mitiq/utils.py` into a **new** file `mitiq/interface/utils.py` since I think they more naturally belong here. WDYT?

q-inho:
@natestemen Sorry for late update. I moved the `compare_cost` related files from `mitiq/utils.py` to new file `mitiq/interface/utils.py`.
I would like to ask whether we should need separate test file for this new `interface/utils.py` file testing.

natestemen (approved):
Fine by me to keep it as is! Thanks for your help here.

PR #2823, opened 2025-08-29
Update calibration screencast in docs

### LINKED ISSUE #2592: Screencast in doc main page shows wrong option for calibration logs
The screencast in the https://mitiq.readthedocs.io/en/stable/index.html#calibration shows that the following option as working: 
```python3
cal = Calibrator(execute, frontend="cirq")
cal.run(log=True)  # <-- bad option for arg log
```
However, the `log` arg of `run()` doesn't accept a Boolean value, but only a value from the [OutputForm enum](https://github.com/unitaryfund/mitiq/blob/fix-zne-doc-asymptote/mitiq/calibration/calibrator.py#L36).

Doc should be fixed accordingly.

bdg221:
## Description
Fixes #2592 

Does what it says on the 🥫!  

The old screencast used an invalid log parameter. Also, the log output is now different.


bdg221:
@natestemen let me know if the delay at the end is long enough to see the best strategy results.

natestemen (approved):
That looks great!

Issue #2815, opened 2025-08-01
upgrade cirq to 1.6


natestemen:
Two minor releases are available (1.5 and 1.6). 1.6 removes support for pyquil, so we may want to hold off on that for now, or at least investigate if there is a way to continue supporting pyquil without the removed `cirq-rigetti` module.

natestemen:
We may be able to continue supporting pyquil by switching these conversions to the qbraid transpiler: https://github.com/qBraid/qBraid/tree/38401204b76f92cdee0d74aef148b9d8bf359dbb/qbraid/transpiler/conversions/pyquil

bdg221:
pyQuil does offer a `pyquil_to_cirq` and `cirq_to_pyquil`. The `cirq_to_pyquil` uses QASM and when running all of the tests, I discovered generated circuits (like w_state, mirror_qv, qpe, and quantum_volume) include global phase operations. 
`qbraid.transpiler.exceptions.ProgramConversionError: Cirq qasm converter doesn't yet support  cirq.global_phase_operation((-0.5584661286763627-0.8295273251203003j)).`

```
from qbraid.transpiler.conversions.cirq import cirq_to_pyquil
from mitiq.benchmarks.mirror_qv_circuits import generate_mirror_qv_circuit
circuit = generate_mirror_qv_circuit(4, 3, return_type="cirq")
pyquil_program = cirq_to_pyquil(circuit)
```

bdg221:
I don't think there is a perfect solution for PyQuil and Cirq 1.6. Currently, QBraid's `cirq_to_pyquil` throws errors if there is a global phase. Technically, that is a QASM 2.0 thing, but Qiskit's conversions just give a warning and remove the global phase automatically. This was found out because Mitiq's benchmarking circuits include global phases. So, we could use QBraid's functions with a warning in the documentation and in the docstring that the circuit must be QASM 2.0 compliant (and modify current tests.)

bdg221:
A potential workaround for this issue would be to use Mitiq's `to_qasm()` and `from_qasm()` as this generate the warning, but does not throw an error.

```
from qbraid.transpiler.conversions.cirq import cirq_to_pyquil

from mitiq.benchmarks.mirror_qv_circuits import generate_mirror_qv_circuit
from mitiq.interface.mitiq_qiskit import from_qasm, to_qasm

circuit = generate_mirror_qv_circuit(4, 3, return_type="cirq")
circuit = from_qasm(to_qasm(circuit))
[...]

PR #2832, opened 2025-09-30
Qbraid integration for pyquil

bdg221:
This PR removes the `cirq-rigetti` dependency which is deprecated in Cirq 1.5 and removed in Cirq 1.6. The functionality is replaced with qBraid's implementation which currently supports Cirq 1.4 and 1.5.

This PR is a necessary step for completing #2815 

### License

### DIFF
--- a/mitiq/interface/mitiq_pyquil/conversions.py
+++ b/mitiq/interface/mitiq_pyquil/conversions.py
@@ -7,10 +7,10 @@
 pyQuil's circuit representation (Quil programs).
 """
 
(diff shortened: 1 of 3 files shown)


natestemen (approved):
Short and sweet! Were you able to run the pyquil tests that require docker locally? I only ask since the pyquil tests only run on ubuntu as part of CI and I wonder if they were also tested on your machine (mac, right?).

bdg221:
The pyquil tests did work on my mac:

image

PR #2833, opened 2025-10-01
Support Cirq 1.5

b-goldsmith:
Updates the `pyproject.toml` to include support for Cirq 1.5. Also, due to a change in the Cirq QasmParser, a Qiskit qft conversion test failed. While the gates and gate counts are still the same, the QasmParser now includes qubit mapping which results in the `_equals()` method from `mitiq.utils` no longer considering the converted circuit to be equal. Therefore, the assertion was changed to check the type of gates and their counts to ensure the circuits were equal.

Running `make install` with the new `pyproject.toml` and `uv.lock` still prefers to install Cirq 1.4. However, I explicitly tested with Cirq 1.5 and confirmed that all tests passed.

### DIFF
--- a/mitiq/interface/mitiq_qiskit/tests/test_conversions_qiskit.py
+++ b/mitiq/interface/mitiq_qiskit/tests/test_conversions_qiskit.py
@@ -6,7 +6,6 @@
 """Unit tests for conversions between Mitiq circuits and Qiskit circuits."""
 
 import copy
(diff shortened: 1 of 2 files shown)


bdg221:
By the way, here is an explanation of the test that was failing and why: https://github.com/unitaryfoundation/mitiq/issues/2815#issuecomment-3339143649

PR #2835, opened 2025-10-03
Bumped `myst_nb` and `nbsphinx` to try and resolve #2723

### LINKED ISSUE #2723: Sphinx 9 build warnings
The docs build generates a warning for each generated page of the documentation. 

https://github.com/unitaryfoundation/mitiq/actions/runs/14456508494/job/40540830340#step:7:1342

It does appear like the issue will be solved on its own when the Sphinx is upgraded to v9. But if not, we should keep track of it.

ACE07-Sev:
Bumped `myst_nb` and `nbsphinx` to try and resolve #2723

### License

- [ x] I license this contribution under the terms of the GNU GPL, version 3 and grant Unitary Foundation the right to provide additional permissions as described in section 7 of the GNU GPL, version 3.


natestemen (approved):
Nice and easy!

PR #2837, opened 2025-10-10
Cirq 1.6 update

### LINKED ISSUE #2815: upgrade cirq to 1.6

b-goldsmith:
Fixes #2815 

This updates Mitiq to use Cirq 1.6 which caused a few issues that needed correction like:

- A qft test sometimes fails during conversion from qiskit, the assert was updated to ensure it the circuits are equal up to a global phase.
- The `pauli.py` and `observables.py` now hit a typing error with the update. 
- The `autoray` dependency is pinned since a newer version exists, but Pennylane does not currently support that version.
- Removes Pyton 3.10 support.

- [X ] I license this contribution under the terms of the GNU GPL, version 3 and grant Unitary Foundation the right to provide additional permissions as described in section 7 of the GNU GPL, version 3.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -87,8 +87,8 @@ jobs:
         run: |
           uv python pin ${{ matrix.python-version }}
           uv python install
-          uv pip install setuptools>=61
           uv sync --all-extras
+          uv pip install setuptools>=61
 
       - name: Run the quilc & qvm Docker images
         if: ${{ matrix.os == 'ubuntu-latest' }}


.github/workflows/build.yml
```diff
@@ -86,8 +86,9 @@ jobs:
       - name: Install Python ${{ matrix.python-version }} and Python dependencies except for docs
         run: |
           uv python pin ${{ matrix.python-version }}
-          uv python install 
+          uv python install
           uv sync --all-extras
+          uv pip install setuptools>=61
```
natestemen, line 91: Were you able to find anything about why setuptools was removed when setting up `uv`? Just curious.

If I cloned mitiq and ran `uv sync --all-extras` and then tried to run the tests (all locally), that should fail since I don't have setuptools. Is that right?

b-goldsmith: I found it!!! The follow outputs are from running `uv tree`:

From #2835 which still worked:
├── pennylane-qiskit v0.36.0 (extra: pennylane)
│   ├── networkx v3.4.2
│   ├── numpy v1.26.4
│   ├── pennylane v0.36.0 (*)
│   ├── qiskit v1.4.2
│   │   ├── dill v0.4.0
│   │   ├── numpy v1.26.4
│   │   ├── python-dateutil v2.9.0.post0 (*)
│   │   ├── rustworkx v0.16.0 (*)
│   │   ├── scipy v1.15.2 (*)
│   │   ├── stevedore v5.4.1
│   │   │   └── pbr v6.1.1
│   │   │       └── setuptools v80.9.0

From this PR:
├── pennylane-qiskit v0.36.0 (extra: pennylane)
│   ├── networkx v3.5
│   ├── numpy v1.26.4
│   ├── pennylane v0.36.0 (*)
│   ├── qiskit v1.4.5
│   │   ├── dill v0.4.0
│   │   ├── numpy v1.26.4
│   │   ├── python-dateutil v2.9.0.post0 (*)
│   │   ├── rustworkx v0.17.1 (*)
│   │   ├── scipy v1.15.2 (*)
│   │   ├── stevedore v5.5.0
│   │   ├── symengine v0.13.0
│   │   ├── sympy v1.14.0 (*)
│   │   └── typing-extensions v4.15.0


I believe this was the result of running `uv sync --upgrade` which tries to find the latest version of all dependencies that are compatible. This was done with this PR, since a new `uv.lock` is being used.

natestemen: For whatever reason I wanted to find evidence that `stevedore` dropped `setuptools` as a dependency, so I went looking in their source. The best I can find is the explicitly drop support for [`pbr`](https://opendev.org/openstack/pbr) in https://opendev.org/openstack/stevedore/commit/3c04bde90db9e6d1ab459ad507b48f3894e6832c and `pbr` does indeed [require `setuptools`](https://opendev.org/openstack/pbr/src/commit/a0d431c5dec60b674ba726323edf442c845de2c2/requirements.txt).

Let's 🚢.

Issue #1503, opened 2022-09-22
Directly support openQASM circuits


nathanshammah:
## Issue Description

Mitiq could support directly as one of its frontends quantum circuits in openQASM format. Mitiq already has some conversion functions for openqasm wrt to Qiskit circuits.

## Additional References

@nickdgardner mentioned this as a useful feature.

nickdgardner:
I've started looking at this (specifically for openQASM 3) and would be happy to collaborate on the issue with others who have more experience with Mitiq and/or openQASM.

andreamari:
Interesting @nickdgardner !  This probably requires some planning/decisions. We can discuss more during a Mitiq call.

natestemen:
Can someone lay out the value prop for this? From my understanding very few users are writing openqasm by hand, and many frontends can import/export from openqasm easily.

nickdgardner:
I don't see a clear one now that I've looked into it more. But maybe @andreamari or @nathanshammah had something in mind? My initial thought was that it would be useful if Mitiq aims to eventually support people who want to apply mitigation methods involving sub-gate-level control, who might be working directly in QASM, and if QASM submodules would be the best way for Mitiq to implement those methods internally. But I see now this would be a huge undertaking, and I don't even know if it's a plausible scenario.

andreamari:
QASM(2) is a [very popular](https://docs.google.com/presentation/d/e/2PACX-1vRjda57eKp7sLZENSJ7x83To2r0ja7XmW5sYaEobmYlvDKw122gagKm1j_qYm820054VbpFL0zZGjug/pub?start=false&loop=false&delayms=3000&slide=id.g14d46be9347_0_101) intermediate representation for circuits . 

Supporting QASM would make Mitiq easy to use with any software library that is able to export QASM circuits and any backend that is able to run QASM circuits. 

**Short term value prop:**
It would simplify the code of the user.
E.g., if Mitiq supported QASM as a fronted, there would be no need to manually convert circuits to Cirq in the [BSQkit tutorial](https://mitiq.readthedocs.io/en/stable/examples/bqskit.html).
Mitiq would do it internally.  So no fundamental advantages, just more user-friendly. 

**Long term value prop:**
In the long term, Mitiq could act directly on some fixed and unique intermediate representation (e.g. QASM2, QASM3 or QIR). This would  imply rewriting almost all the code of Mitiq which is instead Cirq-based, but it has some clear potential advantages. 

[...]

PR #2842, opened 2025-10-26
OpenQASM3 Frontend

ACE07-Sev:
## Description

Initial commit to tackle #1503 

---

### License

### DIFF
--- a/mitiq/interface/conversions.py
+++ b/mitiq/interface/conversions.py
@@ -81,6 +81,11 @@ def convert_to_mitiq(circuit: QPROGRAM) -> tuple[cirq.Circuit, str]:
     """
     conversion_function: Callable[[Any], cirq.Circuit]
 
+    if isinstance(circuit, str) and hasattr(circuit, "__module__") is False:
+        from mitiq.typing import QasmStringType
+
+        circuit = QasmStringType(circuit)
+
     try:
         package = circuit.__module__
     except AttributeError:
--- a/mitiq/interface/utils.py
+++ b/mitiq/interface/utils.py
@@ -138,7 +138,6 @@ def _count_gate_arities_qibo(circuit: Any) -> dict[str, int]:
     return counts
 
 
-# TODO: Temporary solution
 def _count_gate_arities_openqasm(circuit: Any) -> dict[str, int]:
     """Counts gates in a OpenQASM circuit grouped by arity."""
     from mitiq.interface.conversions import convert_to_mitiq
--- a/mitiq/tests/test_conversions.py
+++ b/mitiq/tests/test_conversions.py
@@ -122,7 +122,7 @@ def test_register_from_to_mitiq():
     assert input_type == qasm_circuit.__module__
 
 
-@pytest.mark.parametrize("item", ("circuit", 1, None))
+@pytest.mark.parametrize("item", (1, None))
 def test_to_mitiq_bad_types(item):
     with pytest.raises(
         UnsupportedCircuitError,
@@ -185,6 +185,20 @@ def test_converter(circuit_and_type):
     cirq_scaled = scaling_function(circuit, return_mitiq=True)
     assert isinstance(cirq_scaled, cirq.Circuit)
     assert _equal(cirq_scaled, cirq_circuit)
+
+
+def test_qasm_string_converter():
+    circuit = qiskit.QuantumCircuit(2)
+    circuit.h(0)
+    circuit.cx(0, 1)
+    qasm_str = qiskit.qasm3.dumps(circuit)
+
+    cirq_circuit = cirq.Circuit()
+    qr = cirq.LineQubit.range(2)
+    cirq_circuit.append(cirq.ops.H(qr[0]))
+    cirq_circuit.append(cirq.ops.CX(qr[0], qr[1]))
+
+    assert _equal(cirq_circuit, convert_to_mitiq(qasm_str)[0])
 
 
 @pytest.mark.parametrize("nbits", [1, 10])


mitiq/interface/mitiq_openqasm/conversions.py
```diff
@@ -0,0 +1,37 @@
...
+def from_openqasm(circuit: Program) -> Circuit:
+    """Convert a OpenQASM program to a cirq Circuit.
+    Args:
+        circuit (str): The OpenQASM program to convert.
+    Returns:
+        Circuit: The converted cirq Circuit.
+    """
+    from mitiq.interface.mitiq_qiskit import from_qasm
+
+    return from_qasm(openqasm3_to_qasm3(circuit))
```
natestemen: I think this will require qiskit to be installed to use qasm. Is it possible to only rely on qbraid/cirq functionality or do I have something mixed up here?

mitiq/interface/conversions.py
```diff
@@ -83,6 +83,7 @@ def convert_to_mitiq(circuit: QPROGRAM) -> tuple[cirq.Circuit, str]:
 
     try:
         package = circuit.__module__
+        print(package)
```
natestemen: oops!

mitiq/interface/mitiq_openqasm/conversions.py
```diff
@@ -0,0 +1,37 @@
...
+OpenQASM's circuit representation.
+"""
+
+from cirq import Circuit
+from openqasm3.ast import Program
+from qbraid.transpiler.conversions.openqasm3 import openqasm3_to_qasm3
+from qbraid.transpiler.conversions.qasm3 import qasm3_to_openqasm3
+
+
+def from_openqasm(circuit: Program) -> Circuit:
```
natestemen: I think a more user friendly interface might be to read in strings directly. WDYT about that instead of these `openqasm3.ast.Program` object? From what I've seen folks usually use qasm as strings/files and meeting them there would likely be beneficial.

natestemen (commented):
Awesome start! A few high level comments to align on before we get into the weeds.

mitiq/interface/mitiq_openqasm/conversions.py
ACE07-Sev: I think I tried that initially but it had issues in conversions. Your `from_qasm` was much better at the conversions hence why I've been using that.

mitiq/interface/conversions.py
ACE07-Sev: Oops indeed. I'll remove it.

mitiq/interface/mitiq_openqasm/conversions.py
ACE07-Sev: Sure, I'll try to make it `str`. I mainly just checked the type of what `openqasm3` returns and that was that. I'll have to see if it accepts `str` or not.

ACE07-Sev: Nate, just to confirm, you want the `from_openqasm` to take a string instead of `Program` or just `to_openqasm`? Because then that would just be `from_qasm`.

natestemen: Maybe it's worth a quick github search to see if there are many instances of people working with the `openqasm3.ast.Program` object, because from my experience I've never seen that! I more often see people define circuits in some SDK then export to QASM and pass around files/strings.

mitiq/executor/executor.py
```diff
@@ -346,7 +346,7 @@ def _call_executor(
 
         if self.can_batch:
             self._quantum_results.extend(result)
-            self._executed_circuits.extend(to_run)
+            self._executed_circuits.extend(to_run)  # type: ignore
```
natestemen: What's with all the type ignore comments? Why are they needed the the new QPROGRAM addition?

ACE07-Sev: Oh that's because we assume it's working with `cirq.Circuit` instance in those lines (with the decorator converting from passed circuit type to cirq) but since it is type hinted as `QPROGRAM` it expects all types within `QPROGRAM` to behave the same as `cirq.Circuit` in terms of what dunders they support.

So, when I added `openqasm3.ast.Program` mypy complained about that since it doesn't support certain dunders.

I temporarily added them until it's confirmed whether we're going with `openqasm3.ast.Program` or simply `str`.

mitiq/interface/mitiq_openqasm/conversions.py
ACE07-Sev: Sure thing. I'll go through the libraries that support it and see what dtype they use for it.

ACE07-Sev:
@natestemen Greetings,

Hope you're doing well. Anything I need to do for this PR?

docs/source/apidoc.md
```diff
@@ -392,6 +392,13 @@ See Ref. {cite}`Czarnik_2021_Quantum` for more details on these methods.
    :members:
 ```
 
+#### Pennylane Conversions
```
natestemen, line 395: Oh! Thanks for adding this.

mitiq/interface/conversions.py
```diff
@@ -114,6 +114,12 @@ def convert_to_mitiq(circuit: QPROGRAM) -> tuple[cirq.Circuit, str]:
         input_circuit_type = "qibo"
         conversion_function = from_qibo
 
+    elif "openqasm3.ast" in package:
+        from mitiq.interface.mitiq_openqasm.conversions import from_openqasm
+
+        input_circuit_type = "openqasm"
+        conversion_function = from_openqasm
```
natestemen, line 126: I don't think this `elif` block will ever be entered in the case when `circuit` is a string. In fact won't this function raise an `UnsupportedCircuitError` if you passed a string?

If so, let's make sure to add a test to cover this case.

mitiq/interface/utils.py
```diff
@@ -138,6 +138,14 @@ def _count_gate_arities_qibo(circuit: Any) -> dict[str, int]:
     return counts
 
 
+# TODO: Temporary solution
```
natestemen: Can you leave a more descriptive note as to what needs to be improved here if it's temporary? Looks okay to me tbh.

mitiq/interface/conversions.py
ACE07-Sev: I set the `__module__` to `openqasm3.ast` in https://github.com/unitaryfoundation/mitiq/pull/2842/files#diff-48442767e88b3833575b4ec39d72654d6ec7d89848f8988468d76c6ae26618f9R69-R75. If the user passes a string, it gets converted to that class first, so it allows the code to work with `str`.

mitiq/interface/utils.py
ACE07-Sev: Fine to me as well, just thought maybe you wanted to do it without using `convert_to_mitiq`. If this is okay then I can remove that comment.

mitiq/interface/conversions.py
ACE07-Sev: Copy that.

mitiq/typing.py
```diff
@@ -65,10 +65,28 @@ def keys(cls) -> list[str]:
...
     _QiboCircuit = _Circuit
 
+try:
+    from openqasm3.ast import Program
+
+    class QasmStringType(str):
+        __module__ = Program.__module__
+
+        def copy(self) -> "QasmStringType":
+            return self
```
natestemen, line 75: Why does this class need a copy method?

mitiq/interface/conversions.py
```diff
@@ -81,6 +81,11 @@ def convert_to_mitiq(circuit: QPROGRAM) -> tuple[cirq.Circuit, str]:
     """
     conversion_function: Callable[[Any], cirq.Circuit]
 
+    if isinstance(circuit, str) and hasattr(circuit, "__module__") is False:
```
natestemen, line 84:
```suggestion
    if isinstance(circuit, str):
```

WDYT of simplifying this `if`? Technically what you have is more correct, but I think it's a bit more readable and not much of a safety sacrifice to remove it.

mitiq/typing.py
ACE07-Sev: I think you use `.copy` on `QPROGRAM` somewhere hence why I put it in.

mitiq/interface/conversions.py
ACE07-Sev: Oh, the reason I added that is because without it this fails.
https://github.com/ACE07-Sev/mitiq/blob/openqasm-frontend/mitiq/tests/test_conversions.py#L107C1-L122C49

Failing because it has to assert `"openqasm" == "qasm"`. So, I just convert it from `str` to `QasmStringType` when it's a pure string, not something like `QasmStringType`.

natestemen: Ahh okay good find. Thanks for the explanation.

PR #2843, opened 2025-10-27
Move pennylane import and code inside check for module

### LINKED ISSUE #2812: remove pennylane as a required dependency when using `mitiq.pt`
Pennylane is currently a required dependency when using `mitiq.pt` due to a difference in how noisy simulations are pennylane (https://github.com/unitaryfoundation/mitiq/pull/2397).

Mainn code that needs to be changed is here: https://github.com/unitaryfoundation/mitiq/blob/1f3cecf917a268fd199f5b35ac951e5244372130/mitiq/pt/pt.py#L145-L147

b-goldsmith:
## Description
Fixes #2812 

Previously, `pennylane` was a required module in the Pauli Twirling portion of the Mitiq code (`mitiq.pt`). This PR moved the Pennylane related import and code inside of a check to see if the `pennylane` module is installed. Therefore, this removes `pennylane` from being a required module in order to use Mitiq's Pauli Twirling.

Previously, `pennylane` was a required dependency for Mitiq's Pauli Twirling(mitiq.pt).
This PR moves the PennyLane import and related code inside a runtime check for whether the `pennylane` module is installed.
As a result, `pennylane` is no longer a required dependency to use Mitiq’s Pauli Twirling functionality.

### DIFF
--- a/mitiq/pt/pt.py
+++ b/mitiq/pt/pt.py
@@ -3,14 +3,13 @@
 # This source code is licensed under the GPL license (v3) found in the
 # LICENSE file in the root directory of this source tree.
 
+import importlib.util
 import random
 from collections.abc import Callable
 from functools import singledispatch
 
 import cirq
-import pennylane as qml
 from cirq import Circuit as _Circuit
-from pennylane.tape import QuantumTape
 
 from mitiq import QPROGRAM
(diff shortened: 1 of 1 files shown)


natestemen:
Ahh right okay that makes sense! Merge away.

PR #2875, opened 2025-11-14
0.48.1 release

natestemen:
## Description

Patch release to support newer versions of qiskit and pennylane!


bdg221 (approved):
linkcheck is failing, but that is not surprising.

PR #2915, opened 2026-01-27
chore: updating docs to correctly reflect num_chunks argument in lre.

### LINKED ISSUE #2914: Improve docs for `num_chunks` argument in the layerwise Richardson extrapolation (LRE) module
In many functions of the layerwise Richardson extrapolation module `mitiq.lre` the docstring says that:

```
num_chunks: Number of desired approximately equal chunks. When the
            number of chunks is the same as the layers in the input circuit,
            the input circuit is unchanged.
```

I think it could be improved since it is a bit unclear and the final sentence is actually wrong.  When the number of chunks is the same as the layers in the input circuit, LRE is applied in its default way and the circuit is changed accordingly.

I write here a simple draft, just to explain what `num_chunks` actually means. Feel free to better re-write.
```
num_chunks: The number of approximately equal-depth circuit chunks.
            Each chunk is treated by LRE as an independent elementary layer, i.e., 
            the smallest unit over which noise is uniformly scaled. The minimum
            value is 1 (reducing to the mitiq.zne technique), and the maximum
            value is the full circuit depth (the default).
```

Remark: `num_chunks` appears in many public and private and functions within `mitiq.lre` . So, if improved, it should be consistently changed everywhere.

vprusso:
Closes: #2914 

Updated documentation more clearly explains what `num_chunks` represents (equally-sized circuit chunks) and also how noise-scaling applies to chunks (independently per chunk)
 
Additionally, the docs now explicitly mention the valid range (1 to number of layers) as well as the default behavior (each layer is its own chunk). The factually incorrect statement that the circuit is "unchanged" has been removed.

### DIFF
--- a/mitiq/lre/inference/multivariate_richardson.py
+++ b/mitiq/lre/inference/multivariate_richardson.py
@@ -73,7 +73,7 @@ def sample_matrix(
         fold_multiplier: Scaling gap required by unitary folding.
         num_chunks: The number of equally-sized circuit chunks. Noise
             scaling is applied to each chunk independently. Ranges from 1
-            (all gates in one chunk, similar to global folding in ZNE) to the
+            (all gates in one chunk, similar to ZNE) to the
             number of circuit layers (default, each layer is a separate chunk).
 
     Returns:
--- a/mitiq/lre/inference/multivariate_richardson.py
+++ b/mitiq/lre/inference/multivariate_richardson.py
@@ -71,9 +71,10 @@ def sample_matrix(
         input_circuit: Quantum circuit to be scaled.
         degree: Degree of the multivariate polynomial.
         fold_multiplier: Scaling gap required by unitary folding.
-        num_chunks: Number of desired approximately equal chunks. When the
-            number of chunks is the same as the layers in the input circuit,
-            the input circuit is unchanged.
+        num_chunks: The number of equally-sized circuit chunks. Noise
+            scaling is applied to each chunk independently. Ranges from 1
+            (all gates in one chunk, similar to ZNE) to the number of circuit
+            layers (default, each layer is a separate chunk).
 
     Returns:
         Matrix of the evaluated monomial basis terms from the scale factor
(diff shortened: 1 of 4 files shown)


mitiq/lre/inference/multivariate_richardson.py
```diff
@@ -71,9 +71,10 @@ def sample_matrix(
         input_circuit: Quantum circuit to be scaled.
         degree: Degree of the multivariate polynomial.
         fold_multiplier: Scaling gap required by unitary folding.
-        num_chunks: Number of desired approximately equal chunks. When the
-            number of chunks is the same as the layers in the input circuit,
-            the input circuit is unchanged.
+        num_chunks: The number of equally-sized circuit chunks. Noise
+            scaling is applied to each chunk independently. Ranges from 1
+            (all gates in one chunk, similar to ZNE) to the number of circuit
+            layers (default, each layer is a separate chunk).
```
natestemen:
```suggestion
            (all gates in one chunk, similar to global folding in ZNE) to the
            number of circuit layers (default, each layer is a separate chunk).
```

WDYT?

natestemen (approved):
Thanks for hopping on this @vprusso!

mitiq/lre/inference/multivariate_richardson.py
natestemen: @vprusso should we apply this change in the other docstrings as well?

mitiq/lre/inference/multivariate_richardson.py
```diff
@@ -71,9 +71,10 @@ def sample_matrix(
         input_circuit: Quantum circuit to be scaled.
         degree: Degree of the multivariate polynomial.
         fold_multiplier: Scaling gap required by unitary folding.
-        num_chunks: Number of desired approximately equal chunks. When the
-            number of chunks is the same as the layers in the input circuit,
-            the input circuit is unchanged.
+        num_chunks: The number of equally-sized circuit chunks. Noise
+            scaling is applied to each chunk independently. Ranges from 1
+            (all gates in one chunk, similar to global folding in ZNE) to the
```
andreamari:
```suggestion
            (all gates in one chunk, similar to ZNE) to the
```
I think term "global folding" is too specific. 
One can scale noise over the full circuit via "global folding", "local folding" or any other method that scales noise by a single factor $\lambda$. In fact `folding_method` is an option for both standard ZNE and LRE. So I would go back to `similar to ZNE` , or replace with `similar to standard ZNE` ,   `similar to single-variable ZNE`.

All the rest LGTM.

P.S. Ciao everyone, from grampa Andrea!

natestemen: I never should have messed with grampa Andrea's wording 🤦🏻.

natestemen (approved):
🙏🏻

PR #2934, opened 2026-02-24
feat: trex qem technique

vprusso:
## Description

Implements the TREX QEM protocol as defined in [arXiv:2012.09738](https://arxiv.org/abs/2012.09738). Additionally, this includes docs, tutorials, etc. for how one can use the TREX protocol within mitiq in a similar fashion to how other QEM protocols are implemented.

### License


mitiq/trex/tests/__init__.py
natestemen, line 1: I think this is unnecessary. We don't have init files in our other test repos.

mitiq/trex/trex.py
```diff
@@ -0,0 +1,376 @@
...
+
+
+def construct_circuits(
+    circuit: QPROGRAM,
+    observable: Observable,
+    num_randomizations: int = 32,
+    random_state: int | np.random.RandomState | None = None,
+) -> tuple[
+    list[cirq.Circuit], list[cirq.Circuit], list[npt.NDArray[np.int64]]
+]:
```
natestemen: Circuits should be converted back to user QPROGRAM before returning.

mitiq/trex/trex.py
```diff
@@ -0,0 +1,376 @@
...
+                if abs(calib_factor) > 1e-10:
+                    corrected_values.append(noisy_exp / calib_factor)
+                else:
+                    # Calibration factor too small; skip this randomization.
+                    corrected_values.append(noisy_exp)
+
+            if corrected_values:
+                corrected_exp = float(np.mean(corrected_values))
+            else:
+                corrected_exp = 0.0
```
natestemen: When would we expected `corrected_values` to be empty?

mitiq/experimental/trex/trex_utils.py
```diff
@@ -0,0 +1,129 @@
...
+    new_moments = []
+    x_inserted = False
+    for moment in circuit:
+        has_measurement = any(
+            cirq.is_measurement(op) for op in moment.operations
+        )
+        if has_measurement and not x_inserted:
+            new_moments.append(cirq.Moment(x_ops))
+            x_inserted = True
+        new_moments.append(moment)
```
natestemen, line 62: Does this only work when there is a single measurement? If there are two measurements wouldn't `x_inserted` get set to `True` at the first qubit with a measurement, and then not add $X$ gates before the next measurements?

mitiq/experimental/trex/trex_utils.py
```diff
@@ -0,0 +1,129 @@
...
+    if ops:
+        circuit.append(cirq.Moment(ops))
+    circuit.append(cirq.measure(*sorted(qubits)))
+    return circuit
+
+
+def xor_bitstrings(
+    result: MeasurementResult,
+    bitstring: npt.NDArray[np.int64],
+) -> MeasurementResult:
```
natestemen, line 131: Do you think this would be a useful method to have on `MeasurementResult`?

docs/source/guide/trex.md
```diff
@@ -0,0 +1,25 @@
+# Twirled Readout Error eXtinction (TREX)
```
natestemen:
```suggestion
# Twirled Readout Error eXtinction
```

For consistency with the other techniques.

docs/source/guide/trex-5-theory.md
```diff
@@ -0,0 +1,107 @@
...
+operators $X^{\mathbf{s}}$ immediately before measurement, where
+$\mathbf{s} \in \{0, 1\}^n$ is sampled uniformly at random. After
+measurement, the classical outcome $\mathbf{x}$ is XOR'd with
+$\mathbf{s}$ to undo the bit flips: $\mathbf{y} = \mathbf{x} \oplus \mathbf{s}$.
+
+This creates a *twirled* noise map:
+
+$$
+A_\star = \frac{1}{2^n} \sum_{\mathbf{s}} X^{\mathbf{s}} A X^{\mathbf{s}}
+$$
```
natestemen, line 49: This makes it seem like all $2^n$ bitstrings need to be sampled. I know the protocol is described below and that's not the case, but can you add a note on why there is a difference?

docs/source/guide/trex.md
```diff
@@ -0,0 +1,25 @@
...
+
+Twirled Readout Error eXtinction (TREX) is a model-free readout error
```
natestemen: I'm not sure I understand what model-free means here since we are running calibration circuits which presumably give us some sort of model?

docs/source/guide/trex-5-theory.md
```diff
@@ -0,0 +1,107 @@
...
+
+## Eigenvalue correction
+
+After twirling, each Pauli observable $P$ with support $\mathbf{w}$
+(the set of qubits where $P$ acts non-trivially) has a corresponding
+eigenvalue:
+
+$$
+\lambda_{\mathbf{w}} = \frac{1}{2^n} \sum_{\mathbf{a}, \mathbf{b}} (-1)^{\langle \mathbf{w}, \mathbf{a} + \mathbf{b} \rangle} A_{\mathbf{a}, \mathbf{b}}
+$$
```
natestemen, line 69: What is $A_{\mathbf{a}, \mathbf{b}}$?

docs/source/guide/trex-2-use-case.md
natestemen, line 1: Want to add a link to QEM zoo at the bottom?

natestemen (commented):
First pass looks really good! Thanks a bunch for adding this, and for doing all the work (implementation, documentation, etc) all at the same time.

mitiq/trex/trex.py
vprusso: `construct_circuits` now uses `convert_from_mitiq` to return circuits in the same format as the input. Return type updated from `list[cirq.Circuit]` to `list[QPROGRAM]`.

mitiq/trex/trex.py
vprusso: Kept the guard but added a comment clarifying it only happens when `randomization_strings` is empty.

mitiq/experimental/trex/trex_utils.py
vprusso: Updated the docstring to clarify that it inserts X gates before the *first* measurement moment, and that it's designed for circuits with a single terminal measurement (as produced by `measure_in`).

docs/source/guide/trex-5-theory.md
vprusso: Added a paragraph explaining the sum is mathematical; in practice only N random samples are needed.

docs/source/guide/trex.md
vprusso: Expanded the intro to explain that "model-free" means no pre-characterized noise model is needed as input; calibration circuits estimate correction factors on-the-fly rather than building a reusable model.

docs/source/guide/trex-5-theory.md
vprusso: Added explanation after the eigenvalue equation.

docs/source/guide/trex-2-use-case.md
vprusso: Done! Added at the bottom of trex-2-use-case.md.

docs/source/guide/trex.md
vprusso: Changed.

mitiq/experimental/trex/trex_utils.py
vprusso: I'd say keep it as a standalone util.

docs/source/guide/trex-1-intro.md
```diff
@@ -0,0 +1,122 @@
...
+Here we observe that the application of TREX reduces the readout error when
+compared to the unmitigated result.
+
+```{note}
+TREX is model-free: it does not require knowledge of the readout error
+probabilities or a confusion matrix. The readout error correction is
+estimated automatically from calibration circuits.
+```
+
++++
```
natestemen: I think this note can be removed. It reads a bit redundant with the beginning of the page and section 2 (use cases).

docs/source/guide/trex-2-use-case.md
```diff
@@ -0,0 +1,65 @@
...
+  diagonalizes the readout error channel, which means it works even when
+  readout errors are correlated between qubits.
+
+- **Simple integration**: TREX only requires that the executor returns raw
+  bitstrings (``MeasurementResult``). It can be combined with other
+  error mitigation techniques that address different noise sources (e.g.,
+  [PT](pt.md) for gate errors, [ZNE](zne.md) for noise extrapolation).
+
+- **Mathematically rigorous**: TREX provides provable error bounds on
+  the corrected expectation values.
```
natestemen:
```suggestion
- **Mathematically rigorous**: TREX provides provable error bounds on
  the corrected expectation values when the measurement errors are small.
```

Right?

docs/source/guide/trex-2-use-case.md
```diff
@@ -0,0 +1,65 @@
...
+| Feature | TREX | REM |
+|---------|------|-----|
+| Requires confusion matrix | No | Yes |
+| Handles correlated errors | Yes | Only with full confusion matrix |
+| Model-free | Yes | No |
+| Additional circuits needed | Yes (calibration + twirling) | No |
+| Scaling | Linear in randomizations | Depends on matrix inversion |
+
+For more information on TREX and other error mitigation techniques, see
+the [QEM Zoo](https://qem-zoo.github.io/).
```
natestemen:
```suggestion
Find more information on TREX on the [QEM Zoo](https://qemzoo.com/technique.html?id=trex).
```

docs/source/guide/trex-2-use-case.md
```diff
@@ -0,0 +1,65 @@
...
+  the calibration factors can become small, leading to noisy corrected
+  values. In extreme cases, the correction may amplify statistical noise.
+
+## TREX vs REM
+
+| Feature | TREX | REM |
+|---------|------|-----|
+| Requires confusion matrix | No | Yes |
+| Handles correlated errors | Yes | Only with full confusion matrix |
+| Model-free | Yes | No |
```
natestemen: I think this is redundant of requires confusion matrix? I would lean towards one or the other.

docs/source/guide/trex-3-options.md
```diff
@@ -0,0 +1,102 @@
...
+and the random state for reproducibility. Unlike [REM](rem.md), TREX
+does not require the user to provide a confusion matrix.
+
+## Number of randomizations
+
+The ``num_randomizations`` parameter controls how many random readout
+twirling patterns are used. More randomizations give better accuracy
+but require more circuit executions.
+
+```{code-cell} ipython3
```
natestemen, line 27: The goal of this code cell is to showcase how more randomizations lead to increased performance, but it looks like that's not necessarily the case with this small example. If it's not too slow we can try increasing the problem size, or make a note about applying TREX on small examples.

docs/source/guide/trex-3-options.md
```diff
@@ -0,0 +1,102 @@
...
+    circuit, noisy_executor, observable,
+    num_randomizations=16, random_state=123,
+)
+result2 = execute_with_trex(
+    circuit, noisy_executor, observable,
+    num_randomizations=16, random_state=123,
+)
+print(f"Result 1: {result1:.6f}")
+print(f"Result 2: {result2:.6f}")
+print(f"Reproducible: {result1 == result2}")
```
natestemen, line 99: This is reporting False. There must be some other source of randomness not controlled by the seed. Maybe the readout sampler?

docs/source/guide/trex-3-options.md
```diff
@@ -0,0 +1,102 @@
...
+result, data = execute_with_trex(
+    circuit, noisy_executor, observable,
+    num_randomizations=8, random_state=42,
+    full_output=True,
+)
+
+print(f"TREX value: {result:.4f}")
+print(f"Number of twirled circuits: {len(data['twirled_circuits'])}")
+print(f"Number of calibration circuits: {len(data['calibration_circuits'])}")
+print(f"Number of randomization strings: {len(data['randomization_strings'])}")
```
natestemen: Should we just print the whole object so people can see all the attributes?

docs/source/guide/trex-2-use-case.md
natestemen, line 1: Not sure if this should go in this section or maybe the section on theory, but one question I want to know is: does this technique work well on measurement errors that are not symmetric? E.g. $\mathrm{Pr}(0 \to 1) \neq \mathrm{Pr}(1 \to 0)$.

docs/source/guide/trex-4-low-level.md
```diff
@@ -0,0 +1,131 @@
...
+twirled_results = all_results[:n_twirled]
+calibration_results = all_results[n_twirled:]
+
+# Combine results with TREX correction
+corrected_value = combine_results(
+    twirled_results, calibration_results,
+    randomization_strings, observable,
+)
+print(f"TREX-corrected expectation value: {corrected_value:.4f}")
+print(f"Ideal value: -2.0")
```
natestemen:
```suggestion
print("Ideal value: -2.0")
```

docs/source/guide/trex-4-low-level.md
```diff
@@ -0,0 +1,131 @@
...
+```
+
+## Using mitigate_executor and trex_decorator
```
natestemen:
```suggestion
## Using `mitigate_executor` and `trex_decorator`
```

docs/source/guide/trex-5-theory.md
```diff
@@ -0,0 +1,117 @@
...
+gates before measurement and classically undoing the flips in
+post-processing.
+
+## Readout error model
+
+When measuring $n$ qubits, readout errors are characterized by a
+left-stochastic matrix $A$ where entry $A_{i,j}$ represents the
+probability of measuring state $|i\rangle$ when the true state is
+$|j\rangle$. This transforms the ideal probability distribution
+$\mathbf{p}$ into a noisy one:
```
natestemen, line 28:
```suggestion
When measuring $n$ qubits, readout errors are characterized by a
[left-stochastic matrix](https://en.wikipedia.org/wiki/Stochastic_matrix) $A$ where entry $A_{i,j}$ represents the
probability of measuring state $|i\rangle$ when the true state is
$|j\rangle$. This transforms the ideal probability distribution
$\mathbf{p}$ into a noisy one:
```

modulo formatting.

I would also add a sentence or two about how $\tilde{\mathbf{p}}$ is what we measure in the lab, so we have to invert $A$ to obtain an empirical measure of $\mathbf{p}$.

docs/source/guide/trex-5-theory.md
```diff
@@ -0,0 +1,117 @@
...
+
+Note that the sum is over all $2^n$ possible bitstrings, but this is a
+*mathematical* description of the twirled channel. In practice, we do
+**not** need to enumerate all $2^n$ strings. Instead, we sample a small
+number $N$ of random strings (the ``num_randomizations`` parameter) and
+average the corrected estimates, which converges to the correct value.
+
+The mathematical key insight is that this twirling transforms the
+noise channel so that the computational basis states are eigenvectors
+of $A_\star$, effectively *diagonalizing* the readout error channel.
```
natestemen:
```suggestion
The key insight is that this twirling transforms the
noise channel so that the computational basis states are eigenvectors
of $A_\star$, effectively *diagonalizing* the readout error channel.
```

docs/source/apidoc.md
```diff
@@ -196,6 +196,22 @@ See Ref. {cite}`Czarnik_2021_Quantum` for more details on these methods.
...
+.. automodule:: mitiq.trex.trex
+   :members:
+```
+
+#### TREX Utilities
+
+```{eval-rst}
+.. automodule:: mitiq.trex.trex_utils
+   :members:
+```
```
natestemen: Since two of these functions only work with cirq circuits I'm not sure we want to expose them to users. Unless you have a motivation to do this, I'd rather make these private and not expose them.

mitiq/trex/trex.py
```diff
@@ -0,0 +1,379 @@
...
+    # Convert circuit to Cirq for internal processing.
+    cirq_circuit, input_type = convert_to_mitiq(circuit)
+    all_qubits = sorted(cirq_circuit.all_qubits())
+    n_qubits = len(all_qubits)
+
+    # Generate random twirling patterns.
+    randomization_strings = [
+        rng.randint(0, 2, size=n_qubits).astype(np.int64)
+        for _ in range(num_randomizations)
+    ]
```
natestemen:
```suggestion
    randomization_strings = rng.randint(
        0, 2, size=(num_randomizations, n_qubits), dtype=np.int64
    )
```

We can generate all the bitstrings in one go this way. And you can still iterate them as needed below.

mitiq/trex/trex_utils.py
```diff
@@ -0,0 +1,131 @@
...
+from collections.abc import Sequence
+
+import cirq
+import numpy as np
+import numpy.typing as npt
+
+from mitiq import MeasurementResult
+
+
+def insert_x_before_measurements(
```
natestemen:
```suggestion
def insert_x_before_first_measurement(
```

To better reflect its function?

mitiq/trex/trex.py
```diff
@@ -0,0 +1,379 @@
...
+            twirled = insert_x_before_measurements(
+                meas_circuit, s_group, meas_qubits
+            )
+            twirled_circuits.append(twirled)
+
+    # Build calibration circuits (shared across groups, measure all qubits).
+    calibration_circuits: list[cirq.Circuit] = []
+    for s in randomization_strings:
+        calib = create_calibration_circuit(all_qubits, s)
+        calibration_circuits.append(calib)
```
natestemen: WDYT of moving this logic to `create_calibration_circuit` so that we just need to make one call to fetch all the calibration circuits and they are returned in the users frontend?

mitiq/experimental/trex/trex.py
```diff
@@ -0,0 +1,379 @@
...
+                calib_factor = float(
+                    np.mean([(-1) ** np.sum(b) for b in calib_bits])
+                )
+
+                # Apply TREX correction: divide by calibration factor.
+                if abs(calib_factor) > 1e-10:
+                    corrected_values.append(noisy_exp / calib_factor)
+                else:
+                    # Calibration factor too small; skip this randomization.
+                    corrected_values.append(noisy_exp)
```
natestemen, line 280: When would this branch get hit? When the measurement errors are almost non-existent? Wondering if we should throw a warning to let the user know.

mitiq/trex/trex.py
natestemen: If there are no randomization strings, then does TREX do anything?

natestemen (commented):
Thanks for the quick updates! I have some hopefully small comments, but I think this is pretty close.

docs/source/guide/trex-1-intro.md
vprusso: Good call, removed the redundant note.

docs/source/guide/trex-2-use-case.md
vprusso: Applied the suggestion — added "when the measurement errors are small".

docs/source/guide/trex-2-use-case.md
vprusso: Applied the suggestion — updated the link.

docs/source/guide/trex-2-use-case.md
vprusso: Agreed, removed the "Model-free" row from the table since it's redundant with "Requires confusion matrix".

docs/source/guide/trex-3-options.md
vprusso: Added a note explaining that on small circuits the improvement from additional randomizations may not be monotonic due to finite sampling noise, and that the benefits become clearer on larger circuits.

docs/source/guide/trex-3-options.md
vprusso: Good catch — the `NoisySingleQubitReadoutSampler` has its own internal randomness not controlled by our seed. Fixed by using a noiseless (deterministic) executor for the reproducibility demo and added a note that `random_state` controls the twirling patterns only; for full reproducibility the executor must also be deterministic.

docs/source/guide/trex-3-options.md
vprusso: Updated to print the full data object with all keys and their sizes.

docs/source/guide/trex-2-use-case.md
vprusso: Great question. Yes, TREX handles asymmetric errors naturally — the readout twirling symmetrizes the noise channel by averaging over random X flips, converting any asymmetric readout noise into a symmetric (diagonal) form. The calibration circuits then estimate the resulting eigenvalues. Added a section on this in trex-2-use-case.md.

docs/source/guide/trex-4-low-level.md
vprusso: Applied.

docs/source/guide/trex-4-low-level.md
vprusso: Applied.

docs/source/guide/trex-5-theory.md
vprusso: Applied the suggestion — added the Wikipedia link.

docs/source/guide/trex-5-theory.md
vprusso: Applied — removed "mathematical" from the sentence.

docs/source/apidoc.md
vprusso: Agreed. Removed `trex_utils` from the public API — removed it from `__init__.py` exports and the apidoc page. The utility functions are still used internally but no longer exposed to users.

mitiq/trex/trex.py
vprusso: Applied — now generating all bitstrings in a single `rng.randint` call with `size=(num_randomizations, n_qubits)`.

mitiq/trex/trex_utils.py
vprusso: Renamed to `insert_x_before_first_measurement` and updated all references.

mitiq/trex/trex.py
vprusso: Added a `create_calibration_circuits` (plural) batch helper in `trex_utils.py` that takes all randomization strings and returns all calibration circuits. `construct_circuits` now uses a single call to this function instead of the manual loop.

mitiq/experimental/trex/trex.py
vprusso: This branch gets hit when readout errors are very large (close to 50% bit-flip rate), making the calibration parity average near zero. Added a `warnings.warn` so users are alerted when this fallback is triggered.

mitiq/trex/trex.py
vprusso: You're right — if there are no randomization strings, TREX doesn't do anything meaningful. Changed this to raise a `ValueError` instead of silently returning 0. Updated the corresponding test as well.

mitiq/trex/tests/__init__.py
vprusso: Removed in the previous round — it's no longer in the tree.

docs/source/guide/trex-2-use-case.md
```diff
@@ -0,0 +1,72 @@
...
+| Scaling | Linear in randomizations | Depends on matrix inversion |
+
+## Asymmetric readout errors
+
+TREX handles asymmetric readout errors (where $\mathrm{Pr}(0 \to 1)
+\neq \mathrm{Pr}(1 \to 0)$) naturally. The readout twirling procedure
+symmetrizes the noise channel by averaging over random $X$ flips,
+effectively converting any asymmetric readout noise into a symmetric
+(diagonal) form. The calibration circuits then estimate the resulting
+eigenvalues, so no explicit knowledge of the asymmetry is required.
```
natestemen: This pages typically only has "advantages" and "disadvantages" sections so I think this should either go in one of the bullet points above or in the theory section.

natestemen (commented):
1. Since we are going to merge https://github.com/unitaryfoundation/mitiq/pull/2966 soon, this technique should also go into the `experimental/` directory.
2. The README table needs an update to add this technique.
3. The copilot review looks like it raised some good point.

Once these are all addressed we will merge!

docs/source/guide/trex-2-use-case.md
vprusso: Good point — moved the asymmetric readout errors section to the theory page (trex-5-theory.md) instead.

mitiq/trex/trex.py
```diff
@@ -0,0 +1,376 @@
...
+
+
+def mitigate_executor(
+    executor: Executor | Callable[[QPROGRAM], MeasurementResult],
+    observable: Observable,
+    *,
+    num_randomizations: int = 32,
+    random_state: int | np.random.RandomState | None = None,
+    full_output: bool = False,
+) -> Executor | Callable[[QPROGRAM], float | tuple[float, dict[str, Any]]]:
```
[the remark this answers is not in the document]
vprusso: The return type annotation is `Callable`, not `Executor` — this was already correct in the current code. The function always returns a wrapped callable.

mitiq/trex/trex.py
```diff
@@ -0,0 +1,376 @@
...
+    # Generate random twirling patterns.
+    randomization_strings = [
+        rng.randint(0, 2, size=n_qubits).astype(np.int64)
+        for _ in range(num_randomizations)
+    ]
+
+    # Get measurement circuits for each commuting group.
+    measurement_circuits = []
+    group_qubits = []
+    for group in observable._groups:
```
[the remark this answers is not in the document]
vprusso: Already addressed — the code uses `observable.groups` (public property).

mitiq/trex/trex.py
```diff
@@ -0,0 +1,376 @@
...
+    Returns:
+        The TREX-corrected expectation value.
+    """
+    num_randomizations = len(randomization_strings)
+    num_groups = observable.ngroups
+    all_qubits = sorted(observable._qubits())
+    qubit_to_idx = {q: i for i, q in enumerate(all_qubits)}
+
+    total = 0.0
+    for group_idx, group in enumerate(observable._groups):
```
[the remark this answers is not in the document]
vprusso: Already addressed — the code uses `observable.groups` (public property).

mitiq/trex/trex.py
```diff
@@ -0,0 +1,376 @@
...
+    randomization_strings = [
+        rng.randint(0, 2, size=n_qubits).astype(np.int64)
+        for _ in range(num_randomizations)
+    ]
+
+    # Get measurement circuits for each commuting group.
+    measurement_circuits = []
+    group_qubits = []
+    for group in observable._groups:
+        meas_cirq = group._measure_in(cirq_circuit, group)
```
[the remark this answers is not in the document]
vprusso: Already addressed — the code uses `group.measure_in(cirq_circuit)` (public method).

mitiq/experimental/trex/trex_utils.py
```diff
@@ -0,0 +1,151 @@
...
+        raise ValueError(
+            f"Length of bitstring ({len(bitstring)}) must match "
+            f"number of qubits ({result.nqubits})."
+        )
+
+    flipped = result._bitstrings ^ bitstring
+    return MeasurementResult(
+        result=flipped.tolist(),
+        qubit_indices=result.qubit_indices,
+    )
```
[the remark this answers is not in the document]
vprusso: Good catch — switched to the public `asarray` property.

mitiq/experimental/trex/tests/test_trex_utils.py
```diff
@@ -0,0 +1,232 @@
...
+
+
+def test_xor_all_zeros():
+    """XOR with all zeros leaves bitstrings unchanged."""
+    mr = MeasurementResult(["01", "10", "11"], qubit_indices=(0, 1))
+    bitstring = np.array([0, 0], dtype=np.int64)
+    result = xor_bitstrings(mr, bitstring)
+
+    assert np.array_equal(result._bitstrings, mr._bitstrings)
+
```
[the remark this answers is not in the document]
vprusso: Updated all test assertions to use the public `asarray` property.

mitiq/trex/trex_utils.py
```diff
@@ -0,0 +1,151 @@
...
+            f"Length of bitstring ({len(bitstring)}) must match "
+            f"number of qubits ({len(qubits)})."
+        )
+
+    ops = [cirq.X(qubit) for qubit, bit in zip(qubits, bitstring) if bit == 1]
+
+    circuit = cirq.Circuit()
+    if ops:
+        circuit.append(cirq.Moment(ops))
+    circuit.append(cirq.measure(*sorted(qubits)))
```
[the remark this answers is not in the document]
vprusso: Good point — updated `create_calibration_circuit` to sort qubits internally before both X gate application and measurement, ensuring consistent ordering regardless of caller input.

mitiq/experimental/trex/trex.py
```diff
@@ -0,0 +1,385 @@
...
+    # Build twirled measurement circuits.
+    twirled_circuits: list[cirq.Circuit] = []
+    for meas_circuit, meas_qubits in zip(measurement_circuits, group_qubits):
+        # Map full randomization string to measured qubits.
+        qubit_to_idx = {q: i for i, q in enumerate(all_qubits)}
+
+        for s in randomization_strings:
+            s_group = np.array(
+                [s[qubit_to_idx[q]] for q in meas_qubits], dtype=np.int64
+            )
```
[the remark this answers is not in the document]
vprusso: The qubits passed to `insert_x_before_first_measurement` come from `group._qubits_to_measure()` which returns the qubits that `measure_in` actually measures. Both use `sorted()` so the ordering is consistent. The randomization string `s_group` is mapped from the full randomization string using the same qubit-to-index mapping, so the alignment is correct.

mitiq/experimental/trex/trex.py
```diff
@@ -0,0 +1,385 @@
...
+                # Get the twirled result for this group and randomization.
+                result_idx = group_idx * num_randomizations + rand_idx
+                twirled_res = twirled_results[result_idx]
+
+                # XOR with the randomization bits for measured qubits.
+                s_group = np.array(
+                    [s[qubit_to_idx[q]] for q in measured_qubits],
+                    dtype=np.int64,
+                )
+                flipped_twirled = xor_bitstrings(twirled_res, s_group)
```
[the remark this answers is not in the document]
vprusso: Both `construct_circuits` and `combine_results` derive qubit ordering from `sorted()` — `sorted(cirq_circuit.all_qubits())` in `construct_circuits` and `sorted(observable._qubits())` in `combine_results`. Since the observable qubits are the same as the circuit qubits (both are LineQubits), and both are sorted, the indexing is consistent.

mitiq/experimental/trex/trex.py
```diff
@@ -0,0 +1,388 @@
...
+
+def mitigate_executor(
+    executor: Executor | Callable[[QPROGRAM], MeasurementResult],
+    observable: Observable,
+    *,
+    num_randomizations: int = 32,
+    random_state: int | np.random.RandomState | None = None,
+    full_output: bool = False,
+) -> Callable[[QPROGRAM], float | tuple[float, dict[str, Any]]]:
+    """Returns a modified version of the input ``executor`` which is
```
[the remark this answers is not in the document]
vprusso: This pattern (including the `# type: ignore[misc]`) is consistent with how other mitiq techniques (REM, ZNE, etc.) handle batched vs non-batched executors. Changing it here would create an inconsistency across the codebase. Leaving as-is for now.

mitiq/experimental/trex/trex.py
```diff
@@ -0,0 +1,388 @@
...
+    randomization_strings = rng.randint(
+        0, 2, size=(num_randomizations, n_qubits), dtype=np.int64
+    )
+
+    # Get measurement circuits for each commuting group.
+    measurement_circuits = []
+    group_qubits = []
+    for group in observable.groups:
+        meas_cirq = cast(cirq.Circuit, group.measure_in(cirq_circuit))
+        measured_qubits = sorted(group._qubits_to_measure())
```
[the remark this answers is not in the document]
vprusso: Valid theoretical concern. In practice, TREX is used with standard `LineQubit.range(n)` circuits where canonical and original qubits are identical, and all tests confirm correctness in this case. Supporting non-canonical qubit layouts (e.g., `[LineQubit(1), LineQubit(6)]`) would require a broader refactor — noted as a potential future improvement.

mitiq/experimental/trex/trex_utils.py
```diff
@@ -0,0 +1,155 @@
...
+            f"number of qubits ({len(qubits)})."
+        )
+
+    sorted_qubits = sorted(qubits)
+    ops = [
+        cirq.X(qubit)
+        for qubit, bit in zip(sorted_qubits, bitstring)
+        if bit == 1
+    ]
+
```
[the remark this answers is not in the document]
vprusso: Same as above — the calibration circuits use `sorted(qubits)` consistently with the rest of the pipeline, which works correctly for standard `LineQubit.range(n)` layouts. Non-canonical qubit support would be a broader follow-up.

mitiq/experimental/trex/trex.py
```diff
@@ -0,0 +1,388 @@
...
+            corrected_values = []
+
+            for rand_idx, s in enumerate(randomization_strings):
+                # Get the twirled result for this group and randomization.
+                result_idx = group_idx * num_randomizations + rand_idx
+                twirled_res = twirled_results[result_idx]
+
+                # XOR with the randomization bits for measured qubits.
+                s_group = np.array(
+                    [s[qubit_to_idx[q]] for q in measured_qubits],
```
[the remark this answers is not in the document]
vprusso: This indexing is correct when the observable acts on the same qubits as the circuit (the standard use case). Both `construct_circuits` and `combine_results` use `sorted()` on the same qubit set. The concern about non-contiguous subsets (observable on a strict subset of circuit qubits) is noted as a limitation for a potential follow-up.

mitiq/experimental/trex/trex_utils.py
```diff
@@ -0,0 +1,155 @@
...
+    Returns:
+        A new ``MeasurementResult`` with bits XOR'd by ``bitstring``.
+    """
+    if len(bitstring) != result.nqubits:
+        raise ValueError(
+            f"Length of bitstring ({len(bitstring)}) must match "
+            f"number of qubits ({result.nqubits})."
+        )
+
```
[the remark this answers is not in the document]
vprusso: Already fixed in the previous commit — `xor_bitstrings` now uses the public `asarray` property.

natestemen (commented):
Two leftover comments from https://github.com/unitaryfoundation/mitiq/pull/2934#pullrequestreview-3988174960:

1. Can you move this technique into the `experimental/` directory?
2. Can you add TREX to the readme?

BTW can you also make a note of how you're using AI on this PR just for transparency? No issues with it, but just helpful to know.

natestemen (approved):
Woohoo! Thank you @vprusso!

PR #2939, opened 2026-03-08
Add compare_performance() for PT vs PT+QEM analysis (#2615)

### LINKED ISSUE #2615: Function to show when PT or PT + QEM is helpful
We want to add a function in `mitiq/pt` that spits out a graph showing when PT is useful for some circuit and when combining PT with some QEM technique is helpful.

```py
compare_performance(
  circuit, 
  noisy_backend with coherent noise, 
  num_pauli_twirled_circuits_to_create, 
  {metadata of some QEM if combining PT with QEM}
)
```
This should spit out a plot showing when combining noise tailoring with QEM techniques is useful as a function of noise strength. 

See the example plot below where the difference between ideal and other expectation values are plotted. The closer the curve is to 0.0, the better the technique performs. It is also easy to see the regions when ZNE is useful on its own, PT is useful on its own, combination of PT + ZNE is useful, etc. 

image

mrmoe1985-sketch:
This PR adds a function to compare when Pauli Twirling (PT) alone is useful vs when combining PT with Quantum Error Mitigation (QEM) techniques.

## Features
- `compare_performance(circuit, executor, noise_levels, num_twirled_circuits, qem_strategy)` - Compares PT vs PT+QEM
- `plot_comparison(results)` - Visualizes which strategy works best at different noise levels
- Shows when ZNE alone, PT alone, or PT+ZNE combined is most effective

## Example Usage
```python
from mitiq.pt import compare_performance, plot_comparison

results = compare_performance(
    circuit,
    executor,
    noise_levels=[0.001, 0.005, 0.01, 0.02, 0.05],
    num_twirled_circuits=10,
    qem_strategy='zne'
)
plot_comparison(results)
```

## Testing
- Tested with simulator (no IBM Quantum time used for main development)
- Can be validated on real hardware with user's IBM Quantum connection

Closes #2615

### DIFF
--- a/mitiq/pt/__init__.py
+++ b/mitiq/pt/__init__.py
@@ -8,3 +8,12 @@ from mitiq.pt.pt import (
     twirl_CNOT_gates,
     twirl_CZ_gates,
 )
+from mitiq.pt.compare import compare_performance, plot_comparison
+
+__all__ = [
+    "generate_pauli_twirl_variants",
+    "twirl_CNOT_gates",
+    "twirl_CZ_gates",
+    "compare_performance",
+    "plot_comparison",
+]
--- a/mitiq/pt/compare.py
+++ b/mitiq/pt/compare.py
@@ -1,0 +1,213 @@
+# Copyright (C) Unitary Foundation
+#
+# This source code is licensed under the GPL license (v3) found in the
+# LICENSE file in the root directory of this source tree.
+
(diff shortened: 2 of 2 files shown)


natestemen (changes requested):
Can you show the output for a small example? Also the naming `compare_performance` does not reflect that it only compares performance of ZNE and PT.

PR #2951, opened 2026-03-11
fix cost when ideal executor is being used

### LINKED ISSUE #1716: Support ideal_executor argument in the Calibrator class
## Issue Description

We currently have an `ideal_executor` argument in the calibrator which represents a classical simulator of a perfect quantum computer. In practice however it is not used. 
It would be nice to support this argument, not much for the benchmark problems that we are using now but mostly for more advanced or custom benchmark problems. E.g. small subcircuits of user's circuits, small random circuits, clifford-dominated circuits, etc... where the ideal solution is not known _a priori_ but it can be efficiently simulated.

## Proposed Solution

The first question to ask is: _Do we really need to support this?_ _Is the need of an `ideal_executor` argument sufficiently strong_?

If not, just drop the `ideal_executor` argument.
If yes, add proper support.

natestemen:
## Description

The calibrator uses the ideal executor when a custom circuit is being used, but the cost computation did not account for it.

fixes https://github.com/unitaryfoundation/mitiq/issues/1716

### DIFF
--- a/mitiq/calibration/calibrator.py
+++ b/mitiq/calibration/calibrator.py
@@ -225,7 +225,8 @@ class Calibrator:
         frontend: The executor frontend as a string. For a list of supported
             frontends see ``mitiq.SUPPORTED_PROGRAM_TYPES.keys()``,
         ideal_executor: An optional simulated executor returning the ideal
-            :class:`.MeasurementResult` without noise.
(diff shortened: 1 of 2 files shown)


cosenal (approved):
#1716 sounds more broad than this, but I am assuming most of its scope was already implemented in some other PR.

Issue #2949, opened 2026-03-11
Docs builds burn many GH actions minutes


cosenal:
The UF org is burning GitHub actions minutes at a higher rate than usual, which is resulting in financial costs 💸 

I have been doing a [bit of auditing](https://github.com/unitaryfoundation/mitiq/actions/metrics/usage), and it turns out one of the main culprits is mitiq's [docs-build](https://github.com/unitaryfoundation/mitiq/actions/workflows/.github/workflows/docs-build.yml).

I know we looked into this before (see #2024, #1321, #2738), and that @natestemen created a lighter build* for local development.

This issue is about looking into this again, with the goal of specifically reducing the GH actions minutes burned #frugality

*I wonder if that build lite can be used for actions too.

PR #2953, opened 2026-03-12
Make docs-build workflow more frugal

cosenal:
## Description

Patching the docs-build workflow so that we save on GH actions minutes #2949 
#frugality

- docs-build still run at every PRs update, but only with make docs-lite
- full docs build now run only on push to main, nightly (@natestemen not sure if this is that helpful after all), and manual dispatch
- newer commits cancel older in-progress docs runs for the same PR/branch.
- draft PRs are skipped altogether

---

### License


natestemen (changes requested):
Instead of figuring out docs-lite lets just set it up to do manual dispatch and 1x/day and cancel in progress.

The readthedocs build is good enough check when we want to double check the docs build most of the time.

.github/workflows/docs-build.yml
```diff
@@ -62,10 +61,6 @@ jobs:
           export SKIP_PYQUIL=1
           make docs
 
-      - name: Run linkcheck on 'release' PRs
-        if: github.event_name == 'pull_request'
-        run: |
-          pr_title_lower=$(echo "${{ github.event.pull_request.title }}" | tr '[:upper:]' '[:lower:]')
-          if [[ "$pr_title_lower" == *"release"* ]]; then
```
cosenal, line 69: Removed this over-engineering (by my old self). Now linkcheck runs on the nightly builds only, which was one of the original ideas when we had the linkcheck annoyance.

natestemen (approved):

PR #2958, opened 2026-03-16
`:func:` for function references and Restores the missing Pauli fidelity formula 

axif0:
## Description
- Fixed docstrings where function names are referenced in backticks (plain text) instead of using `Sphinx's :func: cross-reference syntax, which would render as clickable links.
- Restores the missing Pauli fidelity formula in the theory guide, ensuring the math matches the internal implementation for zero-state calibration in `get_single_shot_pauli_fidelity`
 
 (Closes parts of #2110)

---

### License

### DIFF
--- a/mitiq/shadows/shadows.py
+++ b/mitiq/shadows/shadows.py
@@ -35,7 +35,7 @@ def pauli_twirling_calibrate(
     The number of :math:`f_b` is :math:`2^n`, or :math:`\sum_{i=1}^d C_n^i` if
     the locality :math:`d` is given.
 
-    In the notation of :footcite:t:`chen2021robust`, this function estimates
+    In the notation of :cite:`chen2021robust`, this function estimates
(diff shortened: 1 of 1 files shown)


mitiq/shadows/shadows.py
```diff
@@ -53,16 +53,18 @@ def pauli_twirling_calibrate(
...
+            when the ``zero_state_shadow_outcomes`` is None.
         num_total_measurements_calibration: Number of shots per group of
             "median of means" used for calibration. Needs to specify when
-            the `zero_state_shadow_outcomes` is None.
+            the ``zero_state_shadow_outcomes`` is None.
 
     Returns:
         A dictionary containing the calibration outcomes.
+
+    .. footbibliography::
```
natestemen: I don't think this is particularly helpful, especially since we don't use it elsewhere. Can you remove it?

PR #2959, opened 2026-03-16
Refactor Pauli Twirling documentation and remove construction warnings

### LINKED ISSUE #2462: Update Pauli Twirling user guide
With the changes made in #2454, the user guide for PT is still incomplete. 

Below are the things that still require some work before we remove the WIP warning at the top of each page in this user guide.

- [ ] Update the workflow diagram image
- [ ] The `Advantages` and `Disadvantages` section of the `When should I use PT?` are missing details on coherent vs incoherent noise.
- [ ] On `What additional options are available when using PT?` make a table (preferably of the PTM heatmaps for each) showing which noise channels will be tailored by PT.
- [ ] The `What additional options are available when using PT?` is missing a subsection on how the number of generated Pauli twirled circuits affects the outcome.
- [ ] In `What additional options are available when using PT?` add a subsection on how PT could be stacked with some QEM technique.
- [ ] A new workflow diagram for stacking PT and a QEM technique 
- [ ] Portions of `What happens when I use PT?` are commented out.
- [ ] `What is the theory behind Pauli Twirling?` could use additional details and clarifications.

axif0:
## Description

- Removed construction warnings from multiple sections of the Pauli Twirling (PT) documentation.
- Enhanced explanations of PT's functionality, advantages, and disadvantages.
- Added detailed descriptions of the workflow and theoretical background of PT.
- Introduced new content on the effects of PT on various noise channels and the importance of twirling in quantum error mitigation.
- Updated the PT workflow diagram to reflect recent changes and added a new diagram for PT in conjunction with quantum error mitigation techniques.

Closes #2462

---

### License


docs/source/img/pt_qem_workflow.svg
natestemen, line 1: This image does not need to be replaced.

docs/source/guide/pt-2-use-case.md
natestemen, line 1: This page in general is much too long. The advantages/disadvantages sections should have no subsections and should only be 2-3 paragraphs maximum (similar to other user guide "when should I use XYZ?" pages).

docs/source/guide/pt-3-options.md
```diff
@@ -13,9 +13,327 @@ kernelspec:
...
+```{note}
+The exact convergence depends on the circuit, noise model, and noise strength.
+For strong coherent noise, more twirled circuits are needed to average out the
+off-diagonal PTM components. As a rule of thumb, 50--100 variants is
+a reasonable starting point for most applications.
 ```
 
-Currently Pauli Twirling is designed to have relatively few options, in part to make it readily composable with every other Mitiq technique. In the future, we expect a possibility of supporting additional operations as targets (beyond CZ and CNOT gates), with more customization on picking those targets. Stay tuned!
\ No newline at end of file
+## Stacking PT with a quantum error mitigation technique
```
natestemen: A short demo of using PT and ZNE is nice, but it should go on the intro page rather than the "what additional options" page since it's how we expect most people to use PT.

docs/source/guide/pt-3-options.md
natestemen, line 1: Most of the content on this page is quite nice, but does not belong here. This page should answer "what options are available when using PT that I haven't already seen in the user guide".

docs/source/guide/pt-4-low-level.md
```diff
@@ -97,41 +86,45 @@ from mitiq import Executor
...
 
 ```{code-cell} ipython3
-# import numpy as np
-# from typing import cast
+import numpy as np
 
-# average = cast(float, np.average(expvals))
-# print(average)
+average = float(np.average(expvals))
+print(f"Average expectation value over {len(twirled_circuits)} twirled circuits: {average:.6f}")
```
natestemen, line 106: Thanks for cleaning up all this commented out code. Not sure why we had it there to begin with!

docs/source/guide/pt-5-theory.md
```diff
@@ -13,33 +13,181 @@ kernelspec:
...
+  as in [ZNE](zne-5-theory.md), [PEC](pec-5-theory.md), and [CDR](cdr-5-theory.md).
+
+- The error mitigation overhead is minimal: there is no increase in statistical
+  uncertainty in the final result beyond the standard sampling variance.
+
+## Mathematical formulation
+
+### The Pauli twirl of a channel
+
+Let $\mathcal{G}_n = \{I, X, Y, Z\}^{\otimes n}$ be the $n$-qubit Pauli group
```
natestemen: Can we use the more standard notation $\mathcal{P}_n$ for the Pauli group?

docs/source/guide/pt-5-theory.md
```diff
@@ -13,33 +13,181 @@ kernelspec:
...
+\mathcal{T}_P[\mathcal{E}](\rho) = \sum_k p_k \, \sigma_k \rho \sigma_k^\dagger,
+$$(pauli_channel_result)
+
+where the probabilities are
+
+$$
+p_k = \frac{1}{4^n} \sum_j \left| \mathrm{Tr}(\sigma_k K_j) \right|^2.
+$$
+
+**Proof sketch.** Expand each Kraus operator in the Pauli basis:
```
natestemen: No need to provide a proof here, but please provide a reference.

docs/source/guide/pt-5-theory.md
```diff
@@ -13,33 +13,181 @@ kernelspec:
...
+
+The dressed twirling circuit for a two-qubit gate looks like:
+
+$$
+\begin{array}{c}
+-\!- P -\!- \bullet -\!- R -\!- \\
+\qquad\qquad\; | \\
+-\!- Q -\!- \oplus -\!- S -\!-
+\end{array}
+$$
```
natestemen: Replacing this with an image made with quantikz would be nicer. It's currently botched in the rendering.

docs/source/guide/pt-5-theory.md
```diff
@@ -13,33 +13,181 @@ kernelspec:
...
+### Gate twirling (dressed twirling)
+
+Directly applying Eq. {math:numref}`pauli_twirl_def` to a noisy gate
+$\mathcal{E} = \mathcal{N} \circ U$ would destroy the ideal unitary $U$.
+Instead, Mitiq implements **dressed twirling**: for each Pauli $P_i$,
+find $P_i'$ such that $U P_i = P_i' U$ (up to a global phase). The
+twirled circuit is then
+
+$$
+P_i' \circ U \circ P_i = U \quad \text{(ideal effect preserved)},
```
natestemen: `\circ` usually implies function composition but here we're just talking matrix multiplication. Maybe `\cdot`?

docs/source/guide/pt-5-theory.md
```diff
@@ -13,33 +13,181 @@ kernelspec:
...
 
-Pauli Twirling (PT) can be a powerful tool for noise management in quantum systems. By twirling over the Pauli gates, PT transforms complex noise channels into simpler stochastic Pauli noise channels.
+$$
+R_{R_{Y_\theta}} = \begin{pmatrix}
+1 & 0 & 0 & 0 \\
+0 & \cos\theta & 0 & \sin\theta \\
+0 & -\sin\theta & 0 & \cos\theta \\
+0 & 0 & 0 & 1
+\end{pmatrix}.
+$$
```
natestemen: I don't think this is the correct PTM of $R_y$, and it's also a different notation than how the PTM is defined above.

docs/source/guide/pt-5-theory.md
```diff
@@ -13,33 +13,181 @@ kernelspec:
...
+$\theta^2$. This means the worst-case coherent error
+$\sqrt{r(\mathcal{E})} \propto \theta$ is quadratically worse than the
+incoherent error $r(\mathcal{E}) \propto \theta^2$.
 
-For optimal results, Pauli Twirling should be implemented with an understanding of the underlying noise dynamics, and ideally, should be complemented with more direction error mitigation techniques to ensure robust quantum computation.
+After Pauli twirling, the off-diagonal terms vanish, and the remaining
+diagonal errors scale as $\theta^2$ --- the quadratic penalty is removed.
 
-In the context of quantum error mitigation, PT is closer to [DDD](ddd-5-theory.md), but stands apart as a noise tailoring technique. PT's peculiarities include:
+## Relationship to other techniques
```
natestemen: This is perhaps useful, but it does not belong on the theory page.

natestemen (changes requested):
Thanks for taking a stab at this! There's definitely some good changes in here, but it does seem like maybe some of it was AI-generated and maybe not thoroughly checked. This isn't a big issue, but if there are changes here that you aren't comfortable standing behind would you kindly remove them?

For the record, I'm not against AI-assistance here! I just don't want to review/ask questions about code/docs that you don't have a full understanding of.

docs/source/guide/pt-5-theory.md
```diff
@@ -13,33 +13,328 @@ kernelspec:
...
+
+- It does not reduce noise on its own, but tailors the noise so that downstream
+  techniques can mitigate it more effectively.
+
+- It generates multiple circuits with random Pauli modifications and averages
+  over their results. There is no need for linear combinations of noisy results
+  as in [ZNE](zne-5-theory.md), [PEC](pec-5-theory.md), and [CDR](cdr-5-theory.md).
+
+- The error mitigation overhead is minimal: there is no increase in statistical
+  uncertainty in the final result beyond the standard sampling variance.
```
natestemen: This feels like advantages/disadvantages, but this page should just be about the theory of PT.

docs/source/guide/pt-5-theory.md
```diff
@@ -13,33 +13,328 @@ kernelspec:
...
+the conjugation $U P_i U^\dagger$ always yields another Pauli, making this
+straightforward. The valid twirl pairs $(P, Q, R, S)$ for CNOT and CZ gates
+are stored in {attr}`mitiq.pt.pt.CNOT_twirling_gates` and
+{attr}`mitiq.pt.pt.CZ_twirling_gates`.
+
+Each two-qubit gate is sandwiched by a randomly sampled Pauli pair $(P, Q)$
+before and a corresponding pair $(R, S)$ after, chosen so that the ideal
+gate is preserved and only the noise gets twirled. The valid pairs
+$(P, Q, R, S)$ for CNOT and CZ gates are stored in
+{attr}`mitiq.pt.pt.CNOT_twirling_gates` and {attr}`mitiq.pt.pt.CZ_twirling_gates`.
```
natestemen: These two paragraphs are very repetitive.

docs/source/img/pt_qem_workflow.svg
natestemen: Can you edit the existing SVG so it's in the same style rather than creating a new one?

natestemen (commented):
Thanks for continuing to work on this. Can you say if/how you are using AI to assist in this PR? The theory section is a bit scattered and it feels not very cohesive/pedagogical. If you're not comfortable backing up what's in your current rewrite can you remove it?

natestemen (changes requested):
Thanks for your patience @axif0. There's a lot of solid work in here, and I want to find a path that gets the good parts merged without more full-revision cycles. My suggestion: let's trim this PR down to the parts you can fully stand behind, and handle the theory content separately.

**What I'd like to keep:**

- All the `pt-4-low-level.md` changes: the previously commented-out code now runs, the references to the removed `pauli_twirl_circuit()` are fixed, and the docstring now matches the amplitude damping noise actually used. These directly close checklist items from #2462.
- The `add_noise_to_two_qubit_gates` section in `pt-3-options.md`, including the `CIRQ_NOISE_OP` extension example — this is exactly what the issue asked for. The one-line export in `mitiq/pt/__init__.py` is fine to keep here too.
- The `num_circuits` section. One small ask: the "20 to 50 variants" guidance reads as a rule of thumb — either soften the phrasing ("as a starting point...") or add a citation if you have one.
- The `pt-2-use-case.md` advantages/disadvantages expansion, which addresses the coherent-vs-incoherent checklist item.
- Removing the under-construction warnings from the pages covered by the above.

**What I'd like changed:**

1. **PT + ZNE stacking now appears in three places**: the new code example in `pt-1-intro.md`, the new section in `pt-3-options.md`, and the existing [`pt_zne.md` tutorial](https://mitiq.readthedocs.io/en/stable/examples/pt_zne.html). The issue asked for a stacking subsection in `pt-3-options.md` only — let's keep it there, and have `pt-1-intro.md` just link to it and to the tutorial. Note the two copies of the new figure currently share the same label (`name: pt-qem-workflow`), which makes cross-references ambiguous, and the intro example adds a few hundred density-matrix simulations to every docs build.
2. **The PTM heatmap figure needs its generating code.** The physics in the image looks right to me, but a static PNG with no source isn't something we can maintain or verify. Please include the script that produces it — ideally as a code cell in the page (it can be hidden with a `remove-input` tag if it's long). The issue also asked for a table of which channels get tailored; the heatmaps could accompany that.

**What I'd like removed from this PR:**

3. **The `pt-5-theory.md` rewrite.** Following up on our earlier conversation — since you've said parts of this are beyond what you can currently vouch for, I'd rather not merge a rewrite of the theory page in that state. Let's revert this file in this PR; the theory improvements can happen in a follow-up (and I'm happy to help with that one directly).

With those changes I think this is one review away from merging, and it'll knock out most of #2462. Thanks again for the persistence here!

axif0:
Hey, @natestemen   I've updated all the required changes. Feel free to take a look whenever you're free.

docs/source/guide/pt-4-low-level.md
```diff
@@ -13,44 +13,38 @@ kernelspec:
...
+There is no complex inference step --- the final result is a simple average.
 
 ```{note}
-When setting the `num_trials` option to a value larger than one, multiple circuits are actually generated by Mitiq and 
-the associated results are averaged to obtain the final expectation value. This more general case is not shown in the figure since
-it can be considered as an average of independent single-circuit workflows.
+The `num_circuits` parameter in {func}`.generate_pauli_twirl_variants` controls how many
+twirled variants are generated. More variants lead to better convergence of the
+Pauli-twirled noise channel. See [What is the theory behind PT?](pt-5-theory.md)
+for a visualization of how this convergence plays out across different noise types.
```
natestemen: Outdated post section 5 removal.

natestemen (approved):
Just one blocking comment, then let's merge!

natestemen:
Thank you @axif0 for you work here!

PR #2960, opened 2026-03-16
Add calibration benchmark script for average ZNE improvement

### LINKED ISSUE #2828: benchmark mitiq performance on benchmarking circuits
Benchmark mitiq performance by using the Calibrator to specify a single QEM config, and run on ~5 benchmark circuits to compute the average improvement factor.

Can use https://mitiq.readthedocs.io/en/stable/examples/calibration-tutorial.html as a starting point, and modify the the Settings object.

axif0:
## Description
- Added benchmarking_mitiq_performance.py example that benchmarks a single ZNE configuration using Calibrator.
- Run ZNE on 5 standard benchmark circuits (GHZ, W, RB ×2, Mirror) under a depolarizing noise model.
- Compute and report per-circuit and average improvement factors using ExperimentResults data.

image

We can run this using -
```
uv run docs/source/examples/benchmarking_mitiq_performance.py
```

closes #2828

---

### License

### DIFF
--- a/scripts/benchmark.py
+++ b/scripts/benchmark.py
@@ -271,8 +271,7 @@ def main() -> None:
 
     plt.axhline(y=1.0, color="r", linestyle="--", label="No Improvement")
     plt.xlabel(
-        "Exact Resource Cost\n"
-        "(Avg number of operations executed per problem)",
+        "Exact Resource Cost\n(Avg number of operations executed per problem)",
         fontsize=11,
     )
     plt.ylabel("Improvement Factor", fontsize=12)


scripts/benchmark.py
```diff
@@ -0,0 +1,134 @@
...
+            "num_qubits": 2,
+            "circuit_depth": 10,
+        },
+    ],
+    strategies=[
+        {
+            "technique": "zne",
+            "scale_noise": fold_global,
+            "factory": LinearFactory([1.0, 3.0, 5.0]),
+        },
```
natestemen, line 131: Can you add a few more strategies? You can other options for these in the ZNE user guide.

scripts/benchmark.py
```diff
@@ -0,0 +1,134 @@
...
+
+    The Calibrator requires an executor that returns MeasurementResult
+    (bitstrings/counts), not expectation values.
+    """
+    noisy_circuit = circuit.with_noise(cirq.depolarize(NOISE_LEVEL))
+    result = cirq.DensityMatrixSimulator().run(
+        noisy_circuit, repetitions=SHOTS
+    )
+    bitstrings = np.column_stack(list(result.measurements.values()))
+    return MeasurementResult(bitstrings)
```
natestemen, line 72: Just to make this a touch more realistic can we have a separate 2-qubit error rate?

docs/source/examples/benchmarking_mitiq_performance.py
```diff
@@ -0,0 +1,134 @@
...
+#
+# This source code is licensed under the GPL license (v3) found in the
+# LICENSE file in the root directory of this source tree.
+
+"""Benchmark mitiq performance on benchmarking circuits.
+
+Uses the Calibrator with a single ZNE configuration across 5 distinct
+benchmark circuits and computes the average improvement factor.
+
+Reference: https://github.com/unitaryfund/mitiq/issues/2828
```
natestemen:
```suggestion
```
Unnecessary.

scripts/benchmark.py
```diff
@@ -0,0 +1,134 @@
...
+            f"mitigated_err={mitigated_err:.4f}  "
+            f"[{improved}]"
+        )
+
+    avg_factor = np.mean(improvement_factors)
+    print("-" * 50)
+    print(f"Average improvement factor: {avg_factor:.4f}")
+
+    best = cal.best_strategy()
+    print(f"\nBest strategy: {best.to_dict()}")
```
natestemen, line 231: Can you also have it scatter the results where the x axis is "resources used" which should be something like # of gates, and the y axis is improvement factor? It doesn't need to save the plot, but just creating it would be helpful to see how all the techniques perform as a function of cost.

axif0: Ya sure, like -

image

natestemen: I think there should only be one data point per QEM strategy and the performance/overhead should be the averages across all of the problems it ran. Making this interactive would be very cool to see problem/strategy filtering, but that might be another PR.

axif0: Do you mean, like average of the resources and average of the improvement factors across all 5 problems ? @natestemen like this ?


image

natestemen: Yes exactly! It looks a bit silly since everything uses the same number of gates, but if you through in a PEC strategy that should consume a different number of resources and then could be more illuminating.

axif0: Based on the document what I understand, as PEC didn't use depth like ZNE. And I use PEC strategy with 100 samples to the benchmark to compare against ZNE. Instead of everything stacking up at the same point, ZNE resource cost need to based on its total scale factors, while PEC also need to plotted near the base circuit depth but mapped to its own shot overhead ratio. Is my understanding correct ? @natestemen 

Like this -

image

natestemen: Instead of estimating the resource cost it would be more accurate to just track how large the circuits are that get run through the executor function. This way there would be no guessing.

axif0: used monkeypatch internal `Strategy` class, so we can intercept its execution and track exactly how many circuits each strategy is running.

image

scripts/benchmark.py
```diff
@@ -0,0 +1,292 @@
...
+    #             comparable on the same plot.
+    plt.figure(figsize=(10, 6))
+    for strategy_id, strategy in enumerate(cal.strategies):
+        s_dict = strategy.to_dict()
+        tech = s_dict.get("technique", "")
+
+        improvements = []
+        # Calculate exactly how many operations this strategy executed per
+        # problem on average
+        avg_resource = exact_costs.get(strategy_id, 0) / len(cal.problems)
```
natestemen, line 246: Can you help me understand how this is working? When does the length of the circuits get stored?

axif0: When [cal.run()](https://github.com/unitaryfoundation/mitiq/pull/2960/changes/BASE..ab8221737bdc14a75c3f22ba06630582f30bf3cb#diff-c1f82f207bd893fca55d0173c7fe0d91bd1d16a225ca766165de942e4eeaf407R205) starts, monkeypatch sets `current_strategy_id = 0` to indicate the first strategy is beginning its work. Then, the strategy (like ZNE or PEC) processes the circuit and passes it to the noisy_execute function to be simulated; right at the start of that function, it calculates the length of the circuit [(len(list(circuit.all_operations())))](https://github.com/unitaryfoundation/mitiq/pull/2960/changes/BASE..ab8221737bdc14a75c3f22ba06630582f30bf3cb#diff-c1f82f207bd893fca55d0173c7fe0d91bd1d16a225ca766165de942e4eeaf407R44) and immediately adds that number to the running total in the exact_costs dictionary under the current strategy's ID, repeating this accumulation for every single circuit scaled or sampled by that strategy before moving on to the next one. 
Apologize, I am still familiarizing myself with the deep quantum theory behind these mitigation models. Maybe made some mistake unwillingly. :sweat_smile:

natestemen: Got it. I think this is exposing a missing functionality in the calibration module. We shouldn't have to monkey patch this, but it seems like it's need ATM. I've opened https://github.com/unitaryfoundation/mitiq/issues/2972 to resolve this later, but for now we'll keep the monkey patch here.

axif0:
@natestemen thank you for checking out. :smile:

PR #2963, opened 2026-03-18
Shadows user guide

natestemen:
## Description

Completes the shadows user guide.

part of https://github.com/unitaryfoundation/mitiq/issues/2110


purva-thakre:
@natestemen There are merge conflicts...

docs/source/guide/shadows.md
```diff
@@ -11,33 +11,27 @@ kernelspec:
...
-
 # Classical Shadows
 
-Classical shadows protocol {cite}`huang2020predicting` aims to create an approximate classical representation
-of a quantum state using minimal measurements. The protocol is based on the idea of shadow tomography,
-which is a technique for reconstructing a quantum state from a small number of measurements.
-This approach not only characterizes and mitigates noise effectively but also retains
-sample efficiency and demonstrates noise resilience {cite}`chen2021robust`. For more details, see the section
-[Classical Shadow Protocol and its Robust Estimation](shadows-5-theory.md).
+Classical shadows {cite}`huang2020predicting` create an approximate classical representation of a quantum state using minimal measurements.
```
purva-thakre: minimal measurements on what? might want to clarify it a bit further.

docs/source/guide/shadows.md
```diff
@@ -11,33 +11,27 @@ kernelspec:
...
+The protocol is based on shadow tomography, a technique for reconstructing properties of a quantum state from a small number of measurements.
+This approach not only characterizes and mitigates noise effectively but also retains sample efficiency and demonstrates noise resilience {cite}`chen2021robust`.
```
purva-thakre, line 18: add something along the lines of: this is RSE 
 
There are two workflow images on this page but you are only describing what CS is in the intro.

Edit: I accidentally commented instead of starting a review. Will go through the other sections as well.

purva-thakre:
@natestemen Sorry, I completely forgot about this. Will get to reviewing this over the next few days.

natestemen:
@FarLab I wouldn't mind your eyes on this too :)

docs/source/guide/shadows.md
```diff
@@ -11,33 +11,29 @@ kernelspec:
...
 # Classical Shadows
 
-Classical shadows protocol {cite}`huang2020predicting` aims to create an approximate classical representation
-of a quantum state using minimal measurements. The protocol is based on the idea of shadow tomography,
-which is a technique for reconstructing a quantum state from a small number of measurements.
-This approach not only characterizes and mitigates noise effectively but also retains
-sample efficiency and demonstrates noise resilience {cite}`chen2021robust`. For more details, see the section
-[Classical Shadow Protocol and its Robust Estimation](shadows-5-theory.md).
+Classical shadows {cite}`huang2020predicting` create an approximate classical representation of a quantum state using minimal measurements of copies of said state.
+The protocol is based on shadow tomography, a technique for reconstructing properties of a quantum state from a small number of measurements.
```
purva-thakre: add a reference for 'shadow tomography'

docs/source/guide/shadows.md
```diff
@@ -11,33 +11,29 @@ kernelspec:
   name: python3
 ---
 
-```{admonition} Note:
-The documentation for Classical Shadows in Mitiq is still under construction. This users guide will change in the future.
-```
-
```
purva-thakre, line 17: is it worth adding an admonition here that shadows is an experimental technique. Something similar to what you have in the readme to let the reader know the technique is not stable. 

> Experimental techniques are not covered by mitiq's semantic versioning guarantees. A technique graduates to stable once it has broad test coverage, documented user guides, and has seen real-world validation on hardware or well-studied noise models. If you are using an experimental technique and would like to help it graduate, please open an issue or contribute to the discussion on [GitHub](https://github.com/unitaryfoundation/mitiq/issues)

docs/source/guide/shadows-1-intro.md
```diff
@@ -11,66 +11,35 @@ kernelspec:
...
 
 ```{code-cell} ipython3
 import cirq
+import numpy as np
+from mitiq import MeasurementResult, PauliString
+from mitiq.experimental import shadows
+from mitiq.interface.mitiq_cirq.cirq_utils import (
+    sample_bitstrings as cirq_sample_bitstrings,
+)
+```
```
purva-thakre: this code block raises the warning about the technique not being stable. it is also worth adding a note reiterating the same message.

docs/source/guide/shadows-5-theory.md
```diff
@@ -11,186 +11,122 @@ kernelspec:
...
+The protocol works in two steps:
 
-## 1. Classical Shadow
-The "classical shadow" technique, as proposed by {cite}`huang2020predicting`, offers an innovative approach to quantum state approximation. This method is particularly advantageous for predicting the properties of complex, large-scale quantum systems as it requires exponentially fewer measurements. In quantum theory, the quantities of interest are often linear functionals of the quantum state $\rho$, such as the expectation values $o_i$ of a set of self-adjoint operators $\{O_i\}_{i}$:
-\begin{equation}
-    o_i=\mathrm{Tr}(O_i \rho),\qquad i\in\mathbb{N}^+,~~ 1\leq i\leq M.
-\end{equation}
+- **Step 1: Quantum processing.** Apply a random unitary $U$ sampled from a fixed ensemble $\mathcal{U}$, measure in the computational basis, and record the outcome $b$ and the unitary $U$.
```
purva-thakre:
```suggestion
- **Step 1: Quantum processing.** Apply a random unitary $U$ to the unknown state. $U$ is sampled from a fixed ensemble $\mathcal{U}$, measure in the computational basis, and record the outcome $b$ and the unitary $U$.
```
what is `b` here? add some info about that.

docs/source/guide/shadows-5-theory.md
```diff
@@ -11,186 +11,122 @@ kernelspec:
...
 
+- **Step 2: Classical post-processing.** Apply an inverse channel $\mathcal{M}^{-1}$ to each measurement record to obtain a *classical shadow* $\hat{\rho}$, then estimate observables as $\hat{o}_i = \mathrm{Tr}(O_i \hat{\rho})$.
```
purva-thakre: inverse channel of what? you should add some information about this.

docs/source/guide/shadows.md
natestemen: good catch! https://github.com/unitaryfoundation/mitiq/pull/2963/commits/b3171f29fc6da6526b3728ae1217687f6f77d9b3

docs/source/guide/shadows.md
natestemen: I added a warning to every page in https://github.com/unitaryfoundation/mitiq/pull/2963/commits/63ec161d8c3d92bfc5f26e24fc359aee6d8d7df1. It's a bit shorter than your suggestion so let me know if you think it need more info.

docs/source/guide/shadows-5-theory.md
natestemen: I found it hard to write clearly what step two does in plain english. My attempt in https://github.com/unitaryfoundation/mitiq/pull/2963/commits/cec726cb26d5dbc7ab35c76a99abaa50f4c88367 is the best I could come up with. Let me know what you think.

FarLab:
Hey @natestemen, here is a review produced by Claude Fable 5. It is possible some of these points raised were already discussed in this thread. I double checked everything except 5,6 and 7 (I don't think there are hallucinations there, but just so you know).

Claude: 
Nice restructuring — the five-section guide format matches the other techniques, everything is moved onto the `mitiq.experimental.shadows` API, and the cross-file plumbing checks out (imports, kwargs, citations, links, and include targets all match the source). My comments are mostly about technical accuracy and the switch from cached data to live execution. Ordered by severity:

### 1. Calibration-reuse note drops a condition — following it produces silently wrong results

`docs/source/guide/shadows-1-intro.md` (~line 150), duplicated in `shadows-3-options.md` (~line 229):

> You do not need to re-run `pauli_twirling_calibrate` between experiments as long as the number of qubits and the noise channel have not changed.

The old guide also required the new observables to have weight ≤ the calibrated `locality`. That condition matters: `expectation_estimation_shadow` does `fidelities.get(b, np.inf)` (`classical_postprocessing.py:258`), so reusing a `locality=2` calibration (as every example in the guide sets up) for a weight-3 observable divides by infinity and returns exactly **0.0 with no error or warning**. Reproduced at runtime: the same data returns 27.0 uncalibrated and 0.0 with a locality-2 fidelities dict. Please restore the observable-weight condition in both notes.

### 2. The robust-shadows demo injects noise the technique cannot calibrate away

`docs/source/guide/shadows-1-intro.md` (~line 110): `cirq_sample_bitstrings(..., noise_level=(0.05,))` uses `circuit.with_noise(...)`, which applies amplitude damping after **every** moment — including the three GHZ state-prep moments. RSE only calibrates rotation/measurement noise (as this page and `shadows-2-use-case.md` themselves state), so the calibrated estimates converge to the state-prep-noisy ceiling (~0.81 / ~0.92 for the two ZZ observables), not the exact value 1.0. The improvement over uncalibrated (~0.03–0.08) is within shot noise at 5000 measurements, so the rendered output doesn't demonstrate calibration working. The old page avoided this by inserting noise only immediately before measurement — suggest restoring that executor style for this demo.

### 3. `k_shadows` explanation is backwards

`docs/source/guide/shadows-3-options.md` (~line 108):

> A larger `k_shadows` reduces the influence of outlier snapshots at the cost of using more measurements per estimate.

`k_shadows` is pure post-processing over the already-collected, fixed set of snapshots (`shadows_utils.py`: `batch_size = len(bits) // num_batches`). A larger `k_shadows` means **fewer** snapshots per batch (higher per-batch variance); total measurements are unchanged. The cell comment a few lines below ("splits snapshots into 10 batches") already says it correctly. See Algorithm 1 in the paper https://arxiv.org/abs/2002.08953.

### 4. Mis-stated shadow-norm bounds on the theory page

`docs/source/guide/shadows-5-theory.md` (~lines 73, 76):

- "the shadow norm **equals** the Hilbert-Schmidt norm, $\|O\|_{\mathrm{shadow}} \leq 3\mathrm{Tr}[O^2]$" — "equals" followed by an inequality is self-contradictory (and equality doesn't hold).
- Both bounds drop the square from Huang et al. (2020): they should be $\|O\|^2_{\mathrm{shadow}} \leq 3\mathrm{Tr}(O^2)$ and $\|O\|^2_{\mathrm{shadow}} \leq 4^w \|O\|^2_\infty$ (line 76 is also dimensionally inconsistent and omits the $\infty$ subscript). Since the sample-complexity formula earlier on the page correctly uses $\|O_i\|^2_{\mathrm{shadow}}$, the un-squared bounds imply quadratically wrong scaling.

The missing squares are carried over from the old page, but this rewrite seems like the moment to fix them. The exact reference for these inequalities are equations S16 and S17 in the paper https://arxiv.org/abs/2002.08953.
Related nit: $|\sigma_0\rangle\!\rangle$ (~line 109) is now used without the old footnote that defined the double-ket/PTM notation. 

### 5. "Can be passed directly" hides a representation switch

`docs/source/guide/shadows-4-low-level.md` (~line 206): when `fidelities` is passed, `classical_snapshot` / `shadow_state_reconstruction` return a **1-D PTM vector of length $4^n$** (shape `(64,)` for n=3, verified empirically) instead of the $2^n \times 2^n$ density matrix the page promises at line 114. A reader applying the page's own `np.trace(rho_shadow)` check to the calibrated result gets a `ValueError`. The high-level `classical_post_processing(..., state_reconstruction=True, calibration_results=...)` behaves the same way. The sentence needs a caveat about the output representation.

### 6. Build cost + nondeterminism: cached data replaced by ~26k live circuit simulations per docs build

The old intro loaded 5000-shot results from pickles; the new pages run everything live and unseeded: intro 5000+5000, options 5000+5000+3000+1000, low-level 3000+~900. Two consequences:

- Minutes of compute in every CI/nightly docs build. E.g. `shadows-3-options.md` (~line 182) runs a 5000-measurement calibration whose result is immediately overwritten by the next cell's 5000-measurement calibration. Demo counts could be cut ~10× wherever only the API is being illustrated.
- Published numbers change on every build. Passing `random_seed` (which the module supports and the options page documents) would make outputs reproducible.

Also: the now-unreferenced `shadows-1-intro-output1.pkl`, `shadows-1-intro-output2.pkl`, and `shadows-1-intro-PTC-50000.pkl` in `docs/source/examples/resources/` should be deleted in this PR since it removes their last references.

### 7. tqdm monkeypatch duplicated because the API has no opt-out

`shadow_quantum_processing` wraps execution in tqdm unconditionally (`quantum_processing.py:121`), so three guide pages (plus the two tutorials) each carry a hidden `tqdm.tqdm.__init__ = partialmethod(..., disable=True)` cell. Consider a `progress_bar: bool = True` kwarg on the API instead (possibly a follow-up issue). Note the copy in `shadows-4-low-level.md` is dead code either way — that page never calls the tqdm path and the cell can be deleted now.

### 8. The `qubits` option demo is a no-op

`docs/source/guide/shadows-3-options.md` (~line 92): the prose says "restrict measurements to a subset of qubits" but the cell passes `qubits=[qubits[0], qubits[1], qubits[2]]` — all three qubits, identical to the default. Passing `qubits[:2]` would actually demonstrate the option.

### Minor nits

- `shadows-3-options.md` ~line 164: dangling colon — "Increasing it reduces sensitivity to outlier calibration shots:" is followed by the unrelated `locality` paragraph instead of an example.
- Unused imports: `import numpy as np` in `shadows-1-intro.md` and `shadows-3-options.md`; `from functools import partial` in `shadows-1-intro.md` (~line 108, leftover from the old executor pattern).
- `shadows-3-options.md` ~line 194: `import cirq as cirq_` re-imports cirq under an alias although the setup cell already imports it — `cirq.Circuit()` works fine.

natestemen:
In order to address some of those comments @FarLab it seemed like the best course of action was to make code changes rather than just documentation, so this PR got a bit wider in scope. Since this technique is experimental, I'm okay with that.

natestemen:
@FarLab @purva-thakre can someone give a (hopefully) final look through this?

natestemen:
bump

purva-thakre:
Hello, sorry for the delay. I promise to have it done by this weekend. 

BTW is it possible to turn on `coderabbit` reviews for mitiq? It is free for open-source projects. https://www.coderabbit.ai/ IDK who at UF would be the person responsible for this. Alessandro?

purva-thakre:
ignore the noise. I am going to try and use Claude as a reviewer for the first time. 

Edit: never mind, figured out a way around claude directly making comments on the PR

PR #2965, opened 2026-03-18
Clean up shadows code

natestemen:
## Description

Just scratching the surface on cleaning up some code. No functionality is changed, just renaming things and bringing docstrings in line with functionality. Review can just be a sanity check.

part of https://github.com/unitaryfoundation/mitiq/issues/2923


docs/source/examples/shadows_tutorial.md
```diff
@@ -171,12 +171,12 @@ Equation {eq}`eq-label` comes from equation S13 in the paper {cite}`huang2020pre
...
+# minimum number of measurements required for error rate epsilon
+num_measurements = n_measurements_tomography_bound(epsilon, num_qubits)
 
-print("n_total_measurements = {}".format(n_total_measurements))
-shadow_outcomes = shadow_quantum_processing(
-    test_circuits, cirq_executor, n_total_measurements
+print(f"required measurements: {num_measurements}")
+bitstrings, bases = shadow_quantum_processing(
+    test_circuit, cirq_executor, num_measurements
 )
```
purva-thakre, line 184: by `bases`, this means measurement bases, right? I wonder if it's easier to rename it to `measurement_bases` instead? Makes it easy to understand what the function is returning. 

The docstring for this function is a bit confusing in terms of understanding what it returns https://mitiq.readthedocs.io/en/stable/apidoc.html#mitiq.shadows.shadows.shadow_quantum_processing

mitiq/shadows/tests/test_shadows.py
```diff
@@ -77,9 +77,9 @@ def test_shadow_quantum_processing():
     assert isinstance(result, tuple), f"Expected a tuple, got {type(result)}"
 
     # Check that the tuple contains two lists
-    assert (
-        len(result) == 2
-    ), f"Expected two lists in the tuple, got {len(result)}"
+    assert len(result) == 2, (
+        f"Expected two lists in the tuple, got {len(result)}"
+    )
```
purva-thakre:
```suggestion
    assert len(result) == 2
```

Since this is a unit test, I don't see a need to have this. If the test fails, it will fail with AssertionError anyway.

mitiq/shadows/tests/test_shadows.py
```diff
@@ -97,9 +97,9 @@ def test_classical_post_processing():
     )
 
     # Check that the result is a dictionary
-    assert isinstance(
-        result, dict
-    ), f"Expected a dictionary, got {type(result)}"
+    assert isinstance(result, dict), (
+        f"Expected a dictionary, got {type(result)}"
+    )
```
purva-thakre:
```suggestion
    assert isinstance(result, dict)
    )
```
same as my other comment, this will fail with `AssertionError` anyway.

mitiq/shadows/tests/test_shadows_utils.py
```diff
@@ -67,6 +67,6 @@ def test_n_measurements_opts_expectation_bound():
 def test_fidelity():
     state_vector = np.array([0.5, 0.5, 0.5, 0.5])
     rho = np.eye(4) / 4
-    assert np.isclose(
-        fidelity(state_vector, rho), 0.25
-    ), f"Expected 0.25, got {fidelity(state_vector, rho)}"
+    assert np.isclose(fidelity(state_vector, rho), 0.25), (
+        f"Expected 0.25, got {fidelity(state_vector, rho)}"
+    )
```
purva-thakre, line 72:
```suggestion
    assert np.isclose(fidelity(state_vector, rho), 0.25)
```
same as my other comment

mitiq/shadows/quantum_processing.py
```diff
@@ -64,15 +62,15 @@ def get_rotated_circuits(
...
                 rotated_circuit.append(cirq.H(qubit))
             # Pauli Z measurement
-            else:
-                assert (
-                    pauli == "Z"
-                ), f"Pauli must be X, Y, Z. Got {pauli} instead."
+            elif pauli != "Z":
+                raise ValueError(
+                    f"Pauli must be X, Y, or Z. Got {pauli!r} instead."
+                )
```
purva-thakre: I would have expected the entire code block to follow:

- if X
- elif Y
- elif Z
- else ValueError 

the logic for `elif Z` is missing. IDK if there is something for it somewhere else in the code. Haven't checked 😅

purva-thakre (approved):
Took a quick look.

mitiq/shadows/quantum_processing.py
natestemen: Yeah agree it's quite strange. Since you don't need to do modify the circuit when measuring in the Z basis it just skips the "if Z" case. I refactored into something that is hopefully more understandable. I also removed a parameter from the function `add_measurements` since it was always being set to True.

docs/source/examples/shadows_tutorial.md
natestemen: Ah yeah good catch. These changes (since they relate to the tutorials) should not have been part of this PR since there is a separate PR for the tutorials: https://github.com/unitaryfoundation/mitiq/pull/2964/. Removing from here, but point taken about the docstrings of `shadow_quantum_processing`. Actually, since it's a small wrapper around `random_pauli_measurement` I might actually get rid of it.

natestemen: Since I have a few PRs open right now for shadows, removing that function would require changes across many of those PRs, so I'm going to wait for some to land before removing it (and clarifying the docstring).

PR #2966, opened 2026-03-18
Move less tested techniques to experimental module

natestemen:
## Description

Since we are shipping mitiq 1.0 soon, we want to make sure we adhere to semver thereafter. Since some of our techniques are newer, less tested, and potentially prone to changes, @andreamari had the great idea of moving the techniques to an experimental module that would indicate they are prone to breaking changes (even without a new major release). I like this idea as it allows us to keep making changes to these techniques without the heavier burden of a new major release.

The main questions are:
1. which techniques are experimental?
2. is this the "right" way to go about this?


CHANGELOG.md
```diff
@@ -1,5 +1,28 @@
 # Changelog
 
+## Version 1.0.0 (unreleased)
+
+### Breaking Changes
+
+- `mitiq.pea`, `mitiq.shadows`, and `mitiq.vd` are no longer accessible from
+  the top-level `mitiq` namespace. Code using `from mitiq import pea` (or
+  `shadows`/`vd`) will raise an `ImportError` and must be updated.
```
cosenal, line 9: OK, so we are not going to intentionally raise a warning pointing users to the new path, then?

README.md
```diff
@@ -94,12 +94,16 @@ mitiq.qem_methods()
...
-| Robust Shadow Estimation   🚧           | [RSE](https://mitiq.readthedocs.io/en/stable/guide/shadows.html)| [`mitiq.shadows`](https://github.com/unitaryfoundation/mitiq/tree/main/mitiq/shadows) | [2011.09636](https://arxiv.org/abs/2011.09636) <br> [2002.08953](https://arxiv.org/abs/2002.08953)|
 | Layerwise Richardson Extrapolation  | [LRE](https://mitiq.readthedocs.io/en/stable/guide/lre.html) |  [`mitiq.lre`](https://github.com/unitaryfoundation/mitiq/tree/main/mitiq/lre) | [2402.04000](https://arxiv.org/abs/2402.04000) |
-| Probabilistic Error Amplification 🚧  | Coming soon |  `mitiq.pea` | [Nature](https://www.nature.com/articles/s41586-023-06096-3) |
-| Virtual Distillation 🚧  | Coming soon |  `mitiq.vd` | [APS](https://journals.aps.org/prx/abstract/10.1103/PhysRevX.11.041036) |
 
-> 🚧: Technique is currently a work in progress or is untested and may have some rough edges. If you try any of these techniques and have suggestions, please open an issue!
+The following techniques are experimental (`from mitiq import experimental`) and may have unstable APIs:
```
cosenal: It'd be nice to give an explanation of when a technique is considered experimental, in order to provide the user with some expectation on when they will turn into stable.

CHANGELOG.md
natestemen: That's the nice thing to do. Added in https://github.com/unitaryfoundation/mitiq/pull/2966/commits/5a9aa2b6b3005e83f08d6281c07da3e9200ca59f.

README.md
natestemen: I added a note in https://github.com/unitaryfoundation/mitiq/pull/2966/commits/babd9d3ed2ccf2a915c2553c94c45ac519728411. Let me know if this is the direction you were thinking.

cosenal: Yes, excellent phrasing.

cosenal (approved):
🛳️

PR #2967, opened 2026-03-18
Bump wheel to 0.46.3

cosenal:
A new version of #2908, which should pass the builds 🤞 

This closes https://github.com/unitaryfoundation/mitiq/security/dependabot/14

Pinning diastatic-malt is necessary because, without that constraint, `uv lock --upgrade-package wheel==0.46.2` (which we want to do to resolve the security vulnerability) resolves PennyLane onto a broken `diastatic-malt` release 🤷


cosenal:
See also https://github.com/PennyLaneAI/pennylane/pull/9160

natestemen (commented):
Any idea why the lockfile diff so big just for a transitive dependency bump? There are a lot of dependencies that get upgraded here. Maybe they're all reliant on the one you bumped?

natestemen (approved):

PR #2971, opened 2026-03-21
remove remnant of AUTHORS file

natestemen:
## Description

Does what it says on the 🥫.


cosenal (approved):
Approved. 
And thanks to the fact that I reviewed it, I know that mitiq has dropped the AUTHORS file, which is something I had missed in the past.

PR #2974, opened 2026-03-23
add AI contribution policy

natestemen:
## Description

Does what it says on the 🥫. Language taken from [zulips](https://github.com/zulip/zulip/blob/main/CONTRIBUTING.md?plain=1#ai-use-policy-and-guidelines) and [LLVMs](https://llvm.org/docs//AIToolPolicy.html) AI guides/policies..


bachase (approved):
Looks great

1. Would you want to add a statement that maintainers/reviewers might choose to close a PR (or even ban someone?) if they don't follow these standards? I'm particularly thinking of PRs where someone is just piping back and forth to an LLM. Maybe not needed to be explicit.
2. Is there any guidance or requirement on the license-a-bility of contributed code? I didn't see anything already (or in other UF projects) that requires a contributor to attest the code is ok to contribute. AI just might make that harder to monitor.

PR #2984, opened 2026-04-01
Migrate np.polyfit to np.polynomial.polynomial.polyfit

### LINKED ISSUE #2322: Remove usage of `np.polyfit`
We use `np.polyfit` to define the extrapolation methods for ZNE. The related API in `numpy` has since [changed](https://numpy.org/doc/stable/reference/generated/numpy.polyfit.html) to using `numpy.polynomial` instead of `numpy.polyfit`.

https://github.com/unitaryfund/mitiq/blob/9e1cbb5a134983f247f1ed74cbe932a0156e76d1/mitiq/zne/inference.py#L155-L158

The[ transition guide](https://numpy.org/doc/stable/reference/routines.polynomials.html#quick-reference) has a helpful table for adjusting to the newer changes. 

image

danstam:
Fixes #2322

Replaces the deprecated `np.polyfit` API with `np.polynomial.polynomial.polyfit` throughout `mitiq/zne/inference.py`.

### What changed and why

The main challenge was that the new API removed the `cov=True` option. The covariance matrix is now reconstructed manually using the weighted Vandermonde approach:
```python
cov = (residuals / (n - deg - 1)) * inv((W @ V).T @ (W @ V))
```

One non-obvious issue: `full=True` silently suppresses `RankWarning` — numpy only raises it when `full=False`. A manual rank-deficiency check (`rank < deg + 1`) was added to preserve the existing `ExtrapolationWarning` behavior.

Other changes:
- Fixed coefficient ordering reversal across all affected call sites — the new API returns coefficients in ascending order `[a_0, a_1, ..., a_n]` vs the old descending order `[a_n, ..., a_1, a_0]`
- Migrated `np.polyval` → `np.polynomial.polynomial.polyval` (argument order is reversed in the new API)
- Removed `[::-1]` reversals in `_ansatz_unknown` and `_ansatz_known` closures
- Updated hardcoded test assertions to match ascending coefficient order
- Added `cast(npt.NDArray[np.float64], ...)` to satisfy mypy since numpy's type stubs for `full=True` return an imprecise type

All 216 ZNE tests pass. Ruff and mypy clean.

### AI use disclosure
I used Claude Code  to help navigate the codebase and verify the Vandermonde covariance math. All implementation decisions and code reviews were done by me iteratively.

### DIFF
--- a/mitiq/zne/inference.py
+++ b/mitiq/zne/inference.py
@@ -145,8 +145,8 @@ def mitiq_polyfit(
     weights: npt.ArrayLike | None = None,
 ) -> tuple[list[float], npt.NDArray[np.float64] | None]:
     """Fits the ansatz to the (scale factor, expectation value) data using
-    ``numpy.polynomial.polynomial.polyfit``, returning the optimal parameters
-    and covariance matrix of the parameters.
+    polynomial fitting, returning the optimal parameters and covariance matrix
+    of the parameters.
 
     Args:
         scale_factors: The array of noise scale factors.
@@ -194,7 +194,9 @@ def mitiq_polyfit(
         warnings.warn_explicit(
             warn.message, warn.category, warn.filename, warn.lineno
         )
-    return list(opt_params)[::-1], params_cov
+    return list(opt_params)[::-1], (
+        params_cov[::-1, ::-1] if params_cov is not None else None
+    )
 
 
 class Factory(ABC):
@@ -866,7 +868,7 @@ class PolyFactory(BatchedFactory):
 
         if params_cov is not None:
             if params_cov.shape == (order + 1, order + 1):
-                zne_error = np.sqrt(params_cov[0, 0])
+                zne_error = np.sqrt(params_cov[order, order])
 
         def zne_curve(scale_factor: float) -> float:
             return cast(float, np.polyval(opt_params, scale_factor))
@@ -1474,7 +1476,7 @@ class PolyExpFactory(BatchedFactory):
         if params_cov is not None:
             if params_cov.shape == (order + 1, order + 1):
                 zne_error = np.exp(z_coefficients[-1]) * np.sqrt(
-                    params_cov[0, 0]
+                    params_cov[order, order]
                 )
 
         # Parameters from low order to high order
--- a/mitiq/zne/inference.py
+++ b/mitiq/zne/inference.py
@@ -194,7 +194,7 @@ def mitiq_polyfit(
         warnings.warn_explicit(
             warn.message, warn.category, warn.filename, warn.lineno
         )
-    return list(opt_params), params_cov
+    return list(opt_params)[::-1], params_cov
 
 
 class Factory(ABC):
@@ -857,7 +857,7 @@ class PolyFactory(BatchedFactory):
             scale_factors, exp_values, order
         )
 
-        zne_limit = opt_params[0]
+        zne_limit = opt_params[-1]
 
         if not full_output:
             return zne_limit
@@ -869,10 +869,7 @@ class PolyFactory(BatchedFactory):
                 zne_error = np.sqrt(params_cov[0, 0])
 
         def zne_curve(scale_factor: float) -> float:
-            return cast(
-                float,
-                np.polynomial.polynomial.polyval(scale_factor, opt_params),
-            )
+            return cast(float, np.polyval(opt_params, scale_factor))
 
         return zne_limit, zne_error, opt_params, params_cov, zne_curve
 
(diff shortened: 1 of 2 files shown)


natestemen (changes requested):
I don't want this to be a breaking change, so can you make sure the coefficients are returned in the same order as previously?

danstam:
Added a reversal on the return value to keep the original coefficient ordering.

mitiq/zne/tests/test_inference.py
natestemen, line 1: In order to not introduce breaking changes, I think the test files should remain the same, right?

mitiq/zne/inference.py
```diff
@@ -145,8 +145,8 @@ def mitiq_polyfit(
     weights: npt.ArrayLike | None = None,
 ) -> tuple[list[float], npt.NDArray[np.float64] | None]:
     """Fits the ansatz to the (scale factor, expectation value) data using
-    ``numpy.polyfit``, returning the optimal parameters and covariance matrix
-    of the parameters.
+    ``numpy.polynomial.polynomial.polyfit``, returning the optimal parameters
+    and covariance matrix of the parameters.
```
natestemen: Lets remove the use of the specific function that's being used. This shouldn't matter to the user, and if it does they can look at the code :).

mitiq/zne/tests/test_inference.py
danstam: The `opt_params` ordering tests are all reverted. The two covariance matrix assertions changed because the Vandermonde reconstruction naturally produces the matrix in ascending coefficient order — so I reversed `opt_params` before returning but didn't reverse `params_cov` since it was already in the correct internal order for the `zne_error` computation. But you're right — the cleaner fix is to also reverse `params_cov[::-1, ::-1]` on the return so everything matches the original and the tests stay identical to main. Done.

mitiq/zne/inference.py
danstam: Good point — updated.

danstam:
@natestemen Let me know if you need anything else.

natestemen (commented):
Sorry for the delay here!

I think it makes more sense to use scipy here instead of computing some of the needed data more manually. WDYT of something like this?

```py
from scipy.optimize import curve_fit
from numpy.polynomial.polynomial import polyval

def poly_model(x, *coeffs):
    return polyval(x, coeffs)

popt, pcov = curve_fit(
    poly_model, scale_factors, exp_values,
    p0=np.ones(deg + 1),
    sigma=1/w if w is not None else None,
    absolute_sigma=True,
)
```

This saves us having to maintain numerics code which I would prefer.

danstam:
Happy to implement `curve_fit`! Since you asked for my thoughts, I want to highlight a few architectural trade-offs before we finalize this.

The custom Vandermonde logic currently in the branch is a direct linear calculation tailored specifically for this exact scenario. `curve_fit` is a non-linear iterative solver, which means we have to add some 'glue code' to force it to mimic the legacy API:

1. As you noted in your snippet, we have to map weights to uncertainties (`sigma = 1/w`). However, we also have to add explicit, vectorized safeguards to catch `w=0` to prevent `ZeroDivisionError`s.
2. We have to intercept SciPy's `OptimizeWarning` and manually cast it to Mitiq's `ExtrapolationWarning` to keep the test suite green.

More importantly, wrapping `curve_fit` means we are just swapping our dependency risk. The PR was opened because an external API changed. If we use basic linear algebra, we isolate ourselves from future upstream API breaks. If we use `curve_fit`, we are reliant on SciPy not changing its warning handling or solver tolerances in the future.

That said, I completely understand if the project priority is avoiding custom math at all costs. Let me know which architectural direction you prefer, and I can implement the final version!

kawacukennedy (commented):
Thanks for this thorough PR @danstam. The cov matrix reconstruction using the Vandermonde approach is clean, and I appreciate the detailed explanation of the coefficient ordering fix.

A couple of observations after reading through:
1. The manual rank-deficiency check you added to preserve ExtrapolationWarning behavior is a nice touch — that's exactly the kind of edge case that's easy to miss in a migration like this.
2. I see @natestemen suggested exploring scipy's curve_fit as an alternative. Have you had a chance to benchmark both approaches to compare performance/memory? The Vandermonde approach may be lighter for this linear case.

I'm happy to help test either approach locally once the direction is finalized.

---

*Disclosure: I'm a new mitiq contributor and this is my first review. Feedback on my review itself is welcome.*

danstam:
Hey @kawacukennedy, 
 Since you asked for feedback on your review: Your comment doesn't add value to the current technical discussion. If you look at the PR history, you'll see I’ve already addressed @natestemen's points and we're currently in an active exchange. AI can be a great tool, but it's important to use it to add substance rather than just summarizing active conversations, which creates unnecessary noise. To be most helpful, please focus your efforts on unassigned open issues instead.

natestemen:
@danstam thanks for your patience. I've been busy with unitaryHACK issues.

As I've thought about this more, I lean more and more towards not fixing what's not broken. Right now we use deprecated numpy code, but it's working perfectly fine. Until the code is removed, I think we should not make any changes here. I'm just not seeing a huge advantage. If you have arguments for making the change I'm happy to hear it, but for now I'm going to close this PR (and the associated issue) and wait for the breaking change to appear in numpy removing the polynomial functions.

Sorry to close your contribution, but know you're welcome to contribute to other issues!

PR #2989, opened 2026-04-12
Replace deprecated np.polyfit with numpy.polynomial.polynomial.polyfit [fix #2322]

### LINKED ISSUE #2322: Remove usage of `np.polyfit`

hijirii:
## Summary

Fixes #2322: Remove usage of `np.polyfit` (deprecated in NumPy 2.0)

### Changes

- Replaced `np.polyfit` with `np.polynomial.polynomial.polyfit` in the `mitiq_polyfit` function
- The new API does not return covariance matrix directly, so covariance is now computed manually using the formula: `cov = sigma^2 * (X^T X)^-1` where `sigma^2` is the variance of residuals
- Coefficient order is flipped from ascending (new API) to descending (old API) for backward compatibility
- Covariance matrix indices are also flipped to match the descending coefficient order
- Edge case of zero degrees of freedom correctly returns `None` for covariance

### Testing

- Manual tests confirm output matches `np.polyfit(..., cov=True)` for various cases:
  - Basic polynomial fits (degree 1, 2, etc.)
  - Weighted fits
  - Fits with exactly enough DOF (dof > 0)
  - Fits with zero DOF (covariance = None)
- The specific test case `test_get_methods_of_factories` which checks `get_parameters_covariance() = [[3.0, -1.0], [-1.0, 1.0]]` produces the correct result

### Notes

- `np.polyfit` with `cov=True` is deprecated as of NumPy 2.0
- The `numpy.polynomial.polynomial` module is the recommended replacement
- This is a pure refactor that preserves all existing behavior while using the non-deprecated API

### DIFF
--- a/mitiq/zne/inference.py
+++ b/mitiq/zne/inference.py
@@ -169,20 +169,22 @@ def mitiq_polyfit(
 
     with warnings.catch_warnings(record=True) as warn_list:
         warn_list.clear()
-        with warnings.catch_warnings(record=True) as inner_warn_list:
-            inner_warn_list.clear()
-            try:
-                # Use new numpy.polynomial API (replaces deprecated np.polyfit)
-                opt_params_asc, stats = polynomial.polyfit(
-                    scale_factors, exp_values, deg, w=w, full=True
-                )
-            except (ValueError, np.linalg.LinAlgError):
-                # Fall back without covariance if fit fails
+        try:
+            # Use new numpy.polynomial API (replaces deprecated np.polyfit)
+            opt_params_asc, stats = polynomial.polyfit(
+                scale_factors, exp_values, deg, w=w, full=True
+            )
+            # Check for ill-conditioned (singular) Vandermonde matrix
+            # If all scale_factors are the same, the matrix will be singular
+            unique_scales = len(np.unique(np.asarray(scale_factors)))
+            if unique_scales < deg + 1:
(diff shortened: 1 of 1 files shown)


danstam:
Hey @hijirii, quick heads-up on the weighted covariance math here:

Because `polyfit` minimizes weighted residuals, the design matrix is scaled as $WX$ (where $W$ is the diagonal weight matrix). The correct unscaled covariance formula is therefore $((WX)^T WX)^{-1}$, which expands to $(X^T W^2 X)^{-1}$.

The current implementation `X.T @ np.diag(w) @ X` computes $X^T W X$. This misses the required square on the weight matrix, which will return incorrect covariance values for weighted fits.

> **Note for maintainers:** PR #2984 already implements this migration with the correct $W^2$ math and passing CI.

PR #3009, opened 2026-05-13
fix broken qrack docs

natestemen:
## Description

Missed this previously when upgrading to pyqrack 2.0 since we turned off PR docs builds. cc @cosenal

---

### License


docs/source/examples/cdr_qrack.md
WrathfulSpatula, line 1: I suspect you might actually want this:

```
    qsim = QrackSimulator(
        qiskit_circ.width(),
        is_near_clifford_tableau_writer=True,
    )
```

WrathfulSpatula: (Apologies: that should have been `True`, now corrected.)

WrathfulSpatula (approved):

cosenal (approved):

PR #3010, opened 2026-05-14
Use equation references in classical shadows theory guide

matti-sarjala:
## Description

This PR replaces hard-coded equation number references in the Classical Shadows theory guide with MyST/Sphinx equation references.

This avoids references such as `(4)`, `(5)`, `(6)`, `(7)`, `(11)`, and `Eq. (14)` becoming stale when equations are added, removed, or reordered.

## Related issue

Part of #2110.

## Scope

This is intentionally a narrow documentation-only change. It does not rewrite the Classical Shadows theory text or change the mathematical content.


natestemen (approved):
Thanks for making these a bit cleaner! Apologies for the slow turnaround.

PR #3018, opened 2026-05-26
move away from VERSION.txt

natestemen:
## Description

After we moved mitiq to `uv` @b-goldsmith had some helpful notes on further cleanup we could do in https://github.com/unitaryfoundation/mitiq/discussions/2734. This PR implements the removal of a separate `VERSION.txt` file. This also cascaded to being able to remove the `mitiq/_version.py` file as it got compressed to 2 lines which having a separate file for seems unnecessary.


mitiq/__init__.py
```diff
@@ -22,8 +22,9 @@
 from mitiq.observable import PauliString, Observable
 
 # About and version.
+from importlib.metadata import version as _pkg_version
 from mitiq._about import about
-from mitiq._version import __version__
+__version__ = _pkg_version("mitiq")
```
bachase, line 27: I think this will fail for users that just clone the repo and don't actually install mitiq as a package. The old code in `mitiq/_version.py` had a fallback in that case to just read `VERSION.txt`.

I think you could do similar now via something like the below, where you fall back to `pyproject.toml`?

```
from importlib.metadata import PackageNotFoundError, version
from pathlib import Path
import tomllib

try:
    __version__ = version("mitiq")
except PackageNotFoundError:
    with (Path(__file__).resolve().parents[1] / "pyproject.toml").open("rb") as f:
        __version__ = tomllib.load(f)["project"]["version"]
```

bachase (approved):
Nothing blocking.

In addition to the comment I left, it looks like `MANIFEST.in` has `VERSION.txt` listed inside, so can scrub that too

mitiq/__init__.py
natestemen: That's a good point, I was trying to figure out when this `except` block was being triggered. That said, do we want to support the workflow of cloning mitiq, but not installing it? Very surprising workflow IMO.

bachase: Yeah, I'm not sure how common that is as well, so seems fine to not support it.

b-goldsmith: @natestemen if this is the preferred method of getting Mitiq's version, then `mitiq/_about.py` would need to be updated as well. In fact it looks like `_pkg_version()` could be used for the other packages too.

PR #3020, opened 2026-05-28
revert typing change on `QPROGRAM`

natestemen:
## Description

The docs build was failing with:
```
TypeError: unsupported operand type(s) for |: 'ABCMeta' and 'Program'
```
This is caused by the changes I introduced in https://github.com/unitaryfoundation/mitiq/pull/3019 as the `A | B` syntax causes `type(A).__or__(A, B)` to run at runtime (resulting in the error above). `Union` avoids this since there is no runtime call and does not depend on the metaclasses of `A` and `B.


bachase (approved):
👍

PR #3021, opened 2026-06-01
Add tests for observable and pauli error paths (#2365)

phucnguyen1707:
Addresses some of the unchecked items in #2365 for the `mitiq/observable` module.

Adds unit tests for paths that were previously untested:

- `PauliString`: invalid characters in `spec` and a `support`/`spec` length mismatch both raise `ValueError`, plus `__repr__`.
- `Observable`: the `paulis` accessor, and that multiplying by an unsupported operand type returns `NotImplemented` (and raises `TypeError` through the operator).

One related line in `PauliString.can_be_measured_with` — the `if cirq.I in (...): continue` branch — looks unreachable to me, since `cirq.PauliString` drops identity factors so every qubit in the overlap set maps to a non-identity Pauli. Happy to mark it `# pragma: no cover` if you'd prefer that over leaving it uncovered.

Tests pass locally and `ruff check`/`ruff format` are clean.

### DIFF
--- a/mitiq/observable/tests/test_observable.py
+++ b/mitiq/observable/tests/test_observable.py
@@ -352,3 +352,32 @@ def test_pauli_string_deduplication_removal_of_0_coefficients():
     obs = Observable(*pauli_strings)
     assert obs.nterms == 1
     assert obs == Observable(XI.with_coeff(XI.coeff * 2))
+
+
+def test_observable_paulis_property():
+    pauli1 = PauliString(spec="XI")
+    pauli2 = PauliString(spec="ZZ")
+    obs = Observable(pauli1, pauli2)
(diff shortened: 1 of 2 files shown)


natestemen (approved):
LGTM! Thanks for helping increase coverage!

natestemen:
> Happy to mark it # pragma: no cover if you'd prefer that over leaving it uncovered.

Ah missed this on first read, sorry! If you want to open another PR to add the no cover comment, go ahead, but also fine not to IMO.

phucnguyen1707:
Thanks @natestemen, no worries! I'll keep this PR as-is then. Happy to add the `# pragma: no cover` in a small follow-up if you decide you'd like it, otherwise fine to leave it. Thanks for taking a look 🙂

PR #3026, opened 2026-06-04
Add a local CDR worked example

### LINKED ISSUE #1507: Add tutorials with other techniques
## Issue Description

codexbountylab:
Closes #1507.

## Summary

- add a self-contained MyST notebook demonstrating CDR on a three-qubit variational circuit
- define local ideal and depolarizing-noise executors using Cirq
- walk through the essential `execute_with_cdr` arguments and near-Clifford training workflow
- compare ideal, noisy, and CDR-mitigated expectation values with both error metrics and a plot
- add the tutorial to the examples gallery

The example is deterministic and demonstrates a clear improvement:

```text
Ideal expectation value: 0.883
Noisy expectation value: 0.418
CDR-mitigated expectation value: 0.792
Noisy absolute error: 0.465
CDR absolute error:   0.091
Improvement factor:   5.1x
```

## Validation

- executed the complete MyST notebook with Jupytext (`16` cells, no errors)
- `git diff --check`

## AI disclosure

I used OpenAI Codex to help inspect the existing examples, calibrate a deterministic circuit/noise configuration, draft the tutorial, and run local validation. I manually reviewed the prose, code, numerical outputs, and final diff before submission.

### DIFF
--- a/docs/source/examples/cdr_demo.md
+++ b/docs/source/examples/cdr_demo.md
@@ -25,9 +25,10 @@ simulation.
 
 CDR requires the non-Clifford content of the circuit to be contained in
 single-qubit $R_Z$ rotations. This example makes that compilation step
-explicit by translating the $R_Y$ rotations in the original circuit into the
-$\{R_Z, \sqrt{X}, \mathrm{CNOT}\}$ basis before calling
-{func}`.cdr.execute_with_cdr`.
+explicit by translating the unsupported operations in the original circuit
+into Clifford operations and $R_Z$ rotations before calling
+{func}`.cdr.execute_with_cdr`. If a hardware target requires a smaller native
+gate set, that restriction can be applied as a later compilation step.
 
 ## Setup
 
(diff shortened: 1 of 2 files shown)


b-goldsmith (changes requested):
Hi and thanks for the PR. At this point the example is more or less a duplicate of the User Guide for CDR. Could you please do one of the following to differentiate it more and make it a useful example.

1.) Use a different Near-Clifford Simulator (like Unitary Foundation's Clifft) 

2.) Use a different circuit consisting of more gates and then show how to transpile to the appropriate gate set:
```
    The circuit of interest must be compiled in the native basis of the IBM
    quantum computers, that is {Rz, sqrt(X), CNOT}, or such that all the
    non-Clifford gates are contained in the Rz rotations.
```

codexbountylab:
> Hi and thanks for the PR. At this point the example is more or less a duplicate of the User Guide for CDR. Could you please do one of the following to differentiate it more and make it a useful example.
> 
> 1.) Use a different Near-Clifford Simulator (like Unitary Foundation's Clifft)
> 
> 2.) Use a different circuit consisting of more gates and then show how to transpile to the appropriate gate set:
> 
> ```
>     The circuit of interest must be compiled in the native basis of the IBM
>     quantum computers, that is {Rz, sqrt(X), CNOT}, or such that all the
>     non-Clifford gates are contained in the Rz rotations.
> ```

Thanks for the feedback! I reworked the example following option 2: it now starts from a larger four-qubit Cirq circuit, explicitly compiles the RY rotations into the CDR-compatible {Rz, sqrt(X), CNOT} gate set, checks the compiled basis, and then runs CDR on the compiled circuit.

b-goldsmith:
For the transpilation to the appropriate gate set, please use a universal function instead of manually changing the gates. 

For example, Cirq includes [optimize_for_target_gateset](https://quantumai.google/reference/python/cirq/optimize_for_target_gateset).

This makes the CDR example more accessible and applicable to different scenarios.

codexbountylab:
Thanks for the suggestion. I updated the example to use Cirq's general `cirq.decompose` utility instead of manually rewriting each gate. The compilation now uses a CDR-basis predicate that keeps Rz, sqrt(X), and CNOT operations, and the example still checks that the resulting circuit is in the expected basis.

b-goldsmith:
Again, we would like the example to be as universal as possible. This means that your `is_cdr_basis_operation` should be able to take a circuit with any gates and return only Clifford and Rz gates. (I would recommend changing the name of the function to point out the gates are changing to just Clifford and Rz gates.) Then, if the hardware or simulator needs additional restrictions to the gate set, that should be applied next. 

Since `compute_density_matrix` uses Cirq's DensityMatrixSimulator, you do not have to worry about the hardware or simulator restrictions, though it would be helpful to add a note about that.

codexbountylab:
Thanks for the clarification. I updated the example so the compilation predicate now keeps arbitrary `Rz` rotations plus any operation Cirq recognizes as Clifford via `cirq.has_stabilizer_effect`.

I also added a note that hardware-native gate restrictions would only be needed as a later compilation step. Since this example runs locally with Cirq's density-matrix simulator, the Clifford + Rz compilation is enough here.

natestemen:
Thanks for your contribution here, but unfortunately we've decided to go with another hackers contribution (https://github.com/unitaryfoundation/mitiq/pull/3052). I hope you learned something, and know you're always welcome to contribute to mitiq in the future!

PR #3027, opened 2026-06-04
Support direct PEA representations

### LINKED ISSUE #2936: Support representations directly in PEA
## Background

## What needs to change

```python
# BEFORE
pea.construct_circuits(
    circuit,
    epsilon=0.01,
    precision=0.2,
    random_state=1,
)

# AFTER
pea.construct_circuits(
    circuit,
    representations=reps,
    epsilon=0.01,
    precision=0.2,
    random_state=1,
)
```

## Implementation notes

## Acceptance criteria

## Good first steps

Samfresh-ai:
## Description

Fixes #2936.

This updates experimental PEA so `construct_circuits` can take direct `OperationRepresentation` objects instead of forcing the legacy `noise_model` string path.

Changes:

- adds canonical scaling for direct quasi-probability representations
- rejects ambiguous calls where both `representations` and `noise_model` are passed
- keeps `noise_model` backward compatible and emits a deprecation warning
- threads the new representation path through `execute_with_pea`
- adds unit tests for direct representations, validation errors, legacy compatibility, and the base-scale legacy match

Validation run locally:

```bash
uv run --group dev pytest mitiq/experimental/pea/tests/test_pea.py -k 'not braket'
uv run --group dev pytest mitiq/experimental/pea/tests/test_scale_amplifications.py
uv run --extra braket --extra qiskit --group dev pytest mitiq/experimental/pea/tests/test_pea.py
uv run --group dev ruff check mitiq/experimental/pea/pea.py mitiq/experimental/pea/tests/test_pea.py
uv run --group dev ruff format --check mitiq/experimental/pea/pea.py mitiq/experimental/pea/tests/test_pea.py
python3 -m compileall -q mitiq/experimental/pea/pea.py mitiq/experimental/pea/tests/test_pea.py
git diff --check
```

---

### License

### DIFF
--- a/mitiq/experimental/pea/tests/test_pea.py
+++ b/mitiq/experimental/pea/tests/test_pea.py
@@ -354,14 +354,18 @@ def test_pea_data_with_full_output():
         full_output=True,
     )
     # Get num samples from precision
-    _, _, norm = sample_circuit(
-        twoq_circ,
-        amplify_noisy_ops_in_circuit_with_local_depolarizing_noise(
-            twoq_circ, epsilon
-        ),
-        num_samples=1,
-    )
-    num_samples = int((norm / precision) ** 2)
+    scale_factors = [1, 1.2, 1.6]
+    norms = []
+    for scale_factor in scale_factors:
+        _, _, norm = sample_circuit(
+            twoq_circ,
+            scale_circuit_amplifications(
+                twoq_circ, scale_factor, "local_depolarizing", epsilon
+            ),
+            num_samples=1,
+        )
+        norms.append(norm)
+    num_samples = int((max(norms) / precision) ** 2)
 
     # Manually get raw expectation values
     scaled_exp_values = [
--- a/mitiq/experimental/pea/tests/test_pea.py
+++ b/mitiq/experimental/pea/tests/test_pea.py
@@ -159,6 +159,58 @@ def test_direct_representations_match_legacy_at_base_scale():
     assert direct == legacy
 
 
+def test_direct_representations_reject_all_positive_non_base_scale():
+    representations = scale_circuit_amplifications(
+        oneq_circ,
+        scale_factor=1,
+        noise_model="local_depolarizing",
+        epsilon=BASE_NOISE,
+    )
+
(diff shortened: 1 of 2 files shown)


natestemen:
Any idea why the tests are failing for python 3.12 only?

Samfresh-ai:
Yes. The Python 3.12 failure was from a stale test expectation after the sample-count logic changed to use the maximum norm across the requested scale factors. Python 3.12 exposed it as `assert 4 == 3` in `test_pea_data_with_full_output`.

Pushed `63967c8`, which updates that test to compute the expected sample count from the max scaled norm. Local checks passed: `uv run pytest mitiq/experimental/pea/tests/test_pea.py -q` (`25 passed`), ruff check, ruff format check, compileall, and `git diff --check`.

mitiq/experimental/pea/tests/test_pea.py
```diff
@@ -202,14 +354,18 @@ def test_pea_data_with_full_output():
...
-        twoq_circ,
-        amplify_noisy_ops_in_circuit_with_local_depolarizing_noise(
-            twoq_circ, epsilon
-        ),
-        num_samples=1,
-    )
-    num_samples = int((norm / precision) ** 2)
+    scale_factors = [1, 1.2, 1.6]
+    norms = []
+    for scale_factor in scale_factors:
```
natestemen, line 359: Why do we iterate over the scale factors now where as before it was not needed?

Samfresh-ai: We iterate over the requested scale factors because `construct_circuits()` now derives `num_samples` from the largest one-norm across all scaled representations. That matters for the direct `representations` path: different scale factors can produce different norms, so using only the scale-1 norm can under-sample the noisier/larger-norm scale and miss the requested precision target. The loop keeps the test aligned with the production sample-count rule instead of assuming the scale-1 norm is always the limiting case.

natestemen:
Thanks for your work here! I'm going with https://github.com/unitaryfoundation/mitiq/pull/3041 instead, but your work here was good. You're always welcome to contribute to mitiq in the future!

PR #3029, opened 2026-06-04
Add [[4,2,2]] + ZNE tutorial

### LINKED ISSUE #3006: Add tutorial: error detection with the [[4,2,2]] code and ZNE
## Background

Error detection and error mitigation are complementary strategies for dealing with noise on near-term quantum hardware. This tutorial asks for a notebook demonstrating how they can be combined using mitiq: specifically,
using the `[[4,2,2]]` quantum error detecting code to discard shots where an error is known to have occurred, and then applying ZNE to the surviving shots for further mitigation.

The `[[4,2,2]]` code is the smallest quantum error *detecting* code. It encodes 2 logical qubits into 4 physical qubits and can detect (but not correct) any single-qubit error. The key tool on the mitiq side is [`mitiq.rem.post_select`](https://mitiq.readthedocs.io/en/stable/apidoc.html#mitiq.rem.post_select.post_select), which already exists and handles the shot-discarding step cleanly.

## Background: the [[4,2,2]] code

The `[[4,2,2]]` code has two stabilizer generators: `XXXX` and `ZZZZ`.

Any single-qubit error anticommutes with at least one of these stabilizers, flipping its measurement outcome from +1 to -1 and revealing that an error occurred. The four logical codewords are:

$$
\begin{align}
|0_L 0_L\rangle = \frac{1}{\sqrt{2}}(|0000\rangle + |1111\rangle) \\
|0_L 1_L\rangle = \frac{1}{\sqrt{2}}(|0011\rangle + |1100\rangle) \\
|1_L 0_L\rangle = \frac{1}{\sqrt{2}}(|0101\rangle + |1010\rangle) \\
|1_L 1_L\rangle = \frac{1}{\sqrt{2}}(|0110\rangle + |1001\rangle)
\end{align}
$$

A useful reference for the encoding circuit and stabilizer structure is the `[[4,2,2]]` section of the [Quantum Error Correction Zoo](https://errorcorrectionzoo.org/c/stab_4_2_2). The code also appears in [arXiv:2510.01181](https://arxiv.org/abs/2510.01181), which is a good reference for how error detection and mitigation compose.

## Deliverable

A MyST notebook that works through the following narrative:

2. **Implement the [[4, 2, 2]] encoding circuit**: prepare a logical state (e.g. `|0_L 0_L⟩`) and apply a logical operation of your choice
3. **Add syndrome measurement** — measure the XXXX and ZZZZ stabilizers using ancilla qubits and append to the circuit
4. **Post-select with mitiq** — use `mitiq.rem.post_select` to discard shots where either stabilizer measurement is -1 (error detected):
```python
   from mitiq.rem import post_select
   # keep only shots where both syndrome qubits are 0
   post_select(result, lambda bits: bits[-2] == 0 and bits[-1] == 0)
```
5. **Combine with ZNE** — apply `mitiq.zne` to the post-selected results and show the additional improvement
6. **Compare results** — plot all four cases: ideal (noiseless), noisy unmitigated, error-detected only, error-detected + ZNE

## Noise model

Use a local depolarizing noise model on a simulator — no real hardware required. A depolarizing probability of ~1-3% per gate is a realistic and illustrative range.

## Note

The `[[4,2,2]]` code detects errors but cannot correct them — shots where an error is detected are simply discarded. This reduces the number of usable shots, which is a real overhead cost worth discussing in the notebook prose.

axif0:
## Description

Adds a new tutorial to the Mitiq examples gallery demonstrating how to combine quantum error detection with error mitigation, motivated by the hybrid protocol in Zhong et al. ([arXiv:2510.01181](https://arxiv.org/abs/2510.01181)).

Closes #3006 
---

### License


docs/source/examples/combine_422_zne.md
```diff
@@ -0,0 +1,258 @@
...
+print(circuit)
+```
+
+## Noise model and executor
+
+The noise in this example is a local depolarizing channel applied after every gate.
+A depolarizing probability of 1% per gate is a realistic and illustrative value for near-term superconducting devices.
+We use an [executor function](../guide/executors.md) to run the quantum circuit with the noise model applied.
+
+The executor returns a `MeasurementResult` whose bitstrings contain **6 bits each**: bits 0–3 are the data qubits and bits 4–5 are the syndrome ancillae ($XXXX$, $ZZZZ$ respectively).
```
natestemen, line 111: Lets add a note about what happens in the case where one of the ancilla are `1`.

natestemen (changes requested):
This is looking really good! Very clear explanations. Two main comments/questions:

1. I cannot build this locally. Looks like the branch needs to be rebased.
2. Not sure it's needed, but WDYT about using a more complicated circuit/observable on the encoded qubits? Since this is more of a demonstration of how to do QED and QEM together it seems fine keep it simple, but curious your thoughts.

docs/source/guide/shadows-5-theory.md
natestemen, line 1: This file should not be part of the diff.

docs/source/examples/combine_422_zne.md
```diff
@@ -0,0 +1,262 @@
...
+The [[4,2,2]] code detects errors but cannot correct them — shots where an error is detected are simply discarded. This reduces the number of usable shots, which is a real overhead cost. To achieve a target number of *clean* shots, you must collect significantly more *raw* shots than you would without error detection.
+```
+
+```{code-cell} ipython3
+# Keep only shots where both syndrome qubits are 0 (no error detected)
+ps_result = post_select(noisy_mr, lambda bits: bits[4] == 0 and bits[5] == 0)
+
+retained_pct = 100 * ps_result.shots / noisy_mr.shots
+print(f"Shots retained after post-selection: {ps_result.shots} / {noisy_mr.shots} ({retained_pct:.1f}%)")
+print("Mitigated value obtained with error detection:", "{:.5f}".format(expect_zzzz(ps_result)))
```
natestemen, line 180: Can you also add the value obtained if we did not post-select? This should be very close to what we see with `noisy`, but it would be a good sanity check.

docs/source/examples/combine_422_zne.md
```diff
@@ -0,0 +1,262 @@
...
+        bar.get_x() + bar.get_width() / 2,
+        bar.get_height() + 0.01,
+        f"{val:+.3f}",
+        ha="center", va="bottom", fontsize=9, fontweight="bold",
+    )
+plt.tight_layout()
+plt.show()
+```
+
+From this example we can see that each technique affords some improvement, and the combination of [[4,2,2]] error detection and ZNE is more effective in suppressing errors than either technique applied alone.
```
natestemen, line 260: Lets mention here the tradeoff that comes with using error detection as far as the need of additional shots. A "percentage increase" order to obtain the desired shot-count would be a nice wrap up.

natestemen (approved):
Nice changes! Two smaller comments and this should be good to land.

axif0:
@natestemen  Please let me know if there is anything else that needs to be addressed.

docs/source/examples/combine_422_zne.md
```diff
@@ -0,0 +1,280 @@
...
+
+```{code-cell} ipython3
+overhead_factor = 1.0 / (retained_pct / 100)
+overhead_pct = (overhead_factor - 1) * 100
+print(f"Retention rate: {retained_pct:.1f}%")
+print(f"Overhead factor: {overhead_factor:.2f}x  ({overhead_pct:.1f}% more raw shots needed)")
+```
+
+At a gate error rate of 1%, roughly {math}`\sim 10\text{–}20\%` of shots are flagged and discarded, requiring about 10–25% more raw shots to match the clean-shot count of an unprotected run.
+At higher error rates the discarded fraction grows quickly — at 3% per gate, well over half the shots may be rejected — so the shot overhead can become the dominant practical cost of the combined protocol.
```
natestemen: The results don't really match the comments here. Please make sure to review the content before putting it up for review!

natestemen (commented):
I like the summary you added, but it needs to align with the results. Once that is sorted I will merge!

natestemen (approved):
Thanks a bunch for this tutorial! I think it will get lots of eyes on it.

PR #3038, opened 2026-06-07
Add OperationRepresentation support to PEA construct_circuits and execute_with_pea

### LINKED ISSUE #2936: Support representations directly in PEA

Bernardo-R-O-Carvalho:
Closes #2936

## What this PR does

Extends `construct_circuits` and `execute_with_pea` in `mitiq/experimental/pea/pea.py`
to accept `representations` (a `Sequence[OperationRepresentation]`) directly, as an
alternative to the existing `noise_model` + `epsilon` interface.

## Motivation

Previously, users who had already built `OperationRepresentation` objects (e.g. via
`represent_operations_in_circuit_with_local_depolarizing_noise`) had no way to pass
them directly to PEA — they were forced to re-specify the noise model as a string.
This PR removes that friction.

## Changes

- `construct_circuits`: `noise_model` and `epsilon` are now optional keyword arguments.
  A new `representations` parameter is added. Exactly one of `representations` or
  `(noise_model, epsilon)` must be provided; otherwise a `ValueError` is raised.
- `execute_with_pea`: same changes, `extrapolation_method` moved before the new
  optional parameters to keep the most-used arguments first.
- New tests covering: direct use of representations, equivalence with noise_model path,
  and all error cases (both provided, neither provided, noise_model without epsilon).

## Notes

- The two existing `braket` test failures are pre-existing and unrelated to this PR
  (`ModuleNotFoundError: No module named 'braket'`).
- All 20 non-braket tests pass.

## AI disclosure

This PR was developed with AI assistance (Claude).


natestemen (changes requested):
Please remove unrelated files before putting up PRs.

PR #3041, opened 2026-06-07
Support representations directly in PEA (closes #2936)

### LINKED ISSUE #2936: Support representations directly in PEA

zkasuran:
## Description

PEA could only amplify noise through the `"local_depolarizing"` and `"global_depolarizing"` model strings, which makes it unusable on real hardware where the noise is neither. This adds a `representations` parameter (a list of `OperationRepresentation`) to `pea.construct_circuits` and `pea.execute_with_pea`, as requested in #2936. The representations can be learned from hardware characterization, making PEA hardware-agnostic.

The representations are amplified with the canonical noise scaling of Sec. VI D (Eqs. 38-44) of Mari, Shammah and Zeng, [Phys. Rev. A 104, 052607 (2021)](https://doi.org/10.1103/PhysRevA.104.052607) (arXiv:2108.02237), already in the Mitiq bibliography as `Mari_2021_PRA`. The exact rule applied to each representation depends on its sign structure:

- **All-positive representations** (a genuine probability distribution, e.g. an `amplify_*` decomposition or a noise model learned from hardware) are scaled by deviation from the identity: `a0 -> 1 - s*(1 - a0)`, every other `a_k -> s*a_k`. This reproduces the legacy global-depolarizing rebuild exactly at every scale factor, so passing global-depolarizing representations yields the same sampled circuits, signs and norms as the `noise_model` path. That equivalence is the acceptance-criterion regression test, checked at scale factors `[1, 3, 5]` on a two-qubit circuit (not only at `scale_factor=1`).
- **Signed representations** (containing negative coefficients, e.g. a PEC-style `represent_operation_with_*` decomposition) are scaled by the canonical sign-partition rule of Eq. (43). The positive volume `gamma+` is scaled by `(gamma+ - s*gamma-)/gamma+` and the negative volume `gamma-` by `(1 - s)`, valid up to the canonical limit `s <= gamma+/gamma-`. In the amplification band the negative volume vanishes and the representation becomes a genuine probability distribution, so amplifying it adds real noise. Scaling a signed representation by deviation from the identity instead amplifies the error cancellation, the opposite of what PEA needs. The deviation rule above is the `gamma- = 0` special case of this same construction.

This second branch fixes the case nez0b raised on #3058: a representation with mixed-sign coefficients was being amplified the wrong way. The fix commit credits nez0b as co-author.

`representations` and `noise_model` are mutually exclusive: passing both, or neither, raises a clear `ValueError`. `epsilon` only applies to the legacy `noise_model` path, so passing it together with `representations` is also rejected. `noise_model` keeps working unchanged for backward compatibility and now emits a `DeprecationWarning` pointing at the new interface. An empty `representations` list, a representation whose identity term has zero weight, and a signed representation scaled beyond its canonical limit are all rejected with clear messages.

Limitation, documented in the `_scale_positive_representation` docstring: the two-qubit `local_depolarizing` model has a product (quadratic) dependence on the noise level that flat deviation scaling cannot reproduce exactly. The coefficients still stay normalized; single-qubit local and all global depolarizing are exact, so the equivalence test is scoped to global depolarizing.

New tests cover the scale-factor != 1 equivalence, the signed-rep canonical scaling (closed-form match, negative volume dropping to zero for `s >= 1`, the canonical-limit error), the both/neither/missing-epsilon/empty-list/epsilon-with-representations errors, the deprecation warning and the missing/zero identity-term guard. The PEA user guide documents `representations` as the recommended interface, marks `noise_model` legacy, and now describes the sign-dispatch.

Verified locally: `pytest mitiq/experimental/pea` (56 passed), `ruff check`, `ruff format --check`, and `mypy mitiq` (whole package, 109 files) all clean.

### References

- **Canonical noise scaling (the math used here):** A. Mari, N. Shammah, W. J. Zeng, "Extending quantum probabilistic error cancellation by noise scaling", Phys. Rev. A 104, 052607 (2021). [doi:10.1103/PhysRevA.104.052607](https://doi.org/10.1103/PhysRevA.104.052607) (arXiv:2108.02237). Sec. VI D, Eqs. (38)-(44) split a quasi-probability representation into positive and negative volumes and rescale them; Eq. (43) is the signed rule used here, valid up to `s <= gamma+/gamma-`, with the deviation-from-identity rule as the all-positive special case. Already in the Mitiq bibliography as `Mari_2021_PRA`.
- **PEA technique:** Y. Kim et al., "Evidence for the utility of quantum computing before fault tolerance", Nature 618, 500-505 (2023). [doi:10.1038/s41586-023-06096-3](https://doi.org/10.1038/s41586-023-06096-3). Cited in Mitiq as `Kim_2023_Nature`; motivates learning the noise representation from hardware, which is what this PR lets PEA consume.
- **Quasi-probability foundation (PEC):** K. Temme, S. Bravyi, J. M. Gambetta, "Error Mitigation for Short-Depth Quantum Circuits", Phys. Rev. Lett. 119, 180509 (2017). [doi:10.1103/PhysRevLett.119.180509](https://doi.org/10.1103/PhysRevLett.119.180509). Cited in Mitiq as `Temme_2017_PRL`; the `OperationRepresentation` structure this PR scales.

### AI use

This change was implemented with help from Claude (Anthropic), which drafted the code, the tests and this description. I directed the design and reviewed every line. Before implementing the sign-dispatch I read Sec. VI D of the Mari, Shammah and Zeng paper and verified the signed-scaling rule against Eq. (43) (the `s = lambda` identification, the `s <= gamma+/gamma-` limit, and the sign-free amplification band), and I checked the all-positive branch still reproduces the depolarizing rebuild at several scale factors. The signed-rep handling resolves a case nez0b flagged on #3058 and credits him as commit co-author. I ran the test, lint and type-check suites locally. I understand the change and can answer questions about it during review.

---

### License

### DIFF
--- a/mitiq/experimental/pea/pea.py
+++ b/mitiq/experimental/pea/pea.py
@@ -15,7 +15,9 @@ from cirq import Circuit
 from mitiq import QPROGRAM, Executor, Observable, QuantumResult
 from mitiq.experimental.pea.scale_amplifications import (
     scale_circuit_amplifications,
+    scale_representations,
 )
+from mitiq.pec import OperationRepresentation
 from mitiq.pec.pec import (
     _LARGE_SAMPLE_WARN,
     LargeSampleWarning,
@@ -23,11 +25,80 @@ from mitiq.pec.pec import (
 )
 
 
+def _resolve_amplifications(
+    circuit: Circuit,
+    representations: Sequence[OperationRepresentation] | None,
+    noise_model: str | None,
+    epsilon: float | None,
+) -> Callable[[float], Sequence[OperationRepresentation]]:
+    """Validates the noise-amplification arguments and returns a function that
+    builds the amplified representations at a given scale factor.
+
+    Exactly one of ``representations`` or the ``(noise_model, epsilon)`` pair
+    must be supplied. When ``representations`` is given, the returned function
+    scales them with canonical noise scaling. When ``noise_model`` is given
+    (the legacy path), the returned function rebuilds the amplifications at
+    ``scale_factor * epsilon`` and a ``DeprecationWarning`` steers the caller
+    toward ``representations``.
+
+    Args:
+        circuit: The ideal circuit whose operations are represented.
+        representations: User-supplied quasi-probability representations, or
+            None to use the noise-model path.
+        noise_model: The legacy noise-model string, or None.
+        epsilon: The baseline noise level for the legacy path, or None.
+
+    Returns:
+        A function mapping a scale factor to the amplified representations.
+
+    Raises:
+        ValueError: If both ``representations`` and ``noise_model`` are given,
+            or if neither is given.
+    """
+    if representations is not None and noise_model is not None:
+        raise ValueError(
+            "Provide either 'representations' or 'noise_model', not both."
+        )
+    if representations is None and noise_model is None:
+        raise ValueError(
+            "Provide either 'representations' (a list of "
+            "OperationRepresentation) or 'noise_model' together with "
+            "'epsilon'."
+        )
+
+    if representations is not None:
+        if len(representations) == 0:
+            raise ValueError(
+                "'representations' is empty. Provide one "
+                "OperationRepresentation per unique operation of the circuit."
+            )
+        reps = representations
+        return lambda scale_factor: scale_representations(reps, scale_factor)
+
+    if epsilon is None:
+        raise ValueError(
+            "'epsilon' must be given together with 'noise_model'."
+        )
+    assert noise_model is not None  # guaranteed by the checks above
+    model, base_noise = noise_model, epsilon
+    warnings.warn(
+        "The 'noise_model'/'epsilon' interface is legacy. Pass "
+        "'representations' (a list of OperationRepresentation) instead, which "
+        "also works with noise learned from hardware. See the PEA user guide.",
+        DeprecationWarning,
+        stacklevel=3,
+    )
+    return lambda scale_factor: scale_circuit_amplifications(
+        circuit, scale_factor, model, base_noise
+    )
+
+
 def construct_circuits(
     circuit: Circuit,
     scale_factors: list[float],
-    noise_model: str,
-    epsilon: float,
+    representations: Sequence[OperationRepresentation] | None = None,
+    noise_model: str | None = None,
+    epsilon: float | None = None,
     random_state: int | np.random.RandomState | None = None,
     precision: float = 0.1,
     num_samples: int | None = None,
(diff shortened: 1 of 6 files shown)


zkasuran:
Hi @natestemen, this one is ready for review. A short map to make it quicker.

`construct_circuits` and `execute_with_pea` accept a `representations` parameter, mutually exclusive with `noise_model`. The amplification lives in a small dedicated module, `scale_amplifications.py`, and follows the canonical noise scaling of Section D of Mari, Shammah and Zeng (2021), already in the bibliography as `Mari_2021_PRA`: the identity coefficient goes to `1 - s*(1 - a0)` and every other coefficient is multiplied by `s`. That scaling is defined for any quasi-probability representation, so PEA works with representations learned from hardware characterization, not only the two depolarizing strings.

The regression test pins the acceptance criterion from #2936: global-depolarizing representations reproduce the legacy `noise_model` path exactly, same sampled circuits, signs and norms, at scale factors 1, 3 and 5. The PEA test file goes from 23 to 42 collected cases and CI is green on the full matrix (ubuntu, macos and windows on 3.11 and 3.12). The user guide is updated in three places (intro, options and low-level) so the new path is documented.

Happy to adjust the API surface if you would rather have a different entry point.

mitiq/experimental/pea/pea.py
```diff
@@ -15,19 +15,90 @@
...
+        epsilon: The baseline noise level for the legacy path, or None.
+
+    Returns:
+        A function mapping a scale factor to the amplified representations.
+
+    Raises:
+        ValueError: If both ``representations`` and ``noise_model`` are given,
+            or if neither is given.
+    """
+    if representations is not None and noise_model is not None:
```
natestemen, line 60: We should throw in an epsilon check here too.

mitiq/experimental/pea/scale_amplifications.py
```diff
@@ -52,3 +52,100 @@ def scale_circuit_amplifications(
         # TODO allow use of custom noise model
 
     return amp_fn(ideal_circuit, scale_factor * epsilon)
+
+
+def scale_representations(
+    representations: Sequence[OperationRepresentation],
+    scale_factor: float,
+) -> list[OperationRepresentation]:
```
natestemen, line 66: This looks good, but I think there is a flawed which was called out in https://github.com/unitaryfoundation/mitiq/pull/3058. In particular when the signs of the representation are not all positive then this increases error cancellation instead of increasing the noise (which is what we want in PEA). I think there needs to be some branching based on signs in order to properly compute scaling factors.

natestemen (commented):
Looking good. Just one blocking issue for now.

natestemen:
@nez0b I'd encourage you to review this PR for anything you feel like is a must have from your PR that way we don't miss anything. I think the signed reps was the most important, but feel free to add comments to point out other items.

mitiq/experimental/pea/scale_amplifications.py
natestemen: BTW, please add @nez0b as commit author to the commit fixing this issue. You can likely pull code from his PR for this fix.

zkasuran: Fixed in 4970f86. `scale_representations` now branches on sign structure. Signed (PEC-style) representations go through the canonical sign-partition scaling of Sec. VI D, Eq. (43): the positive volume is scaled by `(gamma+ - s*gamma-)/gamma+` and the negative volume by `(1 - s)`, valid up to `s <= gamma+/gamma-`. In the amplification band the negative volume goes to zero, so the representation becomes a real probability distribution and amplifying it adds noise instead of increasing the cancellation. All-positive `amplify_*` reps keep the deviation rule unchanged (it is the `gamma- = 0` special case). New tests check the closed-form match, the negative volume dropping to zero for `s >= 1` and the canonical-limit error.

mitiq/experimental/pea/pea.py
zkasuran: Added in 4970f86. `_resolve_amplifications` now rejects `epsilon` when it is passed together with `representations`, mirroring the `noise_model` mutual-exclusion check (`epsilon` only applies to the legacy `noise_model` path). I read your note as wanting that guard. If you meant a value check on `epsilon` for the `noise_model` path instead, I can switch it.

mitiq/experimental/pea/scale_amplifications.py
zkasuran: Done. Commit 4970f86 credits nez0b as co-author and the signed-rep handling follows the approach from #3058.

zkasuran: Thanks @nez0b. The signed-representation fix in 4970f86 follows your sign-partition approach from #3058, and you are credited as co-author on that commit.

zkasuran:
Hi @natestemen, gentle bump on this. The review fixes from June 21 are in and green (signed-rep scaling, epsilon guard, citation, @nez0b credited). Anything still blocking re-review, or good to take another look?

natestemen (approved):
Sorry for the delay, and thank you both for your hard work!

zkasuran:
Thanks for merging @natestemen, and for the careful review. Good working through the signed-rep scaling with you. @nez0b nice collaborating on this one.

andreamari:
I noticed this great PR only now. 
Happy to see canonical noise scaling implemented in Mitiq!
Thanks @zkasuran, @nez0b and @natestemen

PR #3044, opened 2026-06-08
Add debiasing technique to mitiq.experimental

### LINKED ISSUE #1955: New technique: debiasing/symmetrization/sharpening
## Issue Description

IonQ has recently developed a quantum error mitigation strategy. There is currently no open-source implementation available. Their paper calls it symmetrization and their [online documentation](https://ionq.com/resources/debiasing-and-sharpening) calls it debiasing and sharpening:

> Debiasing is a compiler-level error mitigation strategy that works by creating and running many symmetric variations of a given circuit, and then cleverly aggregating the measured outcomes to mitigate noise and improve the performance of quantum computers without incurring any additional qubit or gate overhead. We call this process sharpening.
>
> Sharpening over the symmetrization variants while debiasing reduces the overall impact of stochastic noise on the computed result — the deterministic inaccuracies largely cancel out while random noise does not get amplified. This approach effectively mitigates the impact of hardware control errors and qubit decoherence, which are major sources of imperfection in modern quantum technologies. By minimizing the impact of these imperfections, debiasing enables quantum computers to produce results at a much higher level of accuracy.

## Proposed Solution

Implement debiasing/symmetrization and sharpening in Mitiq.

## Additional References

Arxiv paper: https://arxiv.org/abs/2301.07233

mnn31:
## Description

Closes #1955.

Adds debiasing (also called symmetrization) as an experimental technique, laid out the way the issue suggested.

The idea is to run a few variants of the same circuit, each one wrapped in a layer of random single-qubit Paulis: the same Pauli goes on a qubit before the circuit and again after it. On a noiseless backend those two layers cancel, so every variant gives back the same distribution. Coherent errors don't cancel that way, so they get scrambled differently across the variants and partly average out once the results are combined. There's also an optional sharpening step that does a per-shot plurality vote instead of averaging, which helps when the answer sits on just a few bitstrings.

What's in the PR:
- `symmetrization.py` with `construct_circuits` to build the variants
- `sharpening.py` with `sharpen` for the plurality vote
- `deb.py` with `execute_with_debiasing` and `execute_with_debiasing_and_sharpening`
- tests for the variant count, the noiseless cancellation, the plurality vote, and an end-to-end run under depolarizing noise
- docs: apidoc, glossary, the README quick tour, a short guide page, and the paper added to `refs.bib`

One thing I'd like your read on: the issue describes `sharpen` taking a list of count dictionaries, but the plurality vote is defined per shot, and counts drop the information about which shot a bitstring came from. So I had it take a list of `MeasurementResult`s lined up shot by shot instead. That seemed necessary to make the vote well defined, but I'm happy to switch back to count dicts if you'd rather keep the signature from the issue.

I used AI assistance while writing this and have gone through and tested all of it, so I can talk through any part during review.

### License


kawacukennedy (approved):
Reviewed the full diff. Implementation is correct:

- `construct_circuits`: Pauli layers correctly cancel in noiseless simulation; skips identity gates; seeded RNG for reproducibility.
- `sharpen`: Shot-wise plurality vote with documented tie-breaking. Using `MeasurementResult` (vs count dicts) is the right call as you explained.
- `execute_with_debiasing`: Clean averaging of variant distributions.
- Tests: Cover noiseless cancellation, reproducibility, sharpening edge cases, and end-to-end under depolarizing noise. Reasonable tolerances.
- Docs: Guide, apidoc, glossary, and bib entry are thorough and match existing patterns.

One tiny observation: `sharpen` silently truncates to `min(result.shots)` when variant shot counts differ — worth documenting explicitly in the docstring that truncation happens, since a user might pass uneven results and get unexpected silent truncation. Not blocking.

mnn31:
Thanks @natestemen, that makes sense. I've updated the PR to use qubit permutations instead of the Pauli layers: each variant relabels the circuit onto a randomly permuted set of qubits (C → πC), and the permutation is undone on the measured bitstrings before the distributions are combined, with a different permutation for each variant. The noiseless cancellation is covered by a test that checks the unscrambled variants reproduce the original distribution. Ready for a review whenever you have a chance.

mitiq/experimental/deb/tests/test_symmetrization.py
natestemen, line 1: Lets add a test here to ensure the unitary equivalence of generated variants.

mitiq/experimental/deb/deb.py
```diff
@@ -0,0 +1,123 @@
...
+    so the unscrambled bit for logical qubit ``i`` is ``shot[permutation[i]]``.
+    """
+    unscrambled = [
+        [int(shot[permutation[i]]) for i in range(len(permutation))]
+        for shot in result.result
+    ]
+    return MeasurementResult(unscrambled)
+
+
+def _probability_vector(
```
natestemen: The `MeasurementResult` class has a `prob_distribution` method. Please use that instead of defining custom methods.

mitiq/experimental/deb/tests/test_sharpening.py
natestemen, line 1: A test is needed for the case where there is no winner and the shot is thrown away.

mitiq/experimental/deb/symmetrization.py
```diff
@@ -0,0 +1,98 @@
...
+    if num_variants < 1:
+        raise ValueError("num_variants must be a positive integer.")
+
+    qubits = sorted(circuit.all_qubits())
+    rng = _random_generator(random_state)
+
+    variants = []
+    for _ in range(num_variants):
+        permutation = list(range(len(qubits)))
+        rng.shuffle(permutation)
```
natestemen: Can we use numpy's built in permutation generator here? https://numpy.org/doc/stable/reference/random/generated/numpy.random.permutation.html

mitiq/experimental/deb/__init__.py
natestemen, line 1: `combine_results` is missing. It should allow the user to use sharpening by plurality voting or averaging.

natestemen (commented):
Looking good! Thanks for getting the guide started too!

mitiq/experimental/deb/symmetrization.py
mnn31: Switched to `numpy.random.default_rng(...).permutation(...)`. `random_state` now accepts a seed or a `numpy.random.Generator`.

mitiq/experimental/deb/deb.py
mnn31: Done. Removed the custom helper; `combine_results` now builds the distribution from `MeasurementResult.prob_distribution()`.

mitiq/experimental/deb/__init__.py
mnn31: Added `combine_results(results, permutations, method=...)` supporting `"averaging"` and `"sharpening"`, and exported it. Both high-level executors now route through it.

mitiq/experimental/deb/tests/test_symmetrization.py
mnn31: Added `test_variants_are_unitarily_equivalent`: it undoes each variant's permutation and checks the recovered unitary matches the original.

mitiq/experimental/deb/tests/test_sharpening.py
mnn31: `sharpen` now discards a shot when there is no plurality winner. Added a test for the tie case and a mixed case where a tie is dropped but a clear winner is kept.

mnn31:
Thanks for the detailed review @natestemen! I've addressed all five points:

- `construct_circuits` now uses `numpy.random.permutation`.
- Added `combine_results` (averaging or sharpening) and exported it; the high-level executors route through it.
- Replaced the custom probability helper with `MeasurementResult.prob_distribution()`.
- `sharpen` now discards shots with no plurality winner, with tests for the tie cases.
- Added a unitary-equivalence test for the generated variants.

Ready for another look whenever you have time.

mitiq/experimental/deb/symmetrization.py
```diff
@@ -0,0 +1,96 @@
...
+
+    return variants
+
+
+def construct_circuits(
+    circuit: cirq.Circuit,
+    num_variants: int = 10,
+    *,
+    random_state: int | np.random.Generator | None = None,
+) -> list[cirq.Circuit]:
```
natestemen: The intended use of `construct_circuits` is with `combine_results`. As things stand that is not possible since the random permutations are not part of the output here. Can you return them as part of a tuple which can then be input to `combine_results`?

mitiq/experimental/deb/deb.py
```diff
@@ -0,0 +1,151 @@
...
+        bitstring: probability / len(distributions)
+        for bitstring, probability in total.items()
+    }
+
+
+def combine_results(
+    results: Sequence[MeasurementResult],
+    permutations: Sequence[Sequence[int]],
+    method: str = "averaging",
+) -> dict[str, float]:
```
natestemen, line 64: I think there is an implicit qubit ordering here. Can you make sure it is reflected in both the docstring, and the user guide page.

mitiq/experimental/deb/deb.py
```diff
@@ -0,0 +1,151 @@
...
+    )
+
+
+def execute_with_debiasing(
+    circuit: cirq.Circuit,
+    executor: Executor,
+    num_variants: int = 10,
+    *,
+    random_state: int | np.random.Generator | None = None,
+) -> dict[str, float]:
```
natestemen, line 113: Instead of having both execute with debiasing and execute with debiasing and sharpening, lets just move the post-processing to a kwarg to simplify the interface.

mitiq/experimental/deb/deb.py
```diff
@@ -0,0 +1,151 @@
...
+    """Combine the measurement results of debiasing variants.
+
+    Each result is first unscrambled with its variant's permutation. The
+    unscrambled results are then combined either by averaging their probability
+    distributions (``"averaging"``) or by a shot-wise plurality vote
+    (``"sharpening"``, see :func:`.sharpen`).
+
+    Args:
+        results: One ``MeasurementResult`` per variant, in the same order as
+            the variants returned by :func:`.construct_circuits`.
```
natestemen, line 78: Lets add a note here (or elsewhere in this docstring) that this function assumes that each measurement result implicitly assumes each variant was run with the same number of shots.

natestemen (commented):
Looking really good! A few more comments, but this is getting there.

mitiq/experimental/deb/symmetrization.py
mnn31: `construct_circuits` now returns `(circuits, permutations)` as a tuple, and the permutations feed directly into `combine_results`. Updated the docstring and guide to show the flow.

mitiq/experimental/deb/deb.py
mnn31: Collapsed the two functions into a single `execute_with_debiasing(..., method="averaging" | "sharpening")` and removed `execute_with_debiasing_and_sharpening`.

mitiq/experimental/deb/deb.py
mnn31: Documented the ordering (bitstrings follow `sorted(circuit.all_qubits())`) in the docstrings of `execute_with_debiasing` and `combine_results`, and in the user guide page.

mitiq/experimental/deb/deb.py
mnn31: Added a note to `combine_results` that sharpening assumes each variant was run with the same number of shots (results are truncated to the smallest count otherwise).

mnn31:
Thanks @natestemen, addressed all four:

- `construct_circuits` now returns `(circuits, permutations)` so it pairs with `combine_results`.
- Collapsed to a single `execute_with_debiasing(..., method=...)` with `"averaging"` / `"sharpening"`.
- Documented the qubit ordering in the docstrings and the guide page.
- Noted the equal-shots assumption on `combine_results`.

All green locally. Ready for another look.

mitiq/experimental/deb/deb.py
```diff
@@ -0,0 +1,140 @@
...
+            :func:`.construct_circuits`).
+        method: Either ``"averaging"`` or ``"sharpening"``.
+
+    Returns:
+        The combined probability distribution over the computational basis,
+        with bitstrings ordered to match ``sorted(circuit.all_qubits())``.
+    """
+    unscrambled = [
+        _unscramble(result, permutation)
+        for result, permutation in zip(results, permutations)
```
natestemen: **Bug: `zip` without `strict=True` silently truncates on length mismatch**

`combine_results` is part of the public API (exported from `__init__.py`), so callers can pass `results` and `permutations` of different lengths. `zip` will silently truncate to the shorter iterable, biasing the combined distribution with no error.

The project already requires Python ≥ 3.11 where `zip(..., strict=True)` is available.

```suggestion
        for result, permutation in zip(results, permutations, strict=True)
```

mitiq/experimental/deb/deb.py
```diff
@@ -0,0 +1,140 @@
...
+
+def _unscramble(
+    result: MeasurementResult, permutation: Sequence[int]
+) -> MeasurementResult:
+    """Undo a variant's qubit permutation on its measured bitstrings.
+
+    Logical qubit ``i`` is measured on the qubit at index ``permutation[i]``,
+    so the unscrambled bit for logical qubit ``i`` is ``shot[permutation[i]]``.
+    """
+    unscrambled = [
```
natestemen: **Simplification: `range(len(...))` pattern + redundant `int()` cast**

`shot` is a row from `result.result`, which stores plain Python `int` values after `MeasurementResult.__post_init__` normalises the input. The `int()` cast is a no-op and the `range(len(permutation))` anti-pattern obscures intent.

```suggestion
        [shot[p] for p in permutation]
```

natestemen (approved):
Two last comments, and then the merge conflict needs resolved, but then we are good to go!

mitiq/experimental/deb/deb.py
mnn31: Fixed — `combine_results` now calls `zip(results, permutations, strict=True)`, so a length mismatch raises a `ValueError` instead of silently truncating to the shorter iterable.

mitiq/experimental/deb/deb.py
mnn31: Dropped the per-element `int()` no-op. `MeasurementResult.__post_init__` normalises every shot to a `list[int]`, but the declared type stays the broader `Bitstring` (`str | list[int]`), so indexing `shot[p]` is statically `str | int` and mypy rejects the result. I narrow once with `cast(list[list[int]], result.result)` and then index directly — this removes the redundant cast while keeping `make check-types` green.

mnn31:
@natestemen this is ready for another look. All review feedback is addressed:

- Replaced the custom distribution helper with `MeasurementResult.prob_distribution()`
- `construct_circuits` now returns `(circuits, permutations)` feeding into `combine_results`
- Single `execute_with_debiasing(..., method="averaging" | "sharpening")` interface
- `combine_results` uses `zip(..., strict=True)`; dropped the redundant per-bit `int()` cast (narrowed once via `cast`)
- Qubit-ordering and equal-shot-count assumptions documented in docstrings + user guide
- Added tests for unitary equivalence of variants and the no-winner/tie sharpening case

Rebased clean on `main` (resolved the `refs.bib` conflict) and CI is green. Mind giving it a merge when you have a moment? Thanks!

PR #3045, opened 2026-06-08
Add a CDR tutorial to the example gallery

### LINKED ISSUE #1507: Add tutorials with other techniques

mnn31:
## Description

Closes #1507.

Adds a worked Clifford data regression (CDR) tutorial to the example gallery. The issue points out that the gallery had tutorials for ZNE but not much for the other techniques, and CDR only had the Qrack-specific example. This one runs on Cirq's local simulators, so it needs no hardware or extra backends.

The tutorial:
- builds a small near-Clifford circuit and a two-term observable,
- defines ideal and noisy (depolarizing) executors,
- compares the noisy estimate against standard CDR and variable-noise CDR,
- and plots the absolute errors.

It also covers variable-noise CDR through the `scale_factors` argument, since that is part of the "(variable-noise) Clifford data regression" technique and is natural to show next to the basic version. Every cell runs during the docs build; on the example circuit both CDR and variable-noise CDR move the noisy value back toward the ideal one, with variable-noise CDR doing better here.

I used AI assistance while writing this and have run and checked the whole notebook, so I can walk through any part during review.

### License

### DIFF
--- a/docs/source/examples/cdr_tutorial.md
+++ b/docs/source/examples/cdr_tutorial.md
@@ -1,0 +1,182 @@
+---
+jupytext:
+  text_representation:
+    extension: .md
+    format_name: myst
+    format_version: 0.13
+    jupytext_version: 1.16.1
+kernelspec:
+  display_name: Python 3 (ipykernel)
+  language: python
+  name: python3
+---
+
+```{tags} cdr, cirq, basic
+```
+
+# Clifford data regression (CDR) with Cirq
+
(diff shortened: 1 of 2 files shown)


natestemen:
Thanks for your contribution here, but unfortunately we've decided to go with another hackers contribution (https://github.com/unitaryfoundation/mitiq/pull/3052). I hope you learned something, and know you're always welcome to contribute to mitiq in the future!

PR #3052, opened 2026-06-13
Add CDR worked example with Clifft as a near-Clifford simulator

### LINKED ISSUE #1507: Add tutorials with other techniques

qnelsonksh:
Closes #1507.

Adds `docs/source/examples/cdr_demo.md` — a worked CDR example running Clifford Data Regression end-to-end on a hardware-native circuit with Clifft as the near-Clifford simulator.

## Why this approach
- **Clifft as the near-Clifford simulator** — CDR's training circuits are near-Clifford by construction, so a specialized simulator is the natural fit (addresses the maintainers' suggestion to use a near-Clifford simulator such as Clifft).
- **Native-basis transpilation** via `qiskit.transpile` to `{rz, sx, cx}`, which confines all non-Clifford content to the `rz` angles — the structural fact CDR exploits.
- **An experiment, not just a demo** — a noise sweep shows CDR tracking the ideal value and degrading gracefully at high noise. (I first tried sweeping circuit depth, but each depth is a different random circuit so the CDR conditioning varied wildly; sweeping noise at fixed depth is the clean comparison.)
- A short **vnCDR** section using `scale_factors` with `fold_global` (the default `fold_gates_at_random` is unseeded, so `fold_global` keeps it reproducible).

## Implementation notes
- Clifft's `R_Z` takes the angle in half-turns, which cirq stores in `gate.exponent`.
- Clifft is little-endian vs cirq's big-endian, so I reverse the statevector axes before building the density matrix.

## Testing
- Executed end-to-end via `jupytext` (7 cells, 0 errors, figure renders).
- Also ran in a clean Google Colab (Linux) environment using clifft's manylinux wheel.

Also adds `clifft==0.4.1` (Python-3.12-marked, matching the `ucc` pattern) to the docs dependencies + `uv.lock`, a gallery entry and thumbnail in `examples.md`/`conf.py`, and a Clifft citation to `refs.bib`.

## AI disclosure
I used Claude (Anthropic) as a co-pilot — to learn the CDR / near-Clifford background, scaffold the notebook structure, and help debug. I wrote the code and prose myself, verified every cell by running the notebook locally and in Colab, and can explain each step.

---

### License

- [ ] I added unit tests for new code. _(N/A — documentation example, executed by the docs build)_
- [ ] I used [Google-style](https://google.github.io/styleguide/pyguide.html#383-functions-and-methods) docstrings for functions. _(N/A — short tutorial cells)_


b-goldsmith:
Thanks for the submission. In the example, you point out:
```
To connect this abstract circuit to real hardware, we transpile it into IBM’s native gate set {rz, sx, cx}.
```

Then the circuit is immediately translated to Cirq. I would imagine if you are taking the time to point out IBM and their gate set, you would use an IBM based simulator or fake provider for the noisy results. If you are only using the Qiskit code for `transpile`, I would recommend sticking with Cirq and using their transpilation methods to either a Clifford + Rz gate set or a more limited gateset for particular hardware.

Per the [User Guide](https://mitiq.readthedocs.io/en/stable/guide/cdr-1-intro.html#define-a-quantum-circuit):
The quantum circuit can be specified as any quantum circuit supported by Mitiq but
**must be compiled into a gateset in which the only non-Clifford gates are
single-qubit rotations around the $Z$ axis: $R_Z(\theta)$**.

qnelsonksh:
Thanks for the feedback — you're right that I was only using Qiskit for the transpile. I've rebuilt the circuit entirely in Cirq, constructed directly in a `{H, Rz, CNOT}` gateset where the only non-Clifford gates are the single-qubit `Rz` rotations (per the user guide), so there's no separate transpilation step and no IBM framing. Clifft remains the near-Clifford simulator. I also added a short note on the regime where CDR's linear correction stays valid. Thanks again!

docs/source/examples/cdr_demo.md
```diff
@@ -0,0 +1,360 @@
...
+        else:
+            raise ValueError(f"unhandled gate: {g!r}")
+    return "\n".join(lines)
+
+
+def clifft_statevector(circuit, qubits):
+    prog = clifft.compile(to_clifft_text(circuit, qubits))
+    state = clifft.State(
+        peak_rank=prog.peak_rank,
+        num_measurements=prog.num_measurements,
```
bachase: Note that I think only peak_rank is needed here in practice.

docs/source/examples/cdr_demo.md
```diff
@@ -0,0 +1,360 @@
...
+        num_detectors=prog.num_detectors,
+        num_observables=prog.num_observables,
+        num_exp_vals=prog.num_exp_vals,
+    )
+    clifft.execute(prog, state)  # required before reading when peak_rank > 0
+    psi = clifft.get_statevector(prog, state)
```
bachase: `get_statevector` is actually a relatively slow pathway for clifft, materializing a $2^n \times  2^n$ matrix to backout the statevector. That might be fine for this demo, but might be misleading for users that want to generalize to much larger $n$.

I'd consider using [`EXP_VAL`]( https://unitaryfoundation.github.io/clifft/stable/guide/simulation/#expectation-values) to calculate `<Z0 Z3>` directly, as it should be much more efficient and generalize more cleanly.

bachase (commented):
Nice to see another clifft use case! As noted below, I think there is a more efficient and clifft idiomatic approach to try, which hopefully wouldn't be a huge lift.

qnelsonksh:
Thanks @bachase — great suggestion, straight from the source! I switched the simulator to read `<Z0 Z3>` directly via `EXP_VAL` / `clifft.sample(...).exp_vals`, so there's no more statevector or density-matrix materialization (and it drops the `State` boilerplate you flagged too). The executors now return the expectation directly, and I call `execute_with_cdr` without a separate observable. Verified `EXP_VAL` matches the exact value and the CDR results are unchanged. Much appreciated!

docs/source/examples/cdr_demo.md
```diff
@@ -0,0 +1,347 @@
...
+
+```{tags} cdr, clifft, cirq, intermediate
+```
+
+# Clifford data regression with the Clifft near-Clifford simulator
+
+Clifford data regression (CDR) is an error mitigation technique that learns a
+correction from noisy circuit outputs to their ideal values by training on
+classically simulable near-Clifford circuits. In this tutorial, we run CDR
+end-to-end on a layered variational circuit, using the Clifft near-Clifford
```
natestemen: Link for first instance of clifft needed.

docs/source/examples/cdr_demo.md
```diff
@@ -0,0 +1,347 @@
...
+generate near-Clifford training circuits by adjusting a small number of
+non-Clifford rotations, and it is precisely the regime where a near-Clifford
+simulator such as Clifft provides an advantage over general-purpose simulation.
+
+```{code-cell}
+def build_ansatz(qubits, n_layers, angles):
+    circuit = cirq.Circuit()
+    for layer in range(n_layers):
+        for i, q in enumerate(qubits):
+            circuit.append([cirq.H(q), cirq.rz(angles[layer][i]).on(q), cirq.H(q)])
```
natestemen, line 87: Why not just use an rx rotation instead of the H Rz H?

docs/source/examples/cdr_demo.md
```diff
@@ -0,0 +1,347 @@
...
+of the lost signal. As the noise becomes more severe, the quality of the learned
+correction gradually degrades. This behavior is expected: CDR is not removing
+noise from the quantum device, but rather learning a model of how noise distorts
+nearby circuits. When that distortion becomes too large or too nonlinear, the
+learned correction becomes less predictive. The graceful breakdown visible in the
+plot is therefore an important part of understanding the method, not a failure of
+it.
+
+```{code-cell}
+noise_levels = [0.005, 0.01, 0.02, 0.04, 0.06, 0.08]
```
natestemen: For the sake of the final visualization I think it would be good to include 0 noise in here.

natestemen (approved):
Looks good! Just a few minor comments.

docs/source/examples/cdr_demo.md
b-goldsmith: The [Users Guide](https://mitiq.readthedocs.io/en/stable/guide/cdr-1-intro.html#define-a-quantum-circuit) requires the quantum circuit **must be compiled into a gateset in which the only non-Clifford gates are
single-qubit rotations around the $Z$ axis: $R_Z(\theta)$**.

qnelsonksh:
Thanks @natestemen and @b-goldsmith! Addressed the comments:

- Linked Clifft on its first mention.
- On `H·Rz·H` vs `rx` — exactly as @b-goldsmith notes, the only non-Clifford gates must be `Rz` (per the user guide), and a bare `rx(θ)` is a non-Clifford X-rotation that wouldn't satisfy that. `H·Rz·H` gives the same `Rx(θ)` while keeping the non-Clifford content in the `Rz`. I made this explicit in the text.
- Added the zero-noise point to the sweep so the plot is anchored where noisy = ideal = CDR.

Thanks both!

natestemen (approved):
Once the merge conflicts are resolved I will merge this!

qnelsonksh:
@natestemen conflicts are resolved — kept both the Clifft citation and the [[4,2,2]] one, and regenerated the lockfile. Should be ready to merge whenever you get a chance. Thanks!

b-goldsmith (approved):

PR #3058, opened 2026-06-17
Support representations directly in PEA (closes #2936)

### LINKED ISSUE #2936: Support representations directly in PEA

nez0b:
## Summary

Probabilistic Error Amplification (PEA) currently amplifies noise only through a `noise_model` string
enum (`"local_depolarizing"` / `"global_depolarizing"`), which limits it to analytically tractable
models and makes it unusable on real hardware. This PR adds a `representations` parameter, a
`list[OperationRepresentation]`, to `pea.construct_circuits` and `pea.execute_with_pea`, so PEA can
amplify any quasi-probability representation, including ones learned from hardware characterization.

- `noise_model` becomes optional; `representations` is its alternative. The two are **mutually
  exclusive** (passing both, or neither, raises a clear `ValueError`). `epsilon` is used only by the
  legacy `noise_model` path and is ignored when `representations` is provided.
- `noise_model` keeps working unchanged for backward compatibility, now emitting a `DeprecationWarning`.
- Representations are amplified with the **canonical noise scaling** of Section D of Mari, Shammah &
  Zeng, *Phys. Rev. A* **104**, 052607 (2021), [arXiv:2108.02237](https://arxiv.org/abs/2108.02237),
  implemented as a new helper `scale_representation` that is correct for *both* signed (PEC-style) and
  all-positive (amplification/learned-noise) representations. Internally, both public input modes resolve
  to one `scale_factor -> representations` provider, so the sampling pipeline has a single interface while
  preserving the exact legacy rebuild semantics.

## Background: PEA in brief

A quasi-probability representation writes an ideal gate as a real linear combination of implementable
noisy operations,

$$\mathcal{G} = \sum_\alpha \eta_\alpha\, \mathcal{O}_\alpha, \qquad \sum_\alpha \eta_\alpha = 1, \qquad \gamma = \sum_\alpha |\eta_\alpha| \ge 1 .$$

PEC uses this to *cancel* noise; PEA uses the same object to *amplify* it in a controlled way: each gate
is re-expressed at a series of noise scale factors $\lambda \ge 1$, the expectation values are estimated
at each $\lambda$, and the results are extrapolated to $\lambda = 0$ (zero-noise), i.e. PEA is the
principled, quasi-probability front end of ZNE.

The math for scaling an arbitrary representation is the **canonical noise scaling** of Section D. Split the
coefficients by sign into positive/negative volumes $\gamma^{+} = \sum_{\eta>0}\eta$ and
$\gamma^{-} = \sum_{\eta<0}|\eta|$ (so $\gamma^{+} - \gamma^{-} = 1$), and rescale:

$$\eta_{+} \;\to\; \eta_{+}\,\frac{\gamma^{+} - \lambda\,\gamma^{-}}{\gamma^{+}}, \qquad \eta_{-} \;\to\; \eta_{-}\,(1 - \lambda),$$

valid up to $\lambda_{\max} = \gamma^{+}/\gamma^{-}$. This preserves $\sum_\alpha \eta_\alpha = 1$ for
every $\lambda$; at $\lambda = 1$ it yields the base noisy channel (one-norm $\gamma = 1$), and for
$1 \le \lambda \le \lambda_{\max}$ the rescaled coefficients are **all non-negative**, a genuine
probability distribution, so amplification is sign-problem-free ($\gamma = 1$, negligible sampling
overhead). For an all-positive representation ($\gamma^{-} = 0$, e.g. a learned noise model) the
canonical map degenerates, and the natural rule is **deviation-from-identity** scaling
$a_0 \to 1 - \lambda(1 - a_0)$, $a_k \to \lambda\, a_k$, which reproduces the analytic depolarizing
rebuild.

## Why the existing approaches are incomplete (#3027, #3041)

Two open PRs each implement only one of the two scaling rules above, and so each fails on half of the
realistic inputs. Numerical evidence for single-qubit depolarizing, base $p = 0.01$, amplification factor
$s = 2$ (only the failing cases are shown):

| PR | Input it mishandles | Result of scaling | Issue |
|---|---|---|---|
| **#3041** (deviation-only) | signed PEC rep $[1.0101,\,-0.00338,\,-0.00338,\,-0.00338]$ ($\gamma=1.0203$) | $[1.0203,\,-0.00676,\,-0.00676,\,-0.00676]$, $\gamma = 1.041$ | **still signed; $\gamma$ grew** $\to$ amplifies error *cancellation*, not noise. Silently wrong (no error raised). |
| **#3027** (canonical-only) | all-positive `amplify`/learned rep $[0.98,\,0.0067,\,0.0067,\,0.0067]$ | raises `ValueError` ($\gamma^{-} = 0$) | **cannot scale all-positive reps** $\to$ rejects learned- and `noise_model`-style representations. |

- **#3041** uses deviation-from-identity scaling: on a signed PEC representation it scales every error
  term by $s$ and pushes the identity term *away* from 1, *increasing* the negativity
  ($\gamma:\ 1.020 \to 1.041$), amplifying the error-cancellation rather than the noise, without
  raising.
- **#3027** uses canonical sign-partition scaling: exact on signed PEC reps, but raises on any
  all-positive representation ($\gamma^{-} = 0$), which breaks the depolarizing-`noise_model` equivalence
  path and rejects learned all-positive noise models.

Both PRs are still open/unmerged, and their scaling rules were re-verified against their current heads:
#3041's `scale_representations` applies deviation-from-identity to every representation (no sign
dispatch), and #3027's `_canonical_scaled_representations` raises `ValueError` on any all-positive
representation away from scale factor 1. The gaps above are therefore unfixed. This PR keeps the
strengths of each.

## Our approach: one interface, two providers (fidelity vs generality)

`scale_representation(representation, scale_factor)` is **convention-aware**: it dispatches on the sign
structure of the representation (with a small tolerance so floating-point dust does not misroute it):

- contains negative coefficients (PEC-style) $\to$ **canonical sign-partition** scaling (Section D), with a
  $\lambda_{\max} = \gamma^{+}/\gamma^{-}$ physicality guard;
- all non-negative (learned-noise / `amplify_*` style) $\to$ **deviation-from-identity** scaling.

This single helper supports signed PEC-style representations (depolarizing, biased, amplitude-damping,
optimal/tomographic) and all-positive amplification/learned-noise representations that include an
explicit identity (bare-ideal) term. (All-positive representations *without* an identity term are
rejected with a clear error.)

`num_samples` (when not given) is deduced from the largest scaled one-norm across the requested
scale factors, so noise-reduction factors below 1 (whose one-norm exceeds 1) are not under-sampled.

**Design choice: two providers behind one interface.** The public API keeps two ways to specify the
amplification (`noise_model` and `representations`); internally both resolve through one
`_make_scaled_representation_factory(...)`, so input validation and the sampling interface live in a
single place. What we deliberately do **not** do is the *literal math merge* (routing `noise_model`
through the `representations` scaling rule), because the two providers compute genuinely different, and
both correct, scaling families:

- **`noise_model` provider**: exact analytic rebuild at the target noise $s\,\epsilon$, a regenerable
  family that retains the model's structure (preserves the precise multi-qubit-local behavior; unchanged,
  so byte-for-byte backward compatible);
- **`representations` provider**: canonical/deviation scaling of a single supplied snapshot, the only
  option for arbitrary/learned reps where no regenerable model exists.

These coincide for global depolarizing (any size) and single-qubit local depolarizing but differ for
multi-qubit *local* noise; the precise reason and the math are in **Known limitation** below. Choosing
between the providers is choosing *factorized exactness* (when the analytic model is known) vs
*generality* (any representation).

## What we test

A new `tests/test_representations_pathway.py` (cirq-only, deterministic) exercises every representation
family and the canonical-scaling math:

- **Noise models / representation types**: depolarizing **local & global** (1- and 2-qubit), local
  **biased noise** (1- and 2-qubit), **amplitude damping** (1-qubit), **optimal/tomographic** (1-qubit,
  via `find_optimal_representation`), all-positive **`amplify_*`** reps, and a hand-built **sparse
  Pauli–Lindblad** representation (later could cross validate with ibm-q). For each: scaling preserves
  $\sum_\alpha \eta_\alpha = 1$, preserves the operation count, dispatches to the correct rule, and
  round-trips through `construct_circuits`.
- **Canonical-scaling correctness**: the headline test is that canonically scaling a signed depolarizing
  PEC rep reproduces the paper's closed-form $\eta_j(\lambda)$ to machine precision; the one-norm law
  $\gamma^{(\lambda)} = \gamma - \lambda(\gamma-1)$ on $[0,1]$ and $\gamma^{(\lambda)} = 1$ on
  $[1, \lambda_{\max}]$; $\lambda = 1 \Rightarrow$ one-norm 1; the $\lambda_{\max}$ physicality guard
  raises; the deviation-path upper bound warns; the sign dispatch tolerates floating-point dust.
- **End-to-end mitigation**: `execute_with_pea(..., representations=…)` with signed PEC reps against a
  depolarizing simulator reduces the error and lands within `atol = 0.1`; plus a learned-$\epsilon$
  workflow and a signed-vs-`noise_model` accuracy comparison.
- **Multi-qubit `noise_model` scope**: a 2-qubit *global* depolarizing test asserts the
  `representations` path equals the `noise_model` path exactly, and a 2-qubit *local* test asserts they
  differ (regression-locking the single-parameter limitation); the signed canonical-vs-factorized
  difference is also regression-tested.
- **Sampling budget**: a test asserting `num_samples` is deduced from the max scaled one-norm across the
  requested scale factors (so sub-unit factors are not under-sampled).
- **Validation**: both / neither / empty `representations` $\to$ `ValueError`; non-positive and non-finite
  scale factors $\to$ `ValueError`; empty `scale_factors` $\to$ `ValueError`; an all-positive rep without
  an identity term $\to$ `ValueError`; floating-point dust in the sign dispatch is tolerated; the
  `noise_model` deprecation warning.

## Known limitation

Both representation-scaling rules produce a family that is **affine (degree 1) in the scale factor**
$s$: every coefficient is a degree-1 polynomial in $s$ (canonical:
$\eta_+ \to \eta_+(1 - s\,\gamma^-/\gamma^+)$ and $\eta_- \to \eta_-(1 - s)$; deviation:
$a_0 \to 1 - s(1 - a_0)$ and $a_k \to s\,a_k$). A named `noise_model`, in contrast, is a *regenerable
family*: it rebuilds the representation at any noise level, retaining the model's structure. A
**factorized $k$-qubit local** channel is a tensor product, so its coefficients are *products* of
per-qubit terms and the family is **degree $k$** in $s$. For 2-qubit local depolarizing the two-Pauli
coefficient is

$$c_{\text{true}}(s) = \Big(\tfrac{s\,\epsilon}{3}\Big)^{2} = s^{2}\Big(\tfrac{\epsilon}{3}\Big)^{2}\ \ (\text{degree }2), \qquad c_{\text{aff}}(s) = s\,\Big(\tfrac{\epsilon}{3}\Big)^{2}\ \ (\text{degree }1).$$

They agree at the pinned points $s = 0$ (ideal) and $s = 1$ (base noise) but diverge for any other $s$
once $k \ge 2$: at $\epsilon = 0.05,\ s = 2$ they are $c_{\text{true}} = 0.001111$ vs
$c_{\text{aff}} = 0.000556$ (ratio $s$), the source of the observed $\max|\Delta| \approx 5\times10^{-3}$.
For $k = 1$ (single-qubit local, and global depolarizing, which is one channel for any qubit count) the
true family is itself degree 1, so the affine scaling is **exact**. Equivalence with `noise_model` is
therefore exact only for **global depolarizing (any size)** and **single-qubit local** depolarizing.

**This is not "the representation has no model."** It is tempting to say a representation has "lost the
noise model," but 2-qubit `local_depolarizing` *has* a named model and is *exactly* where the paths
diverge, so that cannot be the cause. In fact every signed representation *induces* a noise model: by
Section D it defines a canonical noise channel $\Lambda_p = (1 - p)\,\mathcal{G} + p\,\Phi^{-}$, and an
all-positive representation is extended by the deviation-from-identity rule. The point is finer: the
**induced model is single-parameter (affine), hence lossy.** A bare representation fixes only one point
($s = 1$); extending it to a family forces the affine rule, whereas a named model carries enough
structure to define the exact, higher-degree family. The representation-to-model map always exists; it
simply need not equal the true physical noise when that noise scales nonlinearly (degree $> 1$), as
factorized multi-qubit local noise does.

**Practical consequence.** When the analytic model is known, prefer `noise_model` (exact). The
`representations` path is the general fallback for arbitrary/learned noise, where no regenerable model
exists and the Section-D single-parameter scaling is the principled choice. Recovering the higher-degree
behavior from a representation would require carrying factorization / noise-model metadata, which would
change the `OperationRepresentation` contract and defeat #2936's goal of accepting arbitrary learned
representations; we deliberately do not do that.

## Files changed

- `mitiq/experimental/pea/pea.py`: `representations` parameter on `construct_circuits` /
  `execute_with_pea`, mutual-exclusion + empty-list validation, deprecation warning.
- `mitiq/experimental/pea/scale_amplifications.py`: new `scale_representation` (sign-dispatch:
  canonical / deviation, dust-tolerant) plus a resolver that maps either public input mode to
  `scale_factor -> representations`.
- `mitiq/experimental/pea/tests/test_pea.py`,
  `mitiq/experimental/pea/tests/test_scale_amplifications.py`: pathway + validation tests.
- `mitiq/experimental/pea/tests/test_representations_pathway.py`: new comprehensive suite.

Closes #2936

### DIFF
--- a/docs/source/guide/pea-1-intro.md
+++ b/docs/source/guide/pea-1-intro.md
@@ -105,6 +105,13 @@ Here we observe that the application of PEA reduces the estimation error when co
 In the example above, both the noise amplification and extrapolation steps were taken behind the scenes thanks to the default options of {func}`.execute_with_pea`.
 In the following pages, we describe additional options and show how to apply the two steps of PEA independently.
 
+```{note}
+The example above uses the legacy `noise_model` string, which now emits a `DeprecationWarning`. The
+recommended interface is to pass `representations`, a list of {class}`.OperationRepresentation` objects
+that can be learned from hardware characterization, which makes PEA usable on noise that is neither local
+nor global depolarizing. See [What additional options are available in PEA?](pea-3-options.md) for details.
+```
+
 ## Two-stage PEA workflow
 
 If you want more control over the process, you can split PEA into two stages:
--- a/docs/source/guide/pea-3-options.md
+++ b/docs/source/guide/pea-3-options.md
@@ -14,8 +14,59 @@ kernelspec:
 # What additional options are available in PEA?
 
 {func}`.execute_with_pea` exposes several optional arguments beyond the required `circuit`, `executor`,
-`scale_factors`, `noise_model`, `epsilon`, and `extrapolation_method`.
+`scale_factors`, and `extrapolation_method`, together with the noise amplification specified by either
+`representations` or the legacy `noise_model`/`epsilon` pair.
 This page describes each one.
+
+## Choosing how the noise is amplified
+
+PEA needs a quasi-probability representation of each operation in the circuit in order to amplify the
+noise. There are two mutually exclusive ways to provide it.
+
+The recommended way is to pass `representations`, a list of {class}`.OperationRepresentation` objects
+(one per unique operation of the circuit). Because these can be learned from hardware characterization,
+this is the only interface that works on noise that is neither local nor global depolarizing:
+
+```{code} python
+# reps: list[OperationRepresentation], e.g. learned from hardware characterization
+mitigated = pea.execute_with_pea(
+    circuit,
+    executor,
+    scale_factors=[1.0, 1.2, 1.6],
+    extrapolation_method=LinearFactory.extrapolate,
+    representations=reps,
+)
+```
+
+Each representation is amplified with the canonical noise scaling of Section D of {cite}`Mari_2021_PRA`:
+signed (PEC-style) representations use the sign-partition rule, and all-positive representations (such as
+learned noise models) use deviation-from-identity scaling. Both keep the coefficients summing to one.
+
+The legacy alternative is to pass `noise_model` together with `epsilon`; Mitiq then builds the
+depolarizing representations for you. This path still works but emits a `DeprecationWarning`; prefer
+`representations`:
+
+```{code} python
+mitigated = pea.execute_with_pea(
+    circuit,
+    executor,
+    scale_factors=[1.0, 1.2, 1.6],
+    extrapolation_method=LinearFactory.extrapolate,
+    noise_model="local_depolarizing",   # legacy
+    epsilon=0.005,
+)
+```
+
+Passing both `representations` and `noise_model`, or neither, raises a `ValueError`. `epsilon` is used
+only by the `noise_model` path and is ignored when `representations` is given.
+
+```{note}
+For a representation of a *standard depolarizing model*, the two paths agree exactly for global
+depolarizing (any number of qubits) and for single-qubit local depolarizing. For multi-qubit *local*
+depolarizing they differ slightly: the `noise_model` path rebuilds the exact (factorized) amplification,
+whereas a representation is scaled with the single-parameter canonical rule. Prefer `noise_model` when
+the analytic model is known; use `representations` for arbitrary or learned noise.
+```
 
 ## Controlling the sampling budget
 
@@ -105,6 +156,6 @@ If no observable is provided, the executor must return the expectation value dir
 ## Supported noise models
 
 ```{attention}
-The only supported noise models are currently `"local_depolarizing"` and `"global_depolarizing"`.
-Please [open an issue](https://github.com/unitaryfoundation/mitiq/issues/new) to request additional noise models.
+The built-in `noise_model` strings are currently `"local_depolarizing"` and `"global_depolarizing"`.
+To amplify any other noise, pass `representations` directly (see [Choosing how the noise is amplified](#choosing-how-the-noise-is-amplified) above), which is not restricted to these two models.
 ```
--- a/mitiq/experimental/pea/pea.py
+++ b/mitiq/experimental/pea/pea.py
@@ -186,9 +186,7 @@ def execute_with_pea(
     scale_factors: list[float],
     noise_model: str | None = None,
     epsilon: float | None = None,
-    extrapolation_method: Callable[
-        [Sequence[float], Sequence[float]], float
-    ]
+    extrapolation_method: Callable[[Sequence[float], Sequence[float]], float]
     | None = None,
     observable: Observable | None = None,
     random_state: int | np.random.RandomState | None = None,
--- a/mitiq/experimental/pea/pea.py
+++ b/mitiq/experimental/pea/pea.py
@@ -14,8 +14,9 @@ from cirq import Circuit
 
 from mitiq import QPROGRAM, Executor, Observable, QuantumResult
 from mitiq.experimental.pea.scale_amplifications import (
-    scale_circuit_amplifications,
+    _make_scaled_representation_factory,
 )
+from mitiq.pec import OperationRepresentation
 from mitiq.pec.pec import (
     _LARGE_SAMPLE_WARN,
     LargeSampleWarning,
@@ -26,11 +27,12 @@ from mitiq.pec.pec import (
 def construct_circuits(
     circuit: Circuit,
     scale_factors: list[float],
-    noise_model: str,
-    epsilon: float,
+    noise_model: str | None = None,
+    epsilon: float | None = None,
     random_state: int | np.random.RandomState | None = None,
     precision: float = 0.1,
     num_samples: int | None = None,
+    representations: Sequence[OperationRepresentation] | None = None,
 ) -> tuple[list[list[QPROGRAM]], list[list[int]], list[float]]:
     """Samples lists of implementable circuits from the noise-amplified
     representation of the input ideal circuit at each input noise scale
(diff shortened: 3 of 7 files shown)


natestemen:
I'm going to close this for now since we are going with https://github.com/unitaryfoundation/mitiq/pull/3041. Thanks @nez0b for your work here. More contributions to mitiq are always welcome! :)

PS I'll ask the other hacker to add you as commit author so you get credited there.

PR #3059, opened 2026-06-17
Fix TREX combine_results qubit indexing for partial-support observables

ayush268:
combine_results indexed the randomization string using qubit_to_idx built from sorted(observable._qubits()), but the randomization strings are generated in construct_circuits indexed by the circuit's qubits. When the observable does not act on every circuit qubit (e.g. a single-qubit Pauli on a high-index qubit), the readout twirling was undone with the wrong bits, silently producing an incorrect mitigated value.

Derive the circuit's qubit ordering from the calibration results (which measure all circuit qubits in the same order the randomization string is indexed) so the twirl is undone consistently. This needs no public API change.

Add a parametrized regression test covering observables acting on qubits 0, 1, 2 and all three qubits of a 3-qubit |111> state; the qubit-1 and qubit-2 cases fail on main without this fix.

## Description

`combine_results` in `mitiq.experimental.trex` builds its `qubit_to_idx` map
from `sorted(observable._qubits())`, but the randomization strings produced by
`construct_circuits` are indexed by the **circuit's** qubits. When an
observable does not act on every circuit qubit (e.g. a single-qubit Pauli on a
high-index qubit), the readout twirling is undone with the wrong bits and TREX
silently returns an incorrect expectation value.

## Reproduction (on `main`)
A noiseless 3-qubit `|111⟩` circuit, expectation of single-qubit `Z`:
- `Z` on qubit 0 → `-1.0` ✅
- `Z` on qubit 1 → `+0.125` ❌ (should be `-1.0`)
- `Z` on qubit 2 → `+0.25`  ❌ (should be `-1.0`)

## Fix
Recover the circuit's qubit ordering from the calibration results (which
measure all circuit qubits in the same order the randomization string is
indexed) and index `s` consistently. No public API change.

## Tests
Adds a parametrized regression test (`test_trex_observable_on_subset_of_qubits`)
covering `Z` on qubits 0/1/2 and `ZZZ`. The qubit‑1 and qubit‑2 cases fail on
`main` without this change. Full `trex` suite passes (44 tests); `ruff` and
`mypy` clean.

---

### License

### DIFF
--- a/mitiq/experimental/trex/tests/test_trex.py
+++ b/mitiq/experimental/trex/tests/test_trex.py
@@ -129,6 +129,38 @@ def test_trex_single_qubit():
     assert np.isclose(result, -1.0, atol=0.1)
 
 
+@pytest.mark.parametrize(
+    "spec, expected",
+    [
+        ("ZII", -1.0),
+        ("IZI", -1.0),
+        ("IIZ", -1.0),
+        ("ZZZ", -1.0),
+    ],
+)
+def test_trex_observable_on_subset_of_qubits(spec, expected):
+    """TREX must be correct when the observable acts on only a subset of the
+    circuit's qubits, including high-index qubits.
+
+    Regression test: ``combine_results`` previously indexed the randomization
+    string by ``sorted(observable._qubits())`` instead of by the circuit's
+    qubits, so the readout twirling was undone with the wrong bits whenever the
+    observable did not act on qubit 0, silently corrupting the result.
+    """
+    qreg = [cirq.LineQubit(i) for i in range(3)]
+    circuit = cirq.Circuit(cirq.X.on_each(*qreg))  # prepares |111>
+    obs = Observable(PauliString(spec))
+
+    result = execute_with_trex(
+        circuit,
+        noiseless_executor,
+        obs,
+        num_randomizations=16,
+        random_state=0,
+    )
+    assert np.isclose(result, expected, atol=0.1)
+
+
 def test_trex_identity_circuit():
     """TREX on an identity circuit should give expectation +1 for Z."""
     q = cirq.LineQubit(0)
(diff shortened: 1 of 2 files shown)


kawacukennedy (approved):
LGTM. The fix correctly addresses the bug where `combine_results` indexed the randomization string by `sorted(observable._qubits())` instead of the circuit's qubits, which caused silent corruption when the observable acted on a subset of qubits. Tests verify Z on each qubit individually for |111> with correct expected values.

ayush268:
Hi @natestemen @vprusso 
Is anything needed on this change, please let me know and I can push the update

vprusso:
Thanks @ayush268. The fix looks right to me. The randomization strings are indexed by the circuit's sorted qubit order, and recovering that order from the calibration results' `qubit_indices` lines the twirl-undo back up with how the X gates were applied. Good catch on a silent-correctness bug.

One thing I'd like before merging. The regression test only covers contiguous qubits 0/1/2, where a qubit's position in the sorted order equals its `LineQubit.x`. In that case the new `circuit_qubit_indices.index(q.x)` lookup returns the same value a plain `q.x` would, so the part of the fix that actually matters, the position vs label remap, never gets exercised. Could you add a case on a non-contiguous register, for example an observable on `LineQubit(4)` over a circuit acting on qubits {0, 2, 4}? That fails with a naive mapping and would lock in the reindexing.

Minor and optional: `cast(cirq.LineQubit, q).x` assumes line qubits, so it raises `AttributeError` on grid or named qubits, where the old `sorted(observable._qubits())` did not. Fine if TREX is line-qubit only in practice, but worth a short comment or guard.

natestemen (approved):
Updated the PR to take care of non-contiguous registers as suggested by Vincent.

Thank you both for your effort here!

PR #3062, opened 2026-06-19
Fix executor return-type detection under `from __future__ import annotations` (closes #3061)

### LINKED ISSUE #3061: Executor return-type detection breaks under `from __future__ import annotations` (PEP 563)
## Issue Description

`Executor` determines how to interpret an executor's results by inspecting the
executor's **return type annotation** at `executor/executor.py:77`:

```python
executor_annotation = inspect.getfullargspec(executor).annotations
self._executor_return_type = executor_annotation.get("return")
```

`inspect.getfullargspec(...).annotations` returns the **raw** `__annotations__` dict. Under [PEP 563](https://peps.python.org/pep-0563/), if the executor is defined in a module that contains `from __future__ import annotations`, every annotation is stored as a **string**, so `_executor_return_type` becomes the string `"float"` instead of the `float` type.

The membership checks in `evaluate()` (e.g. `self._executor_return_type in FloatLike` at lines 185 and 232) compare against actual type objects (`float`, `Iterable[float]`, ...). The string `"float"` matches none of them, so execution falls through to the else branch and raises:

```python
ValueError: Could not parse executed results from executor with type float.
```

This is misleading: the executor does return a float; mitiq just failed to recognize the (stringified) annotation. The same breakage affects
`MeasurementResult` / density-matrix executors and propagates to any technique that builds an `Executor` internally (`execute_with_pec`, `execute_with_zne`, etc.).

## How to Reproduce

Note the **only** distinguishing factor is the `from __future__ import annotations` line — which is increasingly common in modern codebases.

```python
from __future__ import annotations  # <-- the only thing that matters

import cirq
from mitiq import Executor

def executor(circuit) -> float:
    return 1.0

q = cirq.LineQubit(0)
circuit = cirq.Circuit(cirq.X(q))

# Raises: ValueError: Could not parse executed results from executor with type float.
print(Executor(executor).evaluate(circuit))
```

Removing the `from __future__ import annotations` line makes it return `[1.0]` as expected.

## Expected behavior

`Executor` should resolve string annotations to their concrete types before comparing, so float/measurement/density-matrix executors are detected regardless of whether PEP 563 is active in the module where they are defined.

---

<details>
<summary>Environment Context</summary>

mitiq: 1.0.0
Python: 3.12.13
OS: Linux (also reproducible on any platform; the cause is Python-level)

</details>

ayush268:
Fixes #3061 

`Executor` read the executor's return type via
`inspect.getfullargspec(executor).annotations`, which returns the raw `__annotations__` dict. When the executor is defined in a module using `from __future__ import annotations`, annotations are strings, so the return type became the string `"float"` instead of the `float` type. The membership checks in `evaluate()` then matched nothing and raised `ValueError: Could not parse executed results from executor with type float`, even though the executor returned a valid float. This also affected measurement/density-matrix executors and any technique that builds an `Executor` internally (PEC, ZNE, ...).

Resolve the annotation with `typing.get_type_hints()`, which evaluates stringized (PEP 563) annotations back into concrete types, with a fallback to the raw annotations when a hint cannot be resolved.

Add regression tests in a module that uses
`from __future__ import annotations`.

## Description

`Executor.__init__` reads the executor's return type via `inspect.getfullargspec(executor).annotations`, which returns the raw `__annotations__` dict. When the executor is defined in a module using `from __future__ import annotations` ([PEP 563](https://peps.python.org/pep-0563/)), all annotations are strings, so `_executor_return_type` becomes the string `"float"` instead of the `float` type. The membership checks in `evaluate()` (`... in FloatLike`, etc.) compare against concrete type objects, match nothing, and the method raises:

```python
ValueError: Could not parse executed results from executor with type float.
```

…even though the executor returns a valid float. This also affects measurement/density-matrix executors and any technique that builds an `Executor` internally (PEC, ZNE, …).

## Reproducer

```python
from __future__ import annotations  # the only thing that matters

import cirq
from mitiq import Executor

def executor(circuit) -> float:
    return 1.0

q = cirq.LineQubit(0)
circuit = cirq.Circuit(cirq.X(q))
print(Executor(executor).evaluate(circuit))  # raises ValueError
```

## Fix

Resolve the annotation with `typing.get_type_hints()`, which evaluates stringized annotations back into concrete types using the executor's `__globals__`, with a fallback to the raw annotations if a hint can't be resolved (e.g. an unresolvable forward reference).

## Tests

Adds `mitiq/executor/tests/test_executor_pep563.py` with regression tests in a module that uses `from __future__ import annotations`. They fail on `main` and pass with this change.

---

### License

### DIFF
--- a/mitiq/executor/executor.py
+++ b/mitiq/executor/executor.py
@@ -78,11 +78,14 @@ class Executor:
         # evaluates stringized (PEP 563) annotations back into concrete types,
         # so executors defined in modules using
         # ``from __future__ import annotations`` are still detected correctly.
-        # Fall back to the raw annotations if a hint cannot be resolved (e.g.
-        # an unresolvable forward reference).
+        # It evaluates *all* annotations, so it raises ``NameError`` when any
+        # of them is an unresolvable forward reference (e.g. a name imported
+        # only under ``typing.TYPE_CHECKING``) and ``TypeError`` on some
+        # callables it cannot introspect. In those cases fall back to the raw
+        # (possibly stringized) annotations, preserving the previous behavior.
         try:
             executor_annotation = typing.get_type_hints(executor)
-        except Exception:
+        except (NameError, TypeError):
             executor_annotation = inspect.getfullargspec(executor).annotations
         self._executor_return_type = executor_annotation.get("return")
         self._max_batch_size = max_batch_size
--- a/mitiq/executor/executor.py
+++ b/mitiq/executor/executor.py
@@ -74,7 +74,16 @@ class Executor:
     ) -> None:
         self._executor = executor
 
-        executor_annotation = inspect.getfullargspec(executor).annotations
+        # Resolve the executor's return-type annotation. ``get_type_hints``
+        # evaluates stringized (PEP 563) annotations back into concrete types,
+        # so executors defined in modules using
+        # ``from __future__ import annotations`` are still detected correctly.
+        # Fall back to the raw annotations if a hint cannot be resolved (e.g.
+        # an unresolvable forward reference).
+        try:
+            executor_annotation = typing.get_type_hints(executor)
+        except Exception:
+            executor_annotation = inspect.getfullargspec(executor).annotations
         self._executor_return_type = executor_annotation.get("return")
         self._max_batch_size = max_batch_size
 
--- a/mitiq/executor/tests/test_executor_pep563.py
+++ b/mitiq/executor/tests/test_executor_pep563.py
@@ -11,7 +11,8 @@ an executor is defined, every annotation is stored as a string.
 (e.g. ``"float"`` -> ``float``) so that float / measurement /
 density-matrix executors are detected correctly.
 
-See https://github.com/unitaryfoundation/mitiq/issues/3061 for the originating issue.
+See https://github.com/unitaryfoundation/mitiq/issues/3061 for the
+originating issue.
 """
 
 from __future__ import annotations
--- a/mitiq/executor/tests/test_executor_pep563.py
+++ b/mitiq/executor/tests/test_executor_pep563.py
@@ -1,0 +1,60 @@
+# Copyright (C) Unitary Foundation
+#
+# This source code is licensed under the GPL license (v3) found in the
+# LICENSE file in the root directory of this source tree.
+
+"""Regression tests for executors defined under PEP 563.
+
+When ``from __future__ import annotations`` is active in the module where
+an executor is defined, every annotation is stored as a string.
+``Executor`` must still resolve the return-type annotation
+(e.g. ``"float"`` -> ``float``) so that float / measurement /
+density-matrix executors are detected correctly.
+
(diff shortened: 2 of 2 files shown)


mitiq/executor/executor.py
```diff
@@ -74,7 +74,16 @@ def __init__(
...
-        executor_annotation = inspect.getfullargspec(executor).annotations
+        # Resolve the executor's return-type annotation. ``get_type_hints``
+        # evaluates stringized (PEP 563) annotations back into concrete types,
+        # so executors defined in modules using
+        # ``from __future__ import annotations`` are still detected correctly.
+        # Fall back to the raw annotations if a hint cannot be resolved (e.g.
+        # an unresolvable forward reference).
+        try:
+            executor_annotation = typing.get_type_hints(executor)
+        except Exception:
```
natestemen: What exception gets raised by line 84?

mitiq/executor/tests/test_executor_pep563.py
```diff
@@ -0,0 +1,60 @@
...
+
+"""Regression tests for executors defined under PEP 563.
+
+When ``from __future__ import annotations`` is active in the module where
+an executor is defined, every annotation is stored as a string.
+``Executor`` must still resolve the return-type annotation
+(e.g. ``"float"`` -> ``float``) so that float / measurement /
+density-matrix executors are detected correctly.
+
+See https://github.com/unitaryfoundation/mitiq for the originating issue.
```
natestemen: Add the issue # or remove this line.

ayush268:
The exception can be a `NameError` or `TypeError`, updated with comments in the code as well.

Example code to reproduce:
```python
from __future__ import annotations
from typing import TYPE_CHECKING, get_type_hints

# ---- Case 1: NameError ----
# A name imported ONLY for type-checkers does not exist at runtime, so
# get_type_hints() cannot evaluate the (stringized) annotation.
if TYPE_CHECKING:
    from some_heavy_pkg import CircuitLike  # never imported at runtime

def executor_typecheck_only(circuit: CircuitLike) -> float:
    return 1.0

# ---- Case 2: TypeError ----
# get_type_hints() expects a module/class/method/function. A plain callable
# *object* (an instance with __call__) is not a valid argument.
class CallableExecutor:
    def __call__(self, circuit) -> float:
        return 1.0

callable_obj = CallableExecutor()

get_type_hints(executor_typecheck_only)  # raises NameError
get_type_hints(callable_obj)             # raises TypeError
```

mitiq/executor/tests/test_executor_pep563.py
ayush268: Done

mitiq/executor/executor.py
ayush268: Added detailed comment

ayush268:
Hi @natestemen 
Is anything needed on this change, please let me know and I can push the update

natestemen (approved):
LGTM! Thanks for looking into this.

PR #3064, opened 2026-06-20
Fix nightly docs build

natestemen:
## Summary
- Disambiguate `construct_circuits` / `combine_results` cross-references in the new deb module (#3044) — Sphinx finds 8+ matches and treats the warnings as errors
- Fix stale jupytext link, use `/latest/` for deb guide (not on `/stable/` yet), and ignore `scholar.google.com` in linkcheck (403s from bot detection)

## Test plan
- [x] `make docs` passes locally
- [x] `make linkcheck` passes locally

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/docs/source/conf.py
+++ b/docs/source/conf.py
@@ -241,6 +241,7 @@ linkcheck_ignore = [
     r"https://github.com/unitaryfoundation/mitiq/compare/.*",
     r"https://github.com/unitaryfoundation/mitiq/compare/.*",
     r"https://github.com/unitaryfoundation/mitiq/projects/7",
+    r"https://scholar\.google\.com/.*",
 ]
 
 linkcheck_retries = 3
--- a/docs/source/conf.py
+++ b/docs/source/conf.py
@@ -239,7 +239,6 @@ linkcheck_ignore = [
     r"https://www\.contributor-covenant\.org/.*",
     r"https://www\.sciencedirect\.com/science/article/.*",
     r"https://github.com/unitaryfoundation/mitiq/compare/.*",
-    r"https://github.com/unitaryfoundation/mitiq/compare/.*",
     r"https://github.com/unitaryfoundation/mitiq/projects/7",
     r"https://scholar\.google\.com/.*",
 ]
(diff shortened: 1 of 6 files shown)


natestemen:
In case you're wondering, this was me learning how to use claude to open PRs from the command line. I even asked it to "assign Brian to review", to which it assigned @b-goldsmith and @purva-thakre and then unassigned @purva-thakre.

docs/source/conf.py
```diff
@@ -241,6 +241,7 @@ def setup(app):
     r"https://github.com/unitaryfoundation/mitiq/compare/.*",
     r"https://github.com/unitaryfoundation/mitiq/compare/.*",
```
cosenal: unrelated to the PR, but I see a duplicate here

docs/source/conf.py
```diff
@@ -241,6 +241,7 @@ def setup(app):
...
     r"https://github.com/unitaryfoundation/mitiq/projects/7",
```
cosenal, line 242: interesting, why this very specific link is here, and why does it fail the link check?

cosenal (approved):
only unrelated nits

docs/source/conf.py
natestemen: I think it must be private because I do not have access.

docs/source/conf.py
natestemen: removed in https://github.com/unitaryfoundation/mitiq/pull/3064/commits/b962c4fc8e7afa0d40ba5ce21567ffdd64c46c42.

docs/source/conf.py
cosenal: no, I see a 404 as well. mitiq projects don't even exist, do they? I think it's a legacy link

natestemen: I think we previously had projects and we have sinced turned them off (which you can do in settings).

cosenal: I can only find one occurence of links to projects in the changelog, so it's all good

PR #3077, opened 2026-07-13
Add Calibrator.get_cost_breakdown for per-problem, per-strategy cost

### LINKED ISSUE #2972: Add functionality to fetch more granular cost of calibration
The Calibrator class currently has a method `get_cost` which will tell you how many times the executor (and ideal executor) will be run to run the entire calibration experiment. It would be nice to provide more details here through a new function which would return the breakdown of that total cost by problem and strategy.

alyabouzaid:
Fixes #2972.

`Calibrator.get_cost` only reports total noisy/ideal execution counts. This adds `get_cost_breakdown()`, which itemizes that total by benchmark problem and strategy: one row per (problem, strategy) pair, plus a baseline row per problem covering the unmitigated execution (and the ideal execution for custom problems). `get_cost` now sums the breakdown, so the totals stay consistent by construction; its return value is unchanged.

Verified by running `pytest mitiq/calibration` (60 passed, includes two new tests), `make check-format` and `make check-types` (clean), and by exercising it directly: default ZNE settings still report `{'noisy_executions': 100, 'ideal_executions': 0}` with a 36-row breakdown that sums to those totals, and a custom problem with an ideal executor reports the ideal execution on its baseline row.

### DIFF
--- a/mitiq/calibration/calibrator.py
+++ b/mitiq/calibration/calibrator.py
@@ -7,7 +7,7 @@ import warnings
 from collections.abc import Callable, Sequence
 from enum import Enum
 from operator import itemgetter
-from typing import cast
+from typing import Any, cast
 
 import cirq
 import numpy as np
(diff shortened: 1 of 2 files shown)


natestemen (approved):
Thanks for working on this!

PR #3089, opened 2026-07-22
Add optional return_info when inserting DDD sequences

### LINKED ISSUE #2882: Add logging option when inserting DDD sequences
When applying DDD, it's currently difficult for users to tell whether the pass actually inserted any decoupling sequences. There are several ways DDD can silently "do nothing". For example, if there are no idle windows available, no DDD sequences are inserted and the user receives no indication that this happened.

It would be helpful to provide some form of optional logging, warnings, or structured return information so users can confirm whether DDD ran as expected. Even minimal feedback (e.g., number of idle windows found, number of sequences inserted, or a note when nothing is applied) would improve usability and reduce confusion.

**Two ideas**:

```py
transformed_circuit, ddd_info = ddd.construct_circuit(circuit, return_info=True)

print(ddd_info.num_idle_windows)
print(ddd_info.num_sequences_inserted)
```

```py
ddd.construct_circuit(circuit, verbose=True)
# prints:
# "DDD: found 5 idle windows; inserted 5 sequences"
```

sankalpsthakur:
Fixes #2882

## Description

Add opt-in insertion metadata with `return_info=True`. By default, `insert_ddd_sequences` returns one circuit and `construct_circuits` returns a list of circuits. The opt-in forms return `(circuit, DDDInfo)` and `(circuits, tuple_of_DDDInfo)` respectively.

The Cirq implementation returns the circuit and immutable metadata together; frontend conversion preserves the extra return value. No caller-facing mutable output parameter is used. Literal and boolean fallback overloads describe the supported return forms.

`construct_circuits` logs one INFO summary per trial on `mitiq.ddd`. Low-level insertion does not log, avoiding duplicate output.

This description correction adds no new validation claim; the existing review replies and CI results remain the test record.

---

### License

### AI disclosure

AI tools helped draft code and the original description. ChatGPT assisted with this description correction; the existing license declaration is retained, not newly signed.

### DIFF
--- a/docs/source/guide/ddd-4-low-level.md
+++ b/docs/source/guide/ddd-4-low-level.md
@@ -122,6 +122,24 @@ circuit_with_ddd = ddd.insert_ddd_sequences(circuit, rule=xyxy_rule)
 circuit_with_ddd
 ```
 
+### Checking whether DDD sequences were inserted
+
+By default, {func}`.insert_ddd_sequences()` returns only the modified circuit.
+If the input has no sufficiently long idle windows, DDD may leave the circuit unchanged.
+To inspect what happened without changing the default API, pass ``return_info=True``:
+
+```{code-cell} ipython3
+circuit_with_ddd, ddd_info = ddd.insert_ddd_sequences(
+    circuit, rule=xyxy_rule, return_info=True
+)
+print(ddd_info.num_idle_windows)
+print(ddd_info.num_sequences_inserted)
+print(ddd_info.idle_window_lengths)
+```
+
+The returned {class}`~mitiq.ddd.insertion.DDDInfo` reports how many idle windows were found,
+how many non-empty DDD sequences were inserted, and the length of each candidate window.
+
 ```{note}
 In principle, the function {func}`.insert_ddd_sequences()` is all one needs to apply DDD.
 Indeed, since in DDD there is not a final post-processing step, one can simply insert DDD sequences before running the circuit on a noisy backend.
--- a/docs/source/guide/ddd-4-low-level.md
+++ b/docs/source/guide/ddd-4-low-level.md
@@ -140,6 +140,14 @@ print(ddd_info.idle_window_lengths)
 The returned {class}`~mitiq.ddd.insertion.DDDInfo` reports how many idle windows were found,
 how many non-empty DDD sequences were inserted, and the length of each candidate window.
 
+{func}`.ddd.construct_circuits` logs the same counts at INFO on the ``mitiq.ddd`` logger
+(one line per trial). Enable it with:
+
+```{code-block} python
+import logging
+logging.getLogger("mitiq.ddd").setLevel(logging.INFO)
+```
+
 ```{note}
 In principle, the function {func}`.insert_ddd_sequences()` is all one needs to apply DDD.
 Indeed, since in DDD there is not a final post-processing step, one can simply insert DDD sequences before running the circuit on a noisy backend.
--- a/mitiq/ddd/insertion.py
+++ b/mitiq/ddd/insertion.py
@@ -211,12 +211,15 @@ def _insert_ddd_sequences_with_info(
 ) -> tuple[QPROGRAM, DDDInfo]:
     """Insert DDD sequences and return both the converted circuit and info.
 
-    ``@accept_qprogram_and_validate`` only supports Cirq→Cirq (or one-to-many)
-    returns, so it cannot propagate :class:`DDDInfo` on its own. We apply that
-    decorator to a one-shot circuit transformer that records info via a local
-    nonlocal binding, then return a normal ``(circuit, info)`` tuple. No
-    mutable out-parameter is part of any function signature.
+    Cirq inputs go straight through :func:`_apply_ddd_sequences`, which returns
+    a normal ``(circuit, DDDInfo)`` tuple. Other frontends still need
+    ``@accept_qprogram_and_validate`` for conversion, and that decorator can
+    only return a ``QPROGRAM``, so info is recorded in a local binding and
+    returned as a normal tuple. No caller-facing out-parameter.
     """
+    if isinstance(circuit, Circuit):
+        return _apply_ddd_sequences(circuit, rule)
+
     info: DDDInfo | None = None
 
     def _insert_and_record(circ: Circuit) -> Circuit:
@@ -227,7 +230,6 @@ def _insert_ddd_sequences_with_info(
     circuit_with_ddd = accept_qprogram_and_validate(_insert_and_record)(
         circuit
     )
-    # _insert_and_record always assigns info before returning.
     assert info is not None
     return circuit_with_ddd, info
 
--- a/mitiq/ddd/insertion.py
+++ b/mitiq/ddd/insertion.py
@@ -6,6 +6,7 @@
 """Tools to determine slack windows in circuits and to insert DDD sequences."""
 
 from collections.abc import Callable
+from dataclasses import dataclass
 
 import cirq
 import numpy as np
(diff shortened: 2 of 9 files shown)


natestemen (commented):
WDYT of adding the logging to ddd.construct_circuits?

sankalpsthakur:
done — construct_circuits now takes the same return_info flag and returns (circuits, list[DDDInfo]), one per trial.

sankalpsthakur:
fixed the mypy failure — insert_ddd_sequences isn't overload-typed on return_info, so the unpack was inferring a contaminated type for info. cast to the known runtime shape, no behavior change. mypy clean on all 113 files locally now.

mitiq/ddd/insertion.py
```diff
@@ -100,18 +120,32 @@ def insert_ddd_sequences(
...
+        return circuit_with_ddd, info_out[0]
+    return circuit_with_ddd
 
 
 @accept_qprogram_and_validate
 def _insert_ddd_sequences(
     circuit: Circuit,
     rule: Callable[[int], QPROGRAM],
+    _info_out: list[DDDInfo] | None = None,
 ) -> Circuit:
```
natestemen: seems like a code smell to me that you have to pass a mutable list in to get the info out.

natestemen (changes requested):
Are there any workaround for the mutable list thing?

mitiq/ddd/insertion.py
sankalpsthakur: yeah fair call. dropped the out-param.

the Cirq helper now always returns `(circuit, DDDInfo)`. the decorated path still only returns the circuit (that decorator can't carry extra values through conversion), and when `return_info=True` we wrap a one-shot converter and return a normal tuple. public API is unchanged.

sankalpsthakur (commented):
addressed the mutable list out-param. Cirq insertion returns a real (circuit, info) tuple; conversion stays circuit-only under the decorator, and return_info=True comes back as a normal tuple with no smuggling.

sankalpsthakur:
@natestemen dropped the mutable _info_out list. Core path returns (Circuit, DDDInfo); public insert_ddd_sequences stays the same. 5cc64f9.

sankalpsthakur:
@natestemen dropped the mutable out-param. return_info=True returns a normal (circuit, DDDInfo) tuple. Ready when you can take another look.

sankalpsthakur:
@natestemen agreed on the mutable list. Dropped the out-param.

Public API is now `return_info=True -> (circuit, DDDInfo)`. Cirq goes straight through a tuple-returning helper. Other frontends still need `@accept_qprogram_and_validate`, which can only return a circuit, so info is captured locally there and returned as a normal tuple. No caller-facing mutable out-param.

On logging in `construct_circuits`: went with the same `return_info` flag instead, so callers get structured `DDDInfo` per trial and can log or assert as they like. Happy to add logging too if you still want it.

Ready for another look when you have a minute.

sankalpsthakur:
Dropped the mutable list out-param for return_info in favor of a proper return value. Let me know if you'd still like logging wired into construct_circuits as a follow-up.

sankalpsthakur:
Dropped the leftover local binding too. return_info now comes back through the converter as a normal extra value, so every frontend gets a real (circuit, DDDInfo) tuple.

sankalpsthakur:
@natestemen workaround for the mutable list is a normal tuple (no out-param, no list default). `construct_circuits(..., return_info=True)` now returns `tuple[DDDInfo, ...]` rather than a list.

Logging is on `construct_circuits` only: INFO on `mitiq.ddd`, one line per trial (`DDD: found N idle windows; inserted M sequences`). `insert_ddd_sequences` stays quiet so there is no double noise.

Ready for another look.

sankalpsthakur:
@natestemen good call on the mutable list. The tip (`3493313`) no longer uses an out-param — Cirq helper returns `(circuit, DDDInfo)`, and `return_info=True` returns a normal tuple on the public APIs. Ready for another look when you have time.

mitiq/ddd/insertion.py
```diff
@@ -163,4 +170,100 @@ def cirq_rule(slack_length: int) -> Circuit:
...
+) -> QPROGRAM: ...
+
+
+@overload
+def insert_ddd_sequences(
+    circuit: QPROGRAM,
+    rule: Callable[[int], QPROGRAM],
+    *,
+    return_info: Literal[True],
+) -> tuple[QPROGRAM, DDDInfo]: ...
```
natestemen, line 238: [P2] Please add a fallback overload accepting `return_info: bool` here and in `construct_circuits`. The two literal overloads reject a boolean forwarded from configuration or another function, even though the implementation supports it at runtime. The implementation signature is not considered when resolving overloaded calls.

Concretely, keep the existing literal overloads and add this third overload immediately before the implementation in `insertion.py`:

```python
@overload
def insert_ddd_sequences(
    circuit: QPROGRAM,
    rule: Callable[[int], QPROGRAM],
    *,
    return_info: bool,
) -> QPROGRAM | tuple[QPROGRAM, DDDInfo]: ...
```

Likewise, immediately before the implementation in `ddd.py`, add:

```python
@overload
def construct_circuits(
    circuit: QPROGRAM,
    rule: Callable[[int], QPROGRAM],
    rule_args: dict[str, Any] | None = None,
    num_trials: int = 1,
    *,
    return_info: bool,
) -> list[QPROGRAM] | tuple[list[QPROGRAM], tuple[DDDInfo, ...]]: ...
```

To verify, run mypy against the following small caller. Both calls currently fail with `No overload variant ... matches`; both should type-check after the change. Literal True/False calls should retain their precise return types.

```python
from mitiq import QPROGRAM
from mitiq.ddd import construct_circuits, insert_ddd_sequences
from mitiq.ddd.rules import xx

def apply(circuit: QPROGRAM, include_info: bool) -> None:
    insert_ddd_sequences(circuit, xx, return_info=include_info)
    construct_circuits(circuit, xx, return_info=include_info)
```

docs/source/guide/ddd-4-low-level.md
```diff
@@ -122,6 +122,32 @@ circuit_with_ddd = ddd.insert_ddd_sequences(circuit, rule=xyxy_rule)
...
+
+The returned {class}`~mitiq.ddd.insertion.DDDInfo` reports how many idle windows were found,
+how many non-empty DDD sequences were inserted, and the length of each candidate window.
+
+{func}`.ddd.construct_circuits` logs the same counts at INFO on the ``mitiq.ddd`` logger
+(one line per trial). Enable it with:
+
+```{code-block} python
+import logging
+logging.getLogger("mitiq.ddd").setLevel(logging.INFO)
```
natestemen, line 150: [P3] Please configure a handler as well as the logger level. In a fresh Python process, the current snippet leaves `construct_circuits` silent because there is no handler that emits its INFO records. This replacement adds a default console handler and enables INFO for the DDD logger:

```suggestion
import logging

logging.basicConfig()
logging.getLogger("mitiq.ddd").setLevel(logging.INFO)
```

To verify in a fresh process, run the snippet followed by:

```python
import cirq
from mitiq import ddd

q = cirq.LineQubit(0)
circuit = cirq.Circuit(cirq.H(q), cirq.I(q), cirq.I(q), cirq.H(q))
ddd.construct_circuits(circuit, rule=ddd.rules.xx)
```

The current example emits nothing; with the suggested handler it emits `INFO:mitiq.ddd:DDD: found 1 idle windows; inserted 1 sequences`.

natestemen (changes requested):
Thanks for the changes, and apologies for the slow response.

Please address the two inline items: accept nonliteral boolean values in the public overloads, and make the logging setup example work in a fresh process.

Review prepared with Codex.

mitiq/ddd/insertion.py
sankalpsthakur: Added both `bool` fallback overloads after the literal variants in 5013e85. Exact-head CI run 35256124373 now passes mypy, style checks and all six OS/Python test jobs.

docs/source/guide/ddd-4-low-level.md
sankalpsthakur: Added basicConfig() in 5013e85. The logging snippet emits INFO in a fresh process; the full Cirq example was not run locally.

PR #3090, opened 2026-07-22
Document batched circuit execution in the user guide

### LINKED ISSUE #2537: Add batched circuit to documentation
## Issue Description
We do not report on batched quantum circuits. Batching has been used by @vprusso for example in Ref. [1]. The Research folder notebook related to that publication does not contain a batched example though [2]. 

## Proposed Solution
Add a simple batching snippet and explanation in the documentation, such as in the user guide (could even be in core concepts, given how important it is) or as a new tutorial. 

## Additional References
[1] https://arxiv.org/abs/2210.07194
[2] https://github.com/unitaryfund/research/tree/main/qem-on-hardware

sankalpsthakur:
Fixes #2537

## Summary
Adds a clear, practical treatment of **batched quantum circuit execution** in the user guide. Error mitigation generates many circuits; batching is how those circuits are submitted efficiently on hardware.

### Changes
- **`executors.md`**: Expand the batched-execution section with
  - why batching matters for QEM (round-trips vs quantum resource counts)
  - detection via return-type annotations
  - float-valued serial vs batched comparison (`calls_to_executor`)
  - `max_batch_size` splitting example
  - tip for wiring provider batch/job APIs
  - cite to Russo et al. arXiv:2210.07194 (same ref as the issue)
- **`frontends-backends.md`**: Replace pointer-only note with a minimal batch snippet and link to the full guide
- **`resource-requirements.md`**: Clarify that batching reduces wall-clock/API cost, not circuit/gate counts

Docs-only; no code changes.

## Why this matters

Error-mitigation workflows routinely expand one logical circuit into many sampled variants. Without batching guidance, users pay unnecessary API round-trips and wall-clock time even when gate counts are unchanged. Batching is the main lever for making QEM practical on real hardware backends.

## Related

- Request: #2537
- Reference cited in issue: Russo et al. [arXiv:2210.07194](https://arxiv.org/abs/2210.07194)

---

Assisted-by: Codex (OpenAI)

### DIFF
--- a/docs/source/guide/executors.md
+++ b/docs/source/guide/executors.md
@@ -121,10 +121,18 @@ print("\nQuantum results:\n", *executor.quantum_results, sep="\n")
 
 Notice in the above output that the executor has been called once for each circuit it has executed. This is because `mitiq_cirq.compute_density_matrix` inputs one circuit and outputs one `QuantumResult`.
 
-Several quantum computing services allow running a sequence, or "batch," of circuits at once. This is important for error mitigation when running many circuits to speed up the computation.
+Several quantum computing services allow running a sequence, or "batch," of circuits at once.
+This is especially important for error mitigation: techniques such as ZNE, PEC, and LRE typically generate **many** noise-scaled or sampled circuits, and batching reduces the number of backend round-trips (and often wall-clock time) without changing the quantum resource counts themselves.
+See also [Resource Requirements](resource-requirements.md).
 
-To define a batched executor, annotate it with `Sequence[T]`, `list[T]`, `tuple[T]`, or `Iterable[T]` where `T` is a `QuantumResult`.
-Here is an example:
+To define a batched executor, the function must:
+
+1. Accept a **sequence** of circuits (e.g. `list[QPROGRAM]`), and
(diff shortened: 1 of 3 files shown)


docs/source/guide/executors.md
```diff
@@ -156,6 +164,57 @@ print("\nExecuted circuits:\n", *batched_executor.executed_circuits, sep="\n")
...
+def serial_float_executor(circuit: cirq.Circuit) -> float:
+    """Serial executor: one circuit → one float."""
+    rho = mitiq_cirq.compute_density_matrix(circuit)
+    # Population of |0⟩ as a simple float-valued result.
+    return float(np.real(rho[0, 0]))
+
+
+def batched_float_executor(circuits: list[cirq.Circuit]) -> list[float]:
+    """Batched executor: many circuits → many floats in one call."""
+    return [serial_float_executor(c) for c in circuits]
```
natestemen, line 185: We should make a comment somewhere that this method of creating a batched executor doesn't actually make it batched since we are just repeatedly calling the serial one, but this is just for demonstration purposes.

natestemen (approved):
Really great change here, thank you! Just one nit and then ready to merge.

docs/source/guide/executors.md
sankalpsthakur: added a comment above the return noting it's just looping the serial call for demo purposes, not real batching.

PR #3111, opened 2026-08-30
Fix fidelity between two density matrices (#2129)

tzh476:
### Description

`fidelity` computes `|tr(sqrt(sigma) @ rho @ sqrt(sigma))|` for the density-matrix case, which is missing the inner square root of the Uhlmann formula. That expression reduces to `tr(sigma @ rho)`, so **a state does not have fidelity 1 with itself**:

| case | before | correct |
|---|---|---|
| `F(rho, rho)`, 1-qubit maximally mixed | **0.500** | 1 |
| `F(rho, rho)`, 2-qubit random mixed | **0.530** | 1 |
| `F(rho, rho)`, 3-qubit random mixed | **0.237** | 1 |
| `F(rho_0, I/2)` | 0.500 | 0.500 ✓ |

For equal states the old expression returns the **purity**. Pure states have purity 1, so they were unaffected — which is why this went unnoticed. Shadow tomography reconstructs *mixed* states, and this branch is what the docs use to report reconstruction fidelity:

- `docs/source/guide/shadows-1-intro.md:355` — `fidelity(ghz_true, est_corrs["reconstructed_state"])`
- `docs/source/examples/shadows_tutorial.md:261` — `fidelity(rho_true, rho_shadow)`

The branch was also inconsistent with the function's *own* vector branches, which use the squared convention (`<psi|rho|psi>`). So the result depended on whether a caller happened to hold a pure state as a vector or as a density matrix.

### After the fix

All four dimension branches agree with `cirq.fidelity` to within **1.2e-14** (1–3 qubits, self-fidelity and distinct mixed states, plus all three vector branches). Worth noting because #2129 already observed that `cirq.fidelity` exists and suggested it as the reference for this function — this brings mitiq's implementation into agreement with it while keeping the existing signature and the vector branches untouched.

### Tests

#2129 also notes this function "has some lines flagged by pytest as uncovered" — the three uncovered branches were `matrix/vector`, `matrix/matrix`, and the `ValueError`. Added:

- **self-fidelity axiom** on maximally mixed and random mixed states, 1 to 3 qubits — this is the test that fails on `main`
- known analytic values: `F(|0><0|, |1><1|) = 0`, `F(rho, rho) = 1`, `F(|0><0|, I/2) = 0.5`
- **cross-representation agreement**: one pure state passed as a vector and as its density matrix must give the same answer
- the `Invalid input dimensions` path

Verified that `test_fidelity_matrix_matrix_self_is_one` **fails** on the previous implementation while the other new tests pass either way — so the control isolates exactly the defect being fixed.

### Verified locally

```
pytest mitiq/experimental/shadows/     45 passed
ruff check                             All checks passed!
ruff format --check                    8 files already formatted
```

Assisted-by: Claude (Anthropic). Every number above was measured, and the fix is cross-checked against an independent Uhlmann implementation and against `cirq.fidelity`, per the AI use policy in `CONTRIBUTING.md`.

### DIFF
--- a/mitiq/experimental/shadows/shadows_utils.py
+++ b/mitiq/experimental/shadows/shadows_utils.py
@@ -11,9 +11,9 @@
 
 from collections.abc import Generator
 
+import cirq
 import numpy as np
 import numpy.typing as npt
-from scipy.linalg import sqrtm
 
 import mitiq
 
@@ -78,9 +78,10 @@ def fidelity(
     """
     Calculate the fidelity between two states.
 
-    Uses the squared convention, so the fidelity of a state with itself is
-    1 and the fidelity between a pure state :math:`|\\psi\\rangle` and a
-    density matrix is :math:`\\langle\\psi|\\rho|\\psi\\rangle`.
+    Delegates to :func:`cirq.fidelity`, which uses the squared convention: the
+    fidelity of a state with itself is 1, and the fidelity between a pure state
+    :math:`|\\psi\\rangle` and a density matrix is
+    :math:`\\langle\\psi|\\rho|\\psi\\rangle`.
 
     Args:
         sigma: A state in terms of square matrix or vector.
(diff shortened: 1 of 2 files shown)


natestemen (changes requested):
The formula looks right, but can we delegate this to `cirq.fidelity`? That seems to be the cleanup #2129 was asking for, and avoids maintaining a second implementation.

We still need a small wrapper because calibrated shadow reconstructions are unnormalized PTM vectors, so validation has to stay off. The explicit `qid_shape` also lets Cirq distinguish density matrices from state tensors.

```python
if sigma.ndim not in (1, 2) or rho.ndim not in (1, 2):
    raise ValueError("Invalid input dimensions")

return float(
    cirq.fidelity(
        sigma,
        rho,
        qid_shape=(sigma.shape[0],),
        validate=False,
    )
)
```

The new regression tests can stay.

natestemen (approved):
Thank you for catching this, and for the PR!

Issue #2802, opened 2025-07-07
Replace conversions with qbraid transpiler


natestemen:
Offload as many of the `mitiq.interface` modules to `qbraid.transpile`. We will still need to keep around much of the infra to check if the circuits are "mitigable", but the explicit conversions should be simplified by qbraid (and maintained by them 😄).

tzh476:
I have been working inside both sides of this boundary the last few days (mitiq's `interface/` and qBraid's pyquil transpiler), so here are some measurements that might help scope this. All numbers are from `qbraid 0.13.0.dev` and mitiq `main`; happy to be told any of it is already known.

**1. Only pyquil has actually migrated so far.**

| module | lines | references `qbraid` |
|---|---|---|
| `mitiq_pyquil/conversions.py` | 63 | **2** |
| `mitiq_qiskit/conversions.py` | 302 | 0 |
| `mitiq_braket/conversions.py` | 499 | 0 |
| `mitiq_qibo/conversions.py` | 388 | 0 |
| `mitiq_pennylane/conversions.py` | 93 | 0 |
| `mitiq_openqasm/conversions.py` | 36 | 0 |

pyquil is the shape this issue is aiming at — all four functions are thin wrappers delegating straight to `qbraid.transpiler.conversions`. So braket (499) and qibo (388) are where the bulk of the win is.

**2. Every frontend mitiq supports is reachable through qbraid, but not always directly.** `ConversionGraph` registers 32 edges, and direct `cirq <-> X` edges exist only for braket, qasm2 and pyquil — there is **no direct edge either way for qiskit or qibo**. I initially read that as a blocker, then tested it and was wrong: `transpile` finds multi-hop paths and all seven targets succeed.

```
cirq -> qiskit     OK  QuantumCircuit      cirq -> qibo    OK  Circuit
cirq -> braket     OK  Circuit             cirq -> qasm2   OK  str
cirq -> pennylane  OK  QuantumTape         cirq -> qasm3   OK  str
cirq -> pyquil     OK  Program
```

**3. Round-trip fidelity is good enough to migrate on.** `cirq -> X -> cirq`, phase-invariant max deviation of the unitary:

| target | Bell | rotations | 3q Toffoli |
|---|---|---|---|
| qiskit | 0.0e+00 | 6.8e-11 | 0.0e+00 |
| braket | 0.0e+00 | 0.0e+00 | 0.0e+00 |
[...]

PR #3112, opened 2026-08-30
Support non-LineQubit circuits in to_braket

tzh476:
### Description

`to_braket` reads `op.qubits[0].x`, an attribute that only exists on `cirq.LineQubit`, so every other qubit type raises:

```
AttributeError: 'NamedQubit' object has no attribute 'x'
AttributeError: 'GridQubit'  object has no attribute 'x'
```

This is reachable through the public API — `convert_from_mitiq(circuit, "braket")` on a NamedQubit circuit surfaces it as a `CircuitConversionError` wrapping that `AttributeError`. It also means output from **`cirq.testing.random_circuit`** cannot be converted at all, since that helper builds circuits on `NamedQubit`:

```python
>>> circuit = cirq.testing.random_circuit(3, n_moments=6, op_density=0.8, random_state=5)
>>> to_braket(circuit)
AttributeError: 'NamedQubit' object has no attribute 'x'
```

### Fix

Braket addresses qubits by integer index, so `to_braket` now relabels to `LineQubit` in sorted order at the entry point. Doing it there rather than at the three separate `cast(LineQubit, ...).x` sites keeps it to one place, and the cast annotations become true rather than aspirational.

**Relabelling is skipped when every qubit is already a `LineQubit`**, which matters: a circuit on `LineQubit(10..12)` must still convert to Braket targets `10..12`, because `test_to_from_braket_non_zero_qubit` depends on the indices carrying through. `test_to_braket_preserves_line_qubit_indices` now pins that explicitly.

### Verified

Round-trip `cirq -> braket -> cirq`, phase-invariant max deviation of the unitary:

| qubit type | error |
|---|---|
| `LineQubit.range(3)` | 0.0e+00 |
| `LineQubit(10), (11), (12)` | 0.0e+00 |
| `NamedQubit.range(3)` | 0.0e+00 |
| `GridQubit.rect(1, 3)` | 0.0e+00 |
| `random_circuit(3, 6)` | 0.0e+00 |

```
pytest mitiq/interface/mitiq_braket/    34 passed
ruff check                             All checks passed!
ruff format --check                    2 files already formatted
```

Control: on the previous implementation **3 of the new tests fail** (NamedQubit, GridQubit, random_circuit) while the two `LineQubit` parametrizations pass either way — so the tests cover the new capability without masking the existing behaviour.

### How this came up

While measuring qbraid's transpiler coverage for #2802 I compared mitiq's braket converter against `qbraid`'s on a set of circuits, and `random_circuit` was the one case where they diverged: qbraid round-tripped it exactly, mitiq raised. Worth noting for that discussion that qbraid's converter handles `NamedQubit` but **fails differently on `GridQubit`** (`ValueError: Operator qubit count 2 must be equal to size`), so after this change mitiq's own converter accepts strictly more input than the upstream one.

Assisted-by: Claude (Anthropic). Every number above was measured, per the AI use policy in `CONTRIBUTING.md`.

### DIFF
--- a/mitiq/interface/mitiq_braket/conversions.py
+++ b/mitiq/interface/mitiq_braket/conversions.py
@@ -2,6 +2,7 @@
 #
 # This source code is licensed under the GPL license (v3) found in the
 # LICENSE file in the root directory of this source tree.
+import itertools
 from typing import cast
 from warnings import warn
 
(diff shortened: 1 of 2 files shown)


mitiq/interface/mitiq_braket/conversions.py
```diff
@@ -53,9 +54,30 @@ def from_braket(circuit: BKCircuit) -> Circuit:
...
+    if other_qubits:
+        taken = {
+            qubit.x
+            for qubit in circuit.all_qubits()
+            if isinstance(qubit, LineQubit)
+        }
+        free = (index for index in itertools.count() if index not in taken)
+        circuit = circuit.transform_qubits(
+            {qubit: LineQubit(next(free)) for qubit in other_qubits}
+        )
```
natestemen, line 79: nice!

natestemen (commented):
I like your solution, but it implies that a user would care that line qubit number should match the braket qubit numbering, which I'm not convinced is the case. Do you have a reason to need that? If not, I think we could simplify the code a bit to just use line qubits 0 to n for everything.

PR #3120, opened 2026-08-31
Fix folding gate keys "rx", "ry", "rz" being silently ignored

alyabouzaid:
### Problem

The gate-key table documented for `fold_all(exclude=...)` and for the
`fidelities` argument of the local folding functions lists `"rx"`, `"ry"` and
`"rz"`, but those three keys never match anything. `_cirq_gates_to_string_keys`
maps them to `cirq.rx`/`cirq.ry`/`cirq.rz`, which are factory *functions*, not
gates, so they are never equal to the `Rx`/`Ry`/`Rz` gates in a circuit. Both
features silently no-op, with no error or warning:

```python
q = cirq.LineQubit.range(2)
c = cirq.Circuit([cirq.rx(0.3).on(q[0]), cirq.H(q[1]), cirq.CNOT(*q)])

fold_all(c, 3.0, exclude={"rx"})    # the Rx gate is folded anyway
_create_weight_mask(c, {"rx": 1.0}) # [0.99, 0.99, 0.9801], the Rx weight is the default
```

### Fix

Rotations are parametrized, so a single gate instance cannot stand for them.
Identify them by gate type (`ops.Rx`/`ops.Ry`/`ops.Rz`) in a separate map, match
`exclude` string keys against it, and look up weights through a small helper that
falls back to the gate type. Behaviour for every other gate key is unchanged.

### Verification

- Two new regression tests (`test_fold_all_exclude_rotations_with_strings`,
  `test_create_weight_mask_with_rotation_fidelities`); both fail on `main` and
  pass with this change.
- `pytest mitiq` passes (2570 passed, 2 xfailed; `mitiq_pyquil` skipped, needs a
  local QVM), `ruff check` / `ruff format --check` clean, `mypy mitiq` clean.
- Exercised end to end: `fold_gates_at_random(c, scale_factor=2.0, fidelities={"rx": 1.0, "rz": 1.0})`
  now folds only the CNOTs, leaves the rotations alone, and preserves the circuit
  unitary up to global phase.

### DIFF
--- a/mitiq/zne/scaling/folding.py
+++ b/mitiq/zne/scaling/folding.py
@@ -33,9 +33,6 @@ _cirq_gates_to_string_keys = {
     ops.I: "I",
     ops.S: "S",
     ops.T: "T",
-    ops.rx: "rx",
-    ops.ry: "ry",
-    ops.rz: "rz",
     ops.CNOT: "CNOT",
     ops.CZ: "CZ",
     ops.SWAP: "SWAP",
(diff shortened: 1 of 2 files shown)


natestemen (changes requested):
Thanks for fixing this. Could we use Cirq’s `GateFamily` instead of maintaining separate instance and type maps? It supports both exact-instance and type-based matching, so `exclude` and `fidelities` could share one path. Something along these lines:

```python
_GATE_FAMILIES = {
    "CZ": ops.GateFamily(ops.CZ, ignore_global_phase=False),
    "CNOT": ops.GateFamily(ops.CNOT, ignore_global_phase=False),
    "ISWAP": ops.GateFamily(ops.ISWAP, ignore_global_phase=False),
    "rx": ops.GateFamily(ops.Rx),
    "ry": ops.GateFamily(ops.Ry),
    "rz": ops.GateFamily(ops.Rz),
}
```

This would also make the intended behavior for powered gate families explicit.

natestemen (approved):
Nice clean up! Thanks for fixing this.

PR #3122, opened 2026-09-01
Fix AttributeError when raw OpenQASM strings are passed to error mitigation techniques

### LINKED ISSUE #3121: Raw OpenQASM strings crash ZNE/DDD/LRE/PT/PEC: AttributeError: 'str' object has no attribute '__module__'
## Pre-Report Checklist

- [x] I checked to make sure that this bug has not already been reported

## Issue Description

The OpenQASM 3 frontend (added in v0.49.0, #2842) is listed as a supported frontend in the
README and appears in `mitiq.SUPPORTED_PROGRAM_TYPES` in the user guide. However, passing a
raw OpenQASM string directly to most technique entrypoints crashes with
`AttributeError: 'str' object has no attribute '__module__'` before any conversion happens.

**Root cause:** `convert_to_mitiq` wraps bare `str` inputs in `QasmStringType` (which carries
a `__module__` marker), so direct conversion works. But `accept_qprogram_and_validate.new_function`
in `mitiq/interface/conversions.py` reads `circuit.__module__` on the raw incoming argument
*before* conversion — that wrap is not applied there. Any technique whose circuit-processing
step goes through this decorator crashes on a plain string.

**Verified affected on `main` (3e1833d), for both `OPENQASM 3.0;` and `OPENQASM 2.0;` inputs
— 11 entrypoints:**

`zne.execute_with_zne`, `zne.construct_circuits`, `ddd.execute_with_ddd`,
`ddd.construct_circuits`, `lre.execute_with_lre`, `lre.construct_circuits`,
`pt.generate_pauli_twirl_variants`, `pt.twirl_CNOT_gates`, `pt.twirl_CZ_gates`,
`pec.execute_with_pec`, `pec.construct_circuits`

**Not affected:** `cdr.execute_with_cdr` works on raw QASM strings (it does not route through
this decorator). PEC's existing tests pass only because they pre-convert with
`convert_to_mitiq` before calling — so CI never exercises the raw-string path.

**I have a fix ready and will open a PR shortly:** it applies the same `QasmStringType` wrap
at the decorator's entry (mirroring the existing idiom in `convert_to_mitiq`), plus a 26-test
suite covering all 11 entrypoints in both dialects. The new tests fail 26/26 on current
`main` and pass 26/26 with the patch; the full suite shows no regressions
(2594 passed, 2 xfailed).

### Code Snippet

```python
from mitiq import zne

qasm = """OPENQASM 3.0;
include "stdgates.inc";
qubit[1] q;
h q[0];
h q[0];
"""

def executor(circuit):
    return 0.5

zne.execute_with_zne(qasm, executor)
# Expected: a float (the extrapolated value)
# Actual:   AttributeError
```

### Error Output

```
  File "mitiq/zne/inference.py", line 615, in _generate_circuits
    to_run.append(scale_noise(circuit, scale_factor))
  File "mitiq/interface/conversions.py", line 332, in new_function
    if "qiskit" in circuit.__module__:
AttributeError: 'str' object has no attribute '__module__'
```

## Environment Context

```
Mitiq: A Python toolkit for implementing error mitigation on quantum computers
==============================================================================
Authored by: Mitiq team, 2020 & later (https://github.com/unitaryfoundation/mitiq)

Mitiq Version:  1.0.0

Core Dependencies
-----------------
Cirq Version:   1.6.1
NumPy Version:  2.2.6
SciPy Version:  1.17.1

Optional Dependencies
---------------------
PyQuil Version: 4.17.0
Qiskit Version: 2.1.2
Braket Version: 1.124.0

Python Version: 3.11.9
Platform Info:  Linux (x86_64)
```

Editable install from `main` @ 3e1833d.

---

*Side note, out of scope here:* `experimental.trex.execute_with_trex` fails on converted
OpenQASM input for an unrelated reason (`KeyError: cirq.LineQubit(0)` from a hardcoded qubit
in `trex.py`) — I'll file that separately.

v4xsh:
## Description

Fixes #3121.

Raw OpenQASM strings (both `OPENQASM 3.0;` and `OPENQASM 2.0;`) crash 11 technique
entrypoints across ZNE, DDD, LRE, PT, and PEC with
`AttributeError: 'str' object has no attribute '__module__'`.

**Root cause:** `convert_to_mitiq` wraps bare `str` inputs in `QasmStringType` (which
carries a `__module__` marker), but `accept_qprogram_and_validate.new_function` in
`mitiq/interface/conversions.py` reads `circuit.__module__` on the raw incoming argument
*before* any conversion — so every technique whose circuit-processing step goes through
this decorator crashes on a plain string, while direct `convert_to_mitiq` calls work.
PEC's existing tests pass only because they pre-convert to Cirq before calling, so CI
never exercised the raw-string path.

**The fix** adds a small `_wrap_openqasm_string` helper and calls it at the decorator's
entry, mirroring the exact idiom already used in `convert_to_mitiq`. The wrap propagates
through the pipeline, so both the input-side `__module__` read and the return conversion
work: `construct_circuits`-style calls on a QASM string return `QasmStringType`
instances whose contents re-parse as valid QASM. A helper function (rather than an
inline check) is needed because the `isinstance(circuit, str)` narrowing otherwise
propagates past the branch and breaks mypy on the `.copy()` call in the qiskit
sub-branch.

Before, on `main` (3e1833d):

```
>>> zne.execute_with_zne(qasm_string, executor)
AttributeError: 'str' object has no attribute '__module__'
```

After: returns the extrapolated `float`, for all 11 previously crashing entrypoints, in
both QASM dialects. Invalid strings now surface a `CircuitConversionError` instead of an
`AttributeError`.

## Tests

New file `mitiq/interface/mitiq_openqasm/tests/test_openqasm_pipeline.py` (26 tests):

- all 11 affected entrypoints, parametrized over OpenQASM 3.0 and 2.0 inputs,
- return-type preservation (`construct_circuits` on a QASM string returns QASM strings),
- measurement-bearing and multi-register circuits,
- a negative test asserting garbage input raises `CircuitConversionError`.

The new tests fail 26/26 on current `main` and pass 26/26 with this patch.

## Verified locally

```
uv run pytest -n auto -q --ignore=mitiq/interface/mitiq_pyquil
  2594 passed, 2 xfailed        (main at 3e1833d: 2568 passed, 2 xfailed)
uv run mypy mitiq               Success: no issues found in 113 source files
uv run ruff check               All checks passed!
uv run ruff format --check      3 files already formatted (changed files)
```

Python 3.11.9, cirq-core 1.6.1, qiskit 2.1.2, Linux. `make test-pyquil` was not run
(needs QVM/quilc servers); this change does not touch `mitiq/interface/mitiq_pyquil`.

## Out of scope

- `experimental.trex.execute_with_trex` fails on converted OpenQASM input for an
  unrelated reason (`KeyError: cirq.LineQubit(0)` from a hardcoded qubit in `trex.py`)
  — I'll file that separately.
- Converted QASM-string outputs are currently serialized via the QASM 2 emitter
  regardless of input dialect; that is pre-existing behavior and unchanged here.

## AI use

I used Claude (Anthropic)  while investigating and writing this change.
I chose the approach, reviewed every line, and reran the reproducer, the new tests, the
full suite, and the lint/type checks above myself. I can explain and defend any part of
it. The commit carries an `Assisted-by:` trailer.

---

### License

- [x] I license this contribution under the terms of the GNU GPL, version 3 and grant
  Unitary Foundation the right to provide additional permissions as described in
  section 7 of the GNU GPL, version 3.

- [x] I updated the documentation where relevant. *(No user-facing docs changes needed — behavior now matches what the docs already claim.)*

### DIFF
--- a/mitiq/interface/conversions.py
+++ b/mitiq/interface/conversions.py
@@ -305,6 +305,22 @@ def atomic_one_to_many_converter(
     return qprogram_modifier
 
 
+def _wrap_openqasm_string(circuit: QPROGRAM) -> QPROGRAM:
+    """Wrap a bare OpenQASM ``str`` in :class:`~mitiq.typing.QasmStringType`.
+
+    Downstream code (both here and inside :func:`convert_to_mitiq`) reads
+    ``circuit.__module__`` to dispatch by frontend. A plain ``str`` has no
+    ``__module__`` attribute; ``QasmStringType`` does. Wrapping preserves
+    the string content and lets the rest of the pipeline treat all inputs
+    uniformly.
+    """
+    if isinstance(circuit, str) and not hasattr(circuit, "__module__"):
+        from mitiq.typing import QasmStringType
+
+        return QasmStringType(circuit)
+    return circuit
+
+
 def accept_qprogram_and_validate(
     cirq_circuit_modifier: Callable[..., Any],
     one_to_many: bool = False,
@@ -328,6 +344,11 @@ def accept_qprogram_and_validate(
 
     @wraps(cirq_circuit_modifier)
     def new_function(circuit: QPROGRAM, *args: Any, **kwargs: Any) -> QPROGRAM:
+        # A bare ``str`` (OpenQASM program) has no ``__module__``. Mirror the
+        # wrap done in :func:`convert_to_mitiq` so downstream ``.__module__``
+        # reads on both the input and the returned circuit stay valid.
+        circuit = _wrap_openqasm_string(circuit)
+
         # Pre atomic conversion
         if "qiskit" in circuit.__module__:
             from qiskit.transpiler.passes import RemoveBarriers
--- a/mitiq/interface/mitiq_openqasm/tests/test_openqasm_pipeline.py
+++ b/mitiq/interface/mitiq_openqasm/tests/test_openqasm_pipeline.py
@@ -1,0 +1,207 @@
+# Copyright (C) Unitary Foundation
+#
+# This source code is licensed under the GPL license (v3) found in the
+# LICENSE file in the root directory of this source tree.
+
+"""End-to-end tests that error-mitigation entrypoints accept raw OpenQASM
+program strings (the frontend added in v0.49.0)."""
+
+import numpy as np
+import pytest
+from openqasm3 import parse
+
+from mitiq import ddd, lre, pec, zne
+from mitiq.ddd import rules
+from mitiq.interface import convert_to_mitiq
+from mitiq.pec.representations import (
+    represent_operations_in_circuit_with_local_depolarizing_noise,
+)
+from mitiq.pt import (
+    generate_pauli_twirl_variants,
+    twirl_CNOT_gates,
+    twirl_CZ_gates,
+)
+from mitiq.typing import QasmStringType
+
+QASM3_TWO_QUBIT = """OPENQASM 3.0;
+include "stdgates.inc";
+qubit[2] q;
+h q[0];
+cx q[0], q[1];
+"""
+
+QASM2_TWO_QUBIT = """OPENQASM 2.0;
+include "qelib1.inc";
+qreg q[2];
+h q[0];
+cx q[0], q[1];
+"""
+
+QASM3_MEAS = """OPENQASM 3.0;
+include "stdgates.inc";
+qubit[2] q;
+bit[2] c;
+h q[0];
+cx q[0], q[1];
+c = measure q;
+"""
+
+QASM3_MULTI_REG = """OPENQASM 3.0;
+include "stdgates.inc";
+qubit[1] a;
+qubit[1] b;
+h a[0];
+cx a[0], b[0];
+"""
+
+
+def _executor(_circuit):
+    return 0.5
+
+
+def _assert_qasm_list(circuits, expected_len):
+    """Assert circuits is a list of QASM strings that re-parse as OpenQASM."""
+    assert isinstance(circuits, list)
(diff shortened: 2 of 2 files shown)


mitiq/interface/mitiq_openqasm/tests/test_openqasm_pipeline.py
```diff
@@ -0,0 +1,207 @@
+# Copyright (C) Unitary Foundation
+#
+# This source code is licensed under the GPL license (v3) found in the
+# LICENSE file in the root directory of this source tree.
+
+"""End-to-end tests that error-mitigation entrypoints accept raw OpenQASM
+program strings (the frontend added in v0.49.0)."""
```
natestemen, line 7:
```suggestion
program strings."""
```

mitiq/interface/mitiq_openqasm/tests/test_openqasm_pipeline.py
```diff
@@ -0,0 +1,207 @@
...
+h a[0];
+cx a[0], b[0];
+"""
+
+
+def _executor(_circuit):
+    return 0.5
+
+
+def _assert_qasm_list(circuits, expected_len):
```
natestemen, line 62: Since these are tests I don't think we need to have them be "private". Lets do `executor` and `assert_qasm_list`.

mitiq/interface/conversions.py
natestemen, line 87: should we replace this block with the function you defined?

natestemen (approved):
LGTM! Thanks for catching this. Just three minor comments/suggestions.

PR #3126, opened 2026-09-05
make agents work better with mitiq

natestemen:
## Description

The latest and greatest AI infra... I think? Anything I missed/should add?

---

### License

### DIFF
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -13,13 +13,31 @@ For more information, check the Mitiq style guidelines (https://mitiq.readthedoc
 
 ## Description
 
-<!-- Please explain the changes you made here. -->
-
+<!--
+Please explain the changes you made here.
+Please write this PR's description by hand, even if AI tooling wrote some of the code.
+Explaining why the change is worth making, and where you are unsure, is the part a
+reviewer cannot get from reading the diff.
+-->
 ---
 
 ### License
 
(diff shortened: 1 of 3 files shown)


AGENTS.md
bachase, line 1: Although clifft has an AGENTS.md, I've not stayed on top of it or know how useful it is in practice. I like this one and the list of boundaries, gotchas and things that are unique to mitq.

Workflow wise curious if an AI generated it?

Secondly, have you tried using it with and without an AI generated solution to see if it helps? Not sure that is so important before merging this, more curious.

.github/pull_request_template.md
```diff
@@ -13,14 +13,32 @@ For more information, check the Mitiq style guidelines (https://mitiq.readthedoc
 
 ## Description
 
-<!-- Please explain the changes you made here. -->
-
+<!--
```
bachase, line 16: I'm supportive of this requirement (but admit I'm not following it in my work today). There may be cases where you still want an AI generated section, even if the primary description is human written?

bachase (approved):
Not much I have to add. Honestly, I'll be following this to see how it goes =).

AGENTS.md
natestemen: Indeed it was generated by an LLM. I asked it to collect specifics of how the mitiq project does development that are maybe more unique to the project.

I haven't tried using it, so we'll see!

.github/pull_request_template.md
natestemen: Good point, I'll add a note about that.

PR #3127, opened 2026-09-05
Ship py.typed marker so mitiq's type hints reach downstream users

natestemen:
According to [PEP 561](https://peps.python.org/pep-0561/#packaging-type-information), `py.typed` is a necessary file to indicate to downstream packages that you are providing type hints. TIL.

---

Disclosure: the code in this PR was written by an AI assistant (Claude Opus 5) under my direction.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -159,3 +159,6 @@ exclude_lines = [
 
 [tool.setuptools.packages.find]
 where = ["."]
+
+[tool.setuptools.package-data]
+mitiq = ["py.typed"]


cosenal (approved):
TIL: one can't tag copilot in PR reviews

pyproject.toml
```diff
@@ -159,3 +159,6 @@ exclude_lines = [
 
 [tool.setuptools.packages.find]
 where = ["."]
+
+[tool.setuptools.package-data]
+mitiq = ["py.typed"]
```
purva-thakre, line 164: Could you help me understand why you needed to add this? According to the docs for setuptools, this is included by default in versions beyond v0.69 and I see mitiq relies on a much higher version. https://setuptools.pypa.io/en/latest/userguide/miscellaneous.html

Since the docs state this is an experimental feature, is something downstream broken?

This has always confused me :(

purva-thakre: Ah, never mind. I just saw https://github.com/unitaryfoundation/mitiq/pull/3127#issuecomment-5552835471

PR #3131, opened 2026-09-06
Fix batched Executor.evaluate slicing for multi-group observables

alyabouzaid:
### Problem

When an executor returns `MeasurementResult` and the observable partitions into more than one qubit-wise-commuting group, `Executor.evaluate` runs `ngroups` circuits per input circuit and then regroups the results. The regrouping used a stride of one instead of `ngroups`:

```python
all_results[i : i + result_step] for i in range(len(all_results) // result_step)
```

So every circuit after the first was paired with measurement data belonging to other circuits, and the trailing results were never used.

Reproduction on `main` (`Observable(PauliString("X"), PauliString("Z"))` has 2 groups):

```python
obs = Observable(PauliString("X"), PauliString("Z"))
circuits = [Circuit(I(q)), Circuit(H(q)), Circuit(X(q))]   # <X>+<Z> = 1, 1, -1
Executor(execute).evaluate(circuits, obs)
# main: [0.99, 2.00, 1.01]      <- wrong
# here: [1.01, 1.01, -0.99]
```

Evaluating the circuits one at a time already gave the right answers, since only the `i = 0` slice was correct. This reaches users through `execute_with_zne` and friends, which pass all scaled circuits to `evaluate` in one call.

### Fix

Slice with a stride of `result_step`. One-line change in `mitiq/executor/executor.py`.

### Verification

- Added `test_executor_evaluate_measurements_multiple_groups` (serial and batched executors); confirmed it fails on `main` and passes with the fix. The existing `test_executor_evaluate_measurements` only used a single-group observable, which is why this was not caught.
- `pytest mitiq` (excluding `mitiq_pyquil`, which needs a local QVM): 2583 passed, 2 xfailed.
- `ruff check` + `ruff format --check` clean, `mypy mitiq` clean.
- Exercised end to end with the snippet above.

### DIFF
--- a/mitiq/executor/executor.py
+++ b/mitiq/executor/executor.py
@@ -261,7 +261,7 @@ class Executor:
                 observable._expectation_from_measurements(
                     all_results[i : i + result_step]
                 )
-                for i in range(len(all_results) // result_step)
+                for i in range(0, len(all_results), result_step)
             ]
 
         else:
--- a/mitiq/executor/tests/test_executor.py
+++ b/mitiq/executor/tests/test_executor.py
@@ -261,6 +261,28 @@ def test_executor_evaluate_measurements(execute):
 
 
 @pytest.mark.parametrize(
+    "execute", [executor_measurements_typed, executor_measurements_batched]
+)
+def test_executor_evaluate_measurements_multiple_groups(execute):
+    # X and Z do not qubit-wise commute, so each input circuit is measured
+    # with two different circuits, one per group.
+    obs = Observable(PauliString("X"), PauliString("Z"))
+    assert obs.ngroups == 2
+
+    q = cirq.LineQubit(0)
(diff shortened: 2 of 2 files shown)


natestemen (approved):
Wow, nice find! Looks like this has been a bug for almost five years https://github.com/unitaryfoundation/mitiq/pull/1001!

This makes me wonder if people aren't really using mitiq observables 🤔. Out of curiousity did you stumble on this while using mitiq or did you ask an agent to find bugs? Totally impartial, just curious.

alyabouzaid:
Thanks for the review and merge! To answer your question honestly: the latter. This one came from having Claude Code read through the executor source looking for bugs, not from using mitiq day to day. The reproduction and the failing-on-main regression test were how the finding got confirmed before opening the PR.

PR #3143, opened 2026-09-08
restrict dependabot updates to direct dependencies

natestemen:
## Description

Removes the `allow: dependency-type: all` override from Dependabot’s uv configuration.

This restores Dependabot’s default behavior: routine version-update PRs cover direct dependencies only, while Dependabot security updates can still remediate vulnerable dependencies recorded in `uv.lock`.

---

### License

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -6,8 +6,6 @@ updates:
     directory: "/"
     schedule:
       interval: "daily"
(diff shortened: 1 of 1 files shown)


cosenal (approved):
⚠️ docs build fails on a cached nb for reasons I don't fully understand

